PR #23, opened 2019-10-06
Fix Postgis version check

Krizz:
Fixes Postgis version check to allow Beta/RC versions.


stepankuzmin (approved):

stepankuzmin:
Thanks for the fix, @Krizz!

PR #95, opened 2020-06-05
add tls connection envs 

kapcsandi:
You can set extra tls connection options with theese environment variables:
- CA_ROOT_FILE
- CLIENT_PKCS12_FILE
- CLIENT_PKCS12_PASS

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -1,7 +1,9 @@
-use std::io;
+use std::{env, io};
 use std::str::FromStr;
 
-use native_tls::TlsConnector;
+use native_tls::{Certificate, Identity, TlsConnector};
+use std::fs::File;
+use std::io::{Read};
 use postgres_native_tls::MakeTlsConnector;
 use r2d2::PooledConnection;
 use r2d2_postgres::PostgresConnectionManager;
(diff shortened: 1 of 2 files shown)


src/db.rs
```diff
@@ -15,7 +17,48 @@ pub type Pool = r2d2::Pool<ConnectionManager>;
 pub type Connection = PooledConnection<ConnectionManager>;
 
 fn make_tls_connector(danger_accept_invalid_certs: bool) -> io::Result<MakeTlsConnector> {
-    let connector = TlsConnector::builder()
+    let key = "CA_ROOT_FILE";
+    let ca_file = match env::var_os(key) {
+        Some(s) => s.into_string().unwrap(),
```
stepankuzmin, line 22: Could you please not unwrap values and use `map_err` and `prettify_error` for that?

src/db.rs
```diff
@@ -15,7 +17,48 @@ pub type Pool = r2d2::Pool<ConnectionManager>;
 pub type Connection = PooledConnection<ConnectionManager>;
 
 fn make_tls_connector(danger_accept_invalid_certs: bool) -> io::Result<MakeTlsConnector> {
-    let connector = TlsConnector::builder()
+    let key = "CA_ROOT_FILE";
+    let ca_file = match env::var_os(key) {
```
stepankuzmin, line 21: Could you please move repeating parts to a separate function and use `info!` macros then all provided variables are defined?

src/db.rs
```diff
@@ -15,7 +17,48 @@ pub type Pool = r2d2::Pool<ConnectionManager>;
...
+        None => {
+            println!("{} is not defined in the environment", key);
+            String::default()
+        }
+    };
+
+    let mut builder = TlsConnector::builder();
+
+    if !client_identity_file.is_empty() {
+        let mut file = File::open(&client_identity_file).unwrap();
```
stepankuzmin, line 48: Could you please move repeating parts into a separate function?

stepankuzmin (changes requested):
Hi @kapcsandi and thanks for the PR! I really appreciate your work, but could you please fix these minor comments? You may also add some tests for this if you'd like.

PR #124, opened 2020-09-29
Add bounds to table info

jaspervercnocke:
Use exact bounds in table info based on the sql query: "SELECT ST_Extent(geom)::TEXT as bbox from public.tableName"

**Result**
`{"tilejson":"2.2.0","id":null,"name":"public.tableName","description":null,"version":"1.0.0","attribution":null,"template":null,"legend":null,"scheme":"xyz","tiles":["http://localhost:8080/public. public.tableName/{z}/{x}/{y}.pbf"],"grids":null,"data":null,"minzoom":0,"maxzoom":30,"bounds":[3.8209655,50.95215,3.922962,51.028217],"center":null}`

### DIFF
--- a/src/dev.rs
+++ b/src/dev.rs
@@ -26,6 +26,7 @@ pub fn mock_table_sources() -> Option<TableSources> {
         clip_geom: Some(true),
         geometry_type: None,
         properties: HashMap::new(),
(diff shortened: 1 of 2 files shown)


stepankuzmin:
Hi @jaspervercnocke,

Thanks for the contribution. I've modified your suggestion and did it in #260

PR #168, opened 2021-01-14
Detect Martin server heartbeat for monitoring the service health

yamaszone:
Detect Martin server heartbeat for monitoring the service health:
- Simply returns HTTP `200 OK`

NOTE: we have deployed Martin via Kubernetes. This PR helps to do a lightweight ping by both Kubelet as well as load balancer probes to detect the service health before auto-restart on failure or sending traffic. This is also used by monitors to alert on unexpected failures.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -42,6 +42,11 @@ struct TileRequest {
     format: String,
 }
 
+async fn get_health() -> Result<HttpResponse, Error> {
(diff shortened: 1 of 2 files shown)


stepankuzmin (approved):
LGTM

stepankuzmin:
Hey @yamaszone

Thanks for the PR, merged

PR #216, opened 2021-06-04
fix: escape table and schema in function source tiles url

gbip:
Since the end user controls the query json that is
passed to postgres, it is possible to perform
a SQL injection on the function source tiles endpoint.
By preparing the query we avoid such issues. However
we still need to sanitize by hand the schema and function
parameters since they can't be a part of the prepared query.

Using prepared queries also forces us to use the right
data types when passing parameters to postgres. As such,
the coordinates system that were using unsigned
integers uses signed integers since they map directly
to postgres types.

Thanks to Frédéric Rodrigo for discovering this issue.

Before merging this PR, there are a few things to consider :
1) Does using i32 instead of u32 leads to range issues ?
2) What to do about this cast :  `let res = (max * 2.0) / f64::from(2_i32.pow(z as u32));`

Once this PR is in an acceptable state, we should consider fixing the other endpoints.

### DIFF
--- a/src/function_source.rs
+++ b/src/function_source.rs
@@ -1,9 +1,9 @@
 use postgres::types::{Json, Type};
+use postgres_protocol::escape::escape_identifier;
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::io;
 use tilejson::{TileJSON, TileJSONBuilder};
-use postgres_protocol::escape::escape_identifier;
 
 use crate::db::Connection;
 use crate::source::{Query, Source, Tile, Xyz};
--- a/src/function_source.rs
+++ b/src/function_source.rs
@@ -3,6 +3,7 @@ use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::io;
 use tilejson::{TileJSON, TileJSONBuilder};
+use postgres_protocol::escape::escape_identifier;
 
 use crate::db::Connection;
 use crate::source::{Query, Source, Tile, Xyz};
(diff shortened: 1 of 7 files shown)


src/utils.rs
```diff
@@ -13,7 +13,7 @@ pub fn tilebbox(xyz: &Xyz) -> String {
     let z = xyz.z;
 
     let max = 20_037_508.34;
-    let res = (max * 2.0) / f64::from(2_i32.pow(z));
+    let res = (max * 2.0) / f64::from(2_i32.pow(z as u32));
```
gbip, line 16: Watch out, I'm not sure about this.

gbip:
@stepankuzmin Can you review this please ?

src/function_source.rs
```diff
@@ -43,18 +44,30 @@ impl Source for FunctionSource {
...
             .map_err(|err| io::Error::new(io::ErrorKind::Other, err))?;
 
-        let query = format!(
+        // Query preparation : the schema and function can't be part of a prepared query, so they need to be escaped by hand
+        // The query expects the following arguments :
+        // $1 : x
+        // $2 : y
+        // $3 : z
+        // $4 : query_json
+        let escaped_schema = self.schema.replace("\"", "\"\"");
```
regilero: I think using the Postgres escapeIdentifier() function would be better here

gbip: Let me look into that.

gbip: Found https://docs.rs/postgres-protocol/0.6.1/postgres_protocol/escape/fn.escape_identifier.html

stepankuzmin:
Hey @gbip! Thanks for the PR, it looks awesome. But could you please run rustfmt on this?

gbip:
Sure !

stepankuzmin:
Thanks! Good job 👍

PR #224, opened 2021-07-12
Fix invalid json escaping

### LINKED ISSUE #223: Recent pull request broke parameter passing via json
I have the following function:
```
CREATE OR REPLACE FUNCTION v2.license_geometries(z integer, x integer, y integer, query_params json) RETURNS bytea AS $$
DECLARE
    mvt bytea;
BEGIN
    RAISE NOTICE 'params: %',query_params;
    WITH tile AS (
        SELECT distinct
            geom_id,
            ST_AsMVTGeom(ST_Transform(geom, 3857), TileBBox(z, x, y, 3857), 4096, 64, true) AS geom
        FROM v2.license_geometries g
        INNER JOIN v2.license_metadata m
            USING (geom_id)
        WHERE geom && TileBBox(z, x, y, 4326)
            AND call_sign = (query_params->>'call_sign')::varchar
    )
    SELECT INTO mvt ST_AsMVT(t.*, 'v2.license_geometries', 4096, 'geom')
    FROM tile t;
    RETURN mvt;
END
$$
LANGUAGE plpgsql
IMMUTABLE STRICT
PARALLEL SAFE;
```
Before this [recent pull request](https://github.com/urbica/martin/pull/216), the martin endpoint would return correctly, but now it will not. 

I've traced it down to the json object being escaped somehow. 

When I call it manually in a postgres console: `select v2.license_geometries(7,34,48,'{"call_sign":"WMH528"}');` the debug notice shows: `params: {"call_sign":"WMH528"}`.

However, when martin calls this endpoint: `/rpc/v2.license_geometries/7/34/48.pbf?call_sign=WMH528`, I get this for the notice: `... postgres::config] NOTICE: params: "{\"call_sign\":\"WMH528\"}"`

It appears that the previous version passed in a json object as the fourth parameter, but the latest version passes in a stringified version of the json object, which breaks all json parsing behavior within the function. It's not clear to me if that is intended behavior or not (the PR mentions it addresses potential SQL injection), but if it is, I would think it would be a semver major version bump and a docs change. 

Let me know if you can't replicate (it's hard to clone our database for replication purposes, but the function could be rewritten for debugging this specific issue).

gbip:
This fixes #223.

@saosebastiao Could you report if this fixes your issue ?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,6 +98,7 @@ jobs:
           psql -h $POSTGRES_HOST -p $POSTGRES_PORT -U postgres -d test -f tests/fixtures/points1_source.sql
           psql -h $POSTGRES_HOST -p $POSTGRES_PORT -U postgres -d test -f tests/fixtures/points2_source.sql
           psql -h $POSTGRES_HOST -p $POSTGRES_PORT -U postgres -d test -f tests/fixtures/function_source.sql
+          psql -h $POSTGRES_HOST -p $POSTGRES_PORT -U postgres -d test -f tests/fixtures/function_source_query_params.sql
         env:
           POSTGRES_HOST: localhost
           POSTGRES_PORT: ${{ job.services.postgres.ports[5432] }}
--- a/.github/workflows/grcov.yml
+++ b/.github/workflows/grcov.yml
@@ -33,6 +33,7 @@ jobs:
           psql -h $POSTGRES_HOST -p $POSTGRES_PORT -U postgres -d test -f tests/fixtures/points1_source.sql
           psql -h $POSTGRES_HOST -p $POSTGRES_PORT -U postgres -d test -f tests/fixtures/points2_source.sql
           psql -h $POSTGRES_HOST -p $POSTGRES_PORT -U postgres -d test -f tests/fixtures/function_source.sql
+          psql -h $POSTGRES_HOST -p $POSTGRES_PORT -U postgres -d test -f tests/fixtures/function_source_query_params.sql
         env:
           POSTGRES_HOST: localhost
           POSTGRES_PORT: ${{ job.services.postgres.ports[5432] }}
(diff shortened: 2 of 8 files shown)


saosebastiao:
Yes, this works correctly for our specific function source. 

It might be good to modify the `public.function_source` function in the test suite to test for json parseability. 

```
IF (query_params->>'sample_param')::varchar IS NULL THEN
    RAISE EXCEPTION 'the `sample_param` json parameter does not exist in `query_params`';
END
```

Then a test case could be written that feeds it `'{"sample_param":"hello world"}'::json`. If the query_params is an actual parseable json object, it will not throw an exception.

gbip:
I agree that adding a test case would be a good idea, however I have more urgent tasks right now.

gbip:
> Yes, this works correctly for our specific function source.
> 
> It might be good to modify the `public.function_source` function in the test suite to test for json parseability.
> 
> ```
> IF (query_params->>'sample_param')::varchar IS NULL THEN
>     RAISE EXCEPTION 'the `sample_param` json parameter does not exist in `query_params`';
> END
> ```
> 
> Then a test case could be written that feeds it `'{"sample_param":"hello world"}'::json`. If the query_params is an actual parseable json object, it will not throw an exception.

I have been investigating a way for us to have some kind of unit tests.
However I am lacking sql skills and I could not find a way to write a function that tests if sample_param is not null.
If you happen to have written I can try to integrate it into the test suite, otherwise my PR will be ready for review.

stepankuzmin:
Hey @gbip,

Thanks a lot for the PR! Could you please [allow edits from maintainers](https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork), so I can add tests for this case?

gbip:
It is on :)
2021-07-16-094331_271x264_scrot

stepankuzmin:
Awesome, merged! Thank you, @gbip

frodrigo:
Since this fix a security issue, I suggest making a new release.

PR #274, opened 2021-11-01
CORS - Add GET to the list of allowed_methods

AndreaGiardini:
Thanks to @chanange for the hint

We should add `GET` to the list of allowed_methods otherwise it's `deny all` by default, as documented here: https://actix.rs/actix-extras/actix_cors/struct.Cors.html#method.default

This allows CORS to work correctly when deployed without nginx

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -407,7 +407,7 @@ pub fn new(pool: Pool, config: Config) -> SystemRunner {
         let cors_middleware = Cors::default()
             .allow_any_origin()
             .allowed_methods(vec!["GET"]);
-        
+
         App::new()
             .data(state)
             .wrap(cors_middleware)
--- a/src/server.rs
+++ b/src/server.rs
@@ -404,8 +404,10 @@ pub fn new(pool: Pool, config: Config) -> SystemRunner {
     HttpServer::new(move || {
         let state = create_state(db.clone(), coordinator.clone(), config.clone());
 
-        let cors_middleware = Cors::default().allow_any_origin().allowed_methods(vec!["GET"]);
-
+        let cors_middleware = Cors::default()
+            .allow_any_origin()
+            .allowed_methods(vec!["GET"]);
+        
         App::new()
             .data(state)
             .wrap(cors_middleware)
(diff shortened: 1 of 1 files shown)


AndreaGiardini:
Hello @stepankuzmin 

I am asking myself if we should also add `.allowed_headers(vec![header::AUTHORIZATION, header::ACCEPT]);` to the CORS settings. In our case, we are deploying martin behind authentication (as I believe most of the people out there) and without this header CORS will still fail. What do you think?

stepankuzmin:
Hi @AndreaGiardini,

Thanks for the contribution! That sounds reasonable, but I'm concerned about security in that case. It would be a nice feature to allow users to configure CORS using the config file in the long term. What do you think about this approach?

In the meantime, you can add allowed headers using your proxy (e.g., Nginx).

AndreaGiardini:
Hey Stepan

We are not using nginx. We run martin in Google cloud run.

From the loadbalancer I have the possibility to modify the headers of requests/responses. It seems to be impossible to stip them completely though... I can just put their value to an empty string but actix does not seem to like that unfortunately. Do you know of any workaround?

cduruk:
> Hi @AndreaGiardini,
> 
> Thanks for the contribution! That sounds reasonable, but I'm concerned about security in that case. It would be a nice feature to allow users to configure CORS using the config file in the long term. What do you think about this approach?
> 
> In the meantime, you can add allowed headers using your proxy (e.g., Nginx).

Sorry to jump in. We used to use Martin but as we moved to Render (a PasS provider, similar to Heroku), we ran into this issue so ended up having to put a proxy in front of it and having a config option would have helped a lot.

Another thing to note is browsers will actually check for CORS with a OPTIONS pre-flight request, so that's something to test before calling this and/or related work done.

AndreaGiardini:
> Another thing to note is browsers will actually check for CORS with a OPTIONS pre-flight request, so that's something to test before calling this and/or related work done.

With the patch I posted this seems to work just fine. Martin correctly handles OPTIONS and GET requests

PR #285, opened 2021-11-26
fix: #279 500 error on Uppercase Table name

### LINKED ISSUE #279: Table names with uppercase letters cause http 500 response
When requesting a tile using a table name that has an uppercase letter in it (public.City/{z}/{x}/{y}.pbf) generates the following SQL

`WITH bounds AS (SELECT ST_MakeEnvelope(-5009377.085000001, 5009377.085000001, -0.0000000009313225746154785, 0.0000000009313225746154785, 3857) as mercator, ST_Transform(ST_MakeEnvelope(-5009377.085000001, 5009377.085000001, -0.0000000009313225746154785, 0.0000000009313225746154785, 3857), 4326) as original)

    SELECT ST_AsMVT(tile, 'public.Cities', 4096, 'geom' ) FROM (

      SELECT

        ST_AsMVTGeom(ST_Transform(city_center_point, 3857), bounds.mercator, 4096, 64, true) AS geom , "field1","field2"

      FROM public.Cities, bounds

      WHERE city_center_point && bounds.original

    ) AS tile WHERE geom IS NOT NULL

`
Which causes an SQL error
`ERROR:  relation "public.cities" does not exist
LINE 9:       FROM public.Cities, bounds`

The correct Postgres SQL should be 
`public."Cities"`

I am using version v0.5.0 in docker.

kzima:
Postgres requires capital case Table names to be wrapped in double quotes.


stepankuzmin:
Hi @kzima,

Thanks for the contribution!

PR #300, opened 2022-01-12
Fix get_bound.sql to return Polygon everytime

HackJack-101:
To solve the bug https://github.com/urbica/martin/issues/299 we propose a solution which expands by 1 unit the Point returned by ST_Extent() in case of a table with only one Point or with multiple Points with the same geometry.

The main problem here still that we lie about the real boundaries because of ST_Expand.


stepankuzmin (approved):
Hi @HackJack-101

Thanks for the contribution!

PR #338, opened 2022-05-27
Remove extern, optimize use, fmt str

nyurik:
* get rid of the extern statements
* `use log`
* group use statements
* simplify escaped quotes with raw string
* simplify a few format strings


stepankuzmin (approved):
LGTM 👍

PR #341, opened 2022-05-29
Use tilejson 0.3.1 and simplify

nyurik:
* bounds merging is now built-in into Bounds as Add/AddAssign
* use Bounds::MAX instead of hard-coding it everywhere


stepankuzmin (approved):
👍

PR #352, opened 2022-06-20
Optimize prettify_error function

### LINKED ISSUE #348: Optimize prettify_error
Looking at the code, it seems the `prettify_error` causes some significant slowdowns because it gets called on every request, creating an owned string https://github.com/maplibre/martin/blob/4a26366a606cdb7371828b57105c896552dd66a9/src/utils.rs#L11-L13 I suspect the intention was to call it only on error.  Also, it might be more efficient if there were two functions - one for the static str, and another with a formatter.

sharkAndshark:
Resolves #348 
update prettify_error to one single macro
```rust
#[macro_export]
macro_rules! prettify_error {
    ($error:ident,$info:expr,$($arg:expr),+) => {
        {
            std::io::Error::new(std::io::ErrorKind::Other,format!("{}: {}",
            format!($info,$($arg),*),
            $error))
        }
    };
    ($error:ident,$info:literal) => {
        {
            std::io::Error::new(std::io::ErrorKind::Other,format!("{}: {}",$info,$error))
        }
    }
}

pub(crate) use prettify_error;
```

### DIFF
--- a/src/bin/main.rs
+++ b/src/bin/main.rs
@@ -7,9 +7,8 @@ use serde::Deserialize;
 use martin::config::{read_config, Config, ConfigBuilder};
 use martin::db::{check_postgis_version, get_connection, setup_connection_pool, Pool};
 use martin::function_source::get_function_sources;
-use martin::server;
 use martin::table_source::get_table_sources;
-use martin::utils::prettify_error;
+use martin::{prettify_error, server};
 
 const VERSION: &str = env!("CARGO_PKG_VERSION");
 const REQUIRED_POSTGIS_VERSION: &str = ">= 2.4.0";
@@ -83,7 +82,7 @@ pub fn generate_config(args: Args, pool: &Pool) -> io::Result<Config> {
 }
 
 fn setup_from_config(file_name: String) -> io::Result<(Config, Pool)> {
-    let config = read_config(&file_name).map_err(prettify_error("Can't read config".to_owned()))?;
+    let config = read_config(&file_name).map_err(|e| prettify_error!(e, "Can't read config"))?;
 
     let pool = setup_connection_pool(
         &config.connection_string,
@@ -91,7 +90,7 @@ fn setup_from_config(file_name: String) -> io::Result<(Config, Pool)> {
         Some(config.pool_size),
         config.danger_accept_invalid_certs,
     )
-    .map_err(prettify_error("Can't setup connection pool".to_owned()))?;
+    .map_err(|e| prettify_error!(e, "Can't setup connection pool"))?;
 
     if let Some(table_sources) = &config.table_sources {
         for table_source in table_sources.values() {
@@ -134,11 +133,11 @@ fn setup_from_args(args: Args) -> io::Result<(Config, Pool)> {
         args.flag_pool_size,
         args.flag_danger_accept_invalid_certs,
     )
-    .map_err(prettify_error("Can't setup connection pool".to_owned()))?;
+    .map_err(|e| prettify_error!(e, "Can't setup connection pool"))?;
 
     info!("Scanning database");
     let config =
-        generate_config(args, &pool).map_err(prettify_error("Can't generate config".to_owned()))?;
+        generate_config(args, &pool).map_err(|e| prettify_error!(e, "Can't generate config"))?;
 
     Ok((config, pool))
 }
@@ -190,7 +189,7 @@ fn start(args: Args) -> io::Result<actix::SystemRunner> {
     };
 
     let matches = check_postgis_version(REQUIRED_POSTGIS_VERSION, &pool)
-        .map_err(prettify_error("Can't check PostGIS version".to_owned()))?;
+        .map_err(|e| prettify_error!(e, "Can't check PostGIS version"))?;
 
     if !matches {
         std::process::exit(-1);
@@ -209,7 +208,7 @@ fn main() -> io::Result<()> {
 
     let args = Docopt::new(USAGE)
         .and_then(|d| d.help(false).deserialize::<Args>())
-        .map_err(prettify_error("Can't parse CLI arguments".to_owned()))?;
+        .map_err(|e| prettify_error!(e, "Can't parse CLI arguments"))?;
 
     let args = parse_env(args);
 
--- a/src/function_source.rs
+++ b/src/function_source.rs
@@ -101,17 +101,22 @@ impl Source for FunctionSource {
                 &raw_query,
                 &[Type::INT4, Type::INT4, Type::INT4, Type::JSON],
             )
-            .map_err(prettify_error(
-                "Can't create prepared statement for the tile".to_owned(),
-            ))?;
+            .map_err(|e| prettify_error!(e, "Can't create prepared statement for the tile"))?;
 
         let tile = conn
             .query_one(&query, &[&xyz.x, &xyz.y, &xyz.z, &query_json])
             .map(|row| row.get(self.function.as_str()))
-            .map_err(prettify_error(format!(
-                r#"Can't get "{}" tile at /{}/{}/{} with {:?} params"#,
-                self.id, &xyz.z, &xyz.x, &xyz.z, &query_json
-            )))?;
+            .map_err(|error| {
+                prettify_error!(
+                    error,
+                    r#"Can't get "{}" tile at /{}/{}/{} with {:?} params"#,
+                    self.id,
+                    xyz.z,
+                    xyz.x,
+                    xyz.z,
+                    query_json
+                )
+            })?;
 
         Ok(tile)
     }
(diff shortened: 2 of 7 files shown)


src/utils.rs
```diff
@@ -8,10 +8,24 @@ use tilejson::Bounds;
 
 use crate::source::{Query, Xyz};
 
-pub fn prettify_error<E: std::fmt::Display>(message: String) -> impl Fn(E) -> std::io::Error {
-    move |error| std::io::Error::new(std::io::ErrorKind::Other, format!("{message}: {error}"))
+#[macro_export]
```
nyurik, line 11: not certain, but i think we don't need macro_export - without it, the macro will simply need to be `use`-ed just like any normal fn?

src/bin/main.rs
```diff
@@ -83,15 +82,16 @@ pub fn generate_config(args: Args, pool: &Pool) -> io::Result<Config> {
...
+        read_config(&file_name).map_err(|error| prettify_error!(error, "Can't read config"))?;
 
     let pool = setup_connection_pool(
         &config.connection_string,
         &config.ca_root_file,
         Some(config.pool_size),
         config.danger_accept_invalid_certs,
     )
-    .map_err(prettify_error("Can't setup connection pool".to_owned()))?;
+    .map_err(|error| prettify_error!(error, "Can't setup connection pool"))?;
```
nyurik: I have been thinking that this `|error|` is just far too verbose for the error code, and gets in a way.  Could you maybe rename it to `e` ? Not a blocker, but will keep things a bit cleaner.

src/function_source.rs
```diff
@@ -101,17 +101,24 @@ impl Source for FunctionSource {
...
-            .map_err(prettify_error(format!(
-                r#"Can't get "{}" tile at /{}/{}/{} with {:?} params"#,
-                self.id, &xyz.z, &xyz.x, &xyz.z, &query_json
-            )))?;
+            .map_err(|error| {
+                prettify_error!(
+                    error,
+                    r#"Can't get "{}" tile at /{}/{}/{} with {:?} params"#,
+                    self.id,
+                    &xyz.z,
```
nyurik: wouldn't it work as well without the `&` (i know this is not exactly part of your change, but since these are the lines being changed, might as well?  Also not a blocker.

src/table_source.rs
```diff
@@ -164,10 +164,16 @@ impl Source for TableSource {
...
-            .map_err(utils::prettify_error(format!(
-                r#"Can't get "{}" tile at /{}/{}/{}"#,
-                self.id, &xyz.z, &xyz.x, &xyz.z
-            )))?;
+            .map_err(|error| {
+                utils::prettify_error!(
+                    error,
+                    r#"Can't get "{}" tile at /{}/{}/{}"#,
+                    self.id,
+                    &xyz.z,
```
nyurik: same `&` question

src/utils.rs
```diff
@@ -8,10 +8,24 @@ use tilejson::Bounds;
...
+macro_rules! prettify_error {
+    ($error:ident,$info:expr,$($arg:expr),+) => {
```
nyurik: spaces keep readers happy :)  I know `cargo fmt` doesn't try to reformat macros (yet), so we will have to do the heavy lifting.  Unlike other non-important stuff, good format of the source code will keep other developers much happier when they try to understand why nothing works :)

Also, we want to allow users to have a (single) trailing comma, here and in other macro variants.

```suggestion
    ($error:ident, $info:expr, $($arg:expr),+ $(,)?) => {
```

src/utils.rs
```diff
@@ -8,10 +8,24 @@ use tilejson::Bounds;
...
+        {
+            std::io::Error::new(std::io::ErrorKind::Other,format!("{}: {}",
+            format!($info,$($arg),*),
+            $error))
```
nyurik: When in a macro, you always want to be explicit with the scope just for hygienic purposes -- I think `::std::...` is always preferable to `std::...` -- keeps things in a global space.

src/utils.rs
```diff
@@ -8,10 +8,24 @@ use tilejson::Bounds;
...
+        }
+    };
+    ($error:ident,$info:literal) => {
+        {
+            std::io::Error::new(std::io::ErrorKind::Other,format!("{}: {}",$info,$error))
```
nyurik: If possible, we don't want to do two separate string allocations (one in each `format!`), so maybe do static magic with `concat!()`?   I am not certain this will work, but if its a literal, it should?

```suggestion
            ::std::io::Error::new(::std::io::ErrorKind::Other, format!(concat!($info, ": {}", $error))
```

nyurik (commented):
just some basic thoughts, most of them are ideas, not certainties (except the spaces and the optional trailing commas)

nyurik:
P.S.  Thank you for working on this!!!  Rust could be a challenge at times :)

sharkAndshark:
> P.S. Thank you for working on this!!! Rust could be a challenge at times :)

:D Thanks  for your reviewing.I learn a lot from it @nyurik I will try resolve these stuffs.

src/utils.rs
nyurik: @sharkAndshark so can this line be removed now?

src/utils.rs
```diff
@@ -8,10 +8,24 @@ use tilejson::Bounds;
 
 use crate::source::{Query, Xyz};
 
-pub fn prettify_error<E: std::fmt::Display>(message: String) -> impl Fn(E) -> std::io::Error {
-    move |error| std::io::Error::new(std::io::ErrorKind::Other, format!("{message}: {error}"))
+#[macro_export]
+macro_rules! prettify_error {
```
nyurik, line 12: P.S. I just realized that this macro is nearly identical to the `assert_eq!(a, b, str, params...)` -- see https://github.com/rust-lang/rust/blob/master/library/core/src/macros/mod.rs#L50-L62

nyurik: More specifically this part:

```
($left:expr, $right:expr, $($arg:tt)+) => {
...
$crate::format_args!($($arg)+))
```

src/utils.rs
sharkAndshark: I am sure  this is necessary for crossing crates calling.  I'm trying to find some official doc about this.

src/utils.rs
sharkAndshark: > 

Thanks!You are so nice  :D I am checking it.Macro was something like magic to me ever😂,now i know it.

src/utils.rs
sharkAndshark: Hi @nyurik   
With [RFC1561](https://rust-lang.github.io/rfcs/1561-macro-naming.html) there are no `#[macro_use]` or `#[macro_export]` annotations are required.
But we need to switch to  unstable rust channel.By the way,i didn't find the rust version of martin using in readme.md.
demo codes like:  
*src/lib.rs*
```rust
#![feature(decl_macro)]
```
*src/util.rs*
```rust
pub macro prettify_error{
    ($error:ident, $info:expr, $($arg:expr),+ $(,)?) => {
        {
            ::std::io::Error::new(::std::io::ErrorKind::Other, format!("{}: {}",
            format!($info, $($arg),*),
            $error))
        }
    },
    ($error:ident, $info:literal) => {
        {
            ::std::io::Error::new(::std::io::ErrorKind::Other, format!(concat!($info, ": {}"), $error))
        }
    }
}
```
Should we clarify and uniform the rust version of martin?

nyurik:
Hi @sharkAndshark, any updates on this one?  Thx!

src/utils.rs
nyurik: oh, so weird, somehow i only saw this comment when navigating from the email notification, not when viewing the PR. Apologies for the other ping, reviewing...

src/utils.rs
```diff
@@ -8,10 +8,24 @@ use tilejson::Bounds;
...
+            ::std::io::Error::new(::std::io::ErrorKind::Other, format!("{}: {}",
+            format!($info, $($arg),*),
+            $error))
+        }
+    };
+    ($error:ident, $info:literal) => {
+        {
+            ::std::io::Error::new(::std::io::ErrorKind::Other, format!(concat!($info, ": {}"), $error))
+        }
+    }
```
nyurik: I think this will be more accurate, and it will consume the literal variant first.  I did some digging into the exports and macros -- turns out there is a hidden "bug" we have in the code -- the macro is actually exported twice -- once as a top level public function, and another as a macro inside the utils module, but only for internal usage.  Not a big deal of course, but it did cause some confusion for me :)

Regardless, I think we can even merge it as is, or if you accept my proposed change, we can merge it with it.

```suggestion
    ($error:ident, $info:literal) => {
        ::std::io::Error::new(::std::io::ErrorKind::Other, format!(concat!($info, ": {}"), $error))
    };
    ($error:ident, $($arg:tt)+) => {
        ::std::io::Error::new(::std::io::ErrorKind::Other,
            format!("{}: {}", format_args!($($arg)+), $error))
    };
```

nyurik (approved):
see comment

sharkAndshark:
> Hi @sharkAndshark, any updates on this one? Thx!

Thx! There are no more update :D

nyurik (approved):
lgtm, thanks for following through on this one!

PR #363, opened 2022-07-05
Fix clippy in benchmarks, and adjust CI

nyurik:
Clippy wasn't happy, and we apparently weren't testing all targets.


stepankuzmin (approved):
Good catch 👍

PR #364, opened 2022-07-05
Clean up licensing per on-boarding

nyurik:
* Use standard `MIT OR Apache-2.0` license
* Add `MapLibre contributors`
* Add Apache and MIT license files
  * The original LICENSE (MIT) file is near-identical but had different line breaks.
  * MIT license now also includes copyright maplibre contributors
* removed the `license-file` attribute because it is usually used for custom licenses


stepankuzmin (approved):
Awesome, thanks!

PR #375, opened 2022-08-05
remove a few minor code dups

nyurik:
the logic of getting a tile is nearly identical in two places, so merging it


stepankuzmin (approved):
👍

PR #377, opened 2022-08-05
feat: use latest actix, switch to bb8 (async everywhere), remove actions

### LINKED ISSUE #349: Migrate to Actix-web 4
Actix has made a number of structural/async changes that might require some structural rethinking.

nyurik:
* remove all actions and other low-level magic code, making it more straightforward for the most common usage
* replace r2d2 with bb8 to make it all async
* use first significant version in cargo.toml - this makes it easier to maintain

This fixes #349


src/server.rs
```diff
@@ -14,19 +17,15 @@ use serde::Deserialize;
...
 use crate::table_source::{TableSource, TableSources};
 use crate::utils::parse_x_rewrite_url;
-use crate::worker_actor::WorkerActor;
 
 pub struct AppState {
-    pub db: Addr<DbActor>,
-    pub coordinator: Addr<CoordinatorActor>,
+    pub pool: Pool,
     pub table_sources: Rc<RefCell<Option<TableSources>>>,
     pub function_sources: Rc<RefCell<Option<FunctionSources>>>,
```
stepankuzmin: We can omit the `Rc<RefCell>>` here, and use `Option<TableSources>`/`Option<FunctionSources>` instead. After that, we can avoid cloning the sources in the handlers, and use the `.as_ref()` instead

stepankuzmin:
I think we can drop the refresh functionality from the app for now. Let's notify users that this feature was removed when the watch mode cli arg or env var is detected as `true`.

nyurik:
@stepankuzmin thx, good point, removed both.  I think eventually we may want to revisit the watch support, so I wonder if it would make sense to remove watch support as a separate PR?  this would make it easier to review this PR and easier to revert the watch removal if needed.

stepankuzmin:
@nyurik yep, that would be more convenient, thanks!

nyurik:
> @nyurik yep, that would be more convenient, thanks!

@stepankuzmin see #381 - which should be merged before this one.

nyurik:
@stepankuzmin ok, done with the changes, let me know if this needs any other changes

stepankuzmin (approved):
Looks good to me 👍

Issue #47, opened 2019-12-10
martin config expose only specific schema


Captain-Oski:
Hello ! Thank you for the superbe work ! 

I am wondering if we could expose only a entire schema instead of exposing one table at a time ?

Thank you

stepankuzmin:
Hi @Captain-Oski! Thanks for the idea, but I'm not sure how do you want it to be. Could you please provide an example for this?

Captain-Oski:
Thank you for replying so quickly : 

it could be something like this alongside the config.yaml: 

```
schema_sources:
  public
  mySchema
  myOtherSchema
```

So that it would only targets all the tables and functions into a specific schema.

I founding it extremely verbose and difficult to maintain an config file when you need multiple tables and functions but need to hide from the public numerous tables. 

I guess one could migrate the needed tables and functions into a new database instance but ...

stepankuzmin:
I see the problem with config verbosity, but schema sources don't fit very well yet. I'll think about how to simplify config to expose schemas, thanks!

nyurik:
The #380 makes this possible, but would require us to think how to add it to configuration

nyurik:
Implemented in #510, pending a few more PR megres

nyurik:
@stepankuzmin and I at the last minute realized that we don't have a clear agreement on how to specify schema configuration list for this feature.  See #512 -- can use your feedback on the best way to configure, as well as possible usecases you can think of

nyurik:
added in #546

PR #380, opened 2022-08-08
Support z,x,y and record-returning funcs, table rework

### LINKED ISSUE #65: Set id to feature
Is there a way to set `id` to feature (last param of `ST_AsMVT` function)?

`id` from table is not passed to feature id.

image

Or int value:

image

### LINKED ISSUE #378: Handle more function source cases
There are some function source cases that I think Martin should handle. They already exist, e.g. generated by OpenMapTiles is `getmvt(zoom,x,y)`. Proposed changes:

* Support cases when the parameter is `zoom` rather than `z`.
* The function may only have 3 parameters, without the extra `query_params` json value
* The function may return a table with a single row and two columns rather than a single bytea value. OpenMapTiles's [generate-sqltomvt](https://github.com/openmaptiles/openmaptiles-tools/blob/master/bin/generate-sqltomvt) would create a table where first column is the tile data (bytea), and the second is a string with hash (md5) value of the tile.
* The function may return GZIPed tile. It should be possible to detect such cases, but this will require an initial test.

The current function source search [implementation](https://github.com/maplibre/martin/blob/main/src/scripts/get_function_sources.sql) looks for any function whose parameters contain (!) `z, x, y, query_params` (3 ints and a json). I suspect this is a bug because there could be other params before `z`, and this is not checked in code. Plus this excludes valid cases without query_params and the table.

---

## Implementation ideas
I think we should change the above query to something like this, and possibly do some additional validation by requesting a tile during startup like I did in [tilelive-pgquery](https://github.com/nyurik/tilelive-pgquery/blob/main/lib/PgQuery.js#L390)

```sql
-- Find SQL functions that match these criteria:
--     * The function must have 3 or 4 input parameters,
--       first 3 must be integers and named z,x,y (in that order),
--       with the optional JSON "query_params" as the 4th parameter.
--       The first param could be "zoom" instead of "z".
--     * The function output must be either a single bytea value or a table,
--       with the table row being either [bytea] or [bytea, text] (in that order).
-- Output fields:
--   schema: the schema the function is in
--   name: the function name
--   inputs: a JSON array of input parameters [{name: type}, ...]
--   outputs: a JSON value - either a string or an array:
--            for a single-value functions, this is the function type as a string, e.g. "bytea"
--            if the function returns a table, this is an array of column types, e.g. ["bytea", "text"]
WITH
inputs AS (
    -- list of input parameters for each function, returned as a jsonb array [{name: type}, ...]
    SELECT
        specific_name,
        jsonb_agg(jsonb_build_object(COALESCE(parameter_name::text, '_'), data_type::text) ORDER BY ordinal_position) as input_params
    FROM information_schema.parameters
    WHERE parameter_mode = 'IN'
    GROUP BY specific_name
),
outputs AS (
    -- list of output parameters for each function, returned as a jsonb array [{name: type}, ...]
    SELECT
        specific_name,
        jsonb_agg(data_type::text ORDER BY ordinal_position) as output_params
    FROM information_schema.parameters
    WHERE parameter_mode = 'OUT'
    GROUP BY specific_name
)
SELECT
    routines.specific_schema AS schema,
    routines.routine_name AS name,
    inputs.input_params AS inputs,
    COALESCE(outputs.output_params, to_jsonb(routines.data_type::text)) AS output_type
FROM
    information_schema.routines JOIN inputs ON routines.specific_name=inputs.specific_name
        LEFT JOIN outputs ON routines.specific_name=outputs.specific_name
WHERE
 input_params IN (
        '[{"z": "integer"}, {"x": "integer"}, {"y": "integer"}]'::jsonb,
        '[{"zoom": "integer"}, {"x": "integer"}, {"y": "integer"}]'::jsonb,
        '[{"z": "integer"}, {"x": "integer"}, {"y": "integer"}, {"query_params": "json"}]'::jsonb,
        '[{"zoom": "integer"}, {"x": "integer"}, {"y": "integer"}, {"query_params": "json"}]'::jsonb)
    AND (
        (routines.data_type = 'bytea' AND output_params IS NULL)
        OR (routines.data_type = 'record' AND output_params IN (
        '["bytea"]'::jsonb,
        '["bytea", "text"]'::jsonb
        )));
```

### LINKED ISSUE #389: I'm having trouble referencing camel cased db columns?
I need help or an example on how to configure the settings for camelCased geometry columns. If I use a camel-column, I get the following error. 

`ERROR martin::server] Can't get composite source tile: db error: ERROR: column "someCamelCasedColumn" does not exist`

The column does exist and I have verified that other tile server programs and database visualizers are able to display the associated polygons. Furthermore, the column is detected when martin scans the database.

```
 # Geometry column name (required)
    geometry_column: someCamelCasedColumn 
```

### LINKED ISSUE #466: id column not included in properties
I have noticed following behavior with martin:

the data

```
geodata=> \d poverty 
                                            Table "zambia.poverty"
         Column          |            Type             | Collation | Nullable |            Default             
-------------------------+-----------------------------+-----------+----------+--------------------------------
 id                      | integer                     |           | not null | nextval('zp_id_seq'::regclass)
 country_name            | character varying(6)        |           |          | 
 province                | character varying(254)      |           |          | 
 district                | character varying(254)      |           |          | 
 constituency            | character varying(254)      |           |          | 
 poverty                 | double precision            |           |          | 
 absolute_number_of_poor | double precision            |           |          | 
 geom                    | geometry(MultiPolygon,3857) |           |          | 
Indexes:
    "zp_pkey" PRIMARY KEY, btree (id)
    "zp_geom_geom_idx" gist (geom)


```

the config

```yaml
table_sources:
  zambia.poverty:
    id: zambia.poverty
    schema: zambia
    table: poverty
    bounds: [21.99937075303238, -18.07947347592123, 33.705705741754244, -8.224359676087436]
    srid: 3857
    geometry_column: geom
    id_column: id
    extent: 4096
    buffer: 64
    geometry_type: MULTIPOLYGON
    clip_geometry: true
    properties:
      id: int4
      poverty: float8
      district: varchar
      province: varchar
      constituency: varchar
      country_name: varchar
      absolute_number_of_poor: float8

```

the index.json
image

the featureID and available props (first 10)

```bash

1 {'absolute_number_of_poor': 22101.0, 'district': 'Mpika', 'poverty': 0.82, 'province': 'Muchinga', 'constituency': 'Mfuwe', 'country_name': 'Zambia'}
2 {'absolute_number_of_poor': 36685.0, 'district': 'Kawambwa', 'poverty': 0.77, 'province': 'Luapula', 'constituency': 'Kawambwa', 'country_name': 'Zambia'}
4 {'absolute_number_of_poor': 53678.0, 'district': 'Solwezi', 'poverty': 0.62, 'province': 'North Western', 'constituency': 'Solwezi West', 'country_name': 'Zambia'}
5 {'absolute_number_of_poor': 175504.0, 'district': 'Kapiri-Mposhi', 'poverty': 0.68, 'province': 'Central', 'constituency': 'Kapiri Mposhi', 'country_name': 'Zambia'}
56 {'absolute_number_of_poor': 76280.0, 'district': 'Katete', 'poverty': 0.79, 'province': 'Eastern', 'constituency': 'Mkaika', 'country_name': 'Zambia'}
3 {'absolute_number_of_poor': 70049.0, 'district': 'Nyimba', 'poverty': 0.78, 'province': 'Eastern', 'constituency': 'Nyimba', 'country_name': 'Zambia'}
6 {'absolute_number_of_poor': 64100.0, 'district': 'Lufwanyama', 'poverty': 0.81, 'province': 'Copperbelt', 'constituency': 'Lufwanyama', 'country_name': 'Zambia'}
79 {'absolute_number_of_poor': 55507.0, 'district': 'Lusaka', 'poverty': 0.19, 'province': 'Lusaka', 'constituency': 'Matero', 'country_name': 'Zambia'}
80 {'absolute_number_of_poor': 73578.0, 'district': 'Lusaka', 'poverty': 0.2, 'province': 'Lusaka', 'constituency': 'Kanyama', 'country_name': 'Zambia'}
7 {'absolute_number_of_poor': 41139.0, 'district': 'Kalabo', 'poverty': 0.89, 'province': 'Western', 'constituency': 'Sikongo', 'country_name': 'Zambia'}
8 {'absolute_number_of_poor': 52477.0, 'district': 'Kafue', 'poverty': 0.43, 'province': 'Lusaka', 'constituency': 'Kafue', 'country_name': 'Zambia'}
```
as you can see  the id is not in the list of props.

is this a feature or a bug?

thanks

nyurik:
Can now handle several additional Postgres functions to get a tile, plus tons of small fixes

### Multiple result variants
* `getmvt(z,x,y) -> [bytea,md5]`  (single row with two columns)
* `getmvt(z,x,y) -> [bytea]` (single row with a single column)
* `getmvt(z,x,y) -> bytea` (value)

### Multiple input parameter variants
* `getmvt(z, x, y)` or `getmvt(zoom, x, y)` (all 3 vars must be integers)
* `getmvt(z, x, y, url_query)`, where instead of `url_query` it could be any other name, but must be of type JSON

### Breaking
* srid is now the same type as PG -- `i32`
* renamed config vals `table_sources` and `function_sources` into `tables` and `functions`

### Features and fixes
* if postgis is v3.1+, uses margin parameter to extend the search box by the size of the buffer. I think we should make 3.1 minimal required.
* fixes feature ID issue from #466
* fixes mixed case names for schemas, tables and columns, functions and parameter names per #389
 

### Notes
* More dynamic SQL generation in code instead of using external SQL files. Those should only be used when they are not parametrized.
* The new function/table discovery mechanism: query for all functions in the database, and match up those functions with the ones configured (if any), plus adds all the rest of the un-declared ones if discovery mode is on.
* During table and function discovery, the code generates a map of `(PgSqlInfo, FunctionInfo)` (or table) tupples containing SQL needed to get the tile.
* Auto-discovery mode is currently hidden - the discovery is on only when no tables or functions are configured.  TBD - how to configure it in the future
* The new system allows for an easy way to auto-discover for the specific schemas only, solving #47
* predictable order of table/function instantiation
* bounding boxes computed in parallel for all tables (when not configured)
* proper identifier escaping
* test cleanup

fixes #378
fixes #466
fixes #65
fixes #389


src/pg/function_source.rs
```diff
@@ -1,137 +1,114 @@
...
+            assert_eq!(input_types.len(), input_names.len());
+            match (&output_record_names, &output_record_types) {
+                (Some(n), Some(t)) if n.len() == 1 && n.len() == t.len() => {
+                    assert_eq!(t, &["bytea"]);
+                }
+                (Some(n), Some(t)) if n.len() == 2 && n.len() == t.len() => {
+                    assert_eq!(t, &["bytea", "text"]);
+                }
+                (None, None) => {}
+                _ => panic!("Invalid output record names or types"),
```
stepankuzmin, line 41: Maybe skip such functions and use `warn!` instead of `panic!` here?

stepankuzmin (approved):
Overall this looks good 👍

src/pg/function_source.rs
nyurik: This is a validation that SQL matches Rust code -- and uses lots of asserts for that. In theory the whole thing can be rewritten to use warnings, but I would rather panic early and fix the logic bugs than to make it go unnoticed.  TBD i guess

PR #396, opened 2022-09-13
minor docker fixes

nyurik:
* ports should be in quotes
* added .dockerignore (copy of .gitignore)

PR #398, opened 2022-09-13
Rework dockerfile to use rust-alpine

### LINKED ISSUE #397: Use another Docker build image
It seems [emk/rust-musl-builder](https://github.com/emk/rust-musl-builder) the is [no longer maintained](https://github.com/emk/rust-musl-builder/issues/147), so we may want to migrate the docker image to some other base.

@stepankuzmin can this project be built using the default [rust docker image](https://hub.docker.com/_/rust)? The -alpine variants use musl too.  It would need to install the openssl though.

nyurik:
This includes all changes from #396 - merge that first

* use official rust base image to build docker image
* use latest dependencies

Fixes #397

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,7 @@
+.DS_Store
+/target/
+**/*.rs.bk
+.idea/
+
+pg_data/
+config.yml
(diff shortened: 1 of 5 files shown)


stepankuzmin (approved):
Neat!

PR #407, opened 2022-09-16
Migrate to MapLibre documentation

### LINKED ISSUE #405: Migrate to MapLibre examples (README)
We are now in the MapLibre world, time to migrate examples and documentation to use latest MapLibre code.  Help wanted :)

nyurik:
Merge MB into MapLibre getting started guide.

Fixes #405


stepankuzmin (changes requested):
Thanks for the proposal! I feel strongly that we should not replace Mapbox with MapLibre, but keep MapLibre examples along with Mapbox. I understand this reasoning, but I prefer keeping both examples in the README, upgrading the debug page to the latest Mapbox version, and adding another debug page with MapLibre.

nyurik:
Hmm, OK, i guess we could keep a section on Mapbox too. Though, I am not sure about upgrading MB to the latest non-FOSS version - note that all other examples are open source - so it actually could compromise MapLibre project from the legal perspective because none of us has any right to use non-foss mapbox software (or to test it, etc) without having a service agreement with MB.  We could of course copy public documentation (assuming the documentation is public(?)), but again, I am not a lawyer, and I would prefer not to risk it.

stepankuzmin:
Yes, that makes sense. Let's keep the Mapbox updated to the latest FOSS version only, then. Thanks again!

maxammann:
> Hmm, OK, i guess we could keep a section on Mapbox too. Though, I am not sure about upgrading MB to the latest non-FOSS version - note that all other examples are open source - so it actually could compromise MapLibre project from the legal perspective because none of us has any right to use non-foss mapbox software (or to test it, etc) without having a service agreement with MB. We could of course copy public documentation (assuming the documentation is public(?)), but again, I am not a lawyer, and I would prefer not to risk it.

If we use the old mapbox version then there is probably not really an advantage, as old-mapbox API == MapLibre API as of today.

I think we should be fine with just adding code samples for MapBox, as long as we do not provide an executable sample. In fact, I have a MapBox account with a free tier and also signed their ToS. I don't think its forbidden to post API examples for any proprietary project (given that the API docs are public).

maxammann (approved):

README.md
```diff
@@ -220,6 +213,27 @@ const deckgl = new DeckGL({
 });
 ```
 
+## Using with Mapbox
+
+[Mapbox GL JS](https://github.com/mapbox/mapbox-gl-js) is a proprietary JS and native libraries for interactive, customizable vector maps on the web. The MapBox GL JS v1.x was open source, and it was forked as MapLibre (see [above](#using-with-maplibre)), so using Martin with MapBox is similar to MapLibre.
```
stepankuzmin:
```suggestion
[Mapbox GL JS](https://github.com/mapbox/mapbox-gl-js) is a JavaScript library for interactive, customizable vector maps on the web. Mapbox GL JS v1.x was open source, and it was forked as MapLibre (see [above](#using-with-maplibre)), so using Martin with Mapbox is similar to MapLibre. Mapbox GL JS can accept [MVT vector tiles](https://github.com/mapbox/vector-tile-spec) generated by Martin, and applies [a style](https://docs.mapbox.com/mapbox-gl-js/style-spec/) to them to draw a map using Web GL.
```

stepankuzmin (approved):

PR #408, opened 2022-09-17
Attempt to migrate CI to maplibre/martin

nyurik:
Note that I manually published [maplibre/martin:latest](https://hub.docker.com/r/maplibre/martin) to docker, so it might be able to pass CI ok.  I have not looked deeply into the existing CI workflow - @stepankuzmin any suggestions on changes to that?  The github actions are now setup

P.S. I am not certain what that whole `brew/tap` thing is - don't know enough about Macs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -252,7 +252,7 @@ jobs:
         id: docker_meta
         uses: docker/metadata-action@v4
         with:
-          images: urbica/martin
+          images: maplibre/martin
 
       # https://github.com/docker/setup-qemu-action
       - name: Set up QEMU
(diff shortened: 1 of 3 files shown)


birkskyum:
There is no homebrew package named martin.

`brew install martin` returns 

`Warning: No available formula with the name "martin". Did you mean maturin?`

birkskyum:
For the brew/tap thing to work, we need an updated version of this:

https://github.com/urbica/homebrew-tap/blob/master/HomebrewFormula/martin.rb

birkskyum (approved):

nyurik:
> For the brew/tap thing to work, we need an updated version of this:
> 
> https://github.com/urbica/homebrew-tap/blob/master/HomebrewFormula/martin.rb

@birkskyum i'm not familiar with the Tap -- is that something that we need to transfer to maplibre, or perhaps just move it as a sub-dir somewhere in this repo? That repo seems a bit too small to warrant a separate repository.

stepankuzmin:
@birkskyum @nyurik I can update homebrew tap after release. I'd need a new URL and sha256 for the release.

PR #416, opened 2022-09-26
Clean up actix routes, HEAD support

nyurik:
Use attribute-based routing, and support HEAD requests (same as GET)

attribute routes are better because they keep route declaration (i.e. parameters) together with the code that uses it.


stepankuzmin (approved):
👍

PR #419, opened 2022-09-27
Add missing `libc6-compat` in Dockerfile

stepankuzmin:
Docker image is segfaulting because Alpine runtime is missing the `libc6-compat`, compatibility libraries for glibc.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,6 @@ RUN cargo build --release
 
 FROM alpine:latest
 
-RUN apk update
 RUN apk add --no-cache libc6-compat
 
 COPY --from=builder \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -10,6 +10,9 @@ RUN cargo build --release
 
 FROM alpine:latest
 
+RUN apk update
```
nyurik: you don't need `apk update` with `--no-cache` --- this way you won't need to do `rm -rf /var/cache/apk/*` afterwards.

P.S. See https://stackoverflow.com/questions/49118579/alpine-dockerfile-advantages-of-no-cache-vs-rm-var-cache-apk

```suggestion
```

nyurik (approved):
:+1:

PR #428, opened 2022-10-04
Compatible with case sensitive PostgreSQL identifiers

### LINKED ISSUE #389: I'm having trouble referencing camel cased db columns?

sharkAndshark:


sharkAndshark:
resolves #389

nyurik (approved):
lgtm!

PR #439, opened 2022-10-09
Mostly noop code refactorings

nyurik:
* remove most of the utils:: and dev:: namespace usage
* rename `configure_db_source` to `configure_db_sources`
* use defaults from other table info objects (using the `..obj` syntax)
* use doc comments vs regular comment in a few places
* refactor server_test.rs to make it more readable and much shorter
  * use a macro for identical code
  * use a test_get() for identical GET request
* rename fn test_foo() into fn foo() for simplicity

### DIFF
--- a/benches/sources.rs
+++ b/benches/sources.rs
@@ -1,11 +1,10 @@
-use std::collections::HashMap;
-
 use criterion::{criterion_group, criterion_main, Criterion};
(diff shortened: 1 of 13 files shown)


stepankuzmin (approved):
LGTM 👍

PR #440, opened 2022-10-09
sources implementation

nyurik:
Partial implementation of the #430:

* New endpoint structure:
  * `GET /` -- get a list of available sources, as a json blob, with `id -> { name, description, attribution }` fields (if available)
  * `GET /<id>` -- get tilejson for the given source, or a combination of sources. No `.json` extension
  * `GET /<id>/<z>/<x>/<y>` -- get a tile. No format extension.
* Introduce a new tile format support crate (using code from the maplibre/mbtileserve project)
* All sources are now together, removing the `/rpc/...` routes
* Print auto-generated configuration to the stdout or save it to a file

TODO:
* benchmarks need to be rewritten - they were relying on some internal structures that are no longer there


nyurik:
This is now included in #456, closing

PR #448, opened 2022-10-14
Draft : Start docker-compose stack in CI

gbip:


gbip:
Ok, I'll stop here from now. Here are the following work that is left :

- [ ] There seems to be some kind of issue with the way the database is setup : the postgis extensions does not *seems* to be installed
- [ ] Build the docker image instead of pulling it (easy but will make the testing cycle longer, so I'll keep it as is until the PR is ready)
- [ ] `docker-compose exec` always return 0 which might be an issue

gbip:
cc @stepankuzmin

PR #449, opened 2022-10-16
Migrate tests to use maplibre-gl

### LINKED ISSUE #406: Migrate tests to MapLibre
Currently tests are using ancient MB code with some access token info, and need to be updated. 
* https://github.com/maplibre/martin/blob/main/tests/composite_source.html
* https://github.com/maplibre/martin/blob/main/tests/debug-deckgl.html
* https://github.com/maplibre/martin/blob/main/tests/debug.html

pjsier:
Migrates the HTML test files to use `maplibre-gl` instead of `mapbox-gl`. I kept the same version as in `tests/debug-maplibre.html` and used the Carto Positron style on `tests/debug.html` because it was the closet to Mapbox's light style.

Closes #406. Happy to make any changes to this, thanks!

### DIFF
--- a/tests/composite_source.html
+++ b/tests/composite_source.html
@@ -4,8 +4,8 @@
     <meta charset="utf-8" />
     <title>Martin Debug Page</title>
     <meta name="viewport" content="initial-scale=1,maximum-scale=1,user-scalable=no" />
-    <script src="https://api.tiles.mapbox.com/mapbox-gl-js/v0.52.0/mapbox-gl.js"></script>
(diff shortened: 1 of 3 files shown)


nyurik (approved):
Awesome, thanks!  It seems the debug-deckgl is actually broken because it uses a non-existent `http://localhost:3000/public.points.json`  (instead of points1 or points2), but this should be done as a separate PR -- because it seems deckgl itself breaks if it gets valid data... perhaps we should either remove deckgl example, or fix it

stepankuzmin:
Could we please also add debug pages with Mapbox GL JS?

PR #450, opened 2022-10-18
Remove automated docker builds for PRs

### LINKED ISSUE #443: Stop autopublish pr-* docker images on PR CI builds
It seems some new PRs automatically publish to the docker hub, e.g. #438 showed up in [martin tags](https://hub.docker.com/r/maplibre/martin/tags).  I think we shouldn't publish work in progress as docker images.

pjsier:
Closes #443. Adds a conditional to the `docker` job that won't run any of the Docker build steps if triggered by a PR. Based on the docker/metadata-action docs, this should stop any of the `pr-*` images from being built

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -277,7 +277,7 @@ jobs:
       - name: Build and push the Docker image
         uses: docker/build-push-action@v3
         with:
-          push: ${{ github.actor != 'dependabot[bot]' && !github.event.pull_request.head.repo.fork }}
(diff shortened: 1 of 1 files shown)


nyurik (commented):
but we should still build the docker images on PR - we just shouldn't publish them

pjsier:
Got it, I initially misunderstood. Thanks for clarifying! I updated the PR and commit message so that it runs everything except the push for PRs

nyurik (approved):
thanks!

PR #463, opened 2022-10-25
Consolidate DB init and tests

nyurik:
* Add justfile to simplify running all the tests
* Save all PBF outputs to the text files
* Consolidate all tests to reuse the same code
* Consolidate database initialization
* updated readme with the new instructions

Note that while this PR creates "expected" files, the CI cannot validate the generated results because the output is not stable. Eventually we may try to output just the non-geometry values to have reasonable tests comparing against the expected results.


lseelenbinder (approved):
LGTM!

PR #470, opened 2022-11-01
Use OpenSSL v3 in CI for macOS

stepankuzmin:


nyurik (commented):
Would it make sense to `brew uninstal openssl` on the test vm? This way martin should (?) crash if not statically compiled

stepankuzmin:
Yes, that should crash the run. I want to check if martin will use the OpenSSL v3 (it should) if it's available.

stepankuzmin:
I tried to run it on Apple Silicon, but the binary is built for Intel. Apple Silicon VM is not supported in GH Actions yet https://github.com/actions/runner-images/issues/2187

Probably we might be okay with the Intel version for now. I'll check the Intel version with the OpenSSL in https://github.com/urbica/homebrew-tap

PR #472, opened 2022-11-01
Docker improvements and CI tests

### LINKED ISSUE #436: Test Docker image in CI
The goal of this issue is to catch issues such as #423 before they reach production users.

The biggest issue with this is that the tests are not build as part of the docker image.

nyurik:
* Change docker image to use `entrypoint` -- so that Martin can be used as a command:

```bash
docker run maplibre/martin <parameters>
```

* The docker image is now tested the same way as in the CI tests
* Added a few changes to the justfile

Fixes #436

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,6 +2,10 @@
 .git/
 # Having Dockerfile in .dockerignore lets us develop docker images without full rebuild on each change
 Dockerfile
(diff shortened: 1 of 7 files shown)


lseelenbinder (approved):
LGTM. solid improvements!

PR #474, opened 2022-11-02
Switch to Rustls

### LINKED ISSUE #471: Use rustls crate instead of openssl
The [rustls](https://crates.io/crates/rustls) crate is mature rust implementation of TLS. I think it would make a lot of sense to switch to it instead of the openssl that keeps causing some build and startup issues on different platforms.

Note that the [tokio-postgres-rustls](https://github.com/jbg/tokio-postgres-rustls) is relatively recent, but is now archived. So it will likely be moved to the `rust-postgres` repo - see [related issue](https://github.com/sfackler/rust-postgres/issues/421).

nyurik:
Fixes #471

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -256,6 +256,42 @@ jobs:
           name: build-${{ matrix.target }}
           path: target_releases/*
 
+  cross-build:
+    name: Cross-build for other platforms
+    runs-on: ubuntu-latest
+    env:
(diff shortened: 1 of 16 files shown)


nyurik:
@stepankuzmin I have no idea how to test the SSL certs - but the code is done otherwise, so if you can manually test it and it works, feel free to +1

nyurik:
current code seem to be passing tests, but i suspect it won't allow `verify-ca` mode, or when a root cert is given (we still need tests for that)

sharkAndshark (approved):

PR #485, opened 2022-11-17
Try async_fn_in_trait nightly feature

nyurik:
Try to switch to the nightly async trait function.


nyurik:
Seems like this is a no-go path -- see https://stackoverflow.com/questions/77318891/migrating-async-trait-with-box-to-the-new-async-fn-in-trait-feature -- boxing traits with async fn is not yet supported, and we rely on that to transparently clone sources.  So... keeping async_trait crate dependency for now

PR #488, opened 2022-11-19
Make SSL support optional

nyurik:
By default, Martin is now compiled without openssl, simplifying debugging and simple case usage, whereas the docker build and CI publishing would still use openssl


stepankuzmin (approved):
👍

PR #495, opened 2022-11-27
Clean up error wrapping

nyurik:
* Fix incorrect printing of the xyz value - implements Display trait instead to keep it simpler
* Remove the possibly-incorrect macro capture of a simple string - in case the string contains inline vars
* Handle optional extra comma properly
* Rename it to a shorter name
* Inline vars where possible


lseelenbinder (approved):
LGTM

PR #497, opened 2022-11-30
Require PostGIS 3 and use ST_TileEnvelope

nyurik:
* All tests and internal code now uses ST_TileEnvelope function
* Remove `tile_bbox`
* Rename test function sources for clarity - this will be needed in a subsequent PR to add other function tests


stepankuzmin:
I'm concerned that this is a breaking change. However, I'm in favor of using `ST_TileEnvelope`, but we should communicate that the next version will drop support for PostGIS < 3.x if we don't want to support both 2.x and 3.x.

nyurik:
@stepankuzmin you are right, this is a breaking change, but now is the perfect time because of all the other breaking changes we have merged to `main`. Besides, Postgis 2.5 is over 3 years old and 5 postgis releases behind.  So yes, I think now is a perfect time to migrate.

How do you propose we communicate this?

stepankuzmin (approved):
Approved 👍 

We should at least mention it in the release notes and maybe also mention it on Slack if we also announce releases there.

PR #511, opened 2022-12-11
Implement multi-sourced postgres

nyurik:
* Support `postgres` config key to be either a list or an object
* Merge `SrvConfigBuilder` into `SrvConfig`
* More config tests
* Support env vars in the config file, including default fallbacks using [subst](https://crates.io/crates/subst) crate.
* Params must be either passed via config file, or by a combination of env vars and CLI parameters

### DIFF
--- a/.cargo-husky/hooks/pre-push
+++ b/.cargo-husky/hooks/pre-push
@@ -6,7 +6,7 @@
 set -e
 
 if ! command -v just > /dev/null; then
-    echo "`just` is not installed. See https://github.com/casey/just#readme for installation instructions, or just run"
+    echo "`just` is not installed. See https://github.com/casey/just#readme for installation instructions, or use"
     echo ""
     echo "  cargo install just"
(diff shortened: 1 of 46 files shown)


stepankuzmin (approved):
Overall this looks good to me 👍 

Talking about mixing config, CLI args, and Environment variables, I'd rather not do that. Previously, Martin relayed on the priorities — config should take precedence of CLI and CLI of ENVs. So if both config and CLI are defined, Martin would consider only config and skip CLI args. This approach should allow us to keep things simple and reduce some mental load off users (we should also reflect that behavior in docs and maybe warn users if mixing inputs).

I'd also consider failing fast if one of the PostgreSQL connections is unavailable on start.

nyurik:
oops, accidentally left auto-commit here, so the moment tests passed, it merged.  I think it works, just might have a few rough edges - will clean them in a subsequent PRs.

PR #514, opened 2022-12-12
Catch errors in tests, minor fixes

nyurik:
This extracts some of the code from #511 but without breaking changes

* Use `PathBuf` instead of `String` where dealing with files
* Parse keep_alive as u64
* More config tests to crash if martin output contains warnings or errors

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,6 +99,14 @@ jobs:
           password: test
           database: test
           rights: --superuser
+      - name: Log DATABASE_URL string
+        shell: bash
+        run: |
+          echo "DATABASE_URL=$DATABASE_URL"
+          echo "And in base64 to bypass Github's obfuscation:"
+          echo "$DATABASE_URL" | base64
+        env:
+          DATABASE_URL: ${{ steps.pg.outputs.connection-uri }}
       - name: Init database
         if: matrix.target != 'aarch64-apple-darwin'
         shell: bash
@@ -126,7 +134,7 @@ jobs:
           name: test-output
           path: tests/output/*
           retention-days: 5
-      - name: Test
+      - name: Integration Tests
         if: matrix.target != 'aarch64-apple-darwin'
         shell: bash
         run: |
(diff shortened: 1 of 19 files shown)


.github/workflows/ci.yml
```diff
@@ -99,6 +99,14 @@ jobs:
           password: test
           database: test
           rights: --superuser
+      - name: Log DATABASE_URL string
+        shell: bash
+        run: |
+          echo "DATABASE_URL=$DATABASE_URL"
+          echo "And in base64 to bypass Github's obfuscation:"
+          echo "$DATABASE_URL" | base64
```
stepankuzmin, line 107: What is it for?

README.md
```diff
@@ -464,9 +464,6 @@ You can find an example of a configuration file [here](https://github.com/maplib
 # Database connection string
 connection_string: 'postgres://postgres@localhost:5432/db'
 
-# Trust invalid certificates. This introduces significant vulnerabilities, and should only be used as a last resort.
-danger_accept_invalid_certs: false
```
stepankuzmin: Why remove `danger_accept_invalid_certs`?

.github/workflows/ci.yml
nyurik: Turns out the database_url produced by the test postgres setup is not `postgres://...` but `postgresql://...` -- so my multi-PG code was failing.  I don't really need it anymore.

README.md
nyurik: i guess we can keep it in the readme - as long as it is not in the test (which fails now because this value is unrecognized when compiling without SSL... which might actually be a bad thing - we should probably build WITH ssl when testing in CI)

.github/workflows/ci.yml
nyurik: are you ok to keep it?  this is not leaking any secrets -- if you look at the output of that CI step, it outputs all the passwords in clear text too -- its just that github masks anything that looks like a password inside a standard connection string

README.md
nyurik: actually i just realized that CI does test with SSL, and i reverted the readme, so all's good

stepankuzmin (approved):

Issue #519, opened 2022-12-13
Handle null values returned by function source


gbip:
Hi there !

If I'm correct, martin used to return HTTP 204 response when a function source returned NULL.

However it seems that the latest martin version (0.6.1) doesn't handle null values anymore :

```
thread 'actix-rt|system:0|arbiter:0' panicked at 'error retrieving column iti_tile: error deserializing column 0: a Postgres value was `NULL`', /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/tokio-postgres-0.7.7/src/row.rs:151:25
```

Is it a bug in martin or should we fix our function source ?

nyurik:
I think this is a bug -- i would think a null result should be handled as 204.  I'll take a look, or if you have a quick and simple solution, PRs are welcome - even a PR with just a unit test so I can fix the code and get a ready to use test for it.

nyurik:
fixed in #520, once merged, i'll release it

nyurik:
fixed and released as [0.6.2](https://github.com/maplibre/martin/releases/tag/v0.6.2)

gbip:
Thanks a lot !

PR #520, opened 2022-12-13
Bug: handle functions returning null

nyurik:
If a SQL function returns NULL, treat it as HTTP 204 - no-content result.

Also lock update and switch to v0.6.2

Fixes https://github.com/maplibre/martin/issues/519
Fix v0.6 branch only. Main will be fixed differently.


stepankuzmin (approved):
👍

PR #541, opened 2022-12-25
Include full test into the coverage

nyurik:
This is a test to see if multiple runs in the grcov github workflow would increase the coverage.  So far seems like only the last one is used


CommanderStorm:
closing in favor of https://github.com/maplibre/martin/pull/1753

Issue #512, opened 2022-12-12
Decide how to configure postgres table/function auto-discovery


nyurik:
The upcoming #511 allows multiple postgres connections. Also, the recently merged #510 adds support for per-schema filtering when automatically configuring table and function sources.  Yet, we don't have a way to configure them yet.

----

Proposed schema:

```yaml
# Configure some tables/functions, plus use an extensive discovery system configuration
postgres:
    connection: 'postgres://postgres@localhost:5432/db'
    functions:
        src1:
            schema: public
            function: my_function
            maxzoom: 10
    tables:
        src2:
            schema: public
            table: my_table
    discover:
        # Only search the given PostgreSQL schemas, for both functions and tables.
        # Set them up with maxzoom 10.
        # TBD: Naming: discover, auto, configure, auto-config, ...?
        functions: true
        tables: true
        schemas: ["public", "my-schema"]
        maxzoom: 10
        # How should the new table and function sources be named
        table_naming: {schema}.{table}.{column}
        function:naming: {function}
```

----

This is how currently users can automatically configure all available tables and functions in all schemas:
```yaml
postgres:
    connection: 'postgres://postgres@localhost:5432/db'
    # no additional configuration for tables or funcs
```

---

With this config, users enable just one table as a src1 source, and disable auto-discovery
```yaml
postgres:
    connection: 'postgres://postgres@localhost:5432/db'
    tables:
        src1:
            schema: public
            table: my_table
            srid: 4326
            geometry_column: geom
    functions:
        src2:
            schema: public
            function: my_function
```

---

### Need feedback
Beyond the above, we may want to configure the following scenarios:
* Neither tables nor functions are configured (first example above), but the users want to auto-discover just the functions, but not tables. Or just the tables.
* Same, but autodiscovery should be done only in a given list of schemas (per #47 request)
* Allow pre-configure some tables and additionally enable autodiscovery -- not certain if this is needed or not?

### Other configuration ideas

See the primary idea at the top. Most of these other ideas apply to both tables and functions, but for brevity only one is mentioned.

```yaml
# Configure some functions, plus expose all other functions too
postgres:
    connection: 'postgres://postgres@localhost:5432/db'
    functions:
        src1:
            schema: public
            function: my_function
            maxzoom: 10
    discover_functions: true
```

```yaml
# Configure some functions, plus expose all other functions that are listed in the given schemas
postgres:
    connection: 'postgres://postgres@localhost:5432/db'
    functions:
        src1:
            schema: public
[...]

PR #546, opened 2022-12-29
Implement postgres auto-publish

nyurik:
* NEW: support for #512 - pg table/function auto-discovery
  * can filter schemas
  * can use patterns like `{schema}.{table}.{column}` and `{schema}.{function}`
* NEW: add `disable_bounds` bool flag to allow disabling of the bounds computation
* reworked integration tests to use yaml


src/pg/config.rs
```diff
@@ -27,19 +27,30 @@ pub struct PgConfig {
...
     #[serde(skip_serializing_if = "Option::is_none")]
     #[serde(serialize_with = "sorted_opt_map")]
     pub functions: Option<FuncInfoSources>,
-    #[serde(skip)]
-    pub run_autodiscovery: bool,
+}
+
+#[derive(Clone, Debug, Default, PartialEq, Serialize, Deserialize)]
+pub struct PgCfgPublish {
+    pub from_schema: Option<OneOrMany<String>>,
```
stepankuzmin: Should it be `from_schemas` instead of `from_schema` as in https://github.com/maplibre/martin/issues/512?

src/pg/configurator.rs
```diff
@@ -33,15 +53,16 @@ pub struct PgBuilder {
...
     pub async fn new(config: &PgConfig, id_resolver: IdResolver) -> Result<Self> {
         let pool = Pool::new(config).await?;
-        let auto = config.run_autodiscovery;
+
         Ok(Self {
             pool,
             default_srid: config.default_srid,
-            auto_functions: config.auto_functions.clone().unwrap_or(Schemas::Bool(auto)),
-            auto_tables: config.auto_tables.clone().unwrap_or(Schemas::Bool(auto)),
+            calc_bounds: config.calc_bounds.unwrap_or(true),
```
stepankuzmin: We can name it `disable_bounds` since this flag is intended to disable the default behavior (always expose bounds). We can also invert the default and use `enable_bounds`.

src/pg/configurator.rs
```diff
@@ -169,6 +221,41 @@ impl PgBuilder {
     }
 }
 
+fn new_auto_publish(config: &PgConfig, is_function: bool) -> Option<PgBuilderPublish> {
```
stepankuzmin, line 224: I'm okay to go with the notion that auto publish is true by default, as mentioned here https://github.com/maplibre/martin/issues/512#issuecomment-1368259212
But it still looks confusing and hard to reason about for me. We'd need to make this behavior clear in the docs.

stepankuzmin (approved):

Issue #504, opened 2022-12-04
Implement dynamic tilejson based on Forwarded header


nyurik:
Per [tilejson spec](https://github.com/mapbox/tilejson-spec/tree/master/3.0.0#32-tiles), the `tiles` url must be an absolute URL. If Martin is running behind a proxy (most of the time nowadays), it may be serving multiple domains, and it may also not know the outside domain.

We need to support [Forwarded](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded) header to allow dynamic tilejson that matches the original protocol and host accessed by the user.

nyurik:
Turns out Actix already handles the `Forwarded` and `X-forwarded-host/prot` headers, so this is moot.

PR #549, opened 2023-01-08
Add .mbtiles support

### LINKED ISSUE #494: Implement .mbtiles support
Need to support .mbitiles files as another available backend

nyurik:
Merge after #548 and #517

Adds a new [.mbtiles](https://github.com/mapbox/mbtiles-spec/blob/master/1.3/spec.md) backend, without the grid support.  Uses extensive tile content detection, i.e. if the content is gzipped, png, jpeg, gif, webp.

From CLI, can be as easy as adding a path to a directory that contains a .mbtiles file (works just like pmtiles support)

```bash
# All *.mbtiles files in this dir will be published.
# The filename will be used as the source ID
martin ./tests/fixtures
```

From configuration file, the path can be specified in a number of ways (same as pmtiles)

```yaml
mbtiles:
  paths:
    # scan this whole dir, matching all *.mbtiles files
    - /dir-path
    # specific mbtiles file will be published as mbtiles2 source
    - /path/to/mbtiles2.mbtiles
  sources:
      # named source matching source name to a single file
      pm-src1: /tmp/mbtiles.mbtiles
      # named source, where the filename is explicitly set. This way we will be able to add more options later
      pm-src2:
        path: /tmp/mbtiles.mbtiles
```

Fixes #494


maxammann:
I'm looking right now for a mbtiles server so this comes just in time 😅 

I'll test this out now, for a code review the PR is maybe too large. Maybe you can leave comments where you wish reviews.

nyurik:
Thanks, let me know if you run into any issues.  The reason this pr is so big is because it includes two other prs. If you review them in order, they will be much smaller.  @stepankuzmin said he will try to get to them today

src/file_config.rs
```diff
@@ -0,0 +1,270 @@
...
+    use crate::file_config::{FileConfigEnum, FileConfigSource, FileConfigSrc};
+
+    #[test]
+    fn parse() {
+        let cfg = serde_yaml::from_str::<FileConfigEnum>(indoc! {"
+            paths:
+              - /dir-path
+              - /path/to/pmtiles2.pmtiles
+            sources:
+                pm-src1: /tmp/pmtiles.pmtiles
```
stepankuzmin: Probably we should also add the `*.mbtiles` test here?

stepankuzmin (approved):

maxammann:
Not sure where it comes from but I have an emptry string in the `tiles` array: 

image

mbtiles/src/lib.rs
```diff
@@ -0,0 +1,313 @@
...
+                warn!("Unable to parse metadata {title} value in {name}: {err}");
+                None
+            }
+        }
+    }
+
+    pub async fn get_metadata(&self) -> MbtResult<Metadata> {
+        let mut res = Metadata {
+            tilejson: tilejson! {
+                tiles: vec![String::new()],
```
maxammann: I think this adds an empty string in the tiles array.

nyurik: Thanks, I'll need to fix this, and also handle the x-forwarded-for header (there is a ticket for that). I'll have time tonight to look at it, in a few hrs

src/file_config.rs
nyurik: there is no point because this test is not specific to pmtiles - it decodes a sub-element under `pmtiles:` or `mbtiles:` keys.  I will change the filenames to make it clearer.

mbtiles/src/lib.rs
maxammann: That header works for me already. It places the correct host and protocol

nyurik: which header? We currently handle `x-rewrite-url`, which is a really bad one from the security perspective. See #504

ewagstaff:
@nyurik is there a good source to read more about what this change allows? Would this mean MapLibre can display data from a static .mbtiles file without the need to spin up a map server?

nyurik:
@ewagstaff this change simply allows mbtiles file to be Martin's source of data, just like Postgres and `.pmtiles` files.   Martin itself is a map server, but if your data is in a file format, Martin would not need any additional servers. If you want completely serverless, take a look at pmtiles.

Issue #503, opened 2022-12-04
Consider removing `X_REWRITE_URL` due to potential security issues


nyurik:
Apparently many systems used to support `x-rewrite-url` header, and they were dropped due to security concerns. Should we drop it too, or have a different workaround? I wonder if this is CORS related?

Relevant links:
* https://www.acunetix.com/vulnerabilities/web/url-rewrite-vulnerability/
* https://symfony.com/blog/cve-2018-14773-remove-support-for-legacy-and-risky-http-headers

CC for wider discussion @stepankuzmin @lseelenbinder @HarelM @birkskyum

lseelenbinder:
It could affect CORS, but that's usually the `Origin` header + `Access-Control-*`headers. I haven't ever seen it used in practice in my experience, so I'm not quite sure the intended purpose (other than perhaps providing a base URL)?

Generally speaking, I think martin would be running behind some cache or reverse proxy, which allows users to remove/set the header before it hits martin code. Perhaps the best option is it should live behind a command-line flag?

nyurik:
Note that Martin generates tilejson, which must have an absolute URL to the tiles per [spec](https://github.com/mapbox/tilejson-spec/tree/master/3.0.0#32-tiles).  Which means if Martin lives behind a proxy, it must still know the external endpoint, i.e. by looking at some proxy-provided headers...

lseelenbinder:
> by looking at some proxy-provided headers...

For sure. I've typically done this via `x-forwarded-for`, etc. headers (and then you only allow them set by internal systems).

stepankuzmin:
The `X-Rewrite-URL` header is used to rewrite the tiles and TileJSON endpoints in the reverse proxy. We can drop support for this header, but we must reflect on how to do it with the `X-Forwarded-For` header and communicate it back to the community.

nyurik:
@stepankuzmin turns out actix already handles `forwarded` and `x-forwarded-prot/host` headers - see https://github.com/actix/actix-web/blob/cfd40b4f1547629b2023430da55798fd942a0c42/actix-web/src/info.rs#L71 

So safe to remove the whole hack outright.

nyurik:
I kept thinking about this, and it seems we may need to keep it. For some silly reason, I though `X-Rewrite-URL` covers the same use case as `Forwarded`, which is totally not the case - instead, it is a URL rewrite, not host/port rewrite.

I think we may want to make this an opt-in feature rather than always on:

```shell
# in CLI
[...]

PR #552, opened 2023-01-12
Make x-rewrite-url an opt-in feature

nyurik:
This PR is based on #551

Per #503 make `X-Rewrite-URL` an opt-in feature due to a potential for security issues (unconfirmed).  This will allow only the users who needed it to opt-in into it, either via cli or a config flag.

### DIFF
--- a/src/args/srv.rs
+++ b/src/args/srv.rs
@@ -10,6 +10,10 @@ pub struct SrvArgs {
     /// Number of web server workers
     #[arg(short = 'W', long)]
     pub workers: Option<usize>,
+    /// Process `x-rewrite-url` header to rewrite tile URLs in TileJSON. This is useful when running behind a
+    /// reverse proxy that rewrites URLs, e.g. `/tiles/my_source/...` instead of `/my_source/...`.
+    #[arg(long)]
+    pub allow_url_rewrite: bool,
 }
 
 impl SrvArgs {
@@ -23,6 +27,9 @@ impl SrvArgs {
         }
         if self.workers.is_some() {
             srv_config.worker_processes = self.workers;
+        }
+        if self.allow_url_rewrite {
+            srv_config.allow_url_rewrite = Some(self.allow_url_rewrite);
         }
     }
 }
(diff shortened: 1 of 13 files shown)


maxammann:
I'm basically 100% sure there is no security risk. At least there is no indication that there is one.

nyurik:
@maxammann do you think we should continue to support this out of the box despite all frameworks removing this functionality, or should we let the users opt-in for it, thus hugely reducing the uncertainty?  If i was researching Martin, and saw this function built in, I would do a quick search, and see that it has caused some security issues - and would be less likely to jump on it. But if i have the control of enabling it, each shop can evaluate if this is concerning to them or not.

maxammann:
I see a few options:
1. Remove the feature completely and add add an alternative to statically define a URL prefix in the `config.yml`
2. Partially remove the feature and only support the `Forwarded` header (Using custom paths will no longer be possible)
3. Keep the feature as is and use the HTTP header name `X-Rewrite-Url`
4. Keep the feature as is and use some other HTTP header name for communicating between the proxy and martin

All options are equally secure in our thread model. I think martin just arbitrarily went with the `X-Rewrite-Url` header. There is no relation to the vulnerabilities found online.
I found [here](https://owasp.org/www-project-web-security-testing-guide/latest/4-Web_Application_Security_Testing/05-Authorization_Testing/02-Testing_for_Bypassing_Authorization_Schema#testing-for-special-request-header-handling) a hint on how that header is used sometimes.

I fail to see a connection between how we use the header and the reported vulnerabilities.

nyurik:
@maxammann I am not proposing to remove the Forwarded feature at all (host & protocol forwarding) - that one is totally fine and should stay as is.  I am only concerned with the `X-Rewrite-Url` having some yet-not-understood security concern, and so let the users enable it only when they specifically need it.  I don't think we should enable it for everyone though.

maxammann:
I see it this way: Either it is a security issue or it isn't according to our thread model. If it isn't then we should not introduce some arbitrary config flag for it because we are unsure about security implications. If it is a security issue -> drop it.

Though I can see potential for an other similar config flag: `behind_proxy`.

If we are not behind a proxy (the default case) then we can disable some parts of request handling. E.g. we can also ignore the Forwarded header.

nyurik:
won't merge, thx for feedback

PR #560, opened 2023-01-27
update lock, clippy lint

nyurik:


nyurik:
self-approved to avoid CI issues

Issue #496, opened 2022-11-29
Cant retrieve connection from the pool: Timed out in bb8


IceMan1910:
Hi all. Just started on this mapping tools quest :)
I'm not being able to solve what seems to be a simple issue with martin.

I have my postgres bd running (OSM is the bd name), it is already fed with data coming from osm2pgsql with flex output. No issues there. It has a points, lines, polygons table with geometries.
Now i want to serve my vector tiles with martin :)

I'm on windows 10, my postegres is running on a local server on ubuntu.
**Running on terminal:**
`PS C:\Users\[user]\Desktop\martin> .\martin.exe -c .\config.yaml`

**On yaml file I have:**
```
connection_string: "postgres://postgres:[pass]@[local_ip_for_ubuntu]:5432/OSM"
default_srid: 4326
listen_addresses: "0.0.0.0:4000"
```

I've tried without the config file, using connection string on the terminal. Same output:
```
[2022-11-29T16:13:46Z INFO  martin] Starting Martin v0.6.1
[2022-11-29T16:13:46Z INFO  martin] Using .\config.yaml
[2022-11-29T16:13:46Z INFO  martin::pg::db] Connecting to database
[2022-11-29T16:14:16Z ERROR martin] Can't retrieve connection from the pool: Timed out in bb8
```
I can connect just fine with pgadmin to the bd, and I ran the osm2pgsql process on this windows machine with output going to this same bd.
I can´t seem to make this work. What am I missing?
Thanks in advance.

nyurik:
@IceMan1910 are you able to run `psql "postgres://postgres:[pass]@[local_ip_for_ubuntu]:5432/OSM"` ?  (assuming you have psql installed on your windows box)

IceMan1910:
@nyurik I don't have it instaled locally. Does it make sense to install it, just to test?
With psql on the pgAdmin4 on this machine, I can connect fine.

I also have another ubuntu server on my local network. Should I install martin on ubuntu instead of windows 10?

nyurik:
I think it would be helpful - it seems to be happening on your machine, plus having psql when working with PG is highly useful for any kind of simple querying

nyurik:
P.S. To be honest, I would never use Windows for any kind of server-related activities :)

IceMan1910:
@nyurik OK so... I'm really stuck here.
Steps done:
- On my other linux box, same config file as I had on windows.

`./martin --config config.yaml`

Same old, same old :\

```
[...]

PR #567, opened 2023-02-04
Improve SSL support

nyurik:
This is a partial fix for #496

* BREAKING: Now Martin behaves the same way as `psql` -- by default, if SSL is available on the server, it will be used, even though it will not verify that the server has a valid SSL certificate
* Martin now understands `PGSSLCERT`, `PGSSLKEY`, and `PGSSLROOTCERT` env vars (and corresponding config keys) - same as psql.
* Martin can now process `?sslmode=verify-ca` and `verify-full` (just like psql). The verify modes require root and/or client cert & key.
* remove `danger_accept_invalid_certs` -- turns out that behavior is expected by default unless ssl mode is set to verify - which upstream lib [does not support](https://github.com/sfackler/rust-postgres/issues/768) - PR [submitted](https://github.com/sfackler/rust-postgres/pull/988).
* added connection_timeout_ms option for postgres and set it to 5 seconds by default. This way it will fail out earlier.
* added error reporting to bb8 - but it is currently [broken upstream](https://github.com/djc/bb8/issues/151) - not sure we can fix it easily, so may need to switch to deadpool later.
* added docker-based TLS test (horray!) - wasn't trivial at all, despite ending up fairly simple.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -189,25 +189,43 @@ jobs:
         include:
           # These must match the versions of postgres used in the docker-compose.yml
           - image: postgis/postgis:11-3.0-alpine
+            args: postgres
+            sslmode: disable
           - image: postgis/postgis:14-3.3-alpine
+            args: postgres
+            sslmode: disable
+          # alpine images don't support SSL, so for this we use the debian images
+          - image: postgis/postgis:15-3.3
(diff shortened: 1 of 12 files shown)


stepankuzmin (approved):
Looks good to me 👍 

One nit: could you please not remove the "Using with Heroku PostgreSQL" section from the README but drop the `danger-accept` flag notion?

PR #569, opened 2023-02-06
Issue 540 - missing index warnings

### LINKED ISSUE #540: Warn if a table geometry column has no geometry index
During table detection, check if the table's geometry column has an associated GIST index, and if not, give a warning.

Volunteers: if you only know how to change SQL, that would be a great help too!

*  modify [table searching SQL](https://github.com/maplibre/martin/blob/3a713a026910619c29ccd612496edfdd895330d4/src/pg/scripts/get_table_sources.sql) to add another bool column to it. 
* handle the warning in the query result parser

cthiange:
This adds two fields to `TableInfo`:
* `geom_idx: Option<bool>` to tell if a geo column has a spatial index
* `is_view: Option<bool>` to distinguish views from other relations

Missing spatial index warnings are logged for non-view relations. Views will never have indexed columns and, if referencing a table with a missing index, it will be logged already.

Couldn't figure out how to make `just test` accept the new warning (from missing index), so I have them logged as INFO for now :)

fixes #540


nyurik:
hi @cthiange, thanks for the PR!  I like the query change, that's exactly what I had in mind!  I do not think we should expose it to the server users - the fact that a tile source comes from a table/view vs some other source is an implementation / server configuration detail that the users shouldn't even know about.  Eventually we may want to create an admin UI that would report all these things.

That said, the actual warning should be printed via `warn!(...)` - i will take a look at how to best automate the testing of that.

The `is_view` is the same way, but we **should** include the words `view` vs `table` when printing log messages to the console (in the `info!(...)`).  I guess there is really no good way to detect if the query plan of a view would eventually use an index or not, so i agree that we might as well skip it for the views.

cthiange:
Thanks for the feedback @nyurik. Are you suggesting we keep the new flags out of `TableInfo` and handle the table source query results through an intermediate type?

src/pg/config_table.rs
```diff
@@ -23,6 +23,14 @@ pub struct TableInfo {
     /// Geometry column name
     pub geometry_column: String,
 
+    /// Geometry column has a spatial index
+    #[serde(skip_serializing)]
```
nyurik:
```suggestion
    #[serde(skip_deserializing, skip_serializing)]
```

src/pg/config_table.rs
```diff
@@ -23,6 +23,14 @@ pub struct TableInfo {
...
+    pub geometry_index: Option<bool>,
+
+    /// Flag indicating if table is actually a view (PostgreSQL relkind = 'v')
+    #[serde(skip_serializing)]
```
nyurik:
```suggestion
    #[serde(skip_deserializing, skip_serializing)]
```

src/pg/scripts/get_table_sources.sql
```diff
@@ -8,16 +8,44 @@ WITH columns AS (
...
+FROM annotated_geometry_columns AS gc
 LEFT JOIN columns ON
-  geometry_columns.f_table_schema = columns.table_schema AND
-  geometry_columns.f_table_name = columns.table_name AND
-  geometry_columns.f_geometry_column != columns.column_name
-GROUP BY f_table_schema, f_table_name, f_geometry_column, srid, type;
+  gc.schema = columns.table_schema AND
+  gc.name = columns.table_name AND
+  gc.geom != columns.column_name
+GROUP BY gc.schema, gc.name, gc.geom, gc.srid, gc.type, gc.is_view, gc.geom_idx;
```
nyurik:
```suggestion
GROUP BY gc.schema, gc.name, gc.geom, gc.srid, gc.type, gc.is_view, gc.geom_idx;

```

nyurik (commented):
Dope, thanks, it was my mistake - for some reason I thought you were showing the result in the catalog to the end user.  I should have read it more carefully.

I added a few minor tweaks, and tested it locally -- seems great!

**updated** see another comment in a few min

src/pg/table_source.rs
```diff
@@ -40,6 +42,15 @@ pub async fn get_table_sources(pool: &Pool) -> Result<SqlTableInfoMapMapMap> {
             ..TableInfo::default()
         };
 
+        // Warn for missing geometry indices. Ignore views since those can't have indices
+        // and will generally refer to table columns.
+        if let (Some(false), Some(false)) = (info.geometry_index, info.is_view) {
+            info!(
```
nyurik:
```suggestion
            warn!(
```

nyurik:
Oops, we might have committed code changes at the same time to your branch -- I fixed the testing script - so now we can ensure that the output logs have expected warnings.

cthiange:
~~No worries, nice addition. `vtzero-check` is tripping over the warning isn't it?~~

Never mind, I see your changes now.

nyurik (approved):
Thanks!!!  We may also want to use the `is_view` field for proper logging. For example, instead of

```
Configured source points2 from table public.points2 with geom column (POINT, SRID=4326)
```

we should print this (assuming it is a view)

```
Configured source points2 from view public.points2 with geom column (POINT, SRID=4326)
```

Also, we may want to add a view to a test that simply wraps an existing table.  Just add it to an existing test sql table fixture file, and bless the results.  This could be a separate PR or the same one - up to you.

nyurik:
@cthiange thanks, merged!  Feel free to submit another PR per my previous comment.

PR #576, opened 2023-02-17
Feat: max_feature_count query limit

### LINKED ISSUE #384: Add feature count limit
- With PostgreSQL, martin can serve tiles in a reasonable latency, when tile-size is relatively small.
- Current martin has no limitation in feature count in one tile.
- When you request a tile including large amount of features, response become somewhat heavy. 

Idealy, it is the best to simplify features and drop tiny ones on-the-fly but this seems not to be practical in terms of performance.
Therefore, I would like to propose to add feature count limit option in config. I mean developer set maximum number of features in one tile, like following

```
max_feature_count: 10000
```

Elasticsearch seems to take this approach in `size` parameter (off course it is not totally same and comparable...).
https://www.elastic.co/guide/en/elasticsearch/reference/master/search-vector-tile-api.html

nyurik:
Implements #384 - ability to limit the number of features included in a tile from a Postgres table/view.

This allows zoomed-out view of a table with a reasonable speed because each tile could be limited in size, rather than include millions of features.

If set on a CLI, overrides whatever is set in the config file (if given).

Any naming suggestions?

### DIFF
--- a/README.md
+++ b/README.md
@@ -469,6 +469,8 @@ Options:
           If a spatial table has SRID 0, then this default SRID will be used as a fallback
   -p, --pool-size <POOL_SIZE>
           Maximum connections pool size [DEFAULT: 20]
+  -m, --max-feature-count <MAX_FEATURE_COUNT>
+          Limit the number of features in a tile from a PG table source
   -h, --help
           Print help
   -V, --version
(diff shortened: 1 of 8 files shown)


stepankuzmin (approved):
👍

PR #581, opened 2023-02-20
Fix dead link

louwers:
I was going through the README when I stumbled upon a dead link. I _think_ this is the new URL.


nyurik (approved):
thx!

Issue #343, opened 2022-06-03
support different coordinate reference systems like `EPSG:4326`


ChristianBeilschmidt:
Is it possible to use Maritn to output vector tiles in EPSG:4326 coordinate reference system?

sharkAndshark:
Try [Function Source](https://github.com/maplibre/martin#function-sources) mabe.
Define a database  function  which take x y z as argument and output tile byte .This function should clip  tiles based on    4326， TMS/XYZ schema

sharkAndshark:
Is there a plan for tile schema not 3857? @stepankuzmin

stepankuzmin:
Not at the moment, but we might consider this in the future. Is there a real need for this?

ChristianBeilschmidt:
For us, at least 4326 would be relevant.
For local data, in theory, UTM zones could also be relevant.

Even MapTiler started selling 4326 tiles: https://documentation.maptiler.com/hc/en-us/articles/4405462253073-OpenStreetMap-in-WGS84-EPSG-4326-

So this should have also relevance for others.

sharkAndshark:
> For us, at least 4326 would be relevant. For local data, in theory, UTM zones could also be relevant.
> 
> Even MapTiler started selling 4326 tiles: https://documentation.maptiler.com/hc/en-us/articles/4405462253073-OpenStreetMap-in-WGS84-EPSG-4326-
> 
> So this should have also relevance for others.

We use a lot 4490,4544 by a lightweight tile server based on PostGIS.

jlaura:
This is particularly relevant for us. We work on non-terrestrial bodies (Moon, Mars, etc.). In fact, coupling to the EPSG authority is problematic. The authority for planetary projection codes is the International Astronomical Union (IAU). Being able to define a custom tile matrix set allows us to use appropriate body radii, etc. If and when this work is undertaken, I would strong advise considering non-EPSG authorities.

nyurik:
I will be happy to accept any PRs for this capability. If anyone is interested, please comment on your proposed solution, i.e. how the code should be restructured to support such cases.  I would advise looking at the upcoming new code structure that I'm hacking on in #456 and partially described in the #430 discussion. The goal is to support any data backends, including multiple postgress connections, mbtiles, etc.

exotfboy:
ASFAIK, the tile bounds generated for a given tile coordinate(z,x,y) is the key, then it is the same for encoding the data clipped from the source to MVT.

[...]

PR #611, opened 2023-03-23
Configurable tile systems for projections other than web mercator

burleight:
This is first attempt at addressing #343

Adds a `tile_systems` section to the yml config:

```yaml
tile_systems:
  nztm:
    srid: 2193
    bounds: [1000000, 4700000, 2600000, 6300000]
  wgs84:
    srid: 4326
    bounds: [-180, -90, 180, 90]
```

Then, the tile URL becomes `/{source}:{tile_system}/{z}/{x}/{y}`

Issues:

- This in effect leads to each table source being duplicated in the catalog. E.g. `table1`, `table1:nztm`, `table1:wgs84` etc
- The URL for a multiple source request becomes `/{source}:{tile_system},{source2}:{tile_system}/{z}/{x}/{y}`. 
- Nothing is done to prevent multiple source requests using different tile systems
- Needs test coverage
- Configuration documentation needs updating

For the OpenLayer users out there. the trick to using this is to giving `VectorTile` source a custom tile grid:

```javascript
new ol.source.VectorTile({
      format: new ol.format.MVT(),
      projection: 'EPSG:2193',
      tileGrid: new ol.tilegrid.TileGrid({
          extent: [1000000, 4700000, 2600000, 6300000], // square of width 1600000
          resolutions: (new Array(10)).fill(0).map((e, i) => 1600000 / (256 * Math.pow(2, i)))
      }),
      url: `http://0.0.0.0:3000/${layer}/{z}/{x}/{y}`
  })
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -550,6 +550,18 @@ postgres:
       id_format: 'table.{schema}.{table}.{column}'
       # Add more schemas to the ones listed above
       from_schemas: my_other_schema
+      # Optionally, publish tiles in more than one projection/tiling scheme
+      tile_systems:
+        # the default web mercator tiles, enabled by default 
+        - type: WebMercatorQuad
+        # an additional custom tiling scheme
+        - type: Custom
+          srid: 4326
+          # bounds of tile 0/0/0
+          bounds: [ -180.0, -90.0, 180.0, 90.0 ]
+          # name for the tiling scheme 
+          # auto published table will be given the name {source_id}:{tile_system.identifier}
+          identifier: WGS84Quad
     functions:
       id_format: '{schema}.{function}'
   
@@ -601,6 +613,15 @@ postgres:
       # List of columns, that should be encoded as tile properties (required)
       properties:
         gid: int4
+        
+      # Output tiles with a custom projection and tiling
+      tile_system:
+        # output srid
+        srid: 4326
+        # bounds of tile 0/0/0
+        bounds: [ -180.0, -90.0, 180.0, 90.0 ]
+        # name for the tiling scheme
+        identifier: WGS84Quad
   
   # Associative arrays of function sources
   functions:
--- a/src/config.rs
+++ b/src/config.rs
@@ -16,7 +16,6 @@ use crate::pg::PgConfig;
 use crate::pmtiles::PmtSource;
 use crate::source::{IdResolver, Sources};
 use crate::srv::SrvConfig;
-use crate::tilesystems::TileSystemsConfig;
 use crate::utils::{OneOrMany, Result};
 use crate::Error::{ConfigLoadError, ConfigParseError, NoSources};
 
@@ -33,9 +32,6 @@ pub struct Config {
 
     #[serde(skip_serializing_if = "Option::is_none")]
     pub mbtiles: Option<FileConfigEnum>,
-
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub tile_systems: Option<TileSystemsConfig>,
 
     #[serde(flatten)]
     pub unrecognized: HashMap<String, Value>,
@@ -84,7 +80,7 @@ impl Config {
         let mut sources: Vec<Pin<Box<dyn Future<Output = Result<Sources>>>>> = Vec::new();
         if let Some(v) = self.postgres.as_mut() {
             for s in v.iter_mut() {
-                sources.push(Box::pin(s.resolve(idr.clone(), &self.tile_systems)));
+                sources.push(Box::pin(s.resolve(idr.clone())));
             }
         }
         if let Some(v) = self.pmtiles.as_mut() {
--- a/src/config.rs
+++ b/src/config.rs
@@ -16,6 +16,7 @@ use crate::pg::PgConfig;
 use crate::pmtiles::PmtSource;
 use crate::source::{IdResolver, Sources};
 use crate::srv::SrvConfig;
+use crate::tilesystems::TileSystemsConfig;
 use crate::utils::{OneOrMany, Result};
 use crate::Error::{ConfigLoadError, ConfigParseError, NoSources};
 
@@ -32,6 +33,9 @@ pub struct Config {
 
     #[serde(skip_serializing_if = "Option::is_none")]
     pub mbtiles: Option<FileConfigEnum>,
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub tile_systems: Option<TileSystemsConfig>,
 
     #[serde(flatten)]
     pub unrecognized: HashMap<String, Value>,
@@ -80,7 +84,7 @@ impl Config {
         let mut sources: Vec<Pin<Box<dyn Future<Output = Result<Sources>>>>> = Vec::new();
         if let Some(v) = self.postgres.as_mut() {
             for s in v.iter_mut() {
-                sources.push(Box::pin(s.resolve(idr.clone())));
+                sources.push(Box::pin(s.resolve(idr.clone(), &self.tile_systems)));
             }
         }
         if let Some(v) = self.pmtiles.as_mut() {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -14,6 +14,7 @@ pub mod pg;
 pub mod pmtiles;
 mod source;
 pub mod srv;
+mod tilesystems;
 mod utils;
 
 #[cfg(test)]
--- a/src/pg/config.rs
+++ b/src/pg/config.rs
@@ -8,7 +8,7 @@ use crate::pg::config_table::TableInfoSources;
 use crate::pg::configurator::PgBuilder;
 use crate::pg::utils::Result;
 use crate::source::{IdResolver, Sources};
-use crate::tilesystems::TileSystemsConfig;
+use crate::tilesystems::TileSystem;
 use crate::utils::{sorted_opt_map, BoolOrObject, OneOrMany};
 
 pub trait PgInfo {
@@ -69,6 +69,7 @@ pub struct PgCfgPublish {
 pub struct PgCfgPublishType {
     pub from_schemas: Option<OneOrMany<String>>,
     pub id_format: Option<String>,
+    pub tile_systems: Option<Vec<TileSystem>>,
 }
 
 impl PgConfig {
@@ -92,17 +93,10 @@ impl PgConfig {
         Ok(res)
     }
 
-    pub async fn resolve(
-        &mut self,
-        id_resolver: IdResolver,
-        tile_systems: &Option<TileSystemsConfig>,
-    ) -> crate::Result<Sources> {
+    pub async fn resolve(&mut self, id_resolver: IdResolver) -> crate::Result<Sources> {
         let pg = PgBuilder::new(self, id_resolver).await?;
-        let ((mut tables, tbl_info), (funcs, func_info)) = try_join(
-            pg.instantiate_tables(tile_systems),
-            pg.instantiate_functions(),
-        )
-        .await?;
+        let ((mut tables, tbl_info), (funcs, func_info)) =
+            try_join(pg.instantiate_tables(), pg.instantiate_functions()).await?;
 
         self.tables = Some(tbl_info);
         self.functions = Some(func_info);
@@ -124,6 +118,7 @@ mod tests {
     use crate::pg::config_function::FunctionInfo;
     use crate::pg::config_table::TableInfo;
     use crate::test_utils::some;
+    use crate::tilesystems::TileSystemConfig;
     use crate::utils::OneOrMany::{Many, One};
 
     #[test]
@@ -196,6 +191,11 @@ mod tests {
                   geometry_type: GEOMETRY
                   properties:
                     gid: int4
+                  tile_system:
+                    identifier: WGS84Quad
+                    srid: 4326
+                    bounds: [-180, -90, 180, 90]
+
 
               functions:
                 function_zxy_query:
@@ -229,6 +229,11 @@ mod tests {
                                 "gid".to_string(),
                                 "int4".to_string(),
                             )])),
+                            tile_system: Some(TileSystemConfig {
+                                srid: 4326,
+                                bounds: Bounds::new(-180.0, -90.0, 180.0, 90.0),
+                                identifier: "WGS84Quad".to_string(),
+                            }),
                             ..Default::default()
                         },
                     )])),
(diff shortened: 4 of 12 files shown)


nyurik:
Thanks @burleight for working on this!  How common would you think it would be for the users to want all projections for all table sources?  What do you think as the most common use cases?  We need to address two basic cases: table sources when auto-configuring PG connection, and a case when the user wants to pre-configure a specific source with a given projection. 

I think for auto-config, we should add something to the `postgres.auto_publish.tables`, e.g. the config you gave as an example.  The `id_format` may also need to be expanded to support it, e.g. `table.{proj}.{schema}.{table}.{column}`, but it wouldn't be set to that by default (to avoid empty `proj` var.

The individual per-table configurations may be better off to support only a single projection using `postgres.tables.table_source_id` where these params might be placed

burleight:
>  How common would you think it would be for the users to want all projections for all table sources?

This is our use case, we have a lot of data we display in EPSG:2193, and we'd like to experiment with using martin to serve it. 

There are global coordinate systems, such as EPSG:4326 and EPSG:3857, as well as local coordinate systems, such as EPSG:2193, which is the current New Zealand mainland coordinate system. Global to global, global to local and local to global transformations almost always make sense, but local to local systems don't always apply. I.e. it wouldn't make sense to project from the New Zealand coordinate system into the Swiss coordinate system.

So, I may help if we had a way to filter our transformations that don't make sense. We could configure which transformations to a tile system are allowed:

```yaml
tile_systems:
  swiss:
    srid: 2056
    bounds: ...
  nztm:
    srid: 2193
    bounds: [1000000, 4700000, 2600000, 6300000]
    restrict_to_srids: [4326, 3857, 27200] # 27200 is an older New Zealand local projection
```

If `restrict_to_srids` is specified, then the tile_system is only auto published for tables/geometry columns with those srids. Implicitly, always allow identity transformations `2193 -> 2193`

> I think for auto-config, we should add something to the `postgres.auto_publish.tables`

How about keeping `tile_systems` global, then reference tile system ids in `postgres.auto_publish.tables` and `postgres.tables`

So, by default, if someone specifies the tile systems global, all of them are used by auto_publish but, this could be overriden in `postgres.auto_publish.tables.tile_systems` like this:

```yaml
postgres:
  auto_publish:
    tables:
      # only auto publish the default Web Mercator system
      tile_systems: ['default'] 
      # or allow the default and some custom systems
      tile_systems: ['default', 'nztm'] 
      # only auto publish a custom system
      tile_systems: ['nztm']
  # individual table sources may override which tile systems apply
  tables:
    table_source_id:
      schema: public
      table: my_swiss_data
      tile_systems: ['default', 'swiss']
```

nyurik:
Thanks, but do you think that for **each** table source you would want to publish all possible projections?  Or should the projection simply be a setting on the source itself? At the end of the day, there will still be an internal `source_id => handler` mapping, and that `source_id` could be anything, including `foo:bar` syntax (only simple letters/digits are allowed at the moment though, but that can change).  So in theory `src:swiss` and `src:nztm` is no different than `src_swiss` and `src_nztm` - you just have to specify the same config twice. I just feel weird about having a global switch that enables reprojection on every source... plus we could treat this as a "style" rather than projection --- basically it could be a bunch of settings including projection that re-configure a source.

I propose we break this problem into parts:
* support configuration for just **one pre-configured source**:  currently we already have the `srid` and `bounds` params on a table source, so need to figure out if any extra settings are needed, implement the needed query modification, merge, ship.
* decide if/how we could optimize **multiple pre-configured** sources, e.g. introduce a way to share a bunch of settings across multiple sources.
* decide if/how we should support auto-discovery
* decide if we want to have martin-level (as oppose to just postgres-tables level) re-projection support

burleight:
Our use case is, we have a bunch of datasets, most of them NZTM. We'd be setting up a vector tile server that publishes these datasets. For the apps that we write, we'd be picking the best projection for the job. So, if we're making something which only applies to the NZ mainland, we'd tile the data with [NZTM2000Quad](https://github.com/linz/NZTM2000TileMatrixSet), in other cases we'd use web mercator.

> should the projection simply be a setting on the source itself?

Should martin lift these variants into source setting. Instead of `source_id => handler`, would it be better to have a `source_id => variant_id => handler` mapping? Tile variants could be accessed with query params - `{source_id}/{z}/{x}/{y}?tileSystem={tileSystemId}`. Also, with query params, people could write postgresql functions that support multiple tile schemes and a setting in `postgres.functions` markup, which tile systems a function supports.

> support configuration for just one pre-configured source: currently we already have the srid and bounds params on a table source, so need to figure out if any extra settings are needed

`table.srid` defines the input projection and `table.bounds` restricts the output extent, they're both orthogonal to the tiling method chosen, so the table source configuration would need an extra setting for specifying how the source should be tiled.

> decide if we want to have martin-level (as oppose to just postgres-tables level) re-projection support

I still think the definition of a tile system should sit outside of the postgres context. I'd also like to have a route, similar to `catalog`, which exposes the tile system metadata

nyurik:
* So technically you can already solve the whole thing without a single Martin change -- you can simply create a pg function that accepts table name and projection params, and just return the resulting tile. Not ideal, and the autogenerated tilejson might be messed up, so you may have to have some static hosting of it, but could work.
* `tile system` - does it include re-projection and/or other things?  Per your code, you do transform into a non-3857, and use projection bounds for the tile envelope. Guess its simple enough.  Possibly anything else? What type of a catalog-like output do you see?
* the issue with moving tile system outside of postgres is that we would need to support it in pre-tiled cases. Martin supports pg functions, pmtiles and mbtiles - all of which return web mercator `x,y,z -> tile_blob` tiles, and those tiles could be vector or raster tiles.
* Query params like `?projection=foo` or `tilesystem=foo` could work, but it goes so much against the whole CDN/caching concept, plus it limits the whole thing to pg tables only, and wouldn't  work well with anything else (mixing multiple sources becomes a mess). Plus it codifies params structure for tiles - not ideal either.
* The issue still not answered - can we start small - first the minimal one where a user can specify geometry ST_transform srid and the output envelope bounds as additional params to the pre-configured TABLE settings? No CLI changes, only yaml for now.  We could call it `tile_srid` and `tile_bounds` or some other name.

nyurik:
> > should the projection simply be a setting on the source itself?
> 
> Should martin lift these variants into source setting. Instead of `source_id => handler`, would it be better to have a `source_id => variant_id => handler` mapping? Tile variants could be accessed with query params - `{source_id}/{z}/{x}/{y}?tileSystem={tileSystemId}`. Also, with query params, people could write postgresql functions that support multiple tile schemes and a setting in `postgres.functions` markup, which tile systems a function supports.

No, i think we should not expose this as a parameter at all, at least not in the first iteration. Instead, given a table of "my_geometries", you can configure two sources (in yaml):  `my_3857_geometries` and `my_2193_geometries`, and the client decides dynamically which one to use (you can put srid at the end for simplicity of coding - i wanted to highlight the fact that srid part of the source id is not handled in any special way by martin itself)

so once it is possible to configure two separate data sources, each with its own geometry transofrmation function and its own tiling schema, we could proceed to other optimizations - like making it possible to auto-generate such configurations, etc

raboczi:
Is there anything that should be done to anticipate the needs of the proposed Adaptive Composite Map Projection support https://github.com/maplibre/maplibre/issues/190?  Perhaps particular SRIDs are only required at certain zoom levels, for instance.

nyurik:
> Is there anything that should be done to anticipate the needs of the proposed Adaptive Composite Map Projection support [maplibre/maplibre#190](https://github.com/maplibre/maplibre/issues/190)? Perhaps particular SRIDs are only required at certain zoom levels, for instance.

That's a good question... I am not an expert on projection and globe re-projection, so can't tell for certain - but my understanding was that it is possible to construct a globe view from web mercator ... some-magically-how...  One concern though is that web mercator hides the poles, so it would have to fake them by simply showing them as snow

nyurik:
hi @burleight, do you need help with this PR? Any blockers? Thx!

burleight:
> hi @burleight, do you need help with this PR? Any blockers? Thx!

I believe I got to the point where the code was working but the configuration was a bit verbose/ugly and there wasn't much test coverage of the new changes.

Thank you.

gdnwr:
When can it be used?

raboczi:
> When can it be used?

I'm interested in this as well, so I can use it to generate tiles for non-Earth ellipsoids.

CommanderStorm:
> When can it be used?

As mentioned above, this PR is blocked on testing. If you want to contribute in this area, this would be appreciated 🙏.
You can be the one who pushes this over the finish line..

As for timelines: sorry, we don't have any. This is a volunteer project and things get done when they get done.

sharkAndshark:
Let's back to this work. Actually my team need this feature either. We use epsg:4326 a lot.

A proposal here:

## The configure yaml
Only add a tile_systems field to the yaml. Do nothing to else.
```yml
tile_systems:
  nztm:
    srid: 2193
    bounds: [1000000, 4700000, 2600000, 6300000]
  wgs84:
    srid: 4326
    bounds: [-180, -90, 180, 90]
```

## The API

### get tile

* Keep the source id stuffs still
* User could use other tile systems dynamically with restful request
* Take web mercator as default

|api|description|
|----|----|
|`/{sourceID}/{z}/{x}/{y}` |default to get tile with web mercator|
|`/{sourceID}/{tile_system}/{z}/{x}/{y}`|get tile with specific tile system|
| `/{source1},…,{sourceN}/{z}/{x}/{y}`                 | Composite tile source with the default web mercator  |
| `/{source1},…,{sourceN}/{tile system}/{z}/{x}/{y}`     | Composite tile source with the specific tile system |

###  get resources with /catalog
```json
{
  "fonts": {},
  "spirtes": {},
  "tiles": {},
  "tile_systems": {
    "custom_system1": {
        "extent": [
            1620750.2508,
            4271892.7153,
            1625870.2508,
            4277012.7153
        ],
        "origin": [
          1620750.2508,
          4277012.7153
        ],
        "tileSize": [
          4096,
          4096
        ]
    },
},
}
```

sharkAndshark:
Let's discuss and get a good design. :D

CommanderStorm:
I think `tile_systems` as a name is a bit akward and intimidating (for newbies or people not into geoscience ^^)
Lets rename it to `coordinate_reference_systems` or `spatial_reference_systems`.

The current implementation only deals with pg tables.
Your solution (adding an API) sounds like you have a PG-independent plan => also work for pg functions, pmtiles and mbtiles as well.
=> How would you implement this as a basic idea? (read: what would need to change?)
   Where would the source of truth for `srid`& reprojection live?..

About documenting/catalog-api:
Should this be maybe one option which shows up (in the catalog) only after configuring it and be documented on a different wiki page?
This way we could properly explain why one would choose a different CRS.
I think of this as a more advanced/less self-explainatory option..

Otherwise I would be fine with this, as long as it is well tested.
I don't want to deal with non-mercator things tbh, but as long as it is somewhat out of the way, I am fine with it.

burleight:
`coordinate_reference_systems` or `spatial_reference_systems` don't describe how to tile the plane.

We could adopt the OGC nomenclature and call them tile matrix sets.

> Referencing by Coordinates, a coordinate system is a set of mathematical rules for specifying how coordinates are to be assigned to points in space. A CRS is a coordinate system that is related to the real world by a reference datum. An example of mathematical rules is the application of a sphere or an ellipsoid centered in the datum and the use of a projection to transform the sphere or the ellipsoid into a planar representation of the world. Usually, the resulting planar coordinates are expressed as real numbers that are distances to the origin of the projection. This section introduces a tile scheme called **Tile Matrix Set that is defined on top of a CRS**. A fundamental part of the definition of a Tile Matrix Set is the Tile Matrix.

(emphasis added)

Source https://docs.ogc.org/is/17-083r4/17-083r4.html#toc14

sharkAndshark:
* `coordinate_reference_systems` or `spatial_reference_systems`  is more about the srid part not the whole tile grid part.
* Tile Matrix Set sounds professional
* And the word GeoServer using is [gridset](https://docs.geoserver.org/latest/en/user/geowebcache/webadmin/gridsets.html#gwc-webadmin-gridsets)

nyurik:
My questions/thoughts:

### Configuration vs Per-Request
Do we force site admins to explicitly allow which sources should support which projection, or do we let users do that?
* Allowing users to request "anything" creates much bigger scope for abuse / DDOS
* It **might** not make sense to allow all projections for a specific source
* It will make caching less effective, because now the same resource can be requested with two different URLs - e.g. `/src/z/x/y` and `/src/3857/z/x/y`
I suspect we should really let the site admin pre-configure additional sources due to above, but not certain yet.

### URL pattern
* `/src/proj/z/x/y` or `/src@proj/z/x/y` -- The `@` on one hand is like saying "this source in this projection". Symbol IDs currently cannot contain this symbol, so that's ok.
* no changes?  If we only allow admins to define projections, and the the number of different projections per source is not too big, perhaps we should just keep the existing `/src/z/x/y`, and let the admins define the needed ones explicitly in the config?  E.g. they could create `mysource` and `mysource_4326` - keeping the system far simpler, and solving many unknowns except the configuration one.
* What does OpenLayers use here?

### Config
This one is really tricky, and I don't know a good answer. I think we should adapt OGC work as much as possible if applicable, especially since OpenLayers supports a lot of it - and could use this API out of the box(?).  Defining projections is tricky, so we should:
* allow "named" or "numbered" projections where it makes sense, e.g. `epsg:4326` should "just work".  We may also (eventually) allow to do local custom definitions - `martin:foo_bar` is a custom martin projection defined in the config?
* multiple sources may reuse the same projection spec, same tree-like inheritance is needed here

### Tile storage format
* What are the specifications for MVT - would it even work?
* Raster images should be simpler
* Coordinate with the MLT efforts?

### Re-projections
* Should Martin be able to reproject from one projection to another? Which should it support? Should it use [proj](https://crates.io/crates/proj)? Something else?
* raster vs vector?

### PG support
* How different would the PG SQL queries should be to support this? Would `ST_AsMVT` be able to handle it? (More specifically - `ST_AsMVTGeom`

Great to see discussion on this!

nyurik:
I moved my comment to #1750 - lets discuss projections there -- too many things to discuss, we should use threads instead of piling all of our thoughts in this PR.

burleight:
As far as OGC standards go, there is [OGC API - Tiles](https://ogcapi.ogc.org/tiles/overview.html)

pygeoapi implements the OGC API standards, here is a tiles endpoint:

https://demo.pygeoapi.io/stable/collections/lakes/tiles
https://demo.pygeoapi.io/stable/collections/lakes/tiles/WebMercatorQuad/metadata

sharkAndshark:
At least I think martin should not do the Re-projections. Keep martin lightweight without a proj crate. Let PostGIS do the reprojections.

CommanderStorm:
PR seems fairly like a hard merge, lets close this PR and move the discussion to the discussion or to another PR

PR #653, opened 2023-05-02
Fix merging multi-source tilejson

### LINKED ISSUE #652: Composite Catalog Endpoint Only Showing First Source
Using version 0.8.0

We're using composite sources and the catalog endpoint only seems to be listed the first vector layer vs. all of them. The tiles endpoint with the same composite sources does render all of the correct layers. Below you can see I'm requesting a composite of `tiles.layer_one` and `tiles.layer_two`, it's confirmed Martin sees both layers since they're listed in the `tiles` url in the response however the `vector_layers` array only lists the first composite layer `tiles.layer_one` vs. both.

```
curl localhost:3000/tiles.layer_one,tiles.layer_two

{
    "tilejson": "3.0.0",
    "tiles": [
        "http://localhost:3000/tiles.layer_one,tiles.layer_two/{z}/{x}/{y}"
    ],
    "vector_layers": [
        {
            "id": "tiles.layer_one",
            "fields": {
                ...
            }
        }
    ],
    "bounds": [
        ...
    ],
    "description": "tiles.layer_one.geom",
    "name": "tiles.layer_one"
}
```

When using mapbox this does actually appear to work fine, both layers one and two are available to the user. The downside however is that every time I try to add layer two as a layer to the map, I get the following error:

`Error: Source layer "tiles.layer_two" does not exist on source "my_source" as specified by style layer "layer_two"`

nyurik:
Improve merging of multiple tilejsons

Fix #652

### DIFF
--- a/src/srv/server.rs
+++ b/src/srv/server.rs
@@ -21,7 +21,7 @@ use itertools::Itertools;
 use log::{debug, error};
 use martin_tile_utils::{Encoding, Format, TileInfo};
 use serde::{Deserialize, Serialize};
-use tilejson::TileJSON;
+use tilejson::{tilejson, TileJSON};
 
 use crate::source::{Source, Sources, UrlQuery, Xyz};
 use crate::srv::config::{SrvConfig, KEEP_ALIVE_DEFAULT, LISTEN_ADDRESSES_DEFAULT};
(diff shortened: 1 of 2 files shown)


stepankuzmin (approved):
👍

PR #660, opened 2023-05-10
adds search; adds resize; adds dynamic layer color;

### LINKED ISSUE #536: Improve MapLibre demo to show all available lines/points/polygons from all sources
Current martin [demo](https://github.com/maplibre/martin/blob/main/tests/debug.html) page shows a list of all available sources (by first calling the `/catalog` API), and allows the users to turn on/off each source individually.  There are some limitations that need to be addressed:

* it relies on the tilejson to tell it which type of information is in the data. Instead, it should allow all types (points / lines / ...), and visualize them all
* it shows everything in the same color - each layer should get its own color, e.g. randomly when the layer is enabled, and possibly with a color picker for extra bonus.
* it requires to know the name of each layer in the MVT. Instead, it should use the "inspection" mode, where all layers are shown regardless of their name
   * it could use multiple colors if the MVT has more than one layer
* make the left sidebar resizable
* make the list of the sources substring-filterable with an input box at the top
* ...?

### Requirements
* keep it to a single javascript file
* keep it as simple as possible

### Instructions for those who may want to help with this
* install rust
* install docker and docker-compose
* run `rust install just`
* run `just run`
* open `tests/debug.html` file in your browser by copy/pasting the full path to it

Binabh:
- [ ] show lines and polygons
- [ ] show all layers regardless of their name
- [x] different color for each layer
- [x] make left sidebar resizable
- [x] add search box

- closes #536

### DIFF
--- a/tests/debug.html
+++ b/tests/debug.html
@@ -34,6 +34,7 @@
         width: 120px;
         font-family: 'Open Sans', sans-serif;
         overflow: scroll;
+        resize: horizontal;
       }
 
       #menu a {
@@ -68,14 +69,50 @@
       #menu a.active:hover {
         background: #3074a4;
       }
+      #menu-search {
+        padding: 2px;
+        width: 100%;
+      }
     </style>
   </head>
 
   <body>
-    <nav id="menu"></nav>
+    <nav id="menu">
+      <input
+        oninput="handleSearch()"
+        id="menu-search"
+        type="search"
+        placeholder="Search..."
+      />
+    </nav>
     <div id="map"></div>
 
     <script>
+      function handleSearch() {
+        const search = document.getElementById("menu-search").value;
+        const links = document.querySelectorAll("#menu a");
+        for (const link of links) {
+          if (link.textContent.toLowerCase().includes(search.toLowerCase())) {
+            link.style.display = "block";
+          } else {
+            link.style.display = "none";
+          }
+        }
+      }
+
+      // Reference from: https://stackoverflow.com/a/16348977/11522644
+      const stringToColour = function (str) {
+        let hash = 0;
+        for (let i = 0; i < str.length; i++) {
+          hash = str.charCodeAt(i) + ((hash << 5) - hash);
+        }
+        let colour = "#";
+        for (let i = 0; i < 3; i++) {
+          const value = (hash >> (i * 8)) & 0xff;
+          colour += ("00" + value.toString(16)).substr(-2);
+        }
+        return colour;
+      };
       const map = new maplibregl.Map({
         container: 'map',
         style: 'https://basemaps.cartocdn.com/gl/positron-gl-style/style.json',
@@ -130,7 +167,7 @@
               visibility: 'none'
             },
             paint: {
-              [`${layerType}-color`]: 'red'
+              [`${layerType}-color`]: stringToColour(source.id)
             }
           });
 
--- a/tests/debug.html
+++ b/tests/debug.html
@@ -69,9 +69,25 @@
       #menu a.active:hover {
         background: #3074a4;
       }
+
       #menu-search {
         padding: 2px;
         width: 100%;
+      }
+
+      #menu a.active #colorbox {
+        display: block;
+      }
+
+      #menu a #colorbox {
+        display: none;
+      }
+
+      #colorbox {
+        width: 10px;
+        height: 10px;
+        float: right;
+        margin-left: 5px;
       }
     </style>
   </head>
@@ -175,12 +191,17 @@
             console.log(e.features);
           });
 
+          const colorbox = document.createElement("div");
+          colorbox.id = "colorbox";
+          colorbox.style.backgroundColor = stringToColour(source.id);
+
           // Create a link.
           const link = document.createElement('a');
           link.id = source.id;
           link.href = '#';
           link.textContent = source.id;
           link.title = source.id;
+          link.appendChild(colorbox);
 
           // Show or hide layer when the toggle is clicked.
           link.onclick = function (e) {
--- a/tests/debug.html
+++ b/tests/debug.html
@@ -84,10 +84,24 @@
       }
 
       #colorbox {
-        width: 10px;
-        height: 10px;
+        -webkit-appearance: none;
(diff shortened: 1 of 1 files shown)


nyurik:
This is awesome!!! Thank you! I especially like the search capability.

So some feedback:
* the biggest (and most unknown) aspect is that we need to use the inspection approach. No comments on this just yet
* would be nice (but totally minor, do only if easy) to be able to resize the left panel along any location along the right edge, not just the small bottom right resize handle.
* would be nice to show a colored box next to each layer to indicate the color it is drawn with. Don't show that box if the layer is not active.  Bonus points if that box can be clicked to specify a different color (this may be useful in cases when auto-generated color is too pale/invisible on the screen). An alternative to specifying a color is to randomly generate a new color on each layer-enable. This way if i don't like the color, i simply turn it off and on again.
* A much more "long term" planning (just to think - definitly not part of this PR) -- would be awesome to integrate the entire Maputnik as the UI of Martin, and possibly allow the style to be stored in the local store, etc, etc, etc...

In short, thank you for working on this!!!

Binabh:
@nyurik I am still stuck on this part. This is what I have researched upto now:
To display all types of geometries(points/line/polygon) we need to give proper type [here](https://github.com/maplibre/martin/blob/34d16d28fcfc7add0cd6aaddf9b0a2d4df2a1710/tests/debug.html#L123) Now to detect proper geometry type.
As per your suggestion I looked into [mapbox-gl-inspect](https://github.com/lukasmartinelli/mapbox-gl-inspect) the way they detect type I think [here](https://github.com/lukasmartinelli/mapbox-gl-inspect/blob/dab82f2f83670bfeea93195da5db0f35dae99921/lib/MapboxInspect.js#LL191C11-L191C11). However this works when we want to detect geometry type of rendered features unlike in our case where we need to  know geometry type for rendering.
We also have [querySourceFeatures](https://maplibre.org/maplibre-gl-js-docs/api/map/#map#querysourcefeatures) but this is also returning empty array for un-rendered features and even here also I cannot seem to find geometry type in data.

Now I want to ask
1. Am I even looking at right places?
2. Is it even possible to detect geometry types on client side from vector tiles?
3. Can you point me to any such examples/references?

If there is no way to detect geometry type of vector tiles in client side then I have thought or few workarounds
1. include geometry type in `/catalog` or `/<source-id>` api. For this for postgis we could use spatial function [GeometryType](https://postgis.net/docs/GeometryType.html).
2. define geometry type for each source like 
```json
{
"MixPoints": "POINTS",
"points1": "POINTS"
}
```
This approach obviously defies purpose of getting all layers and plotting them automatically. This is the most easy approach.

This is all I have found till now. Your guidance here will be very much helpful.
Thank You!

nyurik:
@Binabh thanks for all the hard work on this!  Try this:
* open https://maputnik.github.io/editor/#4.2/41.91/-91.38
* in the top menu, change `view` to `Inspect`
* in the top menu, click `Open`, and select `Empty Style` (last)
* in the top menu, click `Data Sources`, and click `OpenMapTiles v3` in the `Choose Public Source` - and close that popup

You are now seeing the Maputnik inspector mode -- all geometries.  If you look at the [tilejson](https://api.maptiler.com/tiles/v3/tiles.json?key=KDhMfHvorAFkFe64wlZb), you will see that it does have the `"geometry"` field that lists all available geometry types - but I do NOT know if this is a requirement or not for the inspector mode.

The issue with the server-side geometry generation is that I am not certain if we can compute the type without a full table scan -- something I would obviously want to avoid, esp on startup.

Binabh:
@nyurik I think I cannot do this in client side with data we are currently getting. About Inspect mode I think it displays layer type like the one given [here](https://github.com/maplibre/martin/blob/bb4370c5fe3dcabb3ae9fb6501337971a42a2689/tests/debug.html#L123). In our case its always circle by [this function](https://github.com/maplibre/martin/blob/bb4370c5fe3dcabb3ae9fb6501337971a42a2689/tests/debug.html#L87) as the value of geometry type is always undefined.
So,
By my current understanding,I am unable to show all types of layers by accurately determining their types.
Now, I can do any of following options:
1.Look into options to determine and send layer geometry type from server side
2.Merge this PR in its current state(since things like choosing colors, search box and resizing are implemented) and move task of showing different geometry types to new issue 
3.Close this PR and let someone else handle this issue as a whole

I am also open to any other suggestions.
Please let me know which way can we head? If 2 or 3 then I will change status of PR to ready for merge or closed accordingly. If 1 I will start looking at code base and ways to do it.

Thank you!

nyurik (approved):
Thanks, looks awesome! Let's merge as is for now. Once we figure out how to visualize everything, we may even make something like this a starting page

nyurik:
P.s. I'll have to look into auto detection a bit more myself. Problem is, we may not always have a way to detect what type of geometry each layer has, eg if it comes from mbtiles or pmtiles source

Binabh:
Thank you @nyurik . I will also continue my research on this. Trying to solve this issue I learned a lot about vectortiles, maplibre and martin. I will try to be involved as much as possible.
Thanks to you and team for this awesome project.

tomeronen:
Hello @nyurik and @Binabh i am thinking about trying to create a wrapper around maputnik to server the martin catalog. Do you know if there is already wip for this, or is somebody working on it now?

nyurik:
@tomeronen the closest is the https://github.com/maplibre/martin/blob/main/tests/debug.html - which I would love to abandon and have a proper Martin UI instead. Feel free to create a new issue and/or just hack on a new PR.  Noone's working on it AFAIK.

tomeronen:
@nyurik i did some work on this and created a basic version that is more like [maputnik](https://maputnik.github.io/) and a basic version that is more like [kepler.gl](https://kepler.gl/demo). Kepler is data focused compared, to maputnik which is more focused on creating a mapbox style. Currently thinking on going the kepler.gl route. I can start with a fork with a basic version and add more functionality along the way

tomeronen:
Created a basic poc (still wip) for a possible UI. 
https://github.com/maplibre/martin/pull/1119
Will continue to work on it, adding functionality and cleanups.

PR #663, opened 2023-05-16
In auto-discover, hide geocolumns from tables with multiple ones

### LINKED ISSUE #627: For PG tables, do not include any geo columns in the output
If a PG table has multiple geometry columns, e.g. `col1` and `col2`, Martin should create two table sources, but both table sources should not include neither `col1` or `col2` in the output as a "regular" field.

sharkAndshark:
Hi @nyurik I'm trying to resolve #627 .But it's a little bit hard  battle against rust😂

### DIFF
--- a/README.md
+++ b/README.md
@@ -303,7 +303,7 @@ curl localhost:3000/points,lines/0/0/0
 
 # Table Sources
 
-Table Source is a database table which can be used to query [vector tiles](https://github.com/mapbox/vector-tile-spec). When started, Martin will go through all spatial tables in the database and build a list of table sources. A table should have at least one geometry column with non-zero SRID. All other table columns will be represented as properties of a vector tile feature.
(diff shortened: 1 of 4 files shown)


nyurik:
@sharkAndshark thanks for contributing!  I think you can make your life much easier - just modify this code:

https://github.com/maplibre/martin/blob/34d16d28fcfc7add0cd6aaddf9b0a2d4df2a1710/src/pg/scripts/get_table_sources.sql

I would suggest creating a local db (e.g. run `just start` and `just print-conn-str` to see the DB conn string), connect to it using some postgres query tool, and modify the code to remove geometry columns from the query result.

Simplest -- just add

`FILTER (WHERE columns.column_name IS NOT NULL AND columns.type_name != 'geometry')`
instead of `FILTER (WHERE columns.column_name IS NOT NULL)`

nyurik:
Thanks, looks good! Can you update the Readme to describe this change?

sharkAndshark:
Yeah  sure.

nyurik (approved):
Thanks, looks good!

PR #681, opened 2023-05-26
Create mdbook documentation

sharkAndshark:


.gitignore
```diff
@@ -10,3 +10,4 @@ test_log*
 pg_data/
 config.yml
 tests/output/
+docs/book
```
nyurik: please enable "auto newline at the end of a file" in your IDE, e.g. https://stackoverflow.com/questions/16761227/how-to-make-intellij-idea-insert-a-new-line-at-every-end-of-file

nyurik (commented):
Looks like a great start, thanks!  I added a justfile target for it to simplify things and pushed it to your branch. Looking forward to having a proper docs soon!

.gitignore
sharkAndshark: Yeah. I didn't know this and enable it on vscode now. I'm trying to understand why it's important by google. Thx  :D

sharkAndshark:
Do you think we need adjust the structure of book? @nyurik

.gitignore
nyurik: why its important to have last new line? Same reason as that we want trailing commas in a list, even though last comma is optional:  if you make a change to the code, e.g. add one more line, you want the `diff` to only show one line change, not two

nyurik:
> Do you think we need adjust the structure of book? @nyurik

Sure, any restructuring is possible - but I think we should start with matching the current readme, migrate to mdbook (e.g. make it auto-publish to the maplibre.org/martin site), and then start changing - otherwise we will have two duplicate places to keep up to date, and that's never good.

sharkAndshark:
Yeah.One PR for one thing

nyurik (approved):
thank you !!  merging

PR #692, opened 2023-05-30
Simplify demo - remove nginx

birkskyum:
This uses the webserver built into vitejs instead of configuring an entirely new nginx server with an additional docker image.


demo/docker-compose.yml
```diff
@@ -6,14 +6,17 @@ services:
...
+    command: ["npm", "run", "preview"]
     depends_on:
       - tiles
     ports:
-      - "80:80"
+      - "80:8080"
 
   tiles:
-    image: urbica/martin
+    image: arm64martinimage
```
birkskyum: This should probably be the latest maplibre/martin image

birkskyum:
How come the workflows doesn't run?

PR #697, opened 2023-06-01
Cleaning up mdbook

nyurik:
A lot of book content cleanup, making it a bit shorter and "logical" (whatever that even means)

To try, run `just mdbook`

image

cc: @sharkAndshark


sharkAndshark:
Looks great! I love this version.  
And I'm thinking something these days. I dont' know. Is it ok to add a `get started` section before `runing` section, to clarify the easiest and common ways to give martin a taste. It's for the person who is interested but has no time or enough skills.
 Like:
image

nyurik:
sure, i'm all for some tutorial. Moreover, i think that tutorial will be far easier to do without docker or any other fancy install - instead simply download the binary from the release dir. That said, we still don't have a  good story on how to serve static assets (e.g. maplibre lib, stylesheet, fonts,...)... So still TBD

nyurik:
I think i will self merge this one and this way you can propose a change to it more easily

PR #699, opened 2023-06-02
Move Martin crate to its own subdir

nyurik:
This cleans up the code a bit, making all crates to be on the same level.


birkskyum (approved):

birkskyum:
So with this there are these three, right?
* martin-mbtiles
* martin
* martin-tile-utils

nyurik:
@birkskyum yep, so far :)

Issue #667, opened 2023-05-18
Expand mbtiles binary


nyurik:
mbtiles crate has some good functionality to handle mbtiles (sqlite), and it would benefit mbtiles users to get some more functionality:

* [x] show mbtiles' metadata
* [x] compute available zoom min/max
* [x] support conversion between "deduplicated data with the view schema" vs "simple schema" for storing tiles. Dedup schema is the one where there are two tables: `z, x, y, hash` and `hash, tile_data`, plus a view to join them.
* [x] support comparing two sqlite files
* [x] support diffing between two sqlite files
* [x] allow to `get` and `set` metadata values, e.g. `mbtiles set-meta <file> <property> <new_value>`

Overall structure could be a single binary, and CLI styled similar to `git`: `mbtiles meta <file.mbtiles>`. Some of these functions could be copied from the [OMT's mbtiles-tools](https://github.com/openmaptiles/openmaptiles-tools/blob/master/bin/mbtiles-tools) (except that it should be far simpler to install with `cargo install martin-mbtiles`  (we may want to rename martin-mbtiles into mbtiles for simplicity?)

The above should probably be done as individual sub-tasks

CommanderStorm:
> compute available zoom min/max

![Image](https://github.com/user-attachments/assets/4f20e155-399a-487d-9f4e-d31d1847e600)

@nyurik is this what you meant? -> that subtask could be checked then..

nyurik:
@CommanderStorm thx for doing all the housekeeping!  The `meta-all` does not compute anything, it simply prints metadata table.  That said, @sharkAndshark (IIRC) did it a while back:

```bash
❯ mbtiles summary /home/nyurik/dev/rust/martin/tests/fixtures/mbtiles/geography-class-jpg-diff.mbtiles
MBTiles file summary for /home/nyurik/dev/rust/martin/tests/fixtures/mbtiles/geography-class-jpg-diff.mbtiles
Schema: normalized
File size: 51.50KiB
Page size: 512B
Page count: 103

 Zoom |   Count   | Smallest  |  Largest  |  Average  | Bounding Box
    1 |         1 |   11.8KiB |   11.8KiB |   11.8KiB | 0,0,180,85
    2 |         1 |   21.6KiB |   21.6KiB |   21.6KiB | 0,0,90,67
  all |         2 |   11.8KiB |   21.6KiB |   16.7KiB | 0,0,180,85
```

nyurik:
[...]

PR #701, opened 2023-06-02
Add `meta-get` functionality

upsicleclown:
* Add functionality to retrieve a metadata value in an mbtiles file by key; simple implementation for one of the items in #667 
* Also, disable TTY in docker-up `just` target

### DIFF
--- a/docs/src/tools.md
+++ b/docs/src/tools.md
@@ -1,8 +1,7 @@
 # Tools
 
-## Mbtiles tools:
-A small binary utility that allows users to interact with mbtiles files from the CLI as follows: `mbtiles <command> <file.mbtiles>`
-- ### `meta-get`:
-    Retrieve a metadata value by key: `mbtiles meta-get <file.mbtiles> <options> <key>`.  
-    #### Options:
-  - `-r, --raw` : return the raw metadata value
+## MBTiles tools
+A small utility that allows users to interact with mbtiles files from the CLI as follows: `mbtiles <command> ...`
+
+#### `meta-get`
+Retrieve a metadata value by its name: `mbtiles meta-get <file.mbtiles> <key>`. See `mbtiles meta-get --help` to see available options.
--- a/docs/src/tools.md
+++ b/docs/src/tools.md
@@ -1,0 +1,8 @@
+# Tools
+
+## Mbtiles tools:
+A small binary utility that allows users to interact with mbtiles files from the CLI as follows: `mbtiles <command> <file.mbtiles>`
+- ### `meta-get`:
+    Retrieve a metadata value by key: `mbtiles meta-get <file.mbtiles> <options> <key>`.  
+    #### Options:
+  - `-r, --raw` : return the raw metadata value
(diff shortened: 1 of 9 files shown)


martin-mbtiles/Cargo.toml
```diff
@@ -21,6 +21,8 @@ serde_json.workspace = true
 sqlx.workspace = true
 thiserror.workspace = true
 tilejson.workspace = true
+tokio = { version = "1.25.0", features = ["full"] }
+anyhow = "1.0"
```
nyurik: These should be moved [here](https://github.com/maplibre/martin/blob/main/Cargo.toml#L87) (keep that list sorted), and here you just want to use `.workspace = true`. Also, move these two lines down to # Bin dependencies right below.

Using workspace deps allow all dependency versions to be managed in one place.

docs/src/tools.md
```diff
@@ -0,0 +1,8 @@
+# Tools
+
+## Mbtiles tools:
```
nyurik:
```suggestion
## MBTiles tools
```

docs/src/tools.md
```diff
@@ -0,0 +1,8 @@
...
+A small binary utility that allows users to interact with mbtiles files from the CLI as follows: `mbtiles <command> <file.mbtiles>`
```
nyurik:
```suggestion
A small utility that allows users to interact with mbtiles files from the CLI as follows: `mbtiles <command> ...`
```

docs/src/tools.md
```diff
@@ -0,0 +1,8 @@
...
+- ### `meta-get`:
```
nyurik:
```suggestion

#### `meta-get`
```

docs/src/tools.md
```diff
@@ -0,0 +1,8 @@
...
+    Retrieve a metadata value by key: `mbtiles meta-get <file.mbtiles> <options> <key>`.
```
nyurik:
```suggestion
Retrieve a metadata value by its name: `mbtiles meta-get <file.mbtiles> <key>`. See `mbtiles meta-get --help` to see available options.
```

docs/src/tools.md
```diff
@@ -0,0 +1,8 @@
+# Tools
+
+## Mbtiles tools:
+A small binary utility that allows users to interact with mbtiles files from the CLI as follows: `mbtiles <command> <file.mbtiles>`
+- ### `meta-get`:
+    Retrieve a metadata value by key: `mbtiles meta-get <file.mbtiles> <options> <key>`.  
+    #### Options:
+  - `-r, --raw` : return the raw metadata value
```
nyurik:
```suggestion
```

Meh not needed i think - there might be other options, and keeping docs in sync with the actual program is kinda tricky... maybe later?

nyurik (commented):
Awesome work, thanks!  I left a few suggestions

nyurik (approved):
looks great, thanks!!!

PR #702, opened 2023-06-03
Refactor mbtiles lib, mbtiles tool tests & CI

nyurik:
* Broke up martin-mbtiles into multiple files
* Made all mbtiles functions take a `SqliteExecutor` -- this way they can be used with any SQLX connection structs - either a pool connection or an individual non-pooled connection.
* Simplified mbtiles bin a bit - I realized there is really no need to pretty print the output for the single value retrieval. Easier to just dump it to console as is.
* Bump martin-mbtiles to v0.2.0
* Minor fixes in tools docs, cargo.toml, and justfile
* MBTiles tool Integration tests and release publishing

Major thanks to the [stackoverflow](https://stackoverflow.com/questions/76394665/how-to-pass-sqlx-connection-a-mut-trait-as-a-fn-parameter-in-rust/76395111) quick reply by @cafce25 on how to use generic sql executor!

cc: @upsicleclown


birkskyum:
Is there changelogs for react e create?

nyurik:
Change log is generated by the releases screen in github

birkskyum (approved):

PR #708, opened 2023-06-07
Update README.md export command

taiebnoe:
Quoting DATABASE_URL content variable prevents an issue from the command export. In case of a string (i.e. password) with given characters (i.e. '('), it can make the command fail. Quoting it prevents that from happening.
The issue happened on MacOS with an M1 chip.


nyurik:
thanks!   The irony of this is that double quotes would still process dollar signs `$`, back slashes `\`, and double quotes `"`, so in theory wrapping with single quotes is somewhat safer... except that in that case you would have to handle the single quote escaping :)

PR #712, opened 2023-06-12
Add basic copying functionality

upsicleclown:
Copy an existing `.mbtiles` file to a new file.

```shell
mbtiles copy  <src_file.mbtiles> <dst_file.mbtiles>
```

Optionally filters by zooms.  Supports de-duplicated and simple mbtiles


martin-mbtiles/src/queries/is_deduplicated_mbtiles.sql
```diff
@@ -0,0 +1,28 @@
...
+        FROM sqlite_master
+        WHERE name='images' AND type='table')
+        JOIN
+        (SELECT 1
+        FROM (SELECT COUNT(*) as col_count
+                     FROM pragma_table_info('images')
+                     WHERE "notnull"=0 AND ((name="tile_id" AND type="TEXT") OR
+                                            (name="tile_data" AND type="BLOB")))
+        WHERE col_count = 2))
+
```
nyurik: I think using JOINs is a bit hard to reason-about, especially as it gets more and more nested.  A simpler might be something like this. Note that SQL is a messy language, so a bit more verbosity is ok IMO.

```suggestion
SELECT (
           -- Has a "map" table
           SELECT COUNT(*) = 1
           FROM sqlite_master
           WHERE name = 'map'
             AND type = 'table'
           --
       ) AND (
           -- "map" table's columns and their types are as expected:
           -- 4 non-null columns (zoom_level, tile_column, tile_row, tile_id).
           -- The order is not important
           SELECT COUNT(*) = 4
           FROM pragma_table_info('map')
           WHERE "notnull" = 0
             AND ((name = "zoom_level" AND type = "INTEGER") OR
                  (name = "tile_column" AND type = "INTEGER") OR
                  (name = "tile_row" AND type = "INTEGER") OR
                  (name = "tile_id" AND type = "TEXT"))
           --
       ) AND (
           -- Has a "images" table
           SELECT COUNT(*) = 1
           FROM sqlite_master
           WHERE name = 'images'
             AND type = 'table'
           --
       ) AND (
           -- "images" table's columns and their types are as expected:
           -- 2 non-null columns (tile_id, tile_data).
           -- The order is not important
           SELECT COUNT(*) = 2
           FROM pragma_table_info('images')
           WHERE "notnull" = 0
             AND ((name = "tile_id" AND type = "TEXT") OR
                  (name = "tile_data" AND type = "BLOB"))
           --
       ) AS is_valid;
```

martin-mbtiles/src/tile_copier.rs
```diff
@@ -0,0 +1,202 @@
...
+    Unknown,
+    StandardCompliant,
+    Deduplicated,
+}
+
+#[derive(Clone, Default, Debug)]
+pub struct TileCopier {
+    src_filepath: PathBuf,
+    dst_filepath: PathBuf,
+    zooms: Vec<u8>,
```
nyurik: You may want to use `HashSet` here to avoid accidental dups

martin-mbtiles/src/tile_copier.rs
```diff
@@ -0,0 +1,202 @@
...
+use crate::MbtError;
+use sqlx::sqlite::SqliteConnectOptions;
+use sqlx::{query, Connection, Row, SqliteConnection, SqliteExecutor};
+use std::fs::File;
+use std::path::PathBuf;
+
+#[derive(Debug)]
+enum StorageType {
+    Unknown,
+    StandardCompliant,
```
nyurik: technically both are "standard-compliant" (standard allows tiles table to be a view (see [here](https://github.com/mapbox/mbtiles-spec/blob/master/1.3/spec.md#database)):
> Note: the schemas outlined are meant to be followed as interfaces. SQLite views that produce compatible results MAY be used instead. For convenience, this specification refers to tables and virtual tables (views) as tables.

I would recommend something like `Tables` and `Deduplicated`. Also, we do not need the `Unknown` enum type.  If we don't detect what we expect, its an error. If you mix type with unknown, you are effectively mixing valid values with a single error condition - which is a big no-no is the Rust land - because that forces every consumer of the enum type to first check if it contains the error or not.

martin-mbtiles/src/tile_copier.rs
```diff
@@ -0,0 +1,202 @@
...
+        self.min_zoom = Some(max_zoom);
+        self
+    }
+
+    pub fn verbose(&mut self, verbose: bool) -> &mut Self {
+        self.verbose = verbose;
+        self
+    }
+
+    pub async fn run(&self) -> MbtResult<()> {
```
nyurik: I think it might be better to move `run` out of the TileCopier implementer -- this way you will have a clear separation between the `options` object - a simple (dumb) struct that only contains settings, and a separate function that takes that object and runs with it, e.g.

`pub async fn copy_tiles(options: TileCopierOptions) {...}`.  The `fn copy_tiles` you have below can then be inlined, no need to do all that magical `SqliteExecutor` stuff.

martin-mbtiles/src/tile_copier.rs
```diff
@@ -0,0 +1,202 @@
...
+                self.copy_standard_compliant_tiles(src_conn).await
+            }
+            Some(StorageType::Deduplicated) => self.copy_deduplicated_tiles(src_conn).await,
+            _ => Err(MbtError::InvalidDataStorageFormat(
+                self.src_filepath.clone(),
+            )),
+        };
+    }
+
+    async fn determine_storage_type<T>(&self, conn: &mut T) -> MbtResult<Option<StorageType>>
```
nyurik: this should not be a member function. I think Clippy would even raise a flag here because you never use `self`. Also, maybe call it `detect_mbtiles_format`?  Slightly more precise, and consistent with other detect fns?

martin-mbtiles/src/tile_copier.rs
```diff
@@ -0,0 +1,202 @@
...
+        for<'e> &'e mut T: SqliteExecutor<'e>,
+    {
+        match File::create(&self.dst_filepath) {
+            Ok(_) => {}
+            Err(_) => return Err(MbtError::CouldNotCreateMBTiles(self.dst_filepath.clone())),
+        };
+        let opt = SqliteConnectOptions::new().filename(&self.dst_filepath);
+        let mut dst_conn = SqliteConnection::connect_with(&opt).await?;
+        //TODO: Q fix this with format string?
+        let schema = query("SELECT sql FROM sqlite_schema WHERE tbl_name IN ('metadata', 'tiles', 'map', 'images')")
```
nyurik: probably better to move this down two lines where you use it

martin-mbtiles/src/tile_copier.rs
```diff
@@ -0,0 +1,202 @@
...
+            .fetch_optional(&mut *conn)
+            .await?
+        {
+            return Ok(Some(StorageType::Deduplicated));
+        }
+
+        Ok(Some(StorageType::Unknown))
+    }
+
+    async fn create_target_db<T>(&self, conn: &mut T) -> MbtResult<()>
```
nyurik: perhaps `copy_mbtiles_schema_to_new`  (naming is hard, there might be a much better name  - i just think create-target-db is ambiguous)

martin-mbtiles/src/tile_copier.rs
```diff
@@ -0,0 +1,202 @@
...
+        self.create_target_db(conn).await?;
+
+        let opt = SqliteConnectOptions::new().filename(&self.dst_filepath);
+        let mut dst_conn = SqliteConnection::connect_with(&opt).await?;
+
+        query(&*format!(
+            "ATTACH DATABASE '{}' AS sourceDb",
+            match self.src_filepath.to_str() {
+                Some(v) => v,
+                None => return Err(MbtError::NoSuchMBTiles(self.src_filepath.clone())),
```
nyurik: this is the wrong error - the reason `to_str()` could return None is due to inability to convert an "OS" string to a "UTF8" string. This is mostly a Windows compatibly issue - windows internally is still not using unicode, so there could be a filename that has invalid Unicode in it. SQLite also handles that internally when you open a connection - https://github.com/launchbadge/sqlx/blob/main/sqlx-sqlite/src/connection/establish.rs#L56 - so just do something similar?

Also, you probably want to do all the fun things like creating a blank db / attaching / creating table structure in some common code, and then just call one of the copying functions with the working connection and nothing else (i.e. no dsttable).

nyurik (commented):
thanks!  I left a few inline comments.

Overall, good approach, and I like that you attach the source rather than dst - might solve a few weird cases. My overall understanding of the process (correct if i miss something)

* open source file in a read-only mode
* detect mbtiles type + any other validations
* close source file

---

* open dst file
* check that `SELECT count(*) = 0 FROM sqlite_schema;` returns `1`, and error that we don't support copying to the non-empty mbtiles yet
* set/update any initial pragmas/vacuum
* attach source file
* copy sql schema
* copy data using one of the copy methods

I will think about this some more. Also, I suspect i might be wrong in overal structure, still thinking how to organize it best. One thing that kinda bothers me is that the API currently does not consume the options when it starts copying - a better path might be 
* a "builder" option struct `CopyOptions`
* a function `pub async fn copy_mbtiles(options: CopyOptions)` (consumes options)
* a bunch of private functions that actually do the copying

martin-mbtiles/src/queries/is_deduplicated_mbtiles.sql
upsicleclown: If the condition is false, using `AND` will return a `NULL` (represented as a newline) and using `JOIN` will simply return nothing. This way, with `JOIN`, I can use `if let Some(_v) = query(...)` easily. 

Using `JOIN` does make the code slightly less intuitively readable than `AND` though, I will add comments as suggested to clarify the parts of the query.

nyurik: @upsicleclown good catch, thx!  I went through a few iterations (you can see them in the history of the above message), and at one point over-optimized a bit :)  I restored the working version - now it always returns 1 or 0

upsicleclown: Thank you!

martin-mbtiles/src/bin/main.rs
```diff
@@ -39,13 +39,26 @@ enum Commands {
...
+    /// Copy tiles from one mbtiles file to another.
+    #[command(name = "copy")]
+    Copy {
+        /// MBTiles file to read from
+        src_file: PathBuf,
+        /// MBTiles file to write to
+        dst_file: PathBuf,
+        /// Display detailed copying information
+        #[arg(short, long)]
+        verbose: bool,
```
nyurik: we probably want to move `verbose` flag before the `command` arg in the `Args` struct above - I'm sure it will apply to more than just copying

martin-mbtiles/src/bin/main.rs
```diff
@@ -56,6 +69,27 @@ async fn main() -> Result<()> {
...
         }
+        Commands::Copy {
+            src_file,
+            dst_file,
+            verbose,
+            min_zoom,
+            max_zoom,
+            zoom_levels,
+        } => {
+            let mut tile_copier_options =
```
nyurik: please rename this var for readability - a bit too long for such a short usage.... maybe `opt` or `copy_opts` or ...?

martin-mbtiles/src/tile_copier.rs
```diff
@@ -0,0 +1,202 @@
...
+        self.max_zoom = Some(max_zoom);
+        self
+    }
+
+    pub fn verbose(mut self, verbose: bool) -> Self {
+        self.verbose = verbose;
+        self
+    }
+}
+impl TileCopier {
```
nyurik:
```suggestion
}

impl TileCopier {
```

martin-mbtiles/src/tile_copier.rs
```diff
@@ -0,0 +1,202 @@
...
+            .execute(&mut conn)
+            .await?;
+
+        match storage_type {
+            Type::TileTables => self.copy_standard_compliant_tiles(&mut conn).await,
+            Type::DeDuplicated => self.copy_deduplicated_tiles(&mut conn).await,
+        }
+    }
+
+    async fn copy_standard_compliant_tiles(&self, conn: &mut SqliteConnection) -> MbtResult<()> {
```
nyurik: they are all standard compliant - need a better name that matches the ones in the Type enum.  BTW, the `Type` enum seems kinda weird tbh, maybe `MbtType` instead for clarity, and consistency with the errors?

martin-mbtiles/src/tile_copier.rs
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    async fn copy_deduplicated_tiles(&self, conn: &mut SqliteConnection) -> MbtResult<()> {
+        query("INSERT INTO map SELECT * FROM sourceDb.map")
+            .execute(&mut *conn)
+            .await?;
+
+        self.run_query_with_options(
+            conn,
+            String::from(
```
nyurik: i think `"...".to_string()` is slightly shorter and more consistent with other code in Martin - and removes the need to track end parenthesis.

That said, in this specific case you may even simplify the `run_query_with_options` to use `sql: &'static str` instead of `mut sql: String`, and just create a string inside the function, adding any needed WHERE clause.  Note that `'static` is technically not needed either.

martin-mbtiles/src/tile_copier.rs
```diff
@@ -0,0 +1,202 @@
...
+
+    async fn run_query_with_options(
+        &self,
+        conn: &mut SqliteConnection,
+        mut sql: String,
+    ) -> MbtResult<()> {
+        let mut params: Vec<String> = vec![];
+
+        if !&self.options.zooms.is_empty() {
+            sql.push_str(
```
nyurik: this is not a good pattern:
* you create a new String (currently its the caller of the `run_query_with_options`, but see my other comment)
* then you create another string with `format!`, but then just copy its content into the first one

Instead, since you seem to only modify the original `sql` param once, change fn signature to take a `sql: &str`, and use `let sql = format!("{sql} WHERE zoom_level ...", ...)`.  Note that the `sql` var does not need to be mutable.

In a more general case (i.e. if you were to keep the original mut String param), you can use the `write!` method - see https://stackoverflow.com/a/65859308/177275

nyurik (commented):
a few minor nits, but looks good!  Need to implement some tests too... I wonder if we should implement some basic info dump first so that we can use them in the integration tests?  Otherwise we won't have a good way to test if the copying actually happened...  So perhaps `mbtiles summary <file>` that prints a nice output of sorts?  It could show something like...

(definitly a separate PR, possibly later, this is not a blocker for this PR!)

```
 file: <file>
 type: deduplicated mbtiles
tiles: 123,456 tiles sized 10..45,455,574, 12,345.678 on average
available zooms:
  0: 2,000bytes
  1: 4 tiles sized 10..45,455,574, 12,345.678 on average
  2: 16 tiles sized 10..45,455,574, 12,345.678 on average
  ...
metadata:
  <all metadata fields here?>
```

nyurik:
P.S. You will need to run `just bless`

martin-mbtiles/src/tile_copier.rs
upsicleclown: `sql` is used to create the query so the query doesn't pick up the reassignment of `sql` in the `if/else` blocks and will always simply use the unmodified function argument. I do create `String`s unnecessarily though, I will make use of `write!`

nyurik: you can do it even simpler, like this:

```rust
        let sql = if !&self.options.zooms.is_empty() {
            for zoom_level in &self.options.zooms {
                params.push(zoom_level.to_string());
            }
            let params = vec!["?"; self.options.zooms.len()].join(",");
            format!("{sql} WHERE zoom_level IN ({})", params)
        } else if let Some(min_zoom) = &self.options.min_zoom {
            if let Some(max_zoom) = &self.options.max_zoom {
                params.push(min_zoom.to_string());
                params.push(max_zoom.to_string());
                format!("{sql} WHERE zoom_level BETWEEN ? AND ?")
            } else {
                params.push(min_zoom.to_string());
                format!("{sql} WHERE zoom_level >= ?")
            }
        } else if let Some(max_zoom) = &self.options.max_zoom {
            params.push(max_zoom.to_string());
            format!("{sql} WHERE zoom_level <= ?")
        } else {
            sql.to_string()
        };
```

upsicleclown: Oh right, duh, thank you!

martin-mbtiles/src/tile_copier.rs
```diff
@@ -0,0 +1,202 @@
...
+                format!(
+                    " WHERE zoom_level IN ({})",
+                    vec!["?"; self.options.zooms.len()].join(",")
+                )
+                .as_str(),
+            );
+            for zoom_level in &self.options.zooms {
```
nyurik: I have seen this common oneliner for this for loop:

```rust
params.extend(self.options.zooms.iter().map(|z| z.to_string()));
```

Technically you can also do it this way, but i find it a bit less readable in some cases:

```rust
params.extend(self.options.zooms.iter().map(u8::to_string));
```

martin-mbtiles/src/tile_copier.rs
```diff
@@ -0,0 +1,192 @@
...
+                vec!["?"; self.options.zooms.len()].join(",")
+            )
+        } else if let Some(min_zoom) = &self.options.min_zoom {
+            if let Some(max_zoom) = &self.options.max_zoom {
+                params.push(min_zoom.to_string());
+                params.push(max_zoom.to_string());
+                format!("{sql} WHERE zoom_level BETWEEN ? AND ?")
+            } else {
+                params.push(min_zoom.to_string());
+                format!("{sql} WHERE zoom_level >= ? ")
```
nyurik:
```suggestion
                format!("{sql} WHERE zoom_level >= ?")
```

martin-mbtiles/src/tile_copier.rs
```diff
@@ -0,0 +1,192 @@
...
+            }
+        } else if let Some(max_zoom) = &self.options.max_zoom {
+            params.push(max_zoom.to_string());
+            format!("{sql} WHERE zoom_level <= ? ")
```
nyurik, line 169:
```suggestion
            format!("{sql} WHERE zoom_level <= ?")
```

nyurik (commented):
nits

nyurik (approved):
looks great, thanks!

PR #716, opened 2023-06-15
Minor refactor and cleanup

nyurik:
* moved `IdResolver` to a separate file
* added `just fmt2` build target to format code using nightly mode
* moved all Actix `Data<AppState>` into individual state objects, e.g. `Data<Sources>` (allows other source types, separate from `Sources` type)
* move all Source-related code to a new `Sources` struct (a simple wrapper over a HashMap)

### DIFF
--- a/martin-mbtiles/src/mbtiles_queries.rs
+++ b/martin-mbtiles/src/mbtiles_queries.rs
@@ -1,5 +1,6 @@
-use crate::errors::MbtResult;
 use sqlx::{query, SqliteExecutor};
+
+use crate::errors::MbtResult;
 
 pub async fn is_deduplicated_type<T>(conn: &mut T) -> MbtResult<bool>
 where
--- a/martin/src/file_config.rs
+++ b/martin/src/file_config.rs
@@ -12,6 +12,7 @@ use crate::config::{copy_unrecognized_config, Unrecognized};
 use crate::file_config::FileError::{InvalidFilePath, InvalidSourceFilePath, IoError};
 use crate::source::{Source, Sources, Xyz};
 use crate::utils::{sorted_opt_map, Error, IdResolver, OneOrMany};
+use crate::OneOrMany::{Many, One};
 
 #[derive(thiserror::Error, Debug)]
 pub enum FileError {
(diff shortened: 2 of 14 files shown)


martin-mbtiles/src/mbtiles_queries.rs
```diff
@@ -1,6 +1,7 @@
-use crate::errors::MbtResult;
 use sqlx::{query, SqliteExecutor};
 
+use crate::errors::MbtResult;
```
upsicleclown, line 3: Is this because the imports of things that are in the crate should be kept separate? It looks like if we use `rustfmt-nightly` we can configure this by default by specifying 
```
reorder_imports_in_group = true
```
in the `rustfmt.toml` file ([link1](https://github.com/rust-lang/rustfmt/issues/298#issuecomment-331481386), [link2](https://docs.rs/crate/rustfmt-nightly/0.3.4/source/Configurations.md)).

nyurik: @upsicleclown do you mean why it has been moved to a separate line separated by a blank line?  We have a few config options already set - https://github.com/maplibre/martin/blob/main/rustfmt.toml - but I don't want to enable it "by default" because the stable rustfmt throws an error if rustfmt has some settings it does not support. So keeping it commented out.

Note that `reorder_imports_in_group` seem to have been removed - try adding it to rustfmt and running `cargo +nightly fmt` - and it will complain.

I added a `fmt2` just target to do it from CLI

martin/src/file_config.rs
```diff
@@ -40,6 +39,22 @@ pub enum FileConfigEnum {
     Config(FileConfig),
 }
 
+impl FileConfigEnum {
+    pub fn extract_file_config(&mut self) -> FileConfig {
+        match self {
+            FileConfigEnum::Path(path) => FileConfig {
+                paths: Some(OneOrMany::One(mem::take(path))),
```
upsicleclown: nit: This looked cleaner when `One` and `Many` were imported directly

martin/src/source.rs
```diff
@@ -28,7 +31,91 @@ impl Display for Xyz {
...
 pub type Tile = Vec<u8>;
 pub type UrlQuery = HashMap<String, String>;
-pub type Sources = HashMap<String, Box<dyn Source>>;
+
+#[derive(Default, Clone)]
+pub struct Sources(HashMap<String, Box<dyn Source>>);
+
+impl Sources {
+    pub fn insert(&mut self, id: String, source: Box<dyn Source>) {
+        self.0.insert(id, source);
```
upsicleclown, line 40: nit: It might be worth giving the `HashMap` in `Sources` a name, so that the accesses are clearer

upsicleclown (approved):

martin/src/source.rs
nyurik: this is a fairly common pattern - esp when there is only one value wrapped. I would rather not change it to a named struct

martin/src/file_config.rs
nyurik: fixed.  This one is kinda tricky because often when you see `One`, it might not give enough context - so i have seen a lot of things like `io::Result` when things are not extra clear.

PR #720, opened 2023-06-20
Add sqlx-rustls support

pka:
I'd like to add optional rustls support for SQLx. This PR implements the required features, but tests using `--all-features` fail, because runtime selection features are not additive. I didn't find a solution apart from avoiding  `--all-features`.

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,2 +1,2 @@
-github: [maplibre]
+github: [ maplibre ]
 open_collective: maplibre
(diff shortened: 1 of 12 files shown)


nyurik:
@pka welcome!!! https://github.com/maplibre/martin/pull/474 tried to do something similar. I would LOVE to switch to rustls, but PostgreSQL is a major pain to switch, sadly...  Some day...

Anyway, I am ok to remove all the `all-features` - we don't really need that. Instead, we should use more specific targets that we know we support

nyurik (approved):

wipfli:
Hi @pka, welcome...

PR #723, opened 2023-06-21
Add force-simple parameter when copying

upsicleclown:
Add `--force-simple` flag to `mbtiles copy` tool.
If supplied, ensures the destination file has a `tiles` _table_ (as opposed to a `tiles` _view_ made up of `images` and `map` tables).

### DIFF
--- a/martin-mbtiles/src/bin/main.rs
+++ b/martin-mbtiles/src/bin/main.rs
@@ -2,11 +2,11 @@ use std::path::{Path, PathBuf};
 
 use anyhow::Result;
 use clap::{Parser, Subcommand};
-use martin_mbtiles::{Mbtiles, TileCopier, TileCopierOptions};
+use martin_mbtiles::{copy_mbtiles_file, Mbtiles, TileCopierOptions};
 use sqlx::sqlite::SqliteConnectOptions;
 use sqlx::{Connection, SqliteConnection};
 
-#[derive(Parser, Debug)]
+#[derive(Parser, PartialEq, Eq, Debug)]
 #[command(
     version,
     name = "mbtiles",
@@ -20,7 +20,7 @@ pub struct Args {
     command: Commands,
 }
 
-#[derive(Subcommand, Debug)]
+#[derive(Subcommand, PartialEq, Eq, Debug)]
 enum Commands {
     // /// Prints all values in the metadata table.
     // #[command(name = "meta-all")]
@@ -44,10 +44,7 @@ enum Commands {
     // },
     /// Copy tiles from one mbtiles file to another.
     #[command(name = "copy")]
-    Copy {
-        #[clap(flatten)]
-        opts: TileCopierOptions,
-    },
+    Copy(TileCopierOptions),
 }
 
 #[tokio::main]
@@ -58,10 +55,8 @@ async fn main() -> Result<()> {
         Commands::MetaGetValue { file, key } => {
             meta_get_value(file.as_path(), &key).await?;
         }
-        Commands::Copy { opts } => {
-            let tile_copier = TileCopier::new(opts)?;
-
-            tile_copier.run().await?;
+        Commands::Copy(opts) => {
+            copy_mbtiles_file(opts).await?;
         }
     }
 
@@ -76,4 +71,145 @@ async fn meta_get_value(file: &Path, key: &str) -> Result<()> {
         println!("{s}")
     }
     Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::Args;
+    use crate::Commands::{Copy, MetaGetValue};
+    use clap::error::ErrorKind;
+    use clap::Parser;
+    use martin_mbtiles::TileCopierOptions;
+    use std::path::PathBuf;
+
+    #[test]
+    fn test_copy_no_arguments() {
+        assert_eq!(
+            Args::try_parse_from(["mbtiles", "copy"])
+                .unwrap_err()
+                .kind(),
+            ErrorKind::MissingRequiredArgument
+        );
+    }
+
+    #[test]
+    fn test_copy_minimal_arguments() {
+        assert_eq!(
+            Args::parse_from(["mbtiles", "copy", "src_file", "dst_file"]),
+            Args {
+                verbose: false,
+                command: Copy(TileCopierOptions::new(
+                    PathBuf::from("src_file"),
+                    PathBuf::from("dst_file")
+                ))
+            }
+        );
+    }
+
+    #[test]
+    fn test_copy_min_max_zoom_arguments() {
+        assert_eq!(
+            Args::parse_from([
+                "mbtiles",
+                "copy",
+                "src_file",
+                "dst_file",
+                "--max-zoom",
+                "100",
+                "--min-zoom",
+                "1"
+            ]),
+            Args {
+                verbose: false,
+                command: Copy(
+                    TileCopierOptions::new(PathBuf::from("src_file"), PathBuf::from("dst_file"))
+                        .min_zoom(Some(1))
+                        .max_zoom(Some(100))
+                )
+            }
+        );
+    }
+
+    #[test]
+    fn test_copy_min_max_zoom_no_arguments() {
+        assert_eq!(
+            Args::try_parse_from([
+                "mbtiles",
+                "copy",
+                "src_file",
+                "dst_file",
+                "--max-zoom",
+                "--min-zoom",
+            ])
+            .unwrap_err()
+            .kind(),
+            ErrorKind::InvalidValue
+        );
+    }
+
+    #[test]
+    fn test_copy_min_max_zoom_with_zoom_levels_arguments() {
+        assert_eq!(
+            Args::try_parse_from([
+                "mbtiles",
+                "copy",
+                "src_file",
+                "dst_file",
+                "--max-zoom",
+                "100",
+                "--min-zoom",
+                "1",
+                "--zoom-levels",
+                "3,7,1"
+            ])
+            .unwrap_err()
+            .kind(),
+            ErrorKind::ArgumentConflict
+        );
+    }
+
+    #[test]
+    fn test_copy_zoom_levels_arguments() {
+        assert_eq!(
+            Args::parse_from([
+                "mbtiles",
+                "copy",
+                "src_file",
+                "dst_file",
+                "--zoom-levels",
+                "3,7,1"
+            ]),
+            Args {
+                verbose: false,
+                command: Copy(
+                    TileCopierOptions::new(PathBuf::from("src_file"), PathBuf::from("dst_file"))
+                        .zoom_levels(vec![1, 3, 7])
+                )
+            }
+        );
+    }
+
+    #[test]
+    fn test_meta_get_no_arguments() {
+        assert_eq!(
+            Args::try_parse_from(["mbtiles", "meta-get"])
+                .unwrap_err()
+                .kind(),
+            ErrorKind::MissingRequiredArgument
+        );
+    }
+
+    #[test]
+    fn test_meta_get_with_arguments() {
+        assert_eq!(
+            Args::parse_from(["mbtiles", "meta-get", "src_file", "key"]),
+            Args {
+                verbose: false,
+                command: MetaGetValue {
+                    file: PathBuf::from("src_file"),
+                    key: "key".to_string(),
+                }
+            }
+        );
+    }
 }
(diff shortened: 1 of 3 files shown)


martin-mbtiles/src/tile_copier.rs
```diff
@@ -70,13 +76,7 @@ impl TileCopier {
         })
     }
 
-    pub async fn run(self) -> MbtResult<()> {
-        let opt = SqliteConnectOptions::new()
```
nyurik, line 74: why remove it? We should still validate that we are not doing a copy operation on a random sqlite file

martin-mbtiles/src/tile_copier.rs
```diff
@@ -98,23 +98,40 @@ impl TileCopier {
...
-        let schema = query("SELECT sql FROM sourceDb.sqlite_schema WHERE tbl_name IN ('metadata', 'tiles', 'map', 'images')")
-            .fetch_all(&mut conn)
-            .await?;
-
-        for row in &schema {
-            let row: String = row.get(0);
-            query(row.as_str()).execute(&mut conn).await?;
-        }
+        if !self.options.force_simple {
+            for row in query("SELECT sql FROM sourceDb.sqlite_schema WHERE tbl_name IN ('metadata', 'tiles', 'map', 'images')")
```
nyurik: would this include copying indexes? If not, you may need a second statement like this to copy index creation

martin-mbtiles/src/tile_copier.rs
```diff
@@ -98,23 +98,40 @@ impl TileCopier {
...
 
-        let schema = query("SELECT sql FROM sourceDb.sqlite_schema WHERE tbl_name IN ('metadata', 'tiles', 'map', 'images')")
-            .fetch_all(&mut conn)
-            .await?;
-
-        for row in &schema {
-            let row: String = row.get(0);
-            query(row.as_str()).execute(&mut conn).await?;
-        }
+        if !self.options.force_simple {
```
nyurik: for slightly better ease-of-read, handle the "true" case first (here and below) -- there is even a lint for this. So just invert the if statement

nyurik (commented):
looks great!  love the `:memory:` usage for the tests, so much cleaner now.  See inline

martin-mbtiles/src/tile_copier.rs
upsicleclown: it does, because I don't specify the type of the schema statement, it will take statements where `type='index'`

martin-mbtiles/src/tile_copier.rs
upsicleclown: Right, conceptually I put the "default" case first, but I will switch them

martin-mbtiles/src/tile_copier.rs
upsicleclown: Right, if we `force-simple` we don't use the `storage_type` which is why I moved it but it's still valuable to verify that it is a valid `storage_type`. Will add it back

martin-mbtiles/src/tile_copier.rs
```diff
@@ -267,18 +279,40 @@ mod tests {
         ));
     }
 
+    #[actix_rt::test]
+    async fn copy_force_simple() {
+        let src_filepath = PathBuf::from("../tests/fixtures/files/world_cities.mbtiles");
```
nyurik: May want to call them same as other tests for consistency

nyurik (commented):
For copying, seems like we may have a bug - when copying schema, we should copy tables before indexes. If we don't have any `order by`, it may fail

martin-mbtiles/src/tile_copier.rs
```diff
@@ -98,23 +104,35 @@ impl TileCopier {
...
 
-        let schema = query("SELECT sql FROM sourceDb.sqlite_schema WHERE tbl_name IN ('metadata', 'tiles', 'map', 'images')")
-            .fetch_all(&mut conn)
-            .await?;
-
-        for row in &schema {
-            let row: String = row.get(0);
-            query(row.as_str()).execute(&mut conn).await?;
-        }
+        if self.options.force_simple {
```
nyurik: another thing - let's use force simple only if we detect that the source is NOT a simple mbtiles

nyurik (commented):
One more thing:  please add this function to the tests (welcome to modify it if needed)

```rust
    async fn get_one<T>(conn: &mut SqliteConnection, sql: &str) -> T
    where
        for<'r> T: Decode<'r, Sqlite> + Type<Sqlite>,
    {
        query(sql).fetch_one(conn).await.unwrap().get::<T, _>(0)
    }
```

This way you can get a single value with simpler code:

```rust
get_one::<u8>(&mut dst_conn, "SELECT COUNT(DISTINCT zoom_level) FROM tiles;").await
```

instead of the one you have

```rust
query("SELECT COUNT(DISTINCT zoom_level) FROM tiles;")
                .fetch_one(&mut dst_conn)
                .await
                .unwrap()
                .get::<u8, _>(0),
```

martin-mbtiles/src/bin/main.rs
```diff
@@ -69,20 +58,8 @@ async fn main() -> Result<()> {
...
-        } => {
-            let copy_opts = TileCopierOptions::new()
-                .verbose(args.verbose)
-                .min_zoom(min_zoom)
-                .max_zoom(max_zoom)
-                .zooms(zoom_levels);
-
-            let tile_copier = TileCopier::new(src_file, dst_file, copy_opts)?;
+        Commands::Copy { opts } => {
+            let tile_copier = TileCopier::new(opts)?;
```
nyurik: why do we need to expose a public constructor, if it is only used to pass in options and run a `run` method on it? I think it would be cleaner to simply have a public method `copy_mbtiles_file(opts) -> Result<...>`, or have a method on the options itself (in a common builder pattern style) like `opts.execute() -> -> Result<...>`. If you need a struct internally - sure, just don't expose it as a public API.

The only difference with your approach is that there is some additional error-checking being done by both `new()` and `run()`, but they both return the same error type, so might as well combine them per above.

martin-mbtiles/src/tile_copier.rs
```diff
@@ -6,37 +6,87 @@ use crate::{MbtError, Mbtiles};
...
+    /// Minimum zoom level to copy
+    #[cfg_attr(feature = "cli", arg(long))]
     min_zoom: Option<u8>,
+    /// Maximum zoom level to copy
+    #[cfg_attr(feature = "cli", arg(long))]
     max_zoom: Option<u8>,
-    //self.bbox = bbox
-    verbose: bool,
+    /// List of zoom levels to copy; if provided, min-zoom and max-zoom will be ignored
+    #[cfg_attr(feature = "cli", arg(long, value_parser(clap::builder::ValueParser::new(HashSetValueParser{}))))]
```
nyurik: for readability, put the `clap::builder::ValueParser` in the use statement above

martin-mbtiles/src/tile_copier.rs
```diff
@@ -6,37 +6,87 @@ use crate::{MbtError, Mbtiles};
...
+impl clap::builder::TypedValueParser for HashSetValueParser {
+    type Value = HashSet<u8>;
+
+    fn parse_ref(
+        &self,
+        _cmd: &clap::Command,
+        _arg: Option<&clap::Arg>,
+        value: &OsStr,
+    ) -> Result<Self::Value, clap::Error> {
+        let mut result = HashSet::<u8>::new();
```
nyurik, line 49: not certain here - do you actually have to specify the hashset type?  I would think Rust should figure it out based on the `.insert()` below

martin-mbtiles/src/tile_copier.rs
```diff
@@ -6,37 +6,87 @@ use crate::{MbtError, Mbtiles};
...
+    pub fn new(src_filepath: PathBuf, dst_filepath: PathBuf) -> Self {
         Self {
-            zooms: HashSet::new(),
+            src_file: src_filepath,
+            zoom_levels: HashSet::new(),
+            force_simple: false,
             min_zoom: None,
             max_zoom: None,
-            verbose: false,
+            dst_file: dst_filepath,
```
nyurik: probably best to order them as in declaration (i think intellij can do that for you in some cases?). src and dst files separately look weird

martin-mbtiles/src/tile_copier.rs
```diff
@@ -6,37 +6,87 @@ use crate::{MbtError, Mbtiles};
...
-    pub fn zooms(mut self, zooms: Vec<u8>) -> Self {
-        for zoom in zooms {
-            self.zooms.insert(zoom);
+    pub fn force_simple(mut self, force_simple: bool) -> Self {
+        self.force_simple = force_simple;
+        self
+    }
+
+    pub fn zoom_levels(mut self, zoom_levels: Vec<u8>) -> Self {
+        for zoom in zoom_levels {
```
nyurik: you might be able to use `.extend(...)` method, but not certain

martin-mbtiles/src/tile_copier.rs
```diff
@@ -98,23 +139,35 @@ impl TileCopier {
...
-            query(row.as_str()).execute(&mut conn).await?;
-        }
+        if force_simple {
+            for statement in &["CREATE TABLE metadata (name text, value text);",
+                "CREATE TABLE tiles (zoom_level integer, tile_column integer, tile_row integer, tile_data blob);",
+                "CREATE UNIQUE INDEX name on metadata (name);",
+                "CREATE UNIQUE INDEX tile_index on tiles (zoom_level, tile_column, tile_row);"] {
+                query(statement).execute(&mut conn).await?;
+            }
+        } else {
```
nyurik, line 150: Lets add a comment for others to explain this moment:

```suggestion
        } else {
          // DB objects must be created in a specific order: tables, views, triggers, indexes.
          // If there are no triggers, in theory views and indexes can go in any order after tables.
```

In theory, triggers and indexes could be created in any order (i think), but trigger could rely on a view or a table.  I have not seen triggers ever defined on an mbtiles, so we could in theory just filter the type column to be `type IN ('table', 'view', 'index')`.

One more thing - I am not certain your order-by would work - ordered DESC, it would be `view, table, index` - and view cannot be created before the table i would think.

see https://www.sqlite.org/schematab.html

martin-mbtiles/src/tile_copier.rs
nyurik: I meant you can call them `src` and `dst` just like other tests, but this is is just a nit

nyurik (commented):
almost there! a few comments

martin-mbtiles/src/bin/main.rs
upsicleclown: I think the `TileCopier` should do the copying while the `TileCopierOptions` should just hold the options and do nothing else. The `TileCopier` object is what is being exposed to the user.

upsicleclown: Also, I think it makes more sense conceptually to create a `TileCopier` with options and _then_ have it `run` and do the copying as opposed to doing it all at once. If we did it all at once, it would seem like the struct is redundant and we could do it in a single function. The `new` function serves to construct the internal `src_mbtiles` struct so I'd prefer to leave it as is.

martin-mbtiles/src/tile_copier.rs
upsicleclown: I did. FYI I could also do that here: https://github.com/maplibre/martin/pull/723/files#diff-df5958e4bb210418bbc3f1ac707f8df93314ad138dfa40d911eb8b0e51b1fc8cR54-R59 but I avoided it because the necessary mapping would make it much less less readable

martin-mbtiles/src/bin/main.rs
nyurik: Per discussion - you can simplify it with 

```rust
    #[command(name = "copy")]
    Copy(TileCopierOptions),
```

martin-mbtiles/src/bin/main.rs
```diff
@@ -100,3 +72,144 @@ async fn meta_get_value(file: &Path, key: &str) -> Result<()> {
...
+    use clap::Parser;
+    use martin_mbtiles::TileCopierOptions;
+    use std::path::PathBuf;
+
+    #[test]
+    fn test_copy_no_arguments() {
+        assert_eq!(
+            Args::try_parse_from(["mbtiles", "copy"])
+                .unwrap_err()
+                .kind(),
```
nyurik, line 90: nice find!

nyurik (approved):
awesome, thank you!

PR #729, opened 2023-06-26
Add diff-with-file option to copy command

upsicleclown:
* Add ability to generate diff file by specifying `--diff-with-file` to the `copy` tool


justfile
```diff
@@ -218,6 +218,6 @@ git-pre-push: stop start
 prepare-sqlite:
     @if ! command -v cargo-sqlx &> /dev/null; then \
         echo "SQLX could not be found. Installing..." ;\
-        cargo install sqlx-cli --no-default-features --features sqlite,native-tls ;\
+        cargo install sqlx-cli --version 0.7.0-alpha.3 --no-default-features --features sqlite,native-tls ;\
```
nyurik: are we ok to use native-tls here, or should it also use rustls?  I am ok with rustls (one day i hope all our code can migrate to it), but we should double check if it will work ok with the certificates installed by the operating system.

martin-mbtiles/src/mbtiles_queries.rs
```diff
@@ -46,7 +47,15 @@ where
        ) AS is_valid;
 "#
     );
-    Ok(sql.fetch_one(&mut *conn).await?.is_valid == 1)
+
+    Ok(sql
+        .fetch_one(&mut *conn)
+        .await?
+        .is_valid
+        .ok_or(MbtError::SqlError(sqlx::Error::ColumnNotFound(
```
nyurik: I'm a bit uncertain why we need explicit error here? Does the new sqlite have some new error response? If not, and this is an option, can we simply assume that its false and treat it as such (i.e. `unwrap_or_default()`)

martin-mbtiles/src/mbtiles_queries.rs
```diff
@@ -76,5 +85,13 @@ where
        ) as is_valid;
 "#
     );
-    Ok(sql.fetch_one(&mut *conn).await?.is_valid == 1)
+
+    Ok(sql
+        .fetch_one(&mut *conn)
+        .await?
+        .is_valid
+        .ok_or(MbtError::SqlError(sqlx::Error::ColumnNotFound(
```
nyurik: same

martin-mbtiles/src/tile_copier.rs
```diff
@@ -28,9 +28,12 @@ pub struct TileCopierOptions {
     /// Maximum zoom level to copy
     #[cfg_attr(feature = "cli", arg(long, conflicts_with("zoom_levels")))]
     max_zoom: Option<u8>,
-    /// List of zoom levels to copy; if provided, min-zoom and max-zoom will be ignored
+    /// List of zoom levels to copy
     #[cfg_attr(feature = "cli", arg(long, value_parser(ValueParser::new(HashSetValueParser{})), default_value=""))]
     zoom_levels: HashSet<u8>,
+    /// If provided this file will be used to generate a diff with the source file
```
nyurik: How about this text?  I don't think we need "if provided" because these are options already?

```suggestion
    /// Compare source file with this file, and only copy non-identical tiles to destination
```

martin-mbtiles/src/tile_copier.rs
```diff
@@ -186,8 +191,39 @@ impl TileCopier {
...
+                            AND sourceDb.tiles.tile_row = newDb.tiles.tile_row
+                        WHERE (sourceDb.tiles.tile_data != newDb.tiles.tile_data
+                            OR sourceDb.tiles.tile_data ISNULL
+                            OR newDb.tiles.tile_data ISNULL)",
+            )
+            .await
+        } else {
+            self.run_query_with_options(
+                conn,
+                "INSERT INTO tiles SELECT * FROM sourceDb.tiles WHERE 1=1",
```
nyurik: please add a comment why you need `WHERE 1=1` (is it because additional clauses can be added later?).
Also, `WHERE TRUE` is cleaner - shows intent

nyurik (commented):
looks good, thanks!  we do need some more testing though - a test that shows that all corner cases are handled as expected:
* same tile data
* a tile is added
* a tile is deleted

We may, as a separate PR, also want the `mbtiles apply-diff src_file.mbtiles diff.mbtiles dst_file.mbtiles` - but no rush on this

nyurik:
P.S. There is a merge conflict - you may want to simply delete and re-create the lock file after you fix the cargo.toml file merge.

upsicleclown:
> looks good, thanks! we do need some more testing though - a test that shows that all corner cases are handled as expected:
> 
> * same tile data
> * a tile is added
> * a tile is deleted

@nyurik all of these test cases are addressed by diffing `geography-class-jpg.mbtiles` against the new file I created called `geography-class-jpg-modified.mbtiles`. 

`geography-class-jpg-modified.mbtiles` has 1 row added, 1 row removed and 1 row modified. This is why I assert that the diff has 3 `tiles` rows. I do realize this `assert_eq!(...,3)` isn't very clear, but I'm not sure how to make it more transparent while also not making it needlessly complex.

justfile
upsicleclown: I think I actually should leave it as `native-tls`; I accidentally changed it to `rustls` in `Cargo.toml` instead of changing it to `runtime-tokio-native-tls`. We can change it to `rustls` at some point but I think it should be a separate PR and not swallowed in this PR since it isn't actually necessary for the changes I want to make here.

upsicleclown:
The failing test is caused by the fact that the CI pipeline doesn't install `sqlite`. Sqlite is needed with these changes to be able to run the corresponding integration tests in `test.sh`. However, I have an upcoming change that makes it so that installing sqlite is not needed anyways: https://github.com/upsicleclown/martin/pull/2. 
So I would prefer merging those changes into this branch and then merging this branch, so that I do not have to make a fix and then inevitably undo it.

nyurik:
agree, lets comment out the apply test in the test.sh for now. Eventually i would still like to have it as part of the test suite - this way our documentation always stays correct (i.e. we can always guarantee that we can apply the delta directly with sqlite3)

tests/fixtures/apply_diff.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+SQL="ATTACH DATABASE @diffDbFilename AS diffDb;
+     DELETE FROM tiles WHERE (zoom_level, tile_column, tile_row) IN (SELECT zoom_level, tile_column, tile_row FROM diffDb.tiles WHERE tile_data ISNULL);
```
nyurik: classical SQL usually uses `IS NULL` (with a space), or `IS NOT NULL`. I am actually surprised sqlite supports the ones without spaces. I'm ok to keep it though if you want

nyurik (commented):
looks good, one minor nit (in addition to commenting out the call to sqlite3).  BTW, you may want to add a "TODO: " comment - to only re-enable it once sqlite3 is installed on windows CI (and possibly mac?)

nyurik (approved):
p.s. feel free to merge it once it passes the ci

tests/fixtures/apply_diff.sh
upsicleclown: `NOTNULL` works as well, I think I will keep it because it's still quite clear what is happening

martin-mbtiles/src/tile_copier.rs
```diff
@@ -186,8 +191,40 @@ impl TileCopier {
     }
 
     async fn copy_tile_tables(&self, conn: &mut SqliteConnection) -> MbtResult<()> {
-        self.run_query_with_options(conn, "INSERT INTO tiles SELECT * FROM sourceDb.tiles")
+        if let Some(diff_file) = &self.options.diff_with_file {
+            let diff_mbtiles = Mbtiles::new(diff_file)?;
+
+            let _ = &self.detect_type(&diff_mbtiles);
```
nyurik: this is a bug:  `detect_type` returns an awaitable, which you must await in order to let it detect the type. But instead you ignore the result, thus it will never even execute.  I have renamed detect_type in a new patch, but left it there because I am not certain why you have it - in theory the diff file shouldn't even exist yet, should it?

martin-mbtiles/src/tile_copier.rs
```diff
@@ -241,6 +280,14 @@ impl TileCopier {
 
         Ok(())
     }
+
+    async fn detect_type(&self, mbtiles: &Mbtiles) -> MbtResult<MbtType> {
```
nyurik: i moved this function out of the `impl` block because it doesn't need `self` (not certain why clippy hasn't complained)

martin-mbtiles/src/tile_copier.rs
upsicleclown: This file we are diffing against does exist and this line was meant to ensure that it was a vaild `mbtiles` type. There is a bug though, I should have added an await; thank you for updating!

PR #743, opened 2023-07-04
Upgrade sqlx to v0.7

nyurik:
This is mostly a noop, just updating to the new way of storing cached queries and a few other changes per https://github.com/launchbadge/sqlx/blob/main/CHANGELOG.md#070---2023-06-30

Note that this uses some of the code from #729 (thanks @upsicleclown !)

A minor unrelated change - a reformat of `martin/release.toml`


birkskyum:
I see a change between actix - tokio wrt the sqlx - are they interchangeable for us?

nyurik:
@birkskyum yes, actix uses tokio internally anyway.  Also, see this sqlx changelog comment:

> The runtime-actix-* features have been deleted. They were previously changed to be aliases of their runtime-tokio-* counterparts for backwards compatibility reasons, but their continued existence is misleading as SQLx has no special knowledge of Actix anymore.
To fix, simply replace the runtime-actix-* feature with its runtime-tokio-* equivalent.

birkskyum (approved):

PR #747, opened 2023-07-05
Add apply-diff command

upsicleclown:
Add command `apply-diff` to apply diff file generated from `copy` command


docs/src/tools.md
```diff
@@ -26,15 +26,10 @@ Copy command can also be used to compare two mbtiles files and generate a diff.
...
 ```
 
-The `diff_file.mbtiles` can be applied to the `src_file.mbtiles` elsewhere to avoid copying/transmitting the entire modified dataset.
-
-One way to apply the diff is to use the `sqlite3` command line tool directly. Here, we assume that the `src_file.mbtiles` is in the simple tables format, and that the `diff_file.mbtiles` is the output of the `mbtiles copy` command above. This SQL will delete all tiles from `src_file.mbtiles` that are set to `NULL` in `diff_file.mbtiles`, and then insert or update all new tiles from `diff_file.mbtiles` into `src_file.mbtiles`. The name of the diff file is passed as a query parameter to the sqlite3 command line tool, and then used in the SQL statements.
-
+### apply-diff
+Apply the diff file generated from `copy` command above to an mbtiles file. The diff file can be applied to the `src_file.mbtiles` elsewhere, to avoid copying/transmitting the entire modified dataset.
 ```shell
-sqlite3 src_file.mbtiles \
```
upsicleclown, line 34: This script should be preserved somewhere but not here, since the `apply-diff` command replaces it

nyurik: I would prefer to keep both of them here. This is documentation, so the more thorough it is, the better

martin-mbtiles/src/tile_copier.rs
```diff
@@ -290,6 +290,45 @@ async fn open_and_detect_type(mbtiles: &Mbtiles) -> MbtResult<MbtType> {
...
+) -> MbtResult<SqliteConnection> {
+    let src_mbtiles = Mbtiles::new(src_file)?;
+    let diff_mbtiles = Mbtiles::new(diff_file)?;
+
+    let opt = SqliteConnectOptions::new().filename(src_mbtiles.filepath());
+    let mut conn = SqliteConnection::connect_with(&opt).await?;
+    let src_type = src_mbtiles.detect_type(&mut conn).await?;
+
+    if src_type != MbtType::TileTables {
+        return Err(MbtError::InvalidDataFormat(
```
nyurik: we need a special error for this. The format is not invalid, it is simply not compatible with the given operation

nyurik (commented):
looks great, a few minor feedbacks

docs/src/tools.md
```diff
@@ -26,6 +26,18 @@ Copy command can also be used to compare two mbtiles files and generate a diff.
...
+One way to apply the diff is to use the `sqlite3` command line tool directly. Here, we assume that the `src_file.mbtiles` is in the simple tables format, and that the `diff_file.mbtiles` is the output of the `mbtiles copy` command above. This SQL will delete all tiles from `src_file.mbtiles` that are set to `NULL` in `diff_file.mbtiles`, and then insert or update all new tiles from `diff_file.mbtiles` into `src_file.mbtiles`. The name of the diff file is passed as a query parameter to the sqlite3 command line tool, and then used in the SQL statements.
+```shell
+sqlite3 src_file.mbtiles \
+  -bail \
+  -cmd ".parameter set @diffDbFilename diff_file.mbtiles" \
+  "ATTACH DATABASE @diffDbFilename AS diffDb;" \
+  "DELETE FROM tiles WHERE (zoom_level, tile_column, tile_row) IN (SELECT zoom_level, tile_column, tile_row FROM diffDb.tiles WHERE tile_data ISNULL);" \
+  "INSERT OR REPLACE INTO tiles (zoom_level, tile_column, tile_row, tile_data) SELECT * FROM diffDb.tiles WHERE tile_data NOTNULL;"
+```
+Another way to apply the diff is to use the below `apply-diff` command.
```
nyurik: thx for re-adding it - lets bring up the `apply-diff` at the top though - it is the preferred method

nyurik (approved):
one minor nit, and lets merge

PR #754, opened 2023-07-09
(breaking) Future-proof source catalog

nyurik:
Modify `/catalog` endpoint to return an object instead of a list.  This allows future expansion of the catalog schema, e.g. adding new types of data.

The new schema:

```yaml
{
  "tiles" {
    "function_zxy_query": {
      "name": "public.function_zxy_query",
      "content_type": "application/x-protobuf"
    },
    "points1": {
      "name": "public.points1.geom",
      "content_type": "image/webp"
    },
    ...
  },
}
```


stepankuzmin (approved):
👍

PR #755, opened 2023-07-09
Implement dynamic font support `/font/<name>/<start>-<end>`

nyurik:
This implements dynamic font protobuf generation, allowing users to request font ranges on the fly, and combining them in any order, e.g. `Font1,Font2,Font3`, same as with sprites and tiles

This is a first iteration, without any multithreading support. In theory, this could be done far faster by generating SDFs with multiple threads.

### Current process
* during init, figure out all glyphs available in each font, and store them as a bitset
* during request:
  * combine requested bitsets to figure out which glyph should come from which font file
  * load those glyphs from files (using a single instance of the freetype lib)
  * convert them to SDFs and package them into a protobuf


martin/src/fonts/mod.rs
```diff
@@ -0,0 +1,359 @@
...
+            let style = face.style_name();
+            if let Some(style) = &style {
+                name.push(' ');
+                name.push_str(style);
+            }
+            // Make sure font name has no slashes or commas, replacing them with spaces and de-duplicating spaces
+            name = name
+                .replace(['/', ','], " ")
+                .replace("  ", " ")
+                .replace("  ", " ");
```
lseelenbinder: So what if the path has `"    "`? 😄

lseelenbinder: Feels like perhaps a loop until the name is valid is a better approach?

martin/src/fonts/mod.rs
```diff
@@ -0,0 +1,359 @@
...
+                }
+                Entry::Vacant(v) => {
+                    let key = v.key();
+                    let Some((codepoints, count, ranges )) = get_available_codepoints(&mut face) else {
+                        warn!("Ignoring font source {key} from {} because it has no available glyphs", path.display());
+                        continue
+                    };
+
+                    let start = ranges.first().map(|(s, _)| *s).unwrap();
+                    let end = ranges.last().map(|(_, e)| *e).unwrap();
```
lseelenbinder: `break` / `return` instead of `.unwrap()`?

lseelenbinder (approved):
A few minor things / nits, but otherwise solid.

sharkAndshark:
This feature is what I looking forward 🤩

nyurik:
Want to make a test or some docs?

sharkAndshark:
> Want to make a test or some docs?

Sure :D

sharkAndshark (approved):

sharkAndshark:
Open Source Fonts Resources For Tests:
||License|Feature|
|----|----|----|
|[source-han-sans](https://github.com/adobe-fonts/source-han-sans)|[license](https://github.com/adobe-fonts/source-han-sans/blob/master/LICENSE.txt)|Japanese&Chinese|
|[RedHat Overpass](https://github.com/RedHatOfficial/Overpass)|[SIL1.1](https://github.com/RedHatOfficial/Overpass/blob/master/OFL.txt)|English fonts|


I believe we can use these two fonts for tests and my concern is the `font file size`, maybe we should download them in `CI` instead of include them in repo?

sharkAndshark:
Hi @nyurik   Have a lovely day   : )
For doc about font feature, I submitted a [PR here](https://github.com/nyurik/martin/pull/3).

sharkAndshark:
I add font test in test.sh in this [PR](https://github.com/nyurik/martin/pull/4)

PR #761, opened 2023-07-16
Add basic support for systemd

sharkAndshark:


nyurik:
thx @sharkAndshark , can you think of a way to automate testing of the debian packages?  Perhaps run some sort of `apt install` from a local file and test that `which martin` points to the right dir?

More specifically about this PR - Martin is clearly a service, so integrating it with a service infrastructure makes perfect sense, but it does not really make any sense without some configuration - so should it be disabled by default?  Not certain what the best options would be.

sharkAndshark:
Hi @nyurik. Something we should consider of per your comment:
## test
How and what to test a .deb file in our CI? There are some [references](https://github.com/kornelski/cargo-deb/blob/main/systemd.md#advanced-example) I'm still researching on.
## the behaviors of martin as a service
* should not automatically start afeter installation
* should have a configuration file by default
* ??

nyurik:
> ## test
> How and what to test a .deb file in our CI? There are some [references](https://github.com/kornelski/cargo-deb/blob/main/systemd.md#advanced-example) I'm still researching on.

Sorry, my mistake - I just realized you already do that in `Integration Tests (with debian package)` part, which looks great!

sharkAndshark:
Hi   @nyurik     :)
  By now, we could use it like this.
```shell
just package-deb
sudo dpkg -i target/debian/martin.deb
just start
```
## Call martin directily

```shell
martin postgres://postgres:postgres@localhost:5411/db
```
We have made a test about this: `Intergration test (with debian package) in CI.yml`
## Call martin as service
```shell
# enable martin automatically start 
systemctl enable martin
vim /usr/local/etc/martin/config.yaml
...
systemctl start martin
systemctl status martin
```
Do you think we need a test for calling marin as a service?

nyurik (approved):
looks good, thanks!

nyurik:
P.S. We could add a simple check if a specific file exists in CI after debian package install - up to you

PR #770, opened 2023-07-17
Demo: Migrate to yarn

nyurik:
`npm install` is having some weird issues, while `yarn` seem to work well. Also, bumped almost all dependencies.


birkskyum (approved):
Will this run, or is an extra initial step to install yarn needed in the Dockerfile?

nyurik:
@birkskyum apparently yarn is already included in the docker node image.  I updated it to the latest just in case.

PR #778, opened 2023-07-21
Add ability to copy to existing mbtiles files

upsicleclown:
* Allow the `dst` in `mbtiles copy <src> <dst>` command to already contain data, merging tilesets
* Add `mbtiles copy --on-duplicate (override|ignore|abort)`

### DIFF
--- a/martin-mbtiles/src/errors.rs
+++ b/martin-mbtiles/src/errors.rs
@@ -28,6 +28,12 @@ pub enum MbtError {
 
     #[error("The destination file {0} is non-empty")]
     NonEmptyTargetFile(PathBuf),
+
+    #[error("The file {0} is does not have the required uniqueness constraint")]
+    NoUniquenessConstraint(String),
+
+    #[error("Unexpected duplicate values found when copying")]
+    DuplicateValues(),
 }
 
 pub type MbtResult<T> = Result<T, MbtError>;
--- a/martin-mbtiles/src/errors.rs
+++ b/martin-mbtiles/src/errors.rs
@@ -33,7 +33,7 @@ pub enum MbtError {
     NoUniquenessConstraint(String),
 
     #[error("Unexpected duplicate tiles found when copying")]
-    DuplicateValues(),
+    DuplicateValues,
 }
 
 pub type MbtResult<T> = Result<T, MbtError>;
--- a/martin-mbtiles/src/errors.rs
+++ b/martin-mbtiles/src/errors.rs
@@ -32,6 +32,9 @@ pub enum MbtError {
     #[error("The file {0} does not have the required uniqueness constraint")]
     NoUniquenessConstraint(String),
 
+    #[error("Could not copy MBTiles file: {reason}")]
+    UnsupportedCopyOperation { reason: String },
+
     #[error("Unexpected duplicate tiles found when copying")]
     DuplicateValues,
 }
--- a/martin-mbtiles/src/errors.rs
+++ b/martin-mbtiles/src/errors.rs
@@ -29,10 +29,10 @@ pub enum MbtError {
     #[error("The destination file {0} is non-empty")]
     NonEmptyTargetFile(PathBuf),
 
-    #[error("The file {0} is does not have the required uniqueness constraint")]
+    #[error("The file {0} does not have the required uniqueness constraint")]
     NoUniquenessConstraint(String),
 
-    #[error("Unexpected duplicate values found when copying")]
+    #[error("Unexpected duplicate tiles found when copying")]
     DuplicateValues(),
 }
 
--- a/martin-mbtiles/src/mbtiles.rs
+++ b/martin-mbtiles/src/mbtiles.rs
@@ -2,6 +2,7 @@
 
 extern crate core;
 
+use std::collections::HashSet;
 use std::ffi::OsStr;
 use std::fmt::Display;
 use std::path::Path;
@@ -11,7 +12,7 @@ use futures::TryStreamExt;
 use log::{debug, info, warn};
 use martin_tile_utils::{Format, TileInfo};
 use serde_json::{Value as JSONValue, Value};
-use sqlx::{query, SqliteExecutor};
+use sqlx::{query, Row, SqliteExecutor};
 use tilejson::{tilejson, Bounds, Center, TileJSON};
 
 use crate::errors::{MbtError, MbtResult};
@@ -26,7 +27,7 @@ pub struct Metadata {
     pub json: Option<JSONValue>,
 }
 
-#[derive(Debug, PartialEq)]
+#[derive(Debug, Clone, PartialEq)]
 pub enum MbtType {
     TileTables,
     DeDuplicated,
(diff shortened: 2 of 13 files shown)


martin-mbtiles/src/errors.rs
```diff
@@ -24,6 +24,12 @@ pub enum MbtError {
 
     #[error("The destination file {0} is non-empty")]
     NonEmptyTargetFile(PathBuf),
+
+    #[error("The file {0} is does not have the required uniqueness constraint")]
```
nyurik:
```suggestion
    #[error("The file {0} does not have the required uniqueness constraint")]
```

martin-mbtiles/src/errors.rs
```diff
@@ -24,6 +24,12 @@ pub enum MbtError {
...
+    NoUniquenessConstraint(String),
+
+    #[error("Unexpected duplicate values found when copying")]
```
nyurik:
```suggestion
    #[error("Unexpected duplicate tiles found when copying")]
```

martin-mbtiles/src/tile_copier.rs
```diff
@@ -146,42 +163,64 @@ impl TileCopier {
...
-        if force_simple {
-            for statement in &["CREATE TABLE metadata (name text, value text);",
-                "CREATE TABLE tiles (zoom_level integer, tile_column integer, tile_row integer, tile_data blob);",
-                "CREATE UNIQUE INDEX name on metadata (name);",
-                "CREATE UNIQUE INDEX tile_index on tiles (zoom_level, tile_column, tile_row);"] {
-                query(statement).execute(&mut conn).await?;
-            }
+        if is_empty {
+            if force_simple {
+                for statement in &["CREATE TABLE metadata (name text, value text);",
```
nyurik, line 174: why not use the syntax as described in https://www.sqlitetutorial.net/sqlite-primary-key/ ?

martin-mbtiles/src/tile_copier.rs
```diff
@@ -222,25 +261,41 @@ impl TileCopier {
             self.run_query_with_options(
                 conn,
                 // Allows for adding clauses to query using "AND"
-                "INSERT INTO tiles SELECT * FROM sourceDb.tiles WHERE TRUE",
+                &format!(
+                    "INSERT {} INTO tiles SELECT * FROM sourceDb.tiles WHERE TRUE",
+                    match &self.options.on_duplicate {
+                        CopyDuplicateMode::Override => "OR REPLACE",
+                        CopyDuplicateMode::Ignore => "OR IGNORE",
+                        _ => "",
```
nyurik: don't use `_` when not required - better be explicit for all available options. This way when a new value is added to enum, it will stop compiling and you will know something needs to be handled

martin-mbtiles/src/tile_copier.rs
```diff
@@ -222,25 +261,41 @@ impl TileCopier {
...
     }
 
     async fn copy_deduplicated(&self, conn: &mut SqliteConnection) -> MbtResult<()> {
-        query("INSERT INTO map SELECT * FROM sourceDb.map")
-            .execute(&mut *conn)
-            .await?;
+        let on_duplicate_sql = match &self.options.on_duplicate {
+            CopyDuplicateMode::Override => "OR REPLACE",
+            CopyDuplicateMode::Ignore => "OR IGNORE",
+            _ => "",
```
nyurik: same

nyurik (commented):
a few minor feedbacks, will review more later

martin-mbtiles/src/tile_copier.rs
upsicleclown: After our conversation yesterday I looked into it more and realized that `CREATE UNIQUE INDEX ....` doesn't just create an index with non-duplicated keys like I thought, but it _enforces_ a unique constraint on those columns that are part of the key. A unique index also automatically gets generated on the primary key. So that the unique constraint enforced by a unique index is enough for our case; `(zoom_level, tile_column, tile_row)` doesn't have to be the primary key.

nyurik: unique index is enough if we are processing someone else's mbtiles, but it is not the preferred method, so if we ourselves create a new schema, we should use primary key.

Primary keys differ from unique index because they usually are also the clustered index - in other words (loosely), the actual data is stored in the order of the primary key, which (in theory) should reduce space requirements and allow access optimizations.

Lastly, we MAY want to use [WITHOUT ROWID](https://www.sqlite.org/withoutrowid.html) - I'll need to test performance and size difference, but in theory it should be more optimal.  See 3rd section of the link for details.

upsicleclown: My mistake, I thought you were referring to the `check_uniqueness` part of the code. When creating a new table, I agree that we can simply create a primary key. I can change this but I don't think it is within the scope of this PR since this is for copying existing mbtiles. I will create a small follow-up PR.

martin-mbtiles/src/mbtiles.rs
```diff
@@ -280,13 +281,60 @@ impl Mbtiles {
...
+    where
+        for<'e> &'e mut T: SqliteExecutor<'e>,
+    {
+        let table_name = match mbt_type {
+            MbtType::TileTables => "tiles",
+            MbtType::DeDuplicated => "map",
+        };
+
+        let mut unique_idx_cols = HashSet::new();
+        let query_string = format!(
```
nyurik: apparently this can actually be done in a "safer" way (the whole point of sqlx as oppose to calling sqlite directly) - https://docs.rs/sqlx-conditional-queries/0.1.3/sqlx_conditional_queries/macro.conditional_query_as.html

Should we try to use that approach? Could be a separate PR though

upsicleclown: This `conditional_query_as!` macro seems to use the `query_as!` macro. This query is a `PRAGMA` query which doesn't work with sqlx macros.

martin-mbtiles/src/mbtiles.rs
```diff
@@ -280,13 +281,64 @@ impl Mbtiles {
...
+        let index_query =
+            format!("SELECT name FROM pragma_index_list('{table_name}') WHERE [unique] = 1;");
+        let indexes = query(&index_query).fetch_all(&mut *conn).await?;
+
+        // Ensure there is some index on tiles that has a unique constraint on (zoom_level, tile_row, tile_column)
+        for index in indexes {
+            let mut unique_idx_cols = HashSet::new();
+            let key_column_query = format!(
+                "SELECT DISTINCT name FROM pragma_index_info('{}')",
+                index.get::<String, _>("name")
```
nyurik: this is a minor SQL injection attack - using user-provided data (column name in many databases can be arbitrary, possibly in sqlite too, or could be in the future).  Please use `query().bind()` or some other similar approach to pass in the query argument.

nyurik (commented):
one minor fix, and should be good to go :)

nyurik (approved):
awesome, thank you!!!

PR #787, opened 2023-07-28
A few minor mbtiles-copier cleanups

nyurik:


martin-mbtiles/src/tile_copier.rs
```diff
@@ -382,29 +398,21 @@ pub async fn apply_mbtiles_diff(
...
-    SELECT * FROM diffDb.tiles WHERE tile_data NOTNULL;",
-    )
-    .execute(&mut conn)
-    .await?;
+        (SELECT zoom_level, tile_column, tile_row FROM diffDb.tiles WHERE tile_data ISNULL)";
+    query(sql).execute(&mut conn).await?;
+
+    let sql = "INSERT OR REPLACE INTO tiles (zoom_level, tile_column, tile_row, tile_data) 
+        SELECT * FROM diffDb.tiles WHERE tile_data NOTNULL";
+    query(sql).execute(&mut conn).await?;
```
upsicleclown, line 409: nit: I don't think this change is necessary; I think it would be cleaner to just leave it as it was

martin-mbtiles/src/tile_copier.rs
```diff
@@ -514,61 +510,40 @@ mod tests {
...
 
-        assert!(
-            query("SELECT 1 FROM sqlite_schema WHERE type='table' AND tbl_name='tiles';")
-                .fetch_optional(&mut dst_conn)
-                .await
-                .unwrap()
-                .is_some()
-        );
+        let sql = "SELECT 1 FROM sqlite_schema WHERE type='table' AND tbl_name='tiles'";
+        assert!(query(sql)
```
upsicleclown, line 522: I would prefer to keep the strings within the assert statements here

martin-mbtiles/src/tile_copier.rs
```diff
@@ -632,11 +607,13 @@ mod tests {
         let dst =
             PathBuf::from("file:copy_to_existing_ignore_mode_mem_db?mode=memory&cache=shared");
 
+        // TODO: need a detailed comment explaining that due to `let _var = copy`,
```
upsicleclown, line 610: Why leave this as a `TODO`?

martin-mbtiles/src/tile_copier.rs
```diff
@@ -632,11 +607,13 @@ mod tests {
...
+        //   connection remains open and the second copy command reuses the same in-memory file.
         let _dst_conn = copy_mbtiles_file(TileCopierOptions::new(dst_file.clone(), dst.clone()))
             .await
             .unwrap();
 
-        let mut dst_conn = copy_mbtiles_file(
+        let mut conn = copy_mbtiles_file(
```
upsicleclown, line 616: The different files we are using here get confusing; I would like to keep the name `dst_conn` to have a clear distinction and the reader does not have to think about it.

PR #790, opened 2023-07-30
adds support for id_column in tables for auto_publish

### LINKED ISSUE #682: Specifying an `id` column with `auto_publish`
Auto-discovered PG tables may have some columns that should be exposed as a feature ID rather than just being a regular feature property.  I propose to add an `id_column` param as shown below.

```yaml
postgres:
  connection_string: ${DATABASE_URL}
  auto_publish:
      tables:
        # Set a custom source ID based on the table name
        id_format: 'table.{schema}.{table}.{column}'
        # A table column to use as the feature ID
        # If a table has no column with this name, `id_column` will not be set for that table.
        # We could even allow a list of strings here - the first found column will be treated as a feature ID.
        id_column: 'id'
```

### Deliverables
  - [ ] allow `id_column` string param to specify the default id column if it exists.  If it does not exist, log an info message. If the column exists, but it is NOT an integer, log a warning.  If ID column is found, it will NOT be exported as a property, only as the ID.
  - [ ] Optional: possibly allow `id_column` to be a list of strings, so if a table has any of the specified columns, use it.
  - [ ] Optional: consider renaming `id_format` into `source_id_format`, with backwards compatibility

Discussed in https://github.com/maplibre/martin/discussions/680

Binabh:
Resolves #682 

- [x] Get id_column string from config.yaml and use for id column
- [x] Support for list of strings
- [x] Add info/warnings if column is not there or is of wrong type
- [x] if column for the feature ID is found, remove it from properties (see inline comment)
- [x] cleanup logging messages
- [x] need more tests to catch other edge cases


martin/src/pg/table_source.rs
```diff
@@ -193,6 +203,38 @@ FROM {schema}.{table};
         .and_then(|p| polygon_to_bbox(&p)))
 }
 
+async fn calc_id_column(
```
nyurik: I don't think you should query DB for the available IDs. `TableInfo.properties` already contains all available columns and their types - you can simply look through that HashMap and see if any of the requested ID columns are present.

martin/src/pg/configurator.rs
```diff
@@ -265,6 +276,21 @@ fn new_auto_publish(config: &PgConfig, is_function: bool) -> Option<PgBuilderPub
...
+                Some(c) => match c {
+                    BoolOrObject::Bool(_) => None,
+                    BoolOrObject::Object(o) => o.id_column,
+                },
+                None => None,
+            },
+        },
+        None => None,
+    }
+}
```
nyurik: I think this would be more readable (technically it can be one big destructuring statement instead of two, but this is more readable I think. I am not certain if this func should return an optional vector of strings or not -- ` OneOrMany` has a number of methods to convert it to a slice, etc.

```rust
fn get_column_id(config: &PgConfig) -> Option<Vec<String>> {
    if let Some(BoolOrObject::Object(v)) = &config.auto_publish {
        if let Some(BoolOrObject::Object(v)) = &v.tables {
            return v.id_column.as_ref().map(|v| v.iter().cloned().collect());
        }
    }
    None
}
```

nyurik (commented):
@Binabh thank you so much for working on this!  Left a few suggestions inline

sharkAndshark:
Oops.🥲I accidentally changed it from a draft state to ready for review. I was on a train

Binabh:
It may take few days for me to complete this. I have converted to draft.

Binabh:
@nyurik Thank you for your guidance. I have made the changes as per your suggestions. Also, I have tested manually with few variations in config files and this works good. Please let me know if anything. I will modify accordingly.

nyurik:
I made a few minor suggestions as a pr to your branch. Feel free to merge or comment on it. I'll make a note thorough read tomorrow

Binabh:
@nyurik I have looked and merged the changes.

Binabh:
I see that the renaming is handled at serde serialization level. That is much cleaner approach. If I understand this correctly now the code that I wrote wrote supporting new name could be discarded right?

nyurik:
I will rebase it shortly -- that PR was the result of looking deeply into your code. I should be able to push it to your branch in a bit

nyurik:
I pushed a few minor changes and merged with the main branch.  My next todo (i'll fix it) is to make it use the `find_kv_ignore_case` (i may need to do some more base code refactoring)

nyurik:
I made a number of significant changes to the logic of this PR as part of the https://github.com/Binabh/martin/pull/2 (see notes).

TODO (if merged):
* make tests pass
* possibly add a few more tests?
* check that the right properties are exposed, e.g. if it found an ID, that ID should no longer be shown in the properties list (need a test for this for sure)

Binabh:
I have merged.Can I give TODO list a try? It may take some days though.

nyurik:
Sorry, i just saw your comment - i made a few base improvements in https://github.com/maplibre/martin/pull/795 -- that gives some base for testing (I discovered a few minor issues in it).

TODO list is still relevant - i will merge the main into your PR once i merge 795, and then won't touch it until you message me :)

tests/expected/given_config.yaml
```diff
@@ -28,6 +29,7 @@ postgres:
       table: auto_table
       srid: 4326
       geometry_column: geom
+      id_column: feat_id
```
nyurik, line 34: the `feat_id: int4` line in the `properties` section on line 43 should not be there -- it indicates that the same value from the `feat_id` column will be included twice - once as the feature ID, and again as a regular propery.

docs/src/config-file.md
```diff
@@ -67,6 +67,10 @@ postgres:
       source_id_format: 'table.{schema}.{table}.{column}'
       # Add more schemas to the ones listed above
       from_schemas: my_other_schema
+      # A table column to use as the feature ID
+      # If a table has no column with this name, `id_column` will not be set for that table.
+      # If a list of strings is given, the first found column will be treated as a feature ID.
+      id_column: feature_id
```
nyurik: This should probably be `id_columns`?  Note that both are allowed because of the Serde's `alias` attribute.

martin/src/pg/configurator.rs
```diff
@@ -232,12 +234,54 @@ impl PgBuilder {
...
+    let Some(props) = inf.properties.as_mut() else { return };
+    let Some(try_columns) = &auto_tables.id_columns else { return };
+
+    for key in try_columns.iter() {
+        let (column, typ) = if let Some(typ) = props.get(key) {
+            (key, typ)
+        } else {
+            match find_kv_ignore_case(props, key) {
+                Ok(Some(result)) => {
+                    info!("For source {id} from table {}.{}, feature ID '{key}' was not found, but found '{result}' instead.", inf.schema, inf.table);
```
nyurik: the logging here and below might need to be cleaned up to make it consistent with the `pg::utilities` logging, and include the source ID in all, etc.

martin/src/pg/configurator.rs
```diff
@@ -232,12 +234,54 @@ impl PgBuilder {
...
+                Ok(None) => continue,
+                Err(multiple) => {
+                    error!("Unable to set feature ID from table {}.{} because `{key}` has no exact match and more than one potential matches: {}", inf.schema, inf.table, multiple.join(", "));
+                    continue;
+                }
+            }
+        };
+        // ID column can be any integer type as defined in
+        // https://github.com/postgis/postgis/blob/559c95d85564fb74fa9e3b7eafb74851810610da/postgis/mvt.c#L387C4-L387C66
+        if typ != "int4" && typ != "int8" && typ != "int2" {
```
nyurik, line 261: need to double check if PG can ever return `int` or some other types here?  Likely not, just a thought.

tests/config.yaml
```diff
@@ -22,6 +22,7 @@ postgres:
   auto_publish:
     tables:
       from_schemas: autodetect
+      id_column: feat_id
```
nyurik:
```suggestion
      id_columns: feat_id
```

nyurik (commented):
a few todos inline

Binabh:
> Sorry, i just saw your comment - i made a few base improvements in #795 -- that gives some base for testing (I discovered a few minor issues in it).
> 
> TODO list is still relevant - i will merge the main into your PR once i merge 795, and then won't touch it until you message me :)

No worries, and thank you for this learning opportunity. I am learning a lot form your commits, commit messages and detailed comments. I will try these TODOs this weekend.
And surely, I will let you know if anything comes up.

nyurik:
@Binabh let me know if you need any help with this PR, seems like it needs just the tiny push to get ready for merge

martin/src/pg/configurator.rs
Binabh: According to : https://www.postgresql.org/docs/current/datatype-numeric.html#DATATYPE-INT We have covered for all integer types.

Binabh:
@nyurik I have addressed almost all TODOs. Let me know if any improvements are needed.

martin/src/pg/configurator.rs
```diff
@@ -232,12 +234,57 @@ impl PgBuilder {
...
+            warn!("Unable to use column `{key}` in table {}.{} as a tile feature ID because it has a non-integer type `{typ}`.", inf.schema, inf.table);
+            continue;
+        }
+
+        inf.id_column = Some(column.to_string());
+        let mut final_props = props.clone();
```
Binabh, line 267: @nyurik as you said earlier that we should avoid clone whenever possible. can this clone be avoided?

nyurik: it is more of a "don't clone unless you have to" :)

I agree that here it would be better here without the clone, but the issue is that you have a mutable ref to a hashmap - `props`, plus you have a readonly ref to one of the keys in that hashmap - `column` - and Rust won't let you have both readonly and mutable refs to the same object at the same time.  If you can figure out a way around that, sure :)

nyurik:
Looks awesome!  I will go over it tonight one last time before merging. I think it would be really good to add one or two more tests, but up to you

martin/src/pg/configurator.rs
Binabh: Yeah, I also searched a little and found that this was only way to do it. So, for now let's resolve it as it is.

Binabh:
@nyurik I have added a test to check for feature id of type bigint. This is my first try writing test in Rust project. So, please let me know about it. I could revert commit if that test is unnecessary.

nyurik (approved):
thanks!!!

Issue #756, opened 2023-07-09
Add small geometry filtering for PG table sources


nyurik:
It would be great to limit low-zoom tiles from including invisible geometries -- the ones that are too small at the low zoom.

## Implementation ideas
* Add a CLI flag, something like `--pg-filter-small` or `--filter-small` or `--visible-only` or ...? (naming is hard)
* Add a [table config](https://maplibre.org/martin/config-file.html) file setting, e.g.  `hide_small_geometries` (we may want naming consistent with CLI though)
  * this could be a boolean for now, but eventually could even introduce some percentage "per pixel", where the size of the pixel is computed from zoom and extent settings
* Should we offer any kind of "clustering" support, e.g. if there is an area with lots of buildings, each building is too small to be seen, but if they are close enough, they would still be visible as a cluster.  OMT did something similar [here](https://github.com/openmaptiles/openmaptiles/blob/master/layers/building/update_building.sql)

> [!NOTE]
> [ST_AsMVTGeom](https://postgis.net/docs/ST_AsMVTGeom.html) does already drop geometries smaller than the resolution as discussed [here](https://github.com/maplibre/martin/pull/792#issuecomment-1666424076). This would thus not be nessesary.

## Discussed in https://github.com/maplibre/martin/discussions/731

<div type='discussions-op-text'>

<sup>Originally posted by **sharkAndshark** June 27, 2023</sup>
Hi. It's just some random ideas.
For automatically discovered data sources, Martin will fetch a large amount of data when the zoom level is low, such as 0 to 4. At these zoom levels, the tiles are large and contain an overload of information on the map.Is it possible to remove polygons or linestrings whose area / length is not bigger than a single pixel in size at the current zoom level? 

| scale | Marin table source | Martin function source with filtering |
| ---------------------- | ---------------------  | ---------------------- |
|1:250000 | ![image](https://github.com/maplibre/martin/assets/42399654/09bd880c-b761-419b-9172-8b50ace7e925)| ![image](https://github.com/maplibre/martin/assets/42399654/061baabb-334f-4f22-93e5-25743678917b)|
|1:100000|![image](https://github.com/maplibre/martin/assets/42399654/e78370e0-52d3-48bd-a7ad-14fa497dc4e3)|![image](https://github.com/maplibre/martin/assets/42399654/b52af32c-fbb4-4406-811c-cd461c0bc543)|

The function source
```sql
CREATE 
[...]

PR #792, opened 2023-08-02
Allow ignore  small geometries 

sharkAndshark:
Implements #756

### DIFF
--- a/martin/src/args/pg.rs
+++ b/martin/src/args/pg.rs
@@ -24,6 +24,9 @@ pub struct PgArgs {
     /// Limit the number of features in a tile from a PG table source.
     #[arg(short, long)]
     pub max_feature_count: Option<usize>,
+    /// Hide geometries which can not fullfill one single pixel
+    #[arg(short, long)]
+    pub exclude_small_geometries: Option<bool>,
 }
 
 impl PgArgs {
@@ -48,6 +51,7 @@ impl PgArgs {
                     None
                 },
                 max_feature_count: self.max_feature_count,
+                exclude_small_geometries: self.exclude_small_geometries,
                 pool_size: self.pool_size,
                 auto_publish: None,
                 tables: None,
@@ -79,6 +83,16 @@ impl PgArgs {
             info!("Overriding maximum feature count to {} on all Postgres connections because of a CLI parameter", self.max_feature_count.unwrap());
             pg_config.iter_mut().for_each(|c| {
                 c.max_feature_count = self.max_feature_count;
+            });
+        }
+
+        if self.exclude_small_geometries.is_some() {
+            info!(
+                "Overriding exclude small geometries to {} because of a CLI parameter",
+                self.exclude_small_geometries.unwrap()
+            );
+            pg_config.iter_mut().for_each(|c| {
+                c.exclude_small_geometries = self.exclude_small_geometries;
             });
         }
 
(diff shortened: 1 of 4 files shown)


nyurik:
just to clarify - do you plan to make this PR remove polygons that are too small for a given zoom level, or do you want to deal with merging "polygon dust" as discussed in #756? Please update the PR description. Thx!

sharkAndshark:
Hi @nyurik  :)  
I'm still thinking of these things and it would be better to build the talk based on code and test so I make this draft PR. Sorry for this late description.
The basic idea is something like impose a  random number to make the cluster of small buildings still can be seen on screen.

```SQL
...
WHERE 
ST_TRANSFORM("geom",3857) && ST_TileEnvelope(3, 1, 1) 
AND  ( ST_AREA ( ST_TRANSFORM ( geom, 3857 ) ) >= ( ( 40075016.68 / ( 2 ^ z * 512 ) ) ^ 2 ) )   
AND RANDOM() > 0.3  
-- the random threshold should be from config maybe
...
```
Yes,It's for #756   
let's talk and test more on this PR :)

nyurik (commented):
looking at the current SQL changes - I looked at [ST_AsMVTGeom](https://postgis.net/docs/ST_AsMVTGeom.html) -- and it has this text:

> From 3.0, Wagyu can be chosen at configure time to clip and validate MVT polygons. This library is faster and produces more correct results than the GEOS default, but it might drop small polygons.

Which means that exactly the same filtering is already being done by SQL?

Note that this is different from merging tiny chunks

sharkAndshark:
I think you are right @nyurik   
The [code in st_asmvtgeom related to this PR](https://github.com/postgis/postgis/blob/af3e63085419434590f2d52963ca89105e52492f/postgis/lwgeom_out_mvt.c#L42) is here:
```c
/* If possible, peek into the bounding box before deserializing it to discard small geometries
	 * We don't check COLLECTIONTYPE since that might be a collection of points */
	if (type == LINETYPE || type == POLYGONTYPE || type == MULTILINETYPE || type == MULTIPOLYGONTYPE)
	{
		GBOX gserialized_box;
		/* We only apply the optimization if the bounding box is available */
		if (gserialized_fast_gbox_p(geom_in, &gserialized_box) == LW_SUCCESS)
		{
			/* Shortcut to drop geometries smaller than the resolution */
			double geom_width = gserialized_box.xmax - gserialized_box.xmin;
			double geom_height = gserialized_box.ymax - gserialized_box.ymin;

			/* We use half of the square height and width as limit: We use this
			 * and not area so it works properly with lines */
			double bounds_width = ((bounds->xmax - bounds->xmin) / extent) / 2.0;
			double bounds_height = ((bounds->ymax - bounds->ymin) / extent) / 2.0;
			if (geom_width < bounds_width && geom_height < bounds_height)
			{
				PG_RETURN_NULL();
			}
		}
	}
```

CommanderStorm:
> ST_AsMVTGeom [...] might drop small polygons

Given that this PR only adresses geometires too small for a given zoom level and this is already handled in ST_AsMVTGeom, I think we need to "go back to the drawing board" on this one.

=> lets re-group in https://github.com/maplibre/martin/issues/756 on the exact configuration options/feature we should support (we can always reopen this PR if it comes to that ^^)

PR #793, opened 2023-08-02
feat(metrics): add actix-web metrics exporter

### LINKED ISSUE #773: feat: OpenMetrics / Prometheus / metrics endpoint (monitoring)
Hello, I will be interested to use a monitoring endpoint on martin. Of course I'm ready to work on it.

The need is to have some sort of monitoring endpoint to retrieve some metrics.
I would be interested for:
- metrics linked to the web server
- metrics linked to the connexion martin has to postgres and some performance insight
- metrics linked to the duration of the tile generating process
- any others that should be discussed

I'm thinking this can be behind a feature flag in order to give the user a choice. 

What are your take on this?

Implementation details:

we have the choice between two prometheus instrumentation libraries
- https://github.com/prometheus/client_rust
- https://github.com/tikv/rust-prometheus (seems to be more popular)

the https://github.com/tikv/rust-prometheus library is compatible with the actix web instrumentation library https://github.com/nlopes/actix-web-prom

lefuturiste:
For a starter, we can only include the basics metrics from the [actix web prom crate](https://github.com/nlopes/actix-web-prom).

This PR includes #1905 - merge that one first

Resolves #773


martin/src/srv/server.rs
```diff
@@ -407,6 +413,17 @@ pub fn router(cfg: &mut web::ServiceConfig) {
 
 /// Create a new initialized Actix `App` instance together with the listening address.
 pub fn new_server(config: SrvConfig, all_sources: AllSources) -> crate::Result<(Server, String)> {
+    // metrics setup
+    let labels = HashMap::new();
+
+    // labels.insert("label_foo".to_string(), "value_barbarbbabbmiles".to_string());
+    let prometheus = PrometheusMetricsBuilder::new("martin")
+        .endpoint("/metrics")
```
nyurik: if we start handling `/metrics`, we would have to declare this as a reserved keyword (we might be missing docs in a martin book for this). One workaround perhaps would be to add everything "magical" to the `/_/metrics` style prefixes because `_` is not a valid source ID i think.

https://github.com/maplibre/martin/blob/a2acd1732ecdd9ded7ce5fade91518119790e9a6/martin/src/srv/server.rs#L32-L37

nyurik (commented):
this looks fine, and I read your discussion in https://github.com/nlopes/actix-web-prom/issues/72 - but still not certain I understood the resolution of the problem -- would it make sense for the middleware to have an extra API to override path param to convert from `/srcA,srcB/1/1/1` to `srcA,srcB` (but at the same time ignore the tilejson request `/srcA,srcB`?  Should this be some magical attribute?

martin/src/srv/server.rs
lefuturiste: > if we start handling `/metrics`, we would have to declare this as a reserved keyword (we might be missing docs in a martin book for this). One workaround perhaps would be to add everything "magical" to the `/_/metrics` style prefixes because `_` is not a valid source ID i think.
> 
> https://github.com/maplibre/martin/blob/a2acd1732ecdd9ded7ce5fade91518119790e9a6/martin/src/srv/server.rs#L32-L37

Right, as you wish, I'm fine with both solutions (the reserved keyword and the /_/metrics workaround). We just have to keep in mind that `/metrics` is the standard path and the default, so the second solution would mean a little bit more config on the scraper side.

lefuturiste:
> this looks fine, and I read your discussion in [nlopes/actix-web-prom#72](https://github.com/nlopes/actix-web-prom/issues/72) - but still not certain I understood the resolution of the problem -- would it make sense for the middleware to have an extra API to override path param to convert from `/srcA,srcB/1/1/1` to `srcA,srcB` (but at the same time ignore the tilejson request `/srcA,srcB`? Should this be some magical attribute?

I don't really what you are saying, the issue is with the variance of `/srcA,srcB`, could also be `/srcB,srcA` (I suppose)
So probably the best thing to do is to have a magical callback that is defined on the maplibre side, that take the request parameter and request path and return the path for metrics. So the business logic stay downstream and not defined in the lib.
But, I don't recall that we settled on a solution yet.

martin/src/srv/server.rs
nyurik: good point about conventions (actually, didn't we talk about it elsewhere?), i'm ok to add `/metrics` to reserved then

nyurik:
> > this looks fine, and I read your discussion in [nlopes/actix-web-prom#72](https://github.com/nlopes/actix-web-prom/issues/72) - but still not certain I understood the resolution of the problem -- would it make sense for the middleware to have an extra API to override path param to convert from `/srcA,srcB/1/1/1` to `srcA,srcB` (but at the same time ignore the tilejson request `/srcA,srcB`? Should this be some magical attribute?
> 
> I don't really what you are saying, the issue is with the variance of `/srcA,srcB`, could also be `/srcB,srcA` (I suppose) So probably the best thing to do is to have a magical callback that is defined on the maplibre side, that take the request parameter and request path and return the path for metrics. So the business logic stay downstream and not defined in the lib. But, I don't recall that we settled on a solution yet.

Are you sure its a big deal to collapse `/srcA,srcB` and `/srcB,srcA` into one? I wouldn't worry too much about this tbh... Worst case - can later add some sort of a collapser, e.g. `/srcA+` would mean everything with `srcA` as the first layer, plus any number of other layers mixed in.  The collapser could even be configurable to how many sources to keep, e.g. 0 -- `/+` means all multi-sourced requests, or 2 would create `/srcA,srcB` and `/srcA,srcB+` metrics

lefuturiste:
> > > this looks fine, and I read your discussion in [nlopes/actix-web-prom#72](https://github.com/nlopes/actix-web-prom/issues/72) - but still not certain I understood the resolution of the problem -- would it make sense for the middleware to have an extra API to override path param to convert from `/srcA,srcB/1/1/1` to `srcA,srcB` (but at the same time ignore the tilejson request `/srcA,srcB`? Should this be some magical attribute?
> > 
> > 
> > I don't really what you are saying, the issue is with the variance of `/srcA,srcB`, could also be `/srcB,srcA` (I suppose) So probably the best thing to do is to have a magical callback that is defined on the maplibre side, that take the request parameter and request path and return the path for metrics. So the business logic stay downstream and not defined in the lib. But, I don't recall that we settled on a solution yet.
> 
> Are you sure its a big deal to collapse `/srcA,srcB` and `/srcB,srcA` into one? I wouldn't worry too much about this tbh... Worst case - can later add some sort of a collapser, e.g. `/srcA+` would mean everything with `srcA` as the first layer, plus any number of other layers mixed in. The collapser could even be configurable to how many sources to keep, e.g. 0 -- `/+` means all multi-sourced requests, or 2 would create `/srcA,srcB` and `/srcA,srcB+` metrics

yes, you right, for now, let's not worry about this issue, will deal with that later.

nyurik:
@lefuturiste i added metrics to reserved in #802 , let me know if you need any help with this PR

nyurik:
@lefuturiste any updates? thx

lefuturiste:
> @lefuturiste any updates? thx

Hi! 
I've updated upstream PR for this, because I really want to have path cardinality (idk really why an obsession may be 😄 ), https://github.com/nlopes/actix-web-prom/pull/73
I awaiting response from actix-web-prom dev.

lefuturiste:
while i'm awaiting for a response from the actix-web-prom developer I will explore the instrumentation of postgresql connexion.

lefuturiste:
The PR at https://github.com/nlopes/actix-web-prom/pull/73 was merged today. So we can go forward from there.

nyurik:
exciting, thanks, let me know if you need any help

lefuturiste:
update : I'm still working on it, 
I'm waiting for another PR upstream to be merged
- https://github.com/nlopes/actix-web-prom/pull/87

It's ok, I'm patient. In 1 year may be this will be merged xD

nyurik:
@CommanderStorm you might combine these efforts with #1720

CommanderStorm:
@nyurik 
Not quite shure what you mean by combinding..
logging and metrics are a bit different..

@lefuturiste 
FYI: I would be very much fine with using a different repo/branch (i.e. your PR), as long as we pin martin to a commit of said PR.
We don't have to wait/be patient.

tests/test.sh
```diff
@@ -97,6 +97,15 @@ test_jsn() {
   $CURL "$URL" | jq --sort-keys -e 'walk(if type == "number" then .+0.0 else . end)' > "$FILENAME"
 }
 
+test_metrics() {
+  FILENAME="$TEST_OUT_DIR/$1"
+  URL="$MARTIN_URL/metrics"
+
+  echo "Testing $(basename "$FILENAME") from $URL"
+  $CURL "$URL" | sed -E 's/^(martin_.*?) [\.0-9]+$/\1 NUMBER/g' > "$FILENAME"
+  $CURL --compressed "$URL" | sed -E 's/^(martin_.*?) [\.0-9]+$/\1 NUMBER/g' > "$FILENAME"
```
CommanderStorm: I added a few testcases to verify this works and came upon this:
- if I request WITH compression, a the tile `metrics`
- if I request WITHOUT compression, I get `/metrics`

Will investigate further later in the week

docs/src/using.md
```diff
@@ -16,6 +16,7 @@ Martin data is available via the HTTP `GET` endpoints:
 | `/font/{font1},…,{fontN}/{start}-{end}`  | [Composite Font source](sources-fonts.md)      |
 | `/style/{style}`                         | [Style source](sources-styles.md)              |
 | `/health`                                | Martin server health check: returns 200 `OK`   |
+| `/metrics`                                | Martin server metrics: returns Prometheus metrics |
```
CommanderStorm:
```suggestion
| `/health`                                | Martin server health check: returns 200 `OK`   |
| `/metrics`                               | Martin server Prometheus metrics               |
```

martin/src/srv/server.rs
```diff
@@ -188,8 +200,9 @@ pub fn new_server(config: SrvConfig, state: ServerState) -> MartinResult<(Server
             cors_middleware.is_some(),
             cors_middleware.unwrap_or_default(),
         ))
-        .wrap(NormalizePath::new(TrailingSlash::MergeOnly))
+        .wrap(prometheus.clone())
         .wrap(Logger::default())
+        .wrap(NormalizePath::new(TrailingSlash::MergeOnly))
```
CommanderStorm: Where NormalizePath is in this path is a bit arbitrairy.

The others have a clean order as we likely want all requests (even /metrics) logged and since we also want metrics from cors-rejections.

I think putting it at the start of the request chain (=> the end of .wrap) makes most sence.
This way we log, cors and generate metrics for the same thing

CommanderStorm:
@lefuturiste to get around the compression-matching bug, I moved the metrics route to `/_/metrics`
According to my research, this should be accepible too..
Would you agree?

This way we can merge now instead of when said PR might get reviewed.
Or am I missing something of how the other PR works?

(I would like to have metrics for #1897 ^^)

tests/test.sh
```diff
@@ -98,6 +98,20 @@ test_jsn() {
...
+  URL="$MARTIN_URL/_/metrics"
+
+  echo "Testing $1 from $URL"
+  $CURL --dump-header  "$FILENAME.headers" "$URL" | sed -E 's/^(martin_.*?) [\.0-9]+$/\1 NUMBER/g' > "$FILENAME"
+  clean_headers_dump "$FILENAME.headers"
+  $CURL --dump-header  "$FILENAME.gzip.headers" --compressed "$URL" | sed -E 's/^(martin_.*?) [\.0-9]+$/\1 NUMBER/g' > "$FILENAME.gzip"
+  clean_headers_dump "$FILENAME.gzip.headers"
+  # due to slight timing differences, these might be slightly different
+  sed -i 's/^content-length: [\.0-9]+$/content-length: NUMBER/g' "$FILENAME.headers"
+  sed -i 's/^content-length: [\.0-9]+$/content-length: NUMBER/g' "$FILENAME.gzip.headers"
```
CommanderStorm: I think CI needs this, lets see

```suggestion
  sed -iE 's/^content-length: [\.0-9]+$/content-length: NUMBER/g' "$FILENAME.headers"
  sed -iE 's/^content-length: [\.0-9]+$/content-length: NUMBER/g' "$FILENAME.gzip.headers"
```

tests/test.sh
```diff
@@ -98,6 +98,20 @@ test_jsn() {
...
+  URL="$MARTIN_URL/_/metrics"
+
+  echo "Testing $1 from $URL"
+  $CURL --dump-header  "$FILENAME.headers" "$URL" | sed -E 's/^(martin_.*?) [\.0-9]+$/\1 NUMBER/g' > "$FILENAME"
+  clean_headers_dump "$FILENAME.headers"
+  $CURL --dump-header  "$FILENAME.gzip.headers" --compressed "$URL" | sed -E 's/^(martin_.*?) [\.0-9]+$/\1 NUMBER/g' > "$FILENAME.gzip"
+  clean_headers_dump "$FILENAME.gzip.headers"
+  # due to slight timing differences, these might be slightly different
+  sed -iE 's/^content-length: [\.0-9]+$/content-length: NUMBER/g' "$FILENAME.headers"
+  sed -iE 's/^content-length: [\.0-9]+$/content-length: NUMBER/g' "$FILENAME.gzip.headers"
```
CommanderStorm: Fun. `-iE` does not do what I thought it does:

> -i[SUFFIX], --in-place[=SUFFIX]
> edit files in place (makes backup if SUFFIX supplied)


```suggestion
  sed --regexp-extended --in-place 's/^content-length: [\.0-9]+$/content-length: NUMBER/g' "$FILENAME.headers"
  sed --regexp-extended --in-place 's/^content-length: [\.0-9]+$/content-length: NUMBER/g' "$FILENAME.gzip.headers"
```

docs/src/config-file.md
```diff
@@ -44,6 +44,12 @@ preferred_encoding: gzip
 # Enable or disable Martin web UI. At the moment, only allows `enable-for-all` which enables the web UI for all connections. This may be undesirable in a production environment. [default: disable]
 web_ui: disable
 
+# More advanced monitoring montoring options
+observability:
+  # Additional static metric labels to be added to every metric reported under `/_/metrics`
+  # example: { environment: production, server: martin }
+  additional_metric_labels: {}
```
nyurik: if the whole system is `metrics` (plural), perhaps this field should be too?  Naming is hard, and `observability.additional_metric_labels` somehow just catches too much attention... ideas:

* `labels`
* `extra_labels`
* `additional_labels`
* `add_labels`
* `add_metrics_labels` (plural)
* ...?

martin/src/srv/config.rs
```diff
@@ -17,6 +18,16 @@ pub struct SrvConfig {
     #[cfg(feature = "webui")]
     pub web_ui: Option<crate::args::WebUiMode>,
     pub cors: Option<CorsConfig>,
+    /// More advanced monitoring montoring options
```
nyurik:
```suggestion
    /// Advanced monitoring options
```

tests/expected/configured/metrics_1.gzip
```diff
@@ -0,0 +1,169 @@
+# HELP martin_http_requests_duration_seconds HTTP request duration in seconds for all requests
```
nyurik: why is there a .gzip file here?

nyurik (commented):
lets stabilize on the config naming, and otherwise looks great!

One other thing:  the .gzip filenames are really confusing. Lets use more descriptive `.txt` or `.log` extensions perhaps? and if it indicates compression, we could name these files as `metrics_1_gzip.txt` and `metrics_1_gzip.headers.txt` or the like?

tests/expected/configured/metrics_1.gzip
CommanderStorm: To ensure that this does not suffer from https://github.com/maplibre/martin/pull/793#discussion_r2025778445

nyurik: at first i thought this was a real .gzip (.gz) file - and got confused. Lets have more recognizable extensions to keep things simpler. If its a compressed log - lets use `.compressed.txt` or similar

martin/src/srv/config.rs
```diff
@@ -17,6 +18,16 @@ pub struct SrvConfig {
...
     pub web_ui: Option<crate::args::WebUiMode>,
     pub cors: Option<CorsConfig>,
+    /// Advanced monitoring options
+    pub observability: Option<ObservabilityConfig>,
+}
+
+/// More advanced monitoring montoring options
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Default)]
+pub struct ObservabilityConfig {
+    /// Additional metric labels to be added to every metric reported under `/metrics`
```
nyurik:
```suggestion
    /// Extra labels to be added to every metric reported under `/_/metrics`
```

docs/src/config-file.md
CommanderStorm: How about the following.

I am not a fan of deleting `metric` from the name.
Gramar wise, I feel (I don't know why) like metrics in this context is singular.

```yml
observability:
  # Configure the metrics reported under `/_/metrics`
  metrics:
    # Additional, static labels for every metric
    #
    # Example: `{ environment: production, server: martin }`
    additional_labels: {}
```

docs/src/config-file.md
```diff
@@ -44,6 +44,12 @@ preferred_encoding: gzip
 # Enable or disable Martin web UI. At the moment, only allows `enable-for-all` which enables the web UI for all connections. This may be undesirable in a production environment. [default: disable]
 web_ui: disable
 
+# More advanced monitoring montoring options
```
```suggestion
# Advanced monitoring options
```

docs/src/config-file.md
```diff
@@ -44,6 +44,12 @@ preferred_encoding: gzip
 # Enable or disable Martin web UI. At the moment, only allows `enable-for-all` which enables the web UI for all connections. This may be undesirable in a production environment. [default: disable]
 web_ui: disable
 
+# Advanced monitoring options
+observability:
+  # Additional static metric labels to be added to every metric reported under `/_/metrics`
+  # example: { environment: production, server: martin }
+  additional_metric_labels: {}
```
nyurik: How about this?

```suggestion
# Advanced monitoring options
observability:
  # Configure metrics reported under `/_/metrics`
  metrics:
    # Add these labels to every metric
    # Example: `{ env: prod, server: martin }`
    add_labels: {}
```

tests/expected/configured/metrics_1.gzip
CommanderStorm: it is not compressed. It is just curled WITH compression.

But fair compressed is better because it might not be gzip

nyurik: perhaps `.with_compression.txt` ?

docs/src/config-file.md
CommanderStorm: Much better

tests/expected/configured/metrics_1.gzip
CommanderStorm: yea. I can myself getting confused by this in a few years.
Lets be extra explicit: `fetched_with_compression`

nyurik:
I did a minor cleanup pass, making this into a conditional feature, and fixing a few unrelated bugs that we had with conditional compilation.  See #1905

This PR is good to go with one possible issue - what's the reason/need for this code?

```rust
app.wrap(middleware::NormalizePath::new(middleware::TrailingSlash::MergeOnly)
```

nyurik (approved):
awesome work!

Issue #822, opened 2023-08-18
Set TileJSON layer metadata for function sources in config


nick-romano:
For full parity with table sources, it would be helpful to have metadata available for function sources within the tilejson response.

Some relevant issues:
- https://github.com/maplibre/martin/issues/749
- https://github.com/maplibre/martin/issues/277

nick-romano:
seems like this could be difficult if the function accepts query params...

nyurik:
We definitely need to be able to set various per-source settings in the config, especially settings that are not auto-discoverable from the source itself.  Let's sketch out some config structures first - my proposal is to allow direct tilejson modification:

```yaml
postgres:
  functions:
    my_func:
      tilejson:
        description: my magical tile source
        # Here goes any fields that should be set in the tilejson,
        # overriding anything that's already being set there automatically.
        # Note that setting field to `null`, e.g. `description: ~` should remove it from output.
```

Note that the above could be used for all other sources, not just Postgres functions (i.e. tables, mbtiles, pmtiles). Moreover, it could be allowed in the `auto_publish` section too.  In auto-publish, it could use the same placeholders as in the `source_id_format`, e.g. `{schema}, {table}, {column}`

nyurik:
We could also introduce an optional tilejson function that gets called on `/source`. It would support two function types:
* `FUNCTION get_meta(source_id TEXT, host TEXT) RETURNS JSON`
* `FUNCTION get_meta(source_id TEXT, host TEXT, query_params JSON) RETURNS JSON`

Passing `source_id` to the function allows the same function to be used for multiple sources.  The `host` param is taken from the `Host` header of the rquest - allows tilejson to be re-configured for another domain request as needed (this is done automatically too)

Moreover, it could be used with a table source too.  The returned JSON has the same content as the `tilejson` config key above.  In the config, it could be specified as `tilejson_function: get_meta`.  Having both `tilejson` and `tilejson_function` would merge `autogenerated + tilejson + function call`.

sharkAndshark:
Could we move layers metadata of function from martin config  to comment on function in PostgreSQL,
Like Let martin discover layer infos from  comment of function.

CREATE DEMO FUNCTION

```SQL
CREATE OR REPLACE FUNCTION osm_trees(z INTEGER, x INTEGER,y INTEGER)
RETURNS INTEGER AS $$
	select z+x+y;
$$ 
LANGUAGE SQL;
```

COMMENT TILE LAYERS INFO ON FUNCTION
[...]

PR #829, opened 2023-08-21
Retrieve function source description from comments

sharkAndshark:
If a PostgreSQL function has an SQL comment, it will try to parse as JSON and use its values to override the auto-generated TileJSON.  It is recommended to use this form when creating comments to ensure valid JSON values.

```sql
DO $do$ BEGIN
    EXECUTE 'COMMENT ON FUNCTION YOUR_FUNCTION (ARG1_TYPE,ARG2_TYPE,..ARGN_TYPE) IS $tj$' || $$
    {
      "description": "description override",
      ...
    }
    $$::json || '$tj$';
END $do$;
```

Partially implements #822

### DIFF
--- a/martin/src/pg/function_source.rs
+++ b/martin/src/pg/function_source.rs
@@ -2,7 +2,7 @@ use std::collections::HashMap;
 use std::fmt::Write;
 use std::iter::zip;
 
-use log::warn;
+use log::{debug, warn};
 use postgres_protocol::escape::escape_identifier;
 use serde_json::Value;
 
@@ -19,6 +19,7 @@ use crate::pg::Result;
 /// Panics if the built-in query returns unexpected results.
 pub async fn query_available_function(pool: &PgPool) -> Result<SqlFuncInfoMapMap> {
     let mut res = SqlFuncInfoMapMap::new();
+
     pool.get()
         .await?
         .query(include_str!("scripts/query_available_function.sql"), &[])
@@ -33,6 +34,18 @@ pub async fn query_available_function(pool: &PgPool) -> Result<SqlFuncInfoMapMap
             let output_record_names = jsonb_to_vec(&row.get("output_record_names"));
             let input_types = jsonb_to_vec(&row.get("input_types")).expect("Can't get input types");
             let input_names = jsonb_to_vec(&row.get("input_names")).expect("Can't get input names");
+            let tilejson = if let Some(text) = row.get("description") {
+                match serde_json::from_str::<Value>(text) {
+                    Ok(v) => Some(v),
+                    Err(e) => {
+                        warn!("Unable to deserialize SQL comment on {schema}.{function} as tilejson, a default description will be used: {e}");
+                        None
+                    }
+                }
+            } else {
+                debug!("Unable to find a SQL comment on {schema}.{function}, a default function description will be used");
+                None
+            };
 
             assert!(input_types.len() >= 3 && input_types.len() <= 4);
             assert_eq!(input_types.len(), input_names.len());
@@ -50,7 +63,7 @@ pub async fn query_available_function(pool: &PgPool) -> Result<SqlFuncInfoMapMap
 
             // Query preparation: the schema and function can't be part of a prepared query, so they
             // need to be escaped by hand.
-            // However schema and function comes from database introspection so they shall be safe.
+            // However schema and function comes from database introspection so they should be safe.
             let mut query = String::new();
             query.push_str(&escape_identifier(&schema));
             query.push('.');
@@ -95,7 +108,7 @@ pub async fn query_available_function(pool: &PgPool) -> Result<SqlFuncInfoMapMap
                                 input_types.join(", ")
                             ),
                         ),
-                        FunctionInfo::new(schema, function),
+                        FunctionInfo::new(schema, function, tilejson)
                     ),
                 )
             {
--- a/martin/src/pg/scripts/query_available_function.sql
+++ b/martin/src/pg/scripts/query_available_function.sql
@@ -29,17 +29,28 @@ WITH inputs AS (
                 jsonb_agg(parameter_name::text ORDER BY ordinal_position) as out_names
          FROM information_schema.parameters
          WHERE parameter_mode = 'OUT'
-         GROUP BY specific_name)
+         GROUP BY specific_name),
+    comments AS (
+        SELECT
+	        n.nspname AS schema,
+	        P.proname AS name,
+	        obj_description(P.oid, 'pg_proc') AS description
+	    FROM
+	    pg_proc P
+        JOIN pg_namespace n ON P.pronamespace = n.oid
+    )
 SELECT routines.specific_schema AS schema,
        routines.routine_name    AS name,
        routines.data_type       AS output_type,
        outputs.out_params       AS output_record_types,
        out_names                AS output_record_names,
        inputs.input_types       AS input_types,
-       inputs.input_names       AS input_names
+       inputs.input_names       AS input_names,
+       comments.description     AS  description
 FROM information_schema.routines
          JOIN inputs ON routines.specific_name = inputs.specific_name
          LEFT JOIN outputs ON routines.specific_name = outputs.specific_name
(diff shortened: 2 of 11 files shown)


martin/src/pg/function_source.rs
```diff
@@ -13,6 +15,14 @@ use crate::pg::pool::PgPool;
 use crate::pg::PgError::PostgresError;
 use crate::pg::Result;
 
+#[derive(Serialize, Deserialize, Debug)]
```
nyurik: use `TileJSON` object here - no point in creating a new struct.  Simply parse `TileJSON` from the comment

martin/src/pg/scripts/query_available_function.sql
```diff
@@ -29,17 +29,28 @@ WITH inputs AS (
                 jsonb_agg(parameter_name::text ORDER BY ordinal_position) as out_names
          FROM information_schema.parameters
          WHERE parameter_mode = 'OUT'
-         GROUP BY specific_name)
+         GROUP BY specific_name),
+    comments AS (
+        SELECT
+	        n.nspname AS schema,
+	        P.proname AS name,
+	        obj_description ( P.oid ) AS description
```
nyurik:
```suggestion
	        obj_description(P.oid, 'pg_proc') AS description
```

tests/fixtures/functions/function_Mixed_Name.sql
```diff
@@ -13,3 +13,25 @@ RETURNS TABLE("mVt" bytea, key text) AS $$
       WHERE "Geom" && ST_Transform(ST_TileEnvelope("Z", x, y), 4326)
   ) as tile WHERE geom IS NOT NULL) src
 $$ LANGUAGE SQL IMMUTABLE STRICT PARALLEL SAFE;
+
+COMMENT ON FUNCTION "MixedCase"."function_Mixed_Name" (int4,INT4,INT4) IS $tilejson$
```
nyurik: See if you can modify these somehow to auto-verified with `'{...}'::json` approach instead of a unverified string

sharkAndshark: Weired. PostgreSQL complained syntax error. Need a little bit research on.
```SQL
COMMENT ON FUNCTION osm_trees(INTEGER,INTEGER,INTEGER) IS ('{"a":1}'::json::TEXT)
```

sharkAndshark: Seems it only receive literal value?  [doc](https://www.postgresql.org/docs/15/sql-comment.html)

nyurik: @sharkAndshark use this method:

```sql
DO $do$ BEGIN
    EXECUTE 'COMMENT ON FUNCTION public.function_null (INT4,INT4,INT4) IS $tj$' || $$
    {
      "minzoom": 0,
      "maxzoom": 18,
      "bounds": [-180, -85, 180, 85],
      "vector_layers": []
    }
    $$::json || '$tj$';
END $do$;
```

sharkAndshark:
Hi @nyurik  Could you give me some suggestions for fixing the errors in SQL files? I've tried many times.
It's weired. The `setup-database` step in CI failed.

Update:
Seems it's due to the case sentive issue.But I think I shouldn't double quoting every sql file.Fix it and do the method you suggested later

nyurik:
> Hi @nyurik Could you give me some suggestions for fixing the errors in SQL files? I've tried many times. It's weired. The `setup-database` step in CI failed.
> 
> Update: Seems it's due to the case sentive issue.But I think I shouldn't double quoting every sql file.Fix it and do the method you suggested later

**usually** you don't need to quote any identifiers (namespaces, tables, functions and their arguments, etc) in SQL, as long as they don't use any special characters or reserved keywords. BUT, in some very rare cases, someone may have those weird cases. So in our testing, we must make sure Martin supports it properly. And for that, I have a few tests like `"MixedCase"` namespaces, as well as functions, etc.

Note that the generated SQL automatically always uses quotes identifiers to avoid ambiguity.

nyurik:
I reworked your patch in https://github.com/sharkAndshark/martin/pull/50 -- please take a look and merge if agree.

Basic idea:
* tilejson from a comment is stored as a simple json value
* when creating the final tilejson, it uses generated tilejson as the "base", and applies all fields from the comment's tilejson as a "patch" (override) using [json-patch](https://crates.io/crates/json-patch) crate

P.S. I don't think you should add tilejson to ALL unit tests. Instead, add it to just a few, with some variation of which fields are added, and possibly even deleted (set them to null).

sharkAndshark:
Agreed.I realized the tests part and I'm in a conference,merge it later.

sharkAndshark:
To achieve isolation, it is recommended to have one unit test for each observable behavior. Therefore, it might be necessary to remove some comments from the function. However, there is a concern that this level of granularity might be excessive?(And we can not see them on `debug.html` then) @nyurik I'm looking forward your guide.

nyurik:
@sharkAndshark lets keep things simple :)  I think we need to cover these cases. And I think you can do this with just two function comments.

### Test 1
* set on `"MixedCase"."function_Mixed_Name"` - to test comments work on weird named functions
* override a single field:  set the `description` in a function comment -- this should override the whatever description is set from the full ID automatically
* add a new field:  set `vector_layers` field with an array of whatever stuff the function is producing

### Test 2
* set comment on `public.function_zxy_query` -- normal name, different param combination
* delet a field - in another function comment test, set `description` to `null`. No need to set any other fields at all
* add some field that is NOT part of tilejson, e.g. `"myvalue": { "foo": "bar" }`  (or whatever similar). This will test that we can add arbitrary complex content

Just doing these two tests, you should be good enough.

sharkAndshark:
I hope the doc is not too bad as English is not my first language and I'm still learning it.

tests/fixtures/functions/function_zxy_query.sql
```diff
@@ -16,3 +16,14 @@ BEGIN
   RETURN mvt;
 END
 $$ LANGUAGE plpgsql IMMUTABLE STRICT PARALLEL SAFE;
+
+DO $do$ BEGIN
+    EXECUTE 'COMMENT ON FUNCTION public.function_zxy_query (INT4, INT4, INT4, JSON) IS $tj$' || $$
+    {
+      "tilejson": "3.0.0",
```
nyurik: Don't set the version or tiles fields here and above

sharkAndshark: I tried before, It's required for `tilejson` crate.

martin/src/pg/function_source.rs
```diff
@@ -33,6 +34,18 @@ pub async fn query_available_function(pool: &PgPool) -> Result<SqlFuncInfoMapMap
             let output_record_names = jsonb_to_vec(&row.get("output_record_names"));
             let input_types = jsonb_to_vec(&row.get("input_types")).expect("Can't get input types");
             let input_names = jsonb_to_vec(&row.get("input_names")).expect("Can't get input names");
+            let tilejson = if let Some(text) = row.get("description") {
+                match serde_json::from_str::<Value>(text) {
```
nyurik, line 38: parsing now works because it now uses `from_str::<Value>` instead of `from_str::<TileJSON>`.  Value allows any valid json.

nyurik (approved):
great work, thanks!

martin/src/pg/function_source.rs
sharkAndshark: Yeah. hehe. I realized that lol. Cool. ✅

nyurik:
Thanks @sharkAndshark , i may do another round of docs cleanup later, but this is good to merge for now.  Do you perhaps want to do the same thing for tables?  I suspect it is possible to add comments to them too.

sharkAndshark:
Yeah sure. I'd love to do that. A issue ticket for discussion first maybe? @nyurik

nyurik:
technically that's the same issue #822 issue - although it is more of an umbrella issue to cover all tilejson modifications by the user.  Eventually I think we should just add an extra field to the yaml config, and that field will work the same way as the function comment

Issue #835, opened 2023-08-23
Better support for different kinds of hashes in mbtiles 


upsicleclown:
* Add support for validating/generating hashes that are not `MD5`
* Add ability to regenerate hashes when copying files that have an existing hash column
cc: @nyurik

nyurik:
Currently, we have:
* de-normalized mbtiles files (i.e. a single map table with `z,x,y,tile` columns). I have not seen these in the wild, but this is what the mbtiles spec originally specified
* normalized mbtiles files - a two table setup with a view to join them.  The joining key is usually an MD5 hash. These files are the most common.

The tile is usually generated by the tile provider (e.g. a tilelive source). That source typically gzips it, and may provide a `key` - something usually used for HTTP E-Tag and mbtiles joining key.

In theory, it is possible for the hash to be:
* non-MD5 hash - any string could work
* the hash to be created **before** the gzip compression (in case the compression is used)

nyurik:
I have updated my [sqlite-hashes crate](https://github.com/nyurik/sqlite-hashes) - it now supports `md5(value)`, `md5(value1, value2, ...)`, and `md5_concat(value, ...)` usage (for various hash functions)

upsicleclown:
If the mbtiles file contains a hashed column, which hash function was used should be stored in the `metadata` table

PR #851, opened 2023-08-30
Add integrity and global hash check to validation

upsicleclown:


martin-mbtiles/src/mbtiles.rs
```diff
@@ -378,32 +388,102 @@ impl Mbtiles {
...
-                "SELECT * FROM tiles_with_hash WHERE tile_hash!=hex(md5(tile_data)) LIMIT 1;"
+        let select_from = match mbttype {
+            Flat => {
+                println!("Cannot generate global hash, no hash column in flat table format. Skipping global_hash generation...");
+                return Ok(());
             }
-            MbtType::Normalized => {
-                "SELECT * FROM images WHERE tile_id!=hex(md5(tile_data)) LIMIT 1;"
+            MbtType::FlatWithHash => "SELECT 'global_hash', hex(md5_concat(tile_hash)) FROM tiles_with_hash ORDER BY zoom_level, tile_column, tile_row;",
+            MbtType::Normalized => "SELECT 'global_hash', hex(md5_concat(images.tile_id)) FROM images JOIN map ON images.tile_id=map.tile_id ORDER BY zoom_level, tile_column, tile_row;"
```
nyurik: I am not sure you need to join two tables here. A quicker query `SELECT 'global_hash', hex(md5_concat(tile_id)) FROM images ORDER BY zoom_level, tile_column, tile_row;` would give you the same result. That said, you may want to validate that all foreign keys (`tile_id`) are valid - exist in both the images and map tables. And this could be part of the FULL validation. But it should be available even if the global hash is not requested.

For now, I think its better to do the simpler hash (without join), but we should add a full normalized table validation to check if there are `tile_id` in `map` that do not exist in `images`, or if there are any `tile_id` in images that are not being used by `map`.  TBH, I am not certain if this should be done with two querie, or if some sort of an outer join query could do this efficiently?

martin-mbtiles/src/mbtiles.rs
```diff
@@ -378,32 +388,102 @@ impl Mbtiles {
...
+
+        }.to_string();
+
+        let rusqlite_conn = RusqliteConnection::open(Path::new(&self.filepath()))?;
+        register_md5_function(&rusqlite_conn)?;
+        rusqlite_conn.execute(
+            format!("INSERT OR REPLACE INTO metadata(name, value) {select_from}").as_str(),
```
nyurik: i think it would be better to pass `global_hash` as a parameter here, and replace it above with a `?`. This way it is only declared once. Besides, I am still thinking of a better name... Perhaps we should name it `md5sum` or `all_tiles_md5` or ... -- point being that we may want to mention the hashing type in its name

martin-mbtiles/src/mbtiles.rs
```diff
@@ -378,32 +388,102 @@ impl Mbtiles {
...
+        for<'e> &'e mut T: SqliteExecutor<'e>,
+    {
+        // SQLite Integrity check
+        if "ok"
+            != match integrity_check {
+                IntegrityCheck::Full => query("PRAGMA integrity_check;")
+                    .fetch_one(&mut *conn)
+                    .await?
+                    .get::<String, _>(0),
+                IntegrityCheck::Quick => query("PRAGMA integrity_check;")
```
nyurik:
```suggestion
                IntegrityCheck::Quick => query("PRAGMA quick_check;")
```

martin-mbtiles/src/mbtiles.rs
```diff
@@ -378,32 +388,102 @@ impl Mbtiles {
...
+                    .fetch_one(&mut *conn)
+                    .await?
+                    .get::<String, _>(0),
+                IntegrityCheck::Off => "ok".to_string(),
             }
+        {
+            return Err(FailedIntegrityCheck(self.filepath().to_string()));
```
nyurik: I think it would be useful to store whatever the query returns (probably a string?) inside this error too. Will help debugging.

martin-mbtiles/src/mbtiles.rs
```diff
@@ -378,32 +388,102 @@ impl Mbtiles {
...
         let rusqlite_conn = RusqliteConnection::open(Path::new(self.filepath()))?;
         register_md5_function(&rusqlite_conn)?;
 
+        // Global hash check
+        if self
+            .get_metadata_value(&mut *conn, "global_hash")
+            .await?
+            .is_none()
+        {
+            println!(
```
nyurik: if the global hash is missing, its an error. We do not want to rely on silent failure. If used programmatically, noone will ever see these print statements, but they will catch non-zero program exit codes

martin-mbtiles/src/mbtiles.rs
```diff
@@ -378,32 +388,102 @@ impl Mbtiles {
...
+                "No value for 'global_hash' key found in metadata, skipping global hash validation step..."
+            );
+        } else {
+            let sql = if mbttype == MbtType::FlatWithHash {
```
nyurik: seems messy - first you query a value from the metadata table, but then you discard it and instead query it again in a more complex query.  Plus you have identical query more than once. Might as well just do something like this:

```rust
if let Some(hash) =  self.get_metadata_value(&mut *conn, "global_hash").await? {
  // run the same query as above - to compute the hash
  // validate against the hash var
} else {
  // error because no hash is available (unless it is passed via a param?)
}

File: martin-mbtiles/src/mbtiles.rs [hunk 7]
```diff
@@ -378,32 +388,102 @@ impl Mbtiles {
...
+                .to_string();
+
+            if rusqlite_conn.prepare(&sql)?.exists(())? {
+                return Err(InvalidTileData(self.filepath().to_string()));
+            }
+        }
+
+        // Per-tile hash check
+        let sql = if mbttype == MbtType::FlatWithHash {
+            "SELECT * FROM tiles_with_hash WHERE tile_hash!=hex(md5(tile_data)) LIMIT 1;"
```
nyurik:
```suggestion
            "SELECT 1 FROM tiles_with_hash WHERE tile_hash != hex(md5(tile_data)) LIMIT 1;"
```

martin-mbtiles/src/mbtiles.rs
```diff
@@ -378,32 +388,102 @@ impl Mbtiles {
...
+        } else {
+            "SELECT * FROM images WHERE tile_id!=hex(md5(tile_data)) LIMIT 1;"
```
nyurik:
```suggestion
            "SELECT 1 FROM images WHERE tile_id != hex(md5(tile_data)) LIMIT 1;"
```

martin-mbtiles/src/mbtiles.rs
```diff
@@ -378,32 +388,102 @@ impl Mbtiles {
...
+        }
+        .to_string();
```
nyurik: you don't need this as a string because you don't modify it. In the expression `let foo = "abc";` -- `foo` is already a `&str` that you should be able to pass to `prepare`
```suggestion
        ;
```

martin-mbtiles/src/tile_copier.rs
```diff
@@ -31,11 +31,11 @@ pub struct TileCopierOptions {
     src_file: PathBuf,
     /// MBTiles file to write to
     dst_file: PathBuf,
-    /// TODO: add documentation Output format of the destination file, ignored if the file exists. if not specified, defaults to the type of source
+    /// Output format of the destination file, ignored if the file exists. if not specified, defaults to the type of source
```
nyurik:
```suggestion
    /// Output format of the destination file, ignored if the file exists. If not specified, defaults to the type of source
```

nyurik (commented):
looks good!  see inline, but its nearly done

martin-mbtiles/src/mbtiles.rs
upsicleclown: I think for now since there is only one hash function, and the column in `FlatWithHash` is plainly called `tile_hash`, it should just stay `global_hash`

martin-mbtiles/src/mbtiles.rs
upsicleclown: I think as of right now, I will just do the simpler hash with no join and that is all

martin-mbtiles/src/mbtiles.rs
nyurik: Per additional discussion, it only makes sense to hash `z,x,y, (tile hash or tile blob or both)`. We will need to do perf test later, but for now lets do `z,x,y,hash`. Note that all integers will need to be converted to strings first - hashing is only available on strings and blobs.

upsicleclown: We can store which hash it is in the metadata table once we allow for different algorithms. Added a comment about this to the relevant 
issue: https://github.com/maplibre/martin/issues/835#issuecomment-1699677255

martin-mbtiles/src/mbtiles.rs
upsicleclown: The query returns multiple rows, one for each issue that was found ([link](https://www.sqlite.org/pragma.html#pragma_integrity_check)). I will put these all into a string with errors separated by newlines

martin-mbtiles/src/mbtiles.rs
upsicleclown: If we fail everytime the `global_hash` value is missing, we can never generate the global hash value for the first time because we will fail before we get to that step. I agree it is good to just fail here but I'm not sure of a cleaner way to actually generate the global hash for the first time.

upsicleclown: I will make the global hash generation happen at the copying level

martin-mbtiles/src/mbtiles.rs
```diff
@@ -378,33 +388,105 @@ impl Mbtiles {
...
+        if let Some(global_hash) = self.get_global_hash(&mut *conn).await? {
+            query!(
+                "INSERT OR REPLACE INTO metadata(name, value) VALUES('global_hash', ?)",
+                global_hash
+            )
+            .execute(conn)
+            .await?;
+        }
+
+        Ok(())
```
nyurik: you don't need that new sqlx file if you do this instead - which is also better because it doesn't introduce duplicated code for the same functionality:

```suggestion
        if let Some(global_hash) = self.get_global_hash(&mut *conn).await? {
            self.set_metadata_value(conn, "global_hash", Some(global_hash))
                .await
        } else {
            Ok(())
        }
```

martin-mbtiles/src/errors.rs
```diff
@@ -22,11 +20,14 @@ pub enum MbtError {
...
+    #[error("Integrity check failed for MBTile file {0} for the following reasons: \n {1:?}")]
+    FailedIntegrityCheck(String, Vec<String>),
+
     #[error("Invalid tile data for MBTile file {0}")]
     InvalidTileData(String),
 
-    #[error("Incorrect data format for MBTile file {0}; expected {1:?} and got {2:?}")]
-    IncorrectDataFormat(String, &'static [MbtType], MbtType),
+    #[error("Expected metadata value for {key} key in MBTiles file {filepath}")]
+    MetadataValueNotFound { key: String, filepath: String },
```
nyurik: I don't think this error should be this generic - technically we should almost never rely on metadata having any specific field.  Instead, lets rename this error to something like `GlobalHashValueNotFound(filepath)` -- and only use it for that case

martin-mbtiles/src/mbtiles.rs
```diff
@@ -378,33 +388,105 @@ impl Mbtiles {
...
-            MbtType::FlatWithHash => {
-                "SELECT * FROM tiles_with_hash WHERE tile_hash!=hex(md5(tile_data)) LIMIT 1;"
+                println!("Cannot generate global hash, no hash column in flat table format. Skipping global_hash generation...");
+                None
             }
-            MbtType::Normalized => {
-                "SELECT * FROM images WHERE tile_id!=hex(md5(tile_data)) LIMIT 1;"
+            MbtType::FlatWithHash => Some(rusqlite_conn.query_row_and_then("SELECT hex(md5_concat(cast(zoom_level AS text), cast(tile_column AS text), cast(tile_row AS text), tile_hash)) FROM tiles_with_hash ORDER BY zoom_level, tile_column, tile_row;",[], |row| row.get::<_, String>(0))?),
+            MbtType::Normalized => Some(rusqlite_conn.query_row_and_then("SELECT hex(md5_concat(cast(zoom_level AS text), cast(tile_column AS text), cast(tile_row AS text), tile_id)) FROM map ORDER BY zoom_level, tile_column, tile_row;",[], |row| row.get::<_, String>(0))?)
+        })
```
nyurik: Why not keep the SQL separate from the actual query?  E.g.

```rust
    async fn get_global_hash<T>(&self, conn: &mut T) -> MbtResult<Option<String>>
    where
        for<'e> &'e mut T: SqliteExecutor<'e>,
    {
        let rusqlite_conn = RusqliteConnection::open(Path::new(&self.filepath()))?;
        register_md5_function(&rusqlite_conn)?;

        let sql = match self.detect_type(&mut *conn).await? {
            MbtType::Flat => {
                println!("Cannot generate global hash, no hash column in flat table format. Skipping global_hash generation...");
                return Ok(None);
            }
            MbtType::FlatWithHash => "SELECT hex(md5_concat(cast(zoom_level AS text), cast(tile_column AS text), cast(tile_row AS text), tile_hash)) FROM tiles_with_hash ORDER BY zoom_level, tile_column, tile_row;",
            MbtType::Normalized => "SELECT hex(md5_concat(cast(zoom_level AS text), cast(tile_column AS text), cast(tile_row AS text), tile_id)) FROM map ORDER BY zoom_level, tile_column, tile_row;"
        };
        Ok(Some(rusqlite_conn.query_row_and_then(sql, [], |row| {
            row.get::<_, String>(0)
        })?))
    }
```

nyurik (commented):
almost there, a few tiny fixes

martin-mbtiles/src/mbtiles.rs
upsicleclown: Sure, I can do that

nyurik (approved):
Looks good, feel free to merge when ready. Thanks!

nyurik (approved):

Issue #845, opened 2023-08-28
Allow to show tile boundary in `debug.html`


sharkAndshark:
Something like this
![image](https://github.com/maplibre/martin/assets/42399654/d38fbe39-0c69-4869-b7c1-507a66b7dc46)

nyurik:
I think @upsicleclown had some experience with this?

P.S. and yes, i agree that the debug screen should show tile boundaries

sharkAndshark:
A [method](https://maplibre.org/maplibre-gl-js/docs/API/classes/maplibregl.Map/#showtileboundaries:~:text=map.ts%3A3347-,showTileBoundaries,-%E2%80%A2%20get%20showTileBoundaries) in `maplibre`:
```javascript
map.showTileBoundaries = true;
```
* The map will render an outline around each tile and the tile ID
* The uncompressed file size of the first vector source is drawn in the top left corner of each tile, next to the tile ID
Seems `maputnik` used this either.

nyurik:
Yes, good idea

PR #858, opened 2023-09-03
Show tile boundaries in debug.html

sharkAndshark:
Try to implement #845 
Based on the  [method](https://maplibre.org/maplibre-gl-js/docs/API/classes/maplibregl.Map/#showtileboundaries:~:text=map.ts%3A3347-,showTileBoundaries) in doc:

* The map will render an outline around each tile and the tile ID
* The uncompressed file size of the first vector source is drawn in the top left corner of each tile, next to the tile ID. 
 
The second one, the file size is only for the first vector source, is this misleading? I mean, will the user think it's the  sum file size of all sources? @nyurik

### DIFF
--- a/tests/debug.html
+++ b/tests/debug.html
@@ -285,6 +285,7 @@
     });
 
     map.on('load', async function () {
+        map.showTileBoundaries = true;
         const catalog = await fetch('http://0.0.0.0:3000/catalog');
         const sources = (await catalog.json()).tiles;
         // Set up the corresponding toggle button for each layer.


tests/debug.html
```diff
@@ -345,6 +345,7 @@
             const layers = document.getElementById('menu');
             layers.appendChild(link);
         }
+        map.showTileBoundaries = true;
```
nyurik: Move this to line 288

nyurik (approved):

nyurik:
I guess it is a bit misleading, but we can always try to fix it in the maplibre lib itself

Issue #856, opened 2023-08-31
Use consistent hashing validation for mbtiles


nyurik:
Currently, we generate `global_hash` on mbtiles files by hashing `(z, x, y, tile_hash)` for all rows, ordered by z,x,y.  This has one important issue: global hash is not consistent across different mbtile formats, e.g. if the mbtiles is in non-normalized (flat) format, there is no tile hash, thus global hash cannot be generated.

### Proposal
Modify algorithm to use all fields of the `tiles` table/view, as defined by [the spec](https://github.com/mapbox/mbtiles-spec/blob/master/1.3/spec.md#schema-1) - the `zoom_level, tile_column, tile_row, tile_data` columns, with the first 3 columns converted to strings first.  This guarantees consistent hash even if mbtiles is copied from one format to another.

I also think the hash should be renamed to be more specific, and not tied to individual tile hashes (as stored in the metadata table). Some naming ideas:
* `tiles_md5_hash`
* `concat_tiles_md5_hash`
* `combined_tiles_hash`
* `combined_tiles_md5_hash`
* `agg_tiles_hash`
* `agg_tiles_hash_md5`
* (something else)?

We could autodetect the hashing type because each hash digest has a unique byte length:

Algorithm | Digest Size (Bytes)
-- | --
MD5 | 16
SHA-1 | 20
SHA-256 | 32
SHA-384 | 48
SHA-512 | 64

CC: @upsicleclown

nyurik:
Fixed, settled on `agg_tiles_hash` metadata value

PR #861, opened 2023-09-04
A few cleanups on tile hashing

nyurik:
* Rename `global_hash` to `agg_tiles_hash`

This is still a big sticking point: what should be the name for the metadata key for this value? The value represents the hash of all `z,x,y,tile` over all rows of the `tiles` table (or view).  Should it include `md5` in its name, or should the hash be auto-detected by its length? (details in #856)

* Generate it based on `tiles` table/view
* validate or generate, but not both (it will always fail otherwise)
* break up logic for per-tile, total, and integrity checks
* delete unused sqlx prep file


sharkAndshark:
If it's true each hash digest has a unique byte length, I prefer the `auto-guessing`.  
Should we allow user to choose algorithms(default `MD5`) when generating hash?

||feature|drawback|
|----|----|----|
|algorithm in name|  explain its algorithm `explicitly` | different name across mbtile files like `tiles_md5_hash` in file1, `tiles_sha1_hash` in file2  |
| guessing |  explain its algorithm `implicitly`; consistent metadata name across mbtile files||

nyurik:
eventually we could let users pick algorithms - not very difficult, but not worth our time at the moment i think - md5 has worked pretty well for this usecase, despite being ... cryptographically poor.  I agree about autoguessing.  I think `agg_tiles_hash` sounds best, as it is both short, and implies aggregation

sharkAndshark:
||feature|drawback|
|----|----|----|
|algorithm in name|  explain its algorithm `explicitly` | different name across mbtile files like `tiles_md5_hash` in file1, `tiles_sha1_hash` in file2  |
| guessing |  explain its algorithm `implicitly`; consistent metadata name across mbtile files||

martin/src/pg/pool.rs
```diff
@@ -47,7 +47,7 @@ impl PgPool {
             .build()
             .map_err(|e| PostgresPoolBuildError(e, id.clone()))?;
 
-        let version = get_conn(&pool, id.as_str())
+        let version: String = get_conn(&pool, id.as_str())
```
sharkAndshark: Seems there's few connection between cleaning up on `pg/pool.rs` and the target of this PR. Not sure as I didn't do much on mbtiles part of martin.

nyurik: sorry, that was a minor cleanup. Fixed

GeoKarim:
Did a quick overview, all looks good to me.

nyurik:
I walked @GeoKarim (my coworker) through the code to make sure there are no obvious oopsies

PR #867, opened 2023-09-06
Optimize log level when couldn't find tile data

### LINKED ISSUE #815: Downgrade log level  when martin couldn't find tile based on given zxy
It's common that  no data in the requested tile grid.  
```
[2023-08-16T09:24:06Z ERROR martin::srv::server] Tile 9/402/210 not found in chegndu
[2023-08-16T09:24:06Z ERROR martin::srv::server] Tile 9/402/209 not found in chegndu
[2023-08-16T09:24:10Z ERROR martin::srv::server] Tile 9/401/210 not found in chegndu
[2023-08-16T09:24:10Z ERROR martin::srv::server] Tile 9/401/209 not found in chegndu
[2023-08-16T09:24:11Z ERROR martin::srv::server] Tile 9/402/211 not found in chegndu
[2023-08-16T09:24:11Z ERROR martin::srv::server] Tile 9/401/211 not found in chegndu
[2023-08-16T09:24:11Z ERROR martin::srv::server] Tile 9/403/211 not found in chegndu
[2023-08-16T09:24:13Z ERROR martin::srv::server] Tile 9/403/212 not found in chegndu
[2023-08-16T09:24:13Z ERROR martin::srv::server] Tile 9/402/212 not found in chegndu
[2023-08-16T09:24:13Z ERROR martin::srv::server] Tile 9/401/212 not found in chegndu
```

sharkAndshark:
Fix #815

### DIFF
--- a/martin/src/mbtiles/mod.rs
+++ b/martin/src/mbtiles/mod.rs
@@ -4,7 +4,7 @@ use std::path::PathBuf;
 use std::sync::Arc;
 
 use async_trait::async_trait;
-use log::debug;
+use log::trace;
 use martin_mbtiles::MbtilesPool;
 use martin_tile_utils::TileInfo;
 use tilejson::TileJSON;
@@ -95,11 +95,14 @@ impl Source for MbtSource {
         {
             Ok(tile)
         } else {
-            debug!(
+            trace!(
                 "Couldn't find tile data in {}/{}/{} of {}",
-                xyz.z, xyz.x, xyz.y, &self.id
+                xyz.z,
+                xyz.x,
+                xyz.y,
+                &self.id
             );
-            Ok(Vec::default())
+            Ok(Vec::new())
         }
     }
 }
--- a/martin/src/mbtiles/mod.rs
+++ b/martin/src/mbtiles/mod.rs
@@ -99,7 +99,7 @@ impl Source for MbtSource {
                 "Couldn't find tile data in {}/{}/{} of {}",
                 xyz.z, xyz.x, xyz.y, &self.id
             );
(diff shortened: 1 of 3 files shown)


martin/src/mbtiles/mod.rs
```diff
@@ -86,11 +87,19 @@ impl Source for MbtSource {
     }
 
     async fn get_tile(&self, xyz: &Xyz, _url_query: &Option<UrlQuery>) -> Result<Tile, Error> {
-        Ok(self
+        if let Some(tile) = self
```
nyurik, line 90: While we **could** change the `Source` trait to return an `Option<Tile>` instead of `Tile`, but in reality we already have that -- we can simply return an empty `Vec::default()` which has zero allocation cost (empty vectors don't cost anything).

This way it could be clearly indicated that a tile is simply missing, and no logging is done (we could still log it with debug level).

An error would then indicate all abnormal conditions like broken connection, IO error, etc, but not an empty tile.

sharkAndshark: Maybe we could just drop the `debug!("Couldn't find tile data in ...")`, as it's too common that couldn't find tile data.

nyurik: sure, or at least switch it to `trace!`?

nyurik: P.S. Sorry, I didn't realize you were already using `Vec::new()` -- its the same as `Vec::default()`, and actually i prefer the `new()` because it is shorter

nyurik (approved):
lgtm, thx!

PR #870, opened 2023-09-07
Fix broken link in run-with-nginx.md

sharkAndshark:


nyurik (approved):

sharkAndshark:
Weird. The action didn't finished.

nyurik:
its because github actions do not run if only a readme is changed - i will force merge it

PR #887, opened 2023-09-18
Add clip_geom, buffer and extent under auto_publish conf

### LINKED ISSUE #872: How to encode all columns as tile properties using config?
As a comment in [config example](https://maplibre.org/martin/config-file.html#config-example) says one must explicitly list columns that should be encoded as tile properties.

> \# List of columns, that should be encoded as tile properties (required)

In my case I don't know columns in advance since they are generated dynamically. I could use `martin postgresql://postgres@localhost/db` to encode all columns automatically but I want to apply `buffer` and `clip_geom` configuration options. 

How can I make Martin encode all columns as tile properties while using config?

sharkAndshark:
Partly fix #872

### DIFF
--- a/martin/src/pg/configurator.rs
+++ b/martin/src/pg/configurator.rs
@@ -28,6 +28,8 @@ pub struct PgBuilderAuto {
     source_id_format: String,
     schemas: Option<HashSet<String>>,
     id_columns: Option<Vec<String>>,
+    clip_geom: Option<bool>,
+    buffer: Option<u32>,
 }
 
 #[derive(Debug)]
@@ -145,7 +147,7 @@ impl PgBuilder {
                             continue;
                         };
                         db_inf.srid = srid;
-                        update_id_column(&id2, &mut db_inf, auto_tables);
+                        update_auto_fields(&id2, &mut db_inf, auto_tables);
                         info!("Discovered source {id2} from {}", summary(&db_inf));
                         pending.push(table_to_query(
                             id2,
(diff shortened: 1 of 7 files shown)


martin/src/pg/configurator.rs
```diff
@@ -420,6 +431,8 @@ mod tests {
             source_id_format: source_id_format.to_string(),
             schemas: schemas.map(|s| s.iter().map(|s| (*s).to_string()).collect()),
             id_columns: None,
+            clip_geom: None,
+            buffer: None,
```
nyurik, line 464: you need to do the same type of check as for `id_columns` for both of these, e.g.

```suggestion
            clip_geom: if item.clip_geom.is_some() && is_function {
                error!("Configuration parameter auto_publish.functions.clip_geom is not supported");
                None
            } else {
                item.clip_geom
            },
            buffer: if item.buffer.is_some() && is_function {
                error!("Configuration parameter auto_publish.functions.buffer is not supported");
                None
            } else {
                item.buffer
            },
```

martin/src/pg/table_source.rs
```diff
@@ -19,7 +19,11 @@ static DEFAULT_EXTENT: u32 = 4096;
 static DEFAULT_BUFFER: u32 = 64;
 static DEFAULT_CLIP_GEOM: bool = true;
 
-pub async fn query_available_tables(pool: &PgPool) -> Result<SqlTableInfoMapMapMap> {
+pub async fn query_available_tables(
+    pool: &PgPool,
+    clip_geom: Option<bool>,
```
nyurik: I am not sure, but it might be better to actually change `buffer` and `clip_geom` here to `None` to indicate that we don't have them yet, and keep them unset until much later as part of the `instantiate_tables` somewhere around `db_inf.srid = srid;` line.

sharkAndshark: @nyurik  I was thinking of this also,how about here:
```rust
...
db_inf.srid = srid;
update_id_column(&id2, &mut db_inf, auto_tables);
impose_clip_geom(&mut db_inf, auto_tables);
impose_buffer(&mut db_inf, auto_tables);
...
```

nyurik: sure, just maybe it will make sense to rename the `update_id_column` to `update_auto_fields` and do all these steps inside one function

sharkAndshark: Hi @nyurik  Have a good day!   :) 
It's still a draft, but could you help me a code review in case of any big misunderstanding?

nyurik:
I made a few fixes in https://github.com/sharkAndshark/martin/pull/52

Once merged, should be good to go

nyurik:
Looks good, needs docs too

nyurik (approved):
thanks!

PR #889, opened 2023-09-26
Rework CI to build ARM64

nyurik:
* break CI into cleaner steps:  build, test/test-legacy, publish
* add arm64 for linux


nyurik:
I'll merge this PR without passing requirements because some step names were changed

PR #892, opened 2023-09-26
Try to use ubuntu for arm aaple cross-compile

nyurik:


nyurik:
apparently doesn't work as easily, not worth it

PR #895, opened 2023-09-26
In deb package, include ssl and mbtiles tool

nyurik:


sharkAndshark:
When reviewing this, I realized that seems we forget to update the `debian/config.yaml` in recent PR, like [Add clip_geom, buffer and extent under auto_publish conf](https://github.com/maplibre/martin/pull/887)

nyurik:
@sharkAndshark sure, that can be updated separately, although I would not want to put too much example code into that file - instead, just add a link to documentation into it.  Another option is to re-use one config example file for everything - for the config and for documentation (possibly using symlinks)

birkskyum (approved):

PR #906, opened 2023-09-29
Hide password in log

### LINKED ISSUE #902: Martin leaks database password in logs
Currently martin will log the entire connection string at `INFO` level and this includes the provided database password.

sharkAndshark:
Try to fix #902

### DIFF
--- a/martin/src/pg/pool.rs
+++ b/martin/src/pg/pool.rs
@@ -28,7 +28,7 @@ pub struct PgPool {
 impl PgPool {
     pub async fn new(config: &PgConfig) -> Result<Self> {
         let conn_str = config.connection_string.as_ref().unwrap().as_str();
-        info!("Connecting to {conn_str}");
+        info!("Connecting to {}", hide_pwd(conn_str));
         let (pg_cfg, ssl_mode) = parse_conn_str(conn_str)?;
 
         let id = pg_cfg.get_dbname().map_or_else(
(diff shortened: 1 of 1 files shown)


martin/src/pg/pool.rs
```diff
@@ -96,3 +96,58 @@ async fn get_conn(pool: &Pool, id: &str) -> Result<Object> {
...
+
+    if let Some(at_idx) = conn_string.find('@') {
+        if let Some(colon_idx) = &conn_string[..at_idx].find("://") {
+            if let Some(pwd_idx) = &conn_string[colon_idx + 3..at_idx].find(':') {
+                let start = pwd_idx + colon_idx + 4;
+                let end = at_idx;
+                new_str.replace_range(start..end, &"*".repeat(end - start));
+            }
+        }
+    } else if let Some(pwd_idx) = conn_string.find("password=") {
```
sharkAndshark, line 111: Seems there's still a bug here

nyurik:
I think #907 is better for this - not as clean output, but more guaranteed

martin/src/pg/pool.rs
nyurik: no no, this is not a good way to do this. Regex is built specifically for this type of jobs, or a dedicated postgres connection string parser (which is part of the postgress connection lib, so hard to reuse in a better way).  I did submit a better solution to this (IMO) - #907

In general for this type of issues, take a look at [documentation](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING) -- I would use something like this, you can try it at https://rustexp.lpil.uk/

`^postgres(?:ql)?://[^:@/?]+:([^:@/?]+)@` -- the password is captured as the first capture group

sharkAndshark: I'm not good at Regex actually I don't like Regex. So glad to see that there is a dedicated parser in other crate lib. Closed.

sharkAndshark: The reason I don't like Regex is it's easy to write hard to read later..

nyurik: it's not that bad, and it does save you in many cases, so highly advise to embrace it ... sparingly! ;)

nyurik: TBH, I don't think the code of this PR is easier to read / understand -- string parsing in code is notoriously non-trivial to understand either

sharkAndshark: Agree. I like your #907 🤜

sharkAndshark: > TBH, I don't think the code of this PR is easier to read / understand -- string parsing in code is notoriously non-trivial to understand either

TBH I'm exhausted to write this PR🤣

Issue #674, opened 2023-05-24
Automate Homebrew update in CI release


nyurik:
The homebrew added in #623 is a static file (still needs to be tested on a Mac). Now we need to automate that file's update process as part of the CI release, possibly using one of the [marketplace actions](https://github.com/marketplace?type=actions&query=homebrew+).

This issue is to modify/extend this github workflow code to automate updating of the [homebrew-formula/martin.rb](https://github.com/maplibre/martin/blob/main/homebrew-formula/martin.rb) file, and either push the change to the main branch, or at least create a pull request.

https://github.com/maplibre/martin/blob/8f020aa6f634143f90f0bc00a5f7c9521089f657/.github/workflows/ci.yml#L176-L184

nyurik:
The auto-publish to the [maplibre/homebrew-martin](https://github.com/maplibre/homebrew-martin) repo now happens automatically as part of the CI process.  Eventually may switch to push instead of PR creation (same token should allow that)

nyurik:
done

Issue #618, opened 2023-03-27
Add Homebrew formulae in Martin repo


stepankuzmin:
We should try hosting Martin Homebrew formulae in the Martin repo and use the `brew tap URL` instead of migrating the tap from the [Urbica](https://github.com/urbica/homebrew-tap) org.

Closes https://github.com/maplibre/martin/issues/616

nyurik:
This was merged, but I have no way to test it

rubenpoppe:
I get `Warning: No available formula or cask with the name "maplibre/martin/martin".`

nyurik:
@rubenpoppe i sadly have no way of testing homebrew - don't have a mac. Do you know much about that system?  Would love to get some help on this, to help users install it without downloading from the release page

sharkAndshark:
I bought a MacBook weeks ago, maybe any help I can for this?

tomeronen:
It worked for me but checked and now its also not working for me. sorry. ll look later today and check what gone wrong

nyurik:
Seems the homebrew is still broken?  I tried it in #917, but it needs someone familiar with homebrew to add proper tests, and eventually perhaps ensure that it is using latest?  Help wanted :)

rubenpoppe:
Seems to work for me now!

nyurik:
Done in https://github.com/maplibre/homebrew-martin

PR #917, opened 2023-10-02
Add a homebrew CI test

nyurik:
Trying to automate homebrew testing, so that eventually we can be certain our mac users are ok.

HELP NEEDED: I (@nyurik) don't have a Mac.

So far I already see that it is loading a lot of un-needed (?) homebrew stuff when adding a tap - so perhaps we need to move it to another repo all-together, like `maplibre/martin-homebrew` ?

This PR temporarily deletes all other CIs to avoid wasting resources -- they will be re-added later.

See #618 #674


sharkAndshark:
I have a Mac but I know nothing about `homebrew` packaging..  Is there any thing I could help before reading a lot of documents about `homebrew`?

nyurik:
Sorry, I know even less about homebrew :)

sharkAndshark:
I'm reading docs then ..😂

nyurik:
So it seems the installation passes now (thanks @sharkAndshark !), so now the questions are:
* do we want to move homebrew formular to a new repo, e.g. `maplibre/martin-homebrew`
* are there any additional tests we can do?
* is there a way to automate formula update on each release

sharkAndshark:
@nyurik 
* I'm happy to move it to a standalone repo as `homebrew` clone whole martin repo when tapping. It can save user's disk space 
* Maybe run an integration test against a PostGIS?  But we need to duplicate some test shell if we are in an new repo:  `maplibre/martin-homebrew`.. 
* Need more research on. A simple idea is use shell to parse `cargo.toml`?

nyurik:
So this is now a fairly low priority, but if you do notice how other repos structure homebrew, it would help. Seems weird if each installation env (homebrew, snap, nix, ...) would need its own repo tbh, but maybe there is a common pattern other repos follow?

sharkAndshark:
How about we stay a single repo until we figure out the best practice?

nyurik:
sure, we can do that too.  I do want to have some CI that tests & updates the code as part of the release process. Esp maybe have a regularly running (cron job) workflow that validates that homebrew still works.

nyurik:
already done in the homebrew-martin repo

PR #924, opened 2023-10-07
Log a warning if PG startup takes too long

### LINKED ISSUE #810: Log a warning if PG startup takes too long
Per discussion in #804, Martin should warn users if PostgreSQL discovery step is taking too long.

For each PostgreSQL connection, e.g. in [`PgConfig::resolve`](https://github.com/maplibre/martin/blob/7b765af89d08bd434ca39abf4fceee6761873d29/martin/src/pg/config.rs#L108C23-L108C23), print a warning if table/function instantiation takes longer than 5 (?) seconds. This could probably be done using [this method](https://stackoverflow.com/questions/61041856/await-for-future-again-after-tokiotimetimeout) waiting for the `instantiate_tables` function.  The `instantiate_functions` could also use some sort of a warning in case we implement retrieving a test tile, but it probably won't have the same message.

So, if `instantiate_tables` takes over 5 seconds, print something like this:

```
WARN: PostgreSQL table discovery is taking too long. Make sure your table geo columns have a GIS index, or use --disabling-bounds.
```

pjsier:
Adds a warning message using the suggested implementation in #810. Thanks for the recommended approach in the ticket! This one seemed straightforward enough that local tests would do, which I ran by adding a delay to `instantiate_tables`.

I'm getting a clippy error locally for having `Deserialize` on a type with methods using `unsafe`, but it looks like that might be a more significant update.

Fixes #810

### DIFF
--- a/martin/src/pg/config.rs
+++ b/martin/src/pg/config.rs
@@ -4,7 +4,6 @@ use log::warn;
 use serde::{Deserialize, Serialize};
 use std::time::Duration;
 use tilejson::TileJSON;
-use tokio::time::timeout;
 
 use crate::config::{copy_unrecognized_config, UnrecognizedValues};
 use crate::pg::config_function::FuncInfoSources;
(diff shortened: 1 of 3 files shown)


martin/src/pg/config.rs
```diff
@@ -110,8 +114,18 @@ impl PgConfig {
 
     pub async fn resolve(&mut self, id_resolver: IdResolver) -> crate::Result<Sources> {
         let pg = PgBuilder::new(self, id_resolver).await?;
+
+        let instantiate_tables = pg.instantiate_tables();
+        pin_mut!(instantiate_tables);
```
nyurik: I think this should go inside the `fn instantiate_tables()` because otherwise you are not simultaneously instantiating tables AND functions. Instead, you instantiate tables for 5 seconds or until complet, and only then start instantiating functions.

pjsier: Sorry for the mixup there! This updated version should work as expected, but I wrapped the entire `instantiate_tables` function because there are multiple futures awaited inside of it and I wasn't sure the best way to go about that

nyurik:
@pjsier I created a PR against your PR - https://github.com/pjsier/martin/pull/1
I think this approach is a bit cleaner and easier to follow, plus i suspect it won't cause Clippy to complain (I have no idea why it was doing that to begin with).
If you agree, simply merge my PR into yours, and I will approve it.

nyurik (approved):
thx!

PR #932, opened 2023-10-10
Update Homebrew for 0.9.0

rubenpoppe:
- Updated version to 0.9.0, removed openssl dependency
- Added architecture checks

### DIFF
--- a/HomebrewFormula/martin.rb
+++ b/HomebrewFormula/martin.rb
@@ -1,17 +1,19 @@
 class Martin < Formula
-  current_version="0.8.7"
+  current_version="0.9.0"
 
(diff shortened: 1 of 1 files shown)


nyurik:
@rubenpoppe thx for contributing!  What's the process of getting the SHA of the releases?  Also, do you have any experience with automating this?

I want to do another release, probably in the next hour or so, before making this change.

rubenpoppe:
Not much experience with automating that unfortunately but I think it shouldn't be too difficult. I've now manually ran `shasum -a 256 martin-{arch}.tar.gz`. If you could add that to the ci workflow to calculate the hashes and update the file accordingly it's maybe achievable.

nyurik (approved):

PR #937, opened 2023-10-12
Add raster zxy function doc

sharkAndshark:
The main goal of this PR is try to add doc about how to support raster in PostGIS with user defined function.
The test in this PR might be remove after it works before merged.


nyurik:
If it is a user-defined function, I don't think we should actually add any tiff-parsing/generation to the integration testing. Unless we can make Martin recognize image tables and automatically create functions for them, I do not think we should add anything beyond a simple test without raster support.

### Testing strategy
* add some tiny dummy png & jpg images to `tests/fixtures/files` dir
* use some `psql` commands to create a table `images`. Columns could be `format TEXT, image BLOB`
* use `psql` to insert image file content into that table
* create a simple function that always returns the same image, e.g.
    `png_source(z,x,y)` -> `SELECT image FROM images WHERE format = 'png'`
* verify that the `png_source` and `jpeg_source` work correctly, and return correct images with the right content and metadata

nyurik:
Could you reduce the size of the tiff please? A 0.7MB file is a bit large for a test like this. Also, do you think we should also include some PNG and JPG samples into PG serving?  Lastly - while this test is good to have, the main use of it is to have detailed description in the doc book - please add some info there, either as a small paragraph, or maybe even a dedicated page with in-depth examples.

sharkAndshark:
Hi @nyurik Have a good day！
* The main goal of this PR is  `document`: How to integrate raster xyz function and martin
* So don't worry the tiff size, the test here would be deleted before merge
* I'm not  familiar with raster world, need some research on this.
* If martin couldn't set correct `content-type` of raster function which take xyz and return `jpeg/png` bytes array, this PR will try to resolve this and the test will be kept( and  definitely the tiff size would be reduced)
  
Any guidance and help would be really appreciated.  : )

sharkAndshark:
I do think we should also include really simple and tiny tests if we guaranteed we could serve raster function.   
And in this PR we can see martin take `application/x-protobuf` as  `content-type` of raster function, we need to fix this

nyurik:
@sharkAndshark if your goal is to simply document serving raster tiles, please re-read https://github.com/maplibre/martin/pull/937#issuecomment-1760336169 -- you do not need anything TIFF related, or anything related to configuring postgis code.  Instead, much easier is to add PNG or JPG files directly into a Postgres table as a bytea value.

nyurik:
@sharkAndshark where did you find this image? I think it might be ok for documentation purposes to have an image-serving setup, and 0.7MB might be ok, but we would need to add proper copyright to it.  Best is to find NASA-generated images (they are public domain), or some image from Wikipedia Commons (they have open licenses).

sharkAndshark:
Thx. I can't remember the website. I will try to find another small and proper licensed image. Try NASA first.

CommanderStorm:
I think this PR is a bit farther from getting merged and does not have momentum to get there -> lets close and regroup.

Thanks for the PR nevertheless 😄

PR #950, opened 2023-10-17
Use sccache for github CI builds

nyurik:


nyurik:
seems like its running slower, not worth it

PR #961, opened 2023-10-23
Seperate supported MBTiles schema doc

### LINKED ISSUE #918: Document MBTiles schema as a separate book page
We currently have a single [tools page](https://maplibre.org/martin/tools.html) that contains both the description of the `mbtiles` tool, as well as some documentation for the mbtiles schema extensions we support (e.g. hash validation and the well defined normalized/non-normalized/... formats).

I think we should split-up the tools page - to create a documentation section for mbtiles, that has a page documenting schema variants, the hashing capabilities, and another page documenting some capabilities of the `mbtiles` tools

sharkAndshark:
Try to fix #918 
I'm not sure maybe add supported schema as an `appindex`?


nyurik:
I think we need a new section for `MBTiles Support` instead of Tools.  That section will have a dedicated page for `SQL Schemas` which will go in depth on each format.  I don't think the schemas should be part of the configuration file - as this is completely different.

nyurik (approved):
thanks, i did another round of fixes, hope the docs are in a better state now

PR #986, opened 2023-11-05
Add tile statistics to mbtiles

### LINKED ISSUE #964: Add tile statistics to `mbtiles`, e.g. bbox, number of tiles, ...
`mbtiles` currently has `meta-all` to show the content of the metadata table. In addition to that, we should have a feature, either a new command like `stats`, or just expand `meta-all` (less favored) to compute the following:
* compute rough bbox based on the available tiles - i.e. compute min/max X and Y for tile index, and convert it into a bbox. It should be able to update the metadata table with the new value
* compute the number of the tiles in the mbtiles, with the breakdown by zoom level

## Example

```shell
> mbtiles stats myfile.mbtiles
```
#### File Info

myfile.mbtiles
Structure: normalized-with-hash
Page size: 4,096B
File size: 34,000KB
... some other useful information about tables/indexes, possibly taking some ideas from `sqlite3_analyzer` utility (I wonder if any of this data is available via standard SQL... if not, we shouldn't show it - no hacks)

#### Tile Info
| zoom | count | smallest | largest | average | bbox |
|---|---|---|---|---|---|
| 0 | 1 | 48KB | 48KB | 48KB | -180,-86,180,86 |
| 1 | 4 | 30KB | 50KB | 38KB | -180,-86,180,86 |
| 2 | 25 | 10KB | 70KB | 15KB | -180,-86,180,86 |
|---|---|---|---|---|---|
| all | 7,247 | 15B | 2.3MB | 50KB | -180,-86,180,86 |

sharkAndshark:
Try to fix #964 
- [x] Add tile statistics to mbtiles tools
- [x] Add test
- [x] Use 4326 instead of 3857 for tile bounds
- [x] Add document
- [x] Use size-format to prettify output 
- [x]  Statistics  struct Refactor
- [x] Cleanup and reformat


mbtiles/src/mbtiles.rs
```diff
@@ -39,6 +39,60 @@ pub struct Metadata {
...
 }
 
+#[derive(Clone, Debug, PartialEq, Serialize)]
+pub struct LevelDetail {
+    pub zoom: String,
+    pub count: u32,
+    pub smallest: f32,
+    pub largest: f32,
+    pub average: f32,
+    pub bounding_box: [f32; 4],
```
nyurik:
```suggestion
    pub bounds: tilejson::Bounds,
```

You may want to add a `use tilejson::Bounds` at the top instead.

mbtiles/src/mbtiles.rs
```diff
@@ -223,7 +277,114 @@ impl Mbtiles {
...
         }
     }
-
+    pub async fn statistics<T>(&self, conn: &mut T) -> MbtResult<Statistics>
+    where
+        for<'e> &'e mut T: SqliteExecutor<'e>,
+    {
+        let file_size =
+            PathBuf::from(&self.filepath).metadata().unwrap().len() as f64 / 1024.0 / 1024.0;
+        let page_size_query = query!("PRAGMA page_size;");
```
nyurik: I think its better to keep `query!` and related `fetch` together:

```rust
let page_size_query = query!("PRAGMA page_size;");
let page_size = page_size_query.fetch_one(&mut *conn).await?.page_size;
```

or even inline it:

```rust
let page_size = query!("PRAGMA page_size;").fetch_one(&mut *conn).await?.page_size;
```

mbtiles/src/mbtiles.rs
```diff
@@ -39,6 +39,60 @@ pub struct Metadata {
...
+    pub largest: f32,
+    pub average: f32,
+    pub bounding_box: [f32; 4],
+}
+
+#[derive(Clone, Debug, PartialEq, Serialize)]
+pub struct Statistics {
+    pub file_path: String,
+    pub file_size: f64,
+    pub schema: String,
```
nyurik: I think it is better to keep the value as original until serialization.  When serializing, simply use the Display variant (i.e. just print it as is, and let MbtType format itself)

```suggestion
    pub schema: MbtType,
```

mbtiles/src/mbtiles.rs
```diff
@@ -39,6 +39,60 @@ pub struct Metadata {
...
+        writeln!(f, "Page size: {} bytes", self.page_size.unwrap()).unwrap();
+        writeln!(
+            f,
+            "|{:^9}|{:^9}|{:^9}|{:^9}|{:^9}|{:^9}|",
+            "zoom", "count", "smallest", "largest", "average", "bbox"
+        )
+        .unwrap();
+
+        for l in &self.level_details {
+            let bbox_str = format!(
```
nyurik: I just added https://github.com/georust/tilejson/pull/22 - the new version of Bounds will handle formatting itself

mbtiles/src/mbtiles.rs
```diff
@@ -39,6 +39,60 @@ pub struct Metadata {
...
+
+impl Display for Statistics {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        writeln!(f, "File: {}", self.file_path).unwrap();
+        writeln!(f, "FileSize: {:.2}MB", self.file_size).unwrap();
+        writeln!(f, "Schema: {}", self.schema).unwrap();
+        writeln!(f, "Page size: {} bytes", self.page_size.unwrap()).unwrap();
+        writeln!(
+            f,
+            "|{:^9}|{:^9}|{:^9}|{:^9}|{:^9}|{:^9}|",
```
nyurik, line 80: don't spend too much time on formatting - there are crates like https://github.com/devashishdxt/cli-table that can simplify this - but that can be done in a separate PR.

mbtiles/src/mbtiles.rs
```diff
@@ -223,7 +277,114 @@ impl Mbtiles {
...
+                avg( length( tile_data ) ) / 1024.0 AS average,
+                min(tile_column) as min_tile_x,
+                min(tile_row) as min_tile_y,
+                max(tile_column) as max_tile_x,
+                max(tile_row) as max_tile_y 
+            FROM tiles
+            GROUP BY zoom_level"#
+        );
+        let page_size = page_size_query.fetch_one(&mut *conn).await?.page_size;
+        let mb_type_string = match self.detect_type(&mut *conn).await? {
```
nyurik: no need for formatting - let `Display` trait handle this

mbtiles/src/mbtiles.rs
```diff
@@ -39,6 +39,60 @@ pub struct Metadata {
...
+                "{:.2}, {:.2}, {:.2}, {:.2}",
+                l.bounding_box[0], l.bounding_box[1], l.bounding_box[2], l.bounding_box[3]
+            );
+
+            writeln!(
+                f,
+                "|{:^9}|{:^9}|{:^9}|{:^9}|{:^9}|{:^9}|",
+                l.zoom,
+                l.count,
+                format!("{:.2}KB", l.smallest),
```
nyurik: this is fine for now, but eventually we will want KB to automatically compute - see https://crates.io/crates/size_format

mbtiles/src/mbtiles.rs
```diff
@@ -223,7 +277,114 @@ impl Mbtiles {
...
+                .unwrap(),
+            largest: level_details
+                .iter()
+                .map(|l| l.largest)
+                .reduce(f32::max)
+                .unwrap(),
+            average: level_details.iter().map(|l| l.average).sum::<f32>()
+                / level_details.len() as f32,
+            bounding_box: [
+                level_details
```
nyurik: Bounds strtuct supports `+=` and `+` operators, so you can use `.map.reduce` too

mbtiles/src/mbtiles.rs
```diff
@@ -849,4 +1010,51 @@ mod tests {
         assert!(matches!(result, Err(MbtError::AggHashMismatch(..))));
         Ok(())
     }
+
+    #[actix_rt::test]
+    async fn stat() -> MbtResult<()> {
+        let (mut conn, mbt) = open("../tests/fixtures/mbtiles/world_cities.mbtiles").await?;
+        let res = mbt.statistics(&mut conn).await?;
+
+        assert_eq!(
```
nyurik: you can simplify all this code -- if the statistics result supports `Serialize` trait, replace all asserts with `assert_toml_snapshot!(res, @"")`  (although I think YAML will be easier to read - but for that you will need to add `"yaml"` to `insta = { workspace = true, features = ["toml"] }` in the `Cargo.toml` file - then you can use `assert_yaml_snapshot!`.  Once you do that, run `just bless-insta-mbtiles`

sharkAndshark:
After `git rebase maplibre/main`, I can't push to my branch. I need to fix this first 😂
```shell
@sharkAndshark ➜ /workspaces/martin (mbstats) $ git push --no-verify -f
...
remote: Resolving deltas: 100% (45/45), completed with 20 local objects.
To https://github.com/sharkAndshark/martin.git
 ! [remote rejected] mbstats -> mbstats (refusing to allow a GitHub App to create or update workflow `.github/workflows/grcov.yml` without `workflows` permission)
error: failed to push some refs to 'https://github.com/sharkAndshark/martin.git'
```

mbtiles/src/mbtiles.rs
```diff
@@ -223,7 +277,94 @@ impl Mbtiles {
...
+            PathBuf::from(&self.filepath).metadata().unwrap().len() as f64 / 1024.0 / 1024.0;
+        let page_size = query!("PRAGMA page_size;")
+            .fetch_one(&mut *conn)
+            .await?
+            .page_size;
+        let tile_infos_query = query!(
+            r#"SELECT
+                zoom_level AS zoom,
+                count( ) AS count,
+                min( length( tile_data ) ) / 1024.0 AS smallest,
```
nyurik: don't divide by 1024 here -- the query should return as accurate data as possible. We will format it to the right units later.

mbtiles/src/mbtiles.rs
```diff
@@ -223,7 +277,94 @@ impl Mbtiles {
...
+                let smallest = r.smallest.unwrap_or(0.0) as f32;
+                let largest = r.largest.unwrap_or(0.0) as f32;
+                let average = r.average.unwrap_or(0.0) as f32;
+
+                let min_tile_x = r.min_tile_x.unwrap();
+                let min_tile_y = r.min_tile_y.unwrap();
+                let max_tile_x = r.max_tile_x.unwrap();
+                let max_tile_y = r.max_tile_y.unwrap();
+
+                let minx = -20037508.34 + min_tile_x as f64 * tile_length;
```
nyurik: bbox should be in longitude/latitude coordinates (-180..180 for longitude, approximately -86..+86 for latitude). You will need to do a proper conversion

mbtiles/src/mbtiles.rs
```diff
@@ -223,7 +277,94 @@ impl Mbtiles {
...
+                    zoom: format!("{zoom}"),
+                    count,
+                    smallest,
+                    largest,
+                    average,
+                    bounding_box: bbox,
+                }
+            })
+            .collect();
+        let details_of_all = LevelDetail {
```
nyurik: summary of all levels will probably have to be a different struct, or better yet - just put those values in the `Statistics` struct itself. The values will be different because
* most of them will be `Option<...>` -- if there is no data, they will be set to None
* bbox might not be needed because most of the time tileset will contain zoom-0 tile, which covers the whole eart. bbox is mostly useful per zoom.
* there is no `zoom` value - so you should keep the `zoom` value inside LevelDetail as `u8`, not `String`.

nyurik:
P.S. It is usually better to use `git merge main`, rather than using rebase.  Rebase tends to loose tracking of the review comments.

sharkAndshark:
> P.S. It is usually better to use `git merge main`, rather than using rebase. Rebase tends to loose tracking of the review comments.

Thx! Didn't know that before : )

mbtiles/src/mbtiles.rs
```diff
@@ -57,6 +57,10 @@ pub struct Statistics {
     pub schema: MbtType,
     pub page_size: u64,
     pub level_details: Vec<LevelDetail>,
+    pub count: u64,
```
nyurik, line 60: all these should be `Option<u64>` / `Option<f64>`

mbtiles/src/mbtiles.rs
```diff
@@ -350,34 +364,29 @@ impl Mbtiles {
...
-                .reduce(|a, b| a + b)
-                .unwrap(),
-        };
-        level_details.push(details_of_all);
+        let count = level_details.iter().map(|l| l.count).sum();
+        let smallest = level_details
+            .iter()
+            .map(|l| l.smallest)
+            .reduce(u64::min)
+            .unwrap();
```
nyurik: remove unwrap here - because it is ok to not have any data (None)

mbtiles/src/mbtiles.rs
```diff
@@ -350,34 +364,29 @@ impl Mbtiles {
...
+        let largest = level_details
+            .iter()
+            .map(|l| l.largest)
+            .reduce(u64::max)
+            .unwrap();
+        let average =
+            level_details.iter().map(|l| l.average).sum::<f64>() / level_details.len() as f64;
```
nyurik: this is incorrect.
```suggestion
            level_details.iter().map(|l| l.average * l.count).sum::<f64>() / (count as f64);
```

mbtiles/src/mbtiles.rs
```diff
@@ -39,6 +40,79 @@ pub struct Metadata {
...
 }
 
+#[derive(Clone, Debug, PartialEq, Serialize)]
+pub struct LevelDetail {
+    pub zoom: u8,
+    pub count: u64,
+    pub smallest: u64,
+    pub largest: u64,
+    pub average: f64,
+    pub bounding_box: Bounds,
```
nyurik: rename this to bbox - its a common term

mbtiles/src/mbtiles.rs
sharkAndshark: Oops Sure

mbtiles/src/mbtiles.rs
sharkAndshark: Thx!

mbtiles/src/mbtiles.rs
sharkAndshark: `Largest` and` smallest` has been Optional, but I kept `count` and `average` still as the returned type of these are not optional, if the vec is empty, they just return zero.
`let count = level_details.iter().map(|l| l.count).sum();`  
`let average = level_details
                .iter()
                .map(|l| l.average * l.count as f64)
                .sum::<f64>()`

mbtiles/src/mbtiles.rs
sharkAndshark: Updated

mbtiles/src/mbtiles.rs
sharkAndshark: I would give it a try in next PR.

mbtiles/src/mbtiles.rs
```diff
@@ -39,6 +40,82 @@ pub struct Metadata {
...
+
+impl Display for Statistics {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        writeln!(f, "File: {}", self.file_path).unwrap();
+        writeln!(
+            f,
+            "FileSize: {:.2}B",
+            SizeFormatterBinary::new(self.file_size)
+        )
+        .unwrap();
```
nyurik: This style makes it a bit more readable. Same for others below
```suggestion
        let file_size = SizeFormatterBinary::new(self.file_size);
        writeln!(f, "FileSize: {file_size:.2}B").unwrap();
```

mbtiles/src/mbtiles.rs
```diff
@@ -39,6 +40,82 @@ pub struct Metadata {
...
+
+        for l in &self.level_details {
+            writeln!(
+                f,
+                "|{:^9}|{:^9}|{:^9}|{:^9}|{:^9}|{:^9}|",
+                l.zoom,
+                l.count,
+                format!("{:.2}B", SizeFormatterBinary::new(l.smallest)),
+                format!("{:.2}B", SizeFormatterBinary::new(l.largest)),
+                format!("{:.2}B", l.average as u64),
```
nyurik: average should also go through the SizeFormatterBinary

mbtiles/src/mbtiles.rs
```diff
@@ -690,10 +862,17 @@ pub async fn attach_hash_fn(conn: &mut SqliteConnection) -> MbtResult<()> {
     Ok(())
 }
 
+fn webmercator_to_wgs84(xy: (f64, f64)) -> (f64, f64) {
```
nyurik: no point in using a tuple here - too confusing.  Use proper args like `lon, lat`. The return can still be a tuple which you will destruct into `let (lon, lat) = ...` or `let (min_lon, min_lat) = ...` or similar

mbtiles/src/mbtiles.rs
```diff
@@ -849,4 +1028,115 @@ mod tests {
...
+        let xy2 = (20037508.34, 20037508.34);
+        let lonlat2 = (179.99999997494382, 85.05112877764508);
+
+        let xy3 = (0.0, 0.0);
+        let lonlat3 = (0.0, 0.0);
+
+        let xy4 = (3000.0, 9000.0);
+        let lonlat4 = (0.026949458523585643, 0.08084834874097371);
+
+        assert_eq!(webmercator_to_wgs84(xy1), lonlat1);
```
nyurik: not a good practice to separate input values with their expected values from the actual `assert` calls.  Inline all variables. 

Also, it is futile to compare floating point numbers exactly. I suggest you add [approx](https://docs.rs/approx/latest/approx/index.html) as a `dev-dependency`, and use it to do the comparison with the epsilon parameter.

mbtiles/src/mbtiles.rs
```diff
@@ -39,6 +40,82 @@ pub struct Metadata {
     pub json: Option<JSONValue>,
 }
 
+#[derive(Clone, Debug, PartialEq, Serialize)]
+pub struct LevelDetail {
```
nyurik: I keep stumbling on the word "level" here.  Lets rename it to `ZoomInfo` or `PerZoomInfo` or something like that?  The `Statistics` could also have a `zoom_info_list: Vec<ZoomInfo>` ?  Naming is hard, no good ideas here.

mbtiles/src/mbtiles.rs
```diff
@@ -39,6 +40,82 @@ pub struct Metadata {
...
+                format!("{:.2}B", SizeFormatterBinary::new(l.largest)),
+                format!("{:.2}B", l.average as u64),
+                l.bbox
+            )
+            .unwrap();
+        }
+        if self.count != 0 {
+            writeln!(
+                f,
+                "|{:^9}|{:^9}|{:^9}B|{:^9}B|{:^9}B|",
```
nyurik: Centering numbers is a bad idea - hard to compare them. Best to right-align.  Also, don't unwrap - will crash on empty (you probably want to have a test with an empty database - easy to create -- see examples in `tests/mbtiles.rs`... something like this:

```rust
let mbt = Mbtiles::new("file:mbtiles_empty_stats?mode=memory&cache=shared").unwrap();
let cn = mbtiles.open().await.unwrap();
create_flat_tables(&mut cn).await.unwrap();
// now you have both an mbtiles struct and a connection to it that you can pass to the statistic function
```

mbtiles/src/mbtiles.rs
nyurik: please don't resolve conversations like this - i didn't notice your response and had to open all of them

mbtiles/src/mbtiles.rs
sharkAndshark: I was struggling here either.  Is `ZoomStats`  and `zoom_stats_list` Ok?  If not, let's pick `ZoomInfo` firstly. I have bad sense of naming.

mbtiles/src/mbtiles.rs
sharkAndshark: Thx! One of my motivation is your Reviewing. Really learn a lot.  : )

mbtiles/src/mbtiles.rs
nyurik: sure, zoom stats works

mbtiles/src/mbtiles.rs
sharkAndshark: Approx added as a dev-dependency now

mbtiles/src/mbtiles.rs
sharkAndshark: Working on this.

docs/src/SUMMARY.md
```diff
@@ -23,7 +23,8 @@
   - [Recipes](45-recipes.md)
 - [Tools](50-tools.md)
   - [MBTiles Metadata](51-mbtiles-meta.md)
-  - [MBTiles Copying / Diffing](52-mbtiles-copy.md)
-  - [MBTiles Validation](53-mbtiles-validation.md)
-  - [MBTiles Schemas](54-mbtiles-schema.md)
+  - [MBTiles statistics](52-mbtiles-stats.md)
```
nyurik: we don't want to break all internet links every time we add a new docs page... we ought to figure out a way not to do that. For now, lets just add thing at the end.

mbtiles/src/lib.rs
```diff
@@ -24,3 +24,7 @@ pub use queries::{
     create_flat_tables, create_flat_with_hash_tables, create_metadata_table,
     create_normalized_tables, is_flat_with_hash_tables_type, is_normalized_tables_type,
 };
+
+#[cfg(test)]
```
nyurik: is this actually needed?  I thought this was only a 2018's edition if not before, and with 2021 you can simply do `use approx::assert_relative_eq;` in the file that uses it?

mbtiles/src/mbtiles.rs
```diff
@@ -39,6 +40,81 @@ pub struct Metadata {
...
+    pub largest: Option<u64>,
+    pub average: f64,
+}
+
+impl Display for Statistics {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        writeln!(f, "File: {}", self.file_path).unwrap();
+
+        let file_size = SizeFormatterBinary::new(self.file_size);
+        writeln!(f, "FileSize: {:.2}B", file_size).unwrap();
```
nyurik: inlining looks a bit better - here and below
```suggestion
        writeln!(f, "FileSize: {file_size:.2}B").unwrap();
```

mbtiles/src/mbtiles.rs
```diff
@@ -690,10 +861,17 @@ pub async fn attach_hash_fn(conn: &mut SqliteConnection) -> MbtResult<()> {
     Ok(())
 }
 
+fn webmercator_to_wgs84(x: f64, y: f64) -> [f64; 2] {
```
nyurik: arrays are usually not good in such cases --  a better pattern is to return as you did before - `-> (f64, f64)`. Arrays tend to mean "some number of identical elements", whereas tuples indicate that the elements are positional and might have different types.  Here, it just happened to be the same type.

mbtiles/src/mbtiles.rs
```diff
@@ -849,4 +1027,122 @@ mod tests {
...
+            [0.0, 0.0, 0.0, 0.0],
+            [3000.0, 9000.0, 0.026949458523585643, 0.08084834874097371],
+        ];
+        for case in cases.iter() {
+            let actual = webmercator_to_wgs84(case[0], case[1]);
+            let exp_lng = case[2];
+            let exp_lat = case[3];
+            assert_relative_eq!(actual[0], exp_lng, epsilon = 1e-4);
+            assert_relative_eq!(actual[1], exp_lat, epsilon = 1e-4);
+        }
```
nyurik: probably not worth it, keep it simple like this:

```suggestion
        let (lng, lat) = webmercator_to_wgs84(-20037508.34, -20037508.34);
        assert_relative_eq!(lng, -179.99999997494382, epsilon = f64::EPSILON);
        assert_relative_eq!(lat, -85.05112877764508, epsilon = f64::EPSILON);
        
        let (lng, lat) = webmercator_to_wgs84(20037508.34, 20037508.34);
        assert_relative_eq!(lng, 179.99999997494382, epsilon = f64::EPSILON);
        assert_relative_eq!(lat, 85.05112877764508, epsilon = f64::EPSILON);
        
        let (lng, lat) = webmercator_to_wgs84(0.0, 0.0);
        assert_relative_eq!(lng, 0.0, epsilon = f64::EPSILON);
        assert_relative_eq!(lat, 0.0, epsilon = f64::EPSILON);
        
        let (lng, lat) = webmercator_to_wgs84(3000.0, 9000.0);
        assert_relative_eq!(lng, 0.026949458523585643, epsilon = f64::EPSILON);
        assert_relative_eq!(lat, 0.08084834874097371, epsilon = f64::EPSILON);
```

nyurik:
its getting almost there, really like it!

I wonder how we can optimize the bounding box printing. I don't like that it prints all those unneeded digits. Instead, we should have some formula to compute how many digits are significant for each zoom level.  For example, you can convert a single tile near the equator, see how big it is in terms of digits, and use that many digits?  Something like this

```rust
let tile_size = 40075016.7 / (2_u32.pow(zoom as u32)) as f64;
let lng_delta = (tile_size / 6378137.0).to_degrees();
let log = log10(lng_delta) - 0.5
let precision = if log > 0 { 0 } else { ceil(-log) };
let bbox = format!("{box:.precision$}")
```

mbtiles/src/mbtiles.rs
sharkAndshark: Simpler and easy to read!

mbtiles/src/mbtiles.rs
sharkAndshark: All inlined now but something like this is still not supported.
```rust
struct Foo { bar: i32 }

fn main() {
    let foo = Foo { bar: 10 };
    let _x = format!("Hello {foo.bar}");
}
error: invalid format string: field access isn't supported
 --> src/main.rs:5:30
  |
5 |     let _x = format!("Hello {foo.bar}");
  |                              ^^^^^^^ not supported in format string
  |
help: consider using a positional formatting argument instead
  |
5 |     let _x = format!("Hello {0}", foo.bar);
  |                              ~  +++++++++
```

mbtiles/src/lib.rs
sharkAndshark: I think it's because my rust version. Updated.

mbtiles/src/mbtiles.rs
nyurik: yep, only simple vars are supported at the moment

docs/src/54-mbtiles-validation.md
```diff
@@ -13,9 +13,9 @@ The `validate` command will run `PRAGMA integrity_check` on the file, and will f
 The `validate` command will verify that the `tiles` table/view exists, and that it has the expected columns and indexes. It will also verify that the `metadata` table/view exists, and that it has the expected columns and indexes.
 
 ## Per-tile validation
-If the `.mbtiles` file uses [flat_with_hash](54-mbtiles-schema.md#flat-with-hash) or [normalized](54-mbtiles-schema.md#normalized) schema, the `validate` command will verify that the MD5 hash of the `tile_data` column matches the  `tile_hash` or `tile_id` columns (depending on the schema).
+If the `.mbtiles` file uses [flat_with_hash](55-mbtiles-schema.md#flat-with-hash) or [normalized](55-mbtiles-schema.md#normalized) schema, the `validate` command will verify that the MD5 hash of the `tile_data` column matches the  `tile_hash` or `tile_id` columns (depending on the schema).
```
nyurik: need to keep these links as before. Not good to change URLs that others have been using

nyurik:
Forgot to mention - I am not certain if you are affected by this, but mbtiles invert the Y value (`tile_row`) -- the Y coordinate of a tile corresponds to `2^zoom - 1 - tile_row` in mbtiles.

sharkAndshark:
Thx for the mention. I did some research and I hope it's right.
[MBTiles spec](https://github.com/mapbox/mbtiles-spec/blob/6643d4daca3f7417e4f2068fb3974aadb0a1154d/1.3/spec.md?plain=1#L117C1-L122) says it follow [TMS](https://wiki.osgeo.org/wiki/Tile_Map_Service_Specification), so the the Y of tile index increase from down to top.




|Schema|Y|
|----|----|
|XYZ|From top to down|
|TMS|From down to top|

### Tile Origin and Bbox of 3857
|Origin|Bbox|
|----|----|
|(-20037508.34, -20037508.34)|[-20037508.34, -20037508.34, 20037508.34, 20037508.34]|


### Length of one single tile

 (20037508.34 * 2) /  (2^zoom_level)
### MinX,MaxX 
MinX =   OriginX + X * TileLength
MaxX = MinX + TileLength
### MinY,MaxY
MinY =  OriginY + Y * TileLength
MaxY = MinY + TileLength
image

mbtiles/src/mbtiles.rs
```diff
@@ -223,7 +303,108 @@ impl Mbtiles {
             self.check_agg_tiles_hashes(&mut conn).await
         }
     }
+    pub async fn statistics<T>(&self, conn: &mut T) -> MbtResult<Statistics>
+    where
+        for<'e> &'e mut T: SqliteExecutor<'e>,
+    {
+        let file_size = query!(
```
sharkAndshark, line 303: Seems that `PathBuf` couldn't work with `file:something?mode=memory` together.  Using sql query for the file size instead of `PathBuf` now.  @nyurik

wipfli:
@bdon you did something similar for pmtiles, right?

mbtiles/Cargo.toml
```diff
@@ -32,15 +32,17 @@ clap = { workspace = true, optional = true }
 env_logger = { workspace = true, optional = true }
 serde_yaml = { workspace = true, optional = true }
 tokio = { workspace = true, features = ["rt-multi-thread"], optional = true }
+size_format = "1.0.2"
```
nyurik: this list should be sorted, and the actual version should be in the root Cargo.toml, same as all others

mbtiles/Cargo.toml
```diff
@@ -32,15 +32,17 @@ clap = { workspace = true, optional = true }
...
 [dev-dependencies]
 # For testing, might as well use the same async framework as the Martin itself
 actix-rt.workspace = true
 ctor.workspace = true
 env_logger.workspace = true
-insta = { workspace = true, features = ["toml"] }
+insta = { workspace = true, features = ["toml", "yaml"] }
 pretty_assertions.workspace = true
 rstest.workspace = true
+approx = "0.5.1"
```
nyurik: sort the list, and also move it to the root Cargo.toml

mbtiles/src/mbtiles.rs
```diff
@@ -39,6 +40,85 @@ pub struct Metadata {
...
+    pub zoom_stats_list: Vec<ZoomStats>,
+    pub count: u64,
+    pub smallest: Option<u64>,
+    pub largest: Option<u64>,
+    pub average: f64,
+}
+
+impl Display for Statistics {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        writeln!(f, "File: {}", self.file_path).unwrap();
```
nyurik: replace all `.unwrap()` with `?`.  Unwrap is discouraged, you do not want to crash in production code.

mbtiles/src/mbtiles.rs
nyurik: yeah, seems reasonable. The alternative would be to have `Option` for the size, and set it to None if unable to get it from the filesystem.

mbtiles/src/mbtiles.rs
```diff
@@ -223,7 +303,108 @@ impl Mbtiles {
...
+                file_path: self.filepath.clone(),
+                file_size,
+                schema: mbt_type,
+                page_size,
+                zoom_stats_list: level_details,
+                count: 0,
+                smallest: None,
+                largest: None,
+                average: 0.0,
+            })
```
nyurik: you don't need this block.  Average will be `NaN`, and everything else will already calculate this way if its empty.

mbtiles/src/mbtiles.rs
```diff
@@ -223,7 +303,108 @@ impl Mbtiles {
...
+                min( length( tile_data ) ) AS smallest,
+                max( length( tile_data ) ) AS largest,
+                avg( length( tile_data ) ) AS average,
+                min(tile_column) as min_tile_x,
+                min(tile_row) as min_tile_y,
+                max(tile_column) as max_tile_x,
+                max(tile_row) as max_tile_y 
+            FROM tiles
+            GROUP BY zoom_level"#
+        );
```
nyurik: i reformatted it a bit to fit with all other SQL statement formattings:

```suggestion
        let tile_infos_query = query!(
            "
    SELECT zoom_level             AS zoom,
           count()                AS count,
           min(length(tile_data)) AS smallest,
           max(length(tile_data)) AS largest,
           avg(length(tile_data)) AS average,
           min(tile_column)       AS min_tile_x,
           min(tile_row)          AS min_tile_y,
           max(tile_column)       AS max_tile_x,
           max(tile_row)          AS max_tile_y
    FROM tiles
    GROUP BY zoom_level"
        );
```

Cargo.toml
```diff
@@ -49,6 +50,7 @@ rustls-native-certs = "0.6"
 rustls-pemfile = "1"
 semver = "1"
 serde = { version = "1", features = ["derive"] }
+size_format = "1.0.2"
```
nyurik: should be sorted

docs/src/52-mbtiles-stats.md
```diff
@@ -0,0 +1,20 @@
...
+
+For the concern of efficiency, you could figure out the page size and file size, tile size and  bounds(represented as WGS 84 latitude and longitude values) of covered area with `mbtiles stats` command.
+
+```shell
+mbtiles stats /path/to/my_file.mbtiles
+File: ./tests/fixtures/mbtiles/world_cities.mbtiles
+FileSize: 48.00KiB
+Schema: flat
+Page size: 4.00KiB
+|  zoom   |  count  |smallest | largest | average |  bbox   |
```
nyurik: has this example changed with the new output changes?

docs/src/SUMMARY.md
```diff
@@ -26,4 +26,5 @@
   - [MBTiles Copying / Diffing](52-mbtiles-copy.md)
   - [MBTiles Validation](53-mbtiles-validation.md)
   - [MBTiles Schemas](54-mbtiles-schema.md)
+  - [MBTiles statistics](55-mbtiles-stats.md)
```
sharkAndshark, line 29: Added in the end as `55-mbtiles-stats.md` now.

docs/src/54-mbtiles-validation.md
sharkAndshark: Added in the end as `55-mbtiles-stats.md` now.
```markdown
- [Tools](50-tools.md)
  - [MBTiles Metadata](51-mbtiles-meta.md)
  - [MBTiles Copying / Diffing](52-mbtiles-copy.md)
  - [MBTiles Validation](53-mbtiles-validation.md)
  - [MBTiles Schemas](54-mbtiles-schema.md)
  - [MBTiles statistics](55-mbtiles-stats.md)
```

mbtiles/src/mbtiles.rs
```diff
@@ -39,6 +40,82 @@ pub struct Metadata {
...
+                l.count,
+                format!("{smallest:.2}B"),
+                format!("{largest:.2}B"),
+                format!("{average:.2}B"),
+                l.bbox
+            )?;
+        }
+        if self.count != 0 {
+            let smallest = SizeFormatterBinary::new(self.smallest.expect("The smallest tile size of all zoom levels shouldn't be None when the tiles count of all zoom level is not 0"));
+            let largest = SizeFormatterBinary::new(self.largest.expect("The largest tile size of all zoom levels shouldn't be None when the tiles count of all zoom level is not 0"));
```
nyurik: no point for all this complex `expect` strings which will never happen. Might as well do a simpler variant:

```suggestion
        if let (Some(smallest), Some(largest)) = (self.smallest, self.largest) {
            let smallest = SizeFormatterBinary::new(smallest);
            let largest = SizeFormatterBinary::new(largest);
```

Cargo.toml
sharkAndshark: In case any misunderstanding, does the `sorted` mean alphabet? So I placed size_format behind serde. @nyurik

mbtiles/src/mbtiles.rs
```diff
@@ -223,7 +300,95 @@ impl Mbtiles {
...
+            .iter()
+            .map(|l| l.average * l.count as f64)
+            .sum::<f64>()
+            / count as f64;
+        Ok(Statistics {
+            file_path: self.filepath.clone(),
+            file_size,
+            schema: mbt_type,
+            page_size,
+            zoom_stats_list: level_details,
```
nyurik: you would want to rename `level_details` to be the same as the field name - this way no need for the `: level_details` part.  Same for the `schema`. Naming things same as in struct usually keeps things a bit cleaner.

mbtiles/src/mbtiles.rs
sharkAndshark: Is `The largest tile size of all zoom levels shouldn't be None` Ok? Or just "Couldn't be None".

nyurik: you don't need to do any asserts (`.expected(...)`) - you can simply check if both values are not None, and if so, use them.

Cargo.toml
nyurik: exactly - the list of dependencies should be sorted alphabetically.  I have a shortcut in my IDE to select a few lines and sort them

mbtiles/src/mbtiles.rs
nyurik: P.S. `mbt_type` is probably a bit better for the `Statistics` struct too.  `schema` is more ambiguous.

mbtiles/Cargo.toml
```diff
@@ -31,14 +31,16 @@ anyhow = { workspace = true, optional = true }
 clap = { workspace = true, optional = true }
 env_logger = { workspace = true, optional = true }
 serde_yaml = { workspace = true, optional = true }
+size_format.workspace = true
```
nyurik: actually this is the wrong place -- size_format is used as part of the core mbtiles library, not just CLI. This sub-section is just for the `Bin dependencies` -- that's why they are all optional.

Cargo.toml
sharkAndshark: Hi @nyurik  Would you like to share it, the shortcut? Are you using IDEA or something else? I'm using VSCODE.

nyurik: you would need to create it for your vscode.  See https://stackoverflow.com/questions/69115477/how-can-i-order-alphabetically-a-list-in-plain-text-in-vs-code how to sort things, and https://stackoverflow.com/questions/33791097/how-can-i-change-keyboard-shortcut-bindings-in-visual-studio-code/33791170 on how to create shortcuts

docs/src/52-mbtiles-stats.md
sharkAndshark: Updated now.

Cargo.toml
sharkAndshark: Sorted now.

mbtiles/src/mbtiles.rs
sharkAndshark: Got it and udpated.

mbtiles/Cargo.toml
sharkAndshark: Moved and sorted.

nyurik (approved):
Awesome job, well done, thank you!!!

sharkAndshark:
> Awesome job, well done, thank you!!!

Thx all your guidance ！Hope I can do better in next : )

PR #1008, opened 2023-11-17
Minor cleanup

sharkAndshark:
- Add separators to long literal.
- Use first() instead of get(0)


nyurik:
looks good, but you may want to make these values `pub const` somewhere. Maybe even move them to `martin-tile-utils`?  Not sure.

nyurik:
P.S. Technically you only need one constant - `EARTH_CIRCUMFERENCE`

sharkAndshark:
Update 
`-20_037_508.34 + min_x as f64 * tile_size`  to `EARTH_CIRCUMFERENCE * 0.5 + min_x as f64 * tile_size`?  
But if we make a pub const for -20_037_508.34, the naming would be hard...  `HALF_EARTH_CIRCUMFERENCE`?

nyurik:
you don't need to have a constant for half - do it as part of the function, and the compiler will optimize it

sharkAndshark:
Do you mean?
```rust
pub const A_CONST = 4000.0;
let result = A_CONST * 0.5 * var_a;
```
Compiler would generate something like `let result = 2000.0 * var_a`?

nyurik:
yes

mbtiles/src/summary.rs
```diff
@@ -206,7 +208,7 @@ fn get_zoom_precision(zoom: u8) -> usize {
 }
 
 fn webmercator_to_wgs84(x: f64, y: f64) -> (f64, f64) {
-    let lng = (x / 6378137.0).to_degrees();
+    let lng = (x / 6_378_137.0).to_degrees();
```
nyurik: this number is `EARTH_CIRCUMFERENCE÷pi÷2` - i.e. `EARTH_RADIUS` - may want to create a const for that too.

nyurik:
oh, i just realized why you are fixing this - my #1006 now forces mbtiles crate to also go through the pedantic check, but because GitHub CI is still on Rust 1.73, it didn't show any warnings. I think we don't want to make it mandatory just yet, so i will relax these requirements. It doesn't mean we shouldn't fix them, only that we may want to relax a few rules.

nyurik:
the linting rules are now relaxed for mbtiles, you shouldn't get any warnings

mbtiles/src/summary.rs
```diff
@@ -182,21 +183,22 @@ impl Mbtiles {
...
     let (min_lng, min_lat) = webmercator_to_wgs84(
-        -20037508.34 + min_x as f64 * tile_size,
-        -20037508.34 + min_y as f64 * tile_size,
+        EARTH_CIRCUMFERENCE * 0.5 * -1.0 + min_x as f64 * tile_size,
+        EARTH_CIRCUMFERENCE * 0.5 * -1.0 + min_y as f64 * tile_size,
     );
     let (max_lng, max_lat) = webmercator_to_wgs84(
-        -20037508.34 + (max_x as f64 + 1.0) * tile_size,
-        -20037508.34 + (max_y as f64 + 1.0) * tile_size,
+        EARTH_CIRCUMFERENCE * 0.5 * -1.0 + (max_x as f64 + 1.0) * tile_size,
```
sharkAndshark: Or just `* -0.5` better?

nyurik (approved):
thanks!  I made a few more cleanups and re-enabled pedantic clippy

PR #1047, opened 2023-12-07
ignore databse_url envvar

p-rutledge:


p-rutledge:
sorry, i meant to merge this to my fork

nyurik:
just curious - is it because you don't want Martin to use any env vars?  Would it make sense to have a `--no-env-vars` flag in general?

PR #1068, opened 2023-12-15
Feature: new `mbtiles diff` command

sharkAndshark:
Implement `mbtiles diff file_a.mbtiles file_b.mbtiles diff.mbtiles`. This should behave exactly the same as `mbtiles copy file_a.mbtiles --diff-with-file file_b.mbtiles diff.mbtiles`.

- [x] Add `mbtiles diff` as a separate command
- [x] Update doc

### DIFF
--- a/docs/src/mbtiles-diff.md
+++ b/docs/src/mbtiles-diff.md
@@ -1,0 +1,13 @@
+# Diffing MBTiles
+
+## `mbtiles diff`
+
+Diff command compares two mbtiles files `file A` and `file B`, and generates a delta (diff) file. If the diff file is [applied](mbtiles-copy.md#mbtiles-apply-patch) to `A`, it will produce `B`.  
+The delta file will contain all tiles that are different between the two files (modifications, insertions, and deletions as `NULL` values), for both the tile and metadata tables.  
+The only exception is `agg_tiles_has` metadata value. It will be renamed to `agg_tiles_hash_in_diff` and a new `agg_tiles_hash` will be generated for the diff file itself.
+
+```shell
+# This command will comapre `file_a.mbtiles` and `file_b.mbtiles`, and generate a new diff file `diff_result.mbtiles`,This command will compares file_a.mbtiles and file_b.mbtiles, and generates a new diff file diff_result.mbtiles 
+# If the diff file is applied to file_a, it will produce file_b. 
+mbtiles diff file_a.mbtiles file_b.mbtiles diff_result.mbtiles // 
+``` 
--- a/docs/src/mbtiles-diff.md
+++ b/docs/src/mbtiles-diff.md
@@ -2,10 +2,10 @@
 
 ## `mbtiles diff`
 
-Diff command compares two mbtiles files `file A` and `file B`, and generates a delta (diff) file. If the diff file
-is [applied](mbtiles-copy.md#mbtiles-apply-patch) to `A`, it will produce `B`.  
-The delta file will contain all tiles that are different between the two files (modifications, insertions, and deletions
-as `NULL` values), for both the tile and metadata tables.  
+Diff command compares two mbtiles files `A` and `B`, and generates a diff (delta) file.
+If the diff file is [applied](mbtiles-copy.md#mbtiles-apply-patch) to `A`, it will produce `B`.  
+The diff file will contain all tiles that are different between the two files
+(modifications, insertions, and deletions as `NULL` values), for both the tile and metadata tables.  
 The only exception is `agg_tiles_has` metadata value. It will be renamed to `agg_tiles_hash_in_diff` and a
 new `agg_tiles_hash` will be generated for the diff file itself.
 
--- a/docs/src/mbtiles-diff.md
+++ b/docs/src/mbtiles-diff.md
@@ -11,6 +11,16 @@ new `agg_tiles_hash` will be generated for the diff file itself.
 
 ```shell
 # This command will comapre `a.mbtiles` and `b.mbtiles`, and generate a new diff file `diff.mbtiles`.
-# If diff.mbtiles is applied to a.mbtiles, it will produce b.mbtiles 
 mbtiles diff a.mbtiles b.mbtiles diff.mbtiles
+
+# If diff.mbtiles is applied to a.mbtiles, it will produce b.mbtiles 
+mbtiles apply-diff a.mbtiles diff.mbtiles b2.mbtiles
+
+# b.mbtiles and b2.mbtiles should now be the same
+# Validate both files and see that their hash values are identical
+mbtiles validate b.mbtiles
+[INFO ] The agg_tiles_hashes=E95C1081447FB25674DCC1EB97F60C26 has been verified for b.mbtiles
+
+mbtiles validate b2.mbtiles
+[INFO ] The agg_tiles_hashes=E95C1081447FB25674DCC1EB97F60C26 has been verified for b2.mbtiles
 ```
--- a/docs/src/mbtiles-diff.md
+++ b/docs/src/mbtiles-diff.md
@@ -10,7 +10,7 @@ The only exception is `agg_tiles_has` metadata value. It will be renamed to `agg
 new `agg_tiles_hash` will be generated for the diff file itself.
 
 ```shell
-# This command will comapre `file_a.mbtiles` and `file_b.mbtiles`, and generate a new diff file `diff_result.mbtiles`,This command will compares file_a.mbtiles and file_b.mbtiles, and generates a new diff file diff_result.mbtiles 
-# If the diff file is applied to file_a, it will produce file_b. 
-mbtiles diff file_a.mbtiles file_b.mbtiles diff_result.mbtiles
+# This command will comapre `a.mbtiles` and `b.mbtiles`, and generate a new diff file `diff.mbtiles`.
+# If diff.mbtiles is applied to a.mbtiles, it will produce b.mbtiles 
+mbtiles diff a.mbtiles b.mbtiles diff.mbtiles
 ```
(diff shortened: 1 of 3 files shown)


mbtiles/src/bin/mbtiles.rs
```diff
@@ -48,6 +48,15 @@ enum Commands {
         /// Value to set, or nothing if the key should be deleted.
         value: Option<String>,
     },
+    #[command(name = "diff")]
+    Diff{
+        /// MBTiles file to read from
+        src_file: PathBuf,
```
nyurik: I think a better pattern is `diff  FILE_A.mbtiles  FILE_B.mbtiles  DIFF.mbtiles`, and the explanation would be that diff will contain all changes going from A to B, i.e. all changes that must be made to A so it becomes B.

mbtiles/src/diff.rs
```diff
@@ -0,0 +1,29 @@
+use std::{path::PathBuf, str::FromStr};
+
+use crate::{Mbtiles, MbtError, MbtResult, is_empty_database};
+
+impl Mbtiles {
+    pub async fn diff(&self, dst_path: PathBuf, diff_path: PathBuf) -> MbtResult<()>{
```
nyurik: I think it is better to have a function outside of the MBtiles that takes all parameters, and produces the result.  Mbtiles is like a wrapper for a single file, not all 3. That said, you will need to at some point to open a file, attach A and B to it, and start writing to it.

mbtiles/src/diff.rs
```diff
@@ -0,0 +1,29 @@
...
+        let is_empty_db = is_empty_database(&mut dst_conn).await?;
+        if !is_empty_db {
+            return Err(MbtError::NonEmptyTargetFile(dst_path.clone()));
+        }
+
+        self.attach_to(&mut dst_conn, "sourceDb").await?;
+
+        todo!()
+    }
+}
```
nyurik: still need that IDE setting

mbtiles/src/diff.rs
```diff
@@ -0,0 +1,29 @@
...
+            return Err(MbtError::SameDiffAndSourceOrDestination(diff_path.clone()));
+        }
+
+        let dst_mbt = Mbtiles::new(&dst_path)?;
+
+        let mut dst_conn = dst_mbt.open_or_new().await?;
+
+        let is_empty_db = is_empty_database(&mut dst_conn).await?;
+        if !is_empty_db {
+            return Err(MbtError::NonEmptyTargetFile(dst_path.clone()));
```
nyurik: it's ok to have a non-empty destination - you simply append to it, just like we do with copying (assuming on-duplicates is set)

nyurik (commented):
good luck, a fun little project :)  I left a few minor comments, but i'm sure lots will change by the end

mbtiles/src/diff.rs
sharkAndshark: I'm using vscode tunnel, I'm not sure it always lost my settings and back to default .🥲

nyurik:
I just did a bit of refactoring to separate Clap args from the internals of the `mbtiles copy` options.  You could simply reuse that directly, without any changes to the diff. This way the new CLI command will be available right away, and then we can do more refactorings later.

nyurik:
@sharkAndshark i merged in the main branch, resolving a few conflicts. Let me know if you need any help with this

sharkAndshark:
@nyurik Thx!! :D Traveling with family, will be back Monday

mbtiles/src/bin/mbtiles.rs
```diff
@@ -212,6 +220,11 @@ async fn main_int() -> anyhow::Result<()> {
             println!("MBTiles file summary for {mbt}");
             println!("{}", mbt.summary(&mut conn).await?);
         }
+        Commands::Diff {
+            file_a: FILE_A,
```
nyurik: don't use caps for variables

mbtiles/src/bin/mbtiles.rs
```diff
@@ -52,6 +53,13 @@ enum Commands {
         /// Value to set, or nothing if the key should be deleted.
         value: Option<String>,
     },
+    /// Generate a diff file which contain all changes going from A to B, all changes that must be made to A so that it becomes B.
```
nyurik:
```suggestion
    /// Compare two files A and B, and generate a new diff file. If the diff file is applied to A, it will produce B.
```

mbtiles/src/bin/mbtiles.rs
```diff
@@ -212,6 +220,11 @@ async fn main_int() -> anyhow::Result<()> {
             println!("MBTiles file summary for {mbt}");
             println!("{}", mbt.summary(&mut conn).await?);
         }
+        Commands::Diff {
+            file_a: FILE_A,
+            file_b: FILE_B,
+            diff_file: DIFF,
+        } => diff(FILE_A, FILE_B, DIFF).await?,
```
nyurik: you do not need to have a dedicated `diff` function. Instead, copy the above code for `Copy` and modify it, e.g.

```rust
            let opts = MbtilesCopier {
                src_file: opts.file_a,
                dst_file: opts.diff_file,
                diff_with_file: opts.file_b,
            };
            opts.run().await?;
```

sharkAndshark: Oh I see.  
*  Will we keep `--diff-with-file` within copy command?
* Should I remove the `diff.rs`?
I'm trying to make a separated `diff` function from `copy.rs`,  seems I'm walking on a totally wrong way? @nyurik

nyurik: Yeah, you don't need a separate diff function -- too complicated, and it will mostly reuse the copy code anyway.  So yes, please remove diff.rs.  In the future, we can hide the `--diff-with-file`, but keep its functionality.  The `CopyArgs` has a field for that, but we can instruct Clap to ignore that field so that users won't be able to specify it from the command line.  For now, I don't think we should remove it just yet.

nyurik:
Looks much better, thx!  TODO: docs and tests

docs/src/mbtiles-diff.md
```diff
@@ -0,0 +1,16 @@
+# Diffing MBTiles
+
+## `mbtiles diff`
+
+Diff command compares two mbtiles files `file A` and `file B`, and generates a delta (diff) file. If the diff file
+is [applied](mbtiles-copy.md#mbtiles-apply-patch) to `A`, it will produce `B`.  
+The delta file will contain all tiles that are different between the two files (modifications, insertions, and deletions
+as `NULL` values), for both the tile and metadata tables.  
+The only exception is `agg_tiles_has` metadata value. It will be renamed to `agg_tiles_hash_in_diff` and a
+new `agg_tiles_hash` will be generated for the diff file itself.
```
nyurik:
```suggestion
Diff command compares two mbtiles files `A` and `B`, and generates a diff (delta) file.
If the diff file is [applied](mbtiles-copy.md#mbtiles-apply-patch) to `A`, it will produce `B`.  
The diff file will contain all tiles that are different between the two files
(modifications, insertions, and deletions as `NULL` values), for both the tile and metadata tables.  
The only exception is `agg_tiles_has` metadata value. It will be renamed to `agg_tiles_hash_in_diff` and a
new `agg_tiles_hash` will be generated for the diff file itself.
```

docs/src/mbtiles-diff.md
```diff
@@ -0,0 +1,16 @@
...
+
+```shell
+# This command will comapre `file_a.mbtiles` and `file_b.mbtiles`, and generate a new diff file `diff_result.mbtiles`,This command will compares file_a.mbtiles and file_b.mbtiles, and generates a new diff file diff_result.mbtiles 
+# If the diff file is applied to file_a, it will produce file_b. 
+mbtiles diff file_a.mbtiles file_b.mbtiles diff_result.mbtiles
```
nyurik:
```suggestion
# This command will comapre `a.mbtiles` and `b.mbtiles`, and generate a new diff file `diff.mbtiles`.
# If diff.mbtiles is applied to a.mbtiles, it will produce b.mbtiles 
mbtiles diff a.mbtiles b.mbtiles diff.mbtiles
```

mbtiles/src/bin/mbtiles.rs
```diff
@@ -524,4 +553,29 @@ mod tests {
...
+            ]),
+            Args {
+                verbose: false,
+                command: Diff {
+                    file_a,
+                    file_b,
+                    diff,
+                }
+            }
+        );
```
nyurik: Simpler is better, plus you don't need to link to the existing files - this is just a CLI parsing test

```suggestion
        assert_eq!(
            Args::parse_from([
                "mbtiles",
                "diff",
                "file-a.mbtiles",
                "file-b.mbtiles",
                "../delta.mbtiles",
            ]),
            Args {
                verbose: false,
                command: Diff {
                    file_a: PathBuf::from("file-a.mbtiles"),
                    file_b: PathBuf::from("file-b.mbtiles"),
                    diff: PathBuf::from("../delta.mbtiles"),
                }
            }
        );
```

nyurik (commented):
looking better and better.  For testing, please also add another portion to https://github.com/maplibre/martin/blob/main/tests/test.sh#L456

Something basic like

```
  $MBTILES_BIN diff \
    ./tests/fixtures/mbtiles/world_cities.mbtiles \
    ./tests/fixtures/mbtiles/world_cities_modified.mbtiles \
    "$TEST_TEMP_DIR/world_cities_diff2.mbtiles" \
    2>&1 | tee "$TEST_OUT_DIR/copy_diff2.txt"
```

nyurik (approved):

nyurik (approved):
Thanks! I think its ready to go.

PR #1070, opened 2023-12-15
Fix bbox_to_xyz

sharkAndshark:
Figure out and fix the bug with `tile-grid` crate or remove the crate and implement it ourself.


martin-tile-utils/src/lib.rs
```diff
@@ -248,13 +258,22 @@ pub fn webmercator_to_wgs84(x: f64, y: f64) -> (f64, f64) {
     (lng, lat)
 }
 
+#[must_use]
+pub fn wgs84_to_webmercator(lon: f64, lat: f64) -> (f64, f64) {
+    let x = PI * 6_378_137.0 * lon / 180.0;
```
nyurik:
```suggestion
    let x = EARTH_CIRCUMFERENCE / 360.0 * lon;
```

martin-tile-utils/src/lib.rs
```diff
@@ -248,13 +258,22 @@ pub fn webmercator_to_wgs84(x: f64, y: f64) -> (f64, f64) {
...
+    let y = ((90.0 + lat) * PI / 360.0).tan().ln() / (PI / 180.0);
+    let y = PI * 6_378_137.0 * y / 180.0;
```
nyurik:
```suggestion
    let y = EARTH_CIRCUMFERENCE / 360.0 * y;
```

martin-tile-utils/src/lib.rs
```diff
@@ -204,6 +204,17 @@ pub fn tile_index(lon: f64, lat: f64, zoom: u8) -> (u32, u32) {
...
 }
 
+#[must_use]
+#[allow(clippy::cast_possible_truncation)]
+#[allow(clippy::cast_sign_loss)]
+pub fn tile_colrow(lng: f64, lat: f64, zoom: u8) -> (u32, u32) {
+    let tile_size = EARTH_CIRCUMFERENCE / f64::from(1_u32 << zoom);
+    let (x, y) = wgs84_to_webmercator(lng, lat);
+    let col = ((x - (EARTH_CIRCUMFERENCE * -0.5)).abs() / tile_size).trunc() as u32;
+    let row = (((EARTH_CIRCUMFERENCE * 0.5) - y).abs() / tile_size).trunc() as u32;
```
nyurik: You don't need to call `trunc` here because that's exactly what `as u32` does - drops the floating point portion.

```suggestion
    let col = ((x - (EARTH_CIRCUMFERENCE * -0.5)).abs() / tile_size) as u32;
    let row = (((EARTH_CIRCUMFERENCE * 0.5) - y).abs() / tile_size) as u32;
```

martin-tile-utils/src/lib.rs
```diff
@@ -248,13 +258,22 @@ pub fn webmercator_to_wgs84(x: f64, y: f64) -> (f64, f64) {
     (lng, lat)
 }
 
+#[must_use]
+pub fn wgs84_to_webmercator(lon: f64, lat: f64) -> (f64, f64) {
+    let x = EARTH_CIRCUMFERENCE / 360.0 * lon;
+    let y = ((90.0 + lat) * PI / 360.0).tan().ln() / (PI / 180.0);
+    let y = EARTH_CIRCUMFERENCE / 360.0 * y;
```
nyurik: i merged the two suggestions here, but as the result, it no longer matches the expected result due to rounding (?).   I suspect that you would need to call `.min((1<<zoom) -1)` somewhere

nyurik: oh, and another suspicious thing - are you sure y should cover 360?  maybe it does, i don't know.  Make sure you read https://wiki.openstreetmap.org/wiki/Slippy_map_tilenames

martin-tile-utils/Cargo.toml
```diff
@@ -21,3 +21,4 @@ tile-grid.workspace = true
 
 [dev-dependencies]
 approx.workspace = true
+insta = { workspace = true, features = ["toml", "yaml"] }
```
nyurik:
```suggestion
insta.workspace = true
```

nyurik:
I made a few minor cleanups and rebased this on main. You may have to reset your local repo though, sorry about that.  At least now i see the "more correct" values (I do not know if they are perfectly ok, or just "somewhat" ok).

If tile_grid has issues like this, we probably should setup a new project for this to have it cleaned up

martin-tile-utils/src/lib.rs
sharkAndshark: This code is from my private server. TBH I don't remember the source....   

I've adopted some code from [esri](https://github.com/Esri/arcgis-osm-editor/blob/e4b9905c264aa22f8eeb657efd52b12cdebea69a/src/OSMWeb10_1/Utils/WebMercator.cs). Sadly i'm bad at math😔,any help?

martin-tile-utils/src/lib.rs
```diff
@@ -294,61 +315,75 @@ mod tests {
...
 
     #[test]
-    fn test_tile_index() {
-        assert_eq!((0, 0), tile_index(-180.0, 85.0511, 0));
+    fn test_tile_colrow() {
+        assert_eq!((0, 0), tile_colrow(-180.0, 85.0511, 0));
     }
 
     #[test]
     fn test_xyz_to_bbox() {
```
sharkAndshark, line 319: Due to the inherent precision limitations of floating point numbers, performing back-and-forth transformations between XYZ and bounding box coordinates within a single unit test  seems not a good idea?    
 I removed the transform back test code. And actually if you add them back, you can see some failed. Any idea? I can't see logic error in these methods. Is it our const not accurate. My mind get lost now.

martin-tile-utils/src/lib.rs
```diff
@@ -248,13 +251,31 @@ pub fn webmercator_to_wgs84(x: f64, y: f64) -> (f64, f64) {
     (lng, lat)
 }
 
+/// transform WGS84 to WebMercator
+// from https://github.com/Esri/arcgis-osm-editor/blob/e4b9905c264aa22f8eeb657efd52b12cdebea69a/src/OSMWeb10_1/Utils/WebMercator.cs
```
sharkAndshark, line 255: It's from esri.

martin-tile-utils/src/lib.rs
sharkAndshark: Or maybe we should reserve `tile-grid` in this?

martin-tile-utils/src/lib.rs
```diff
@@ -196,35 +188,46 @@ impl Display for TileInfo {
...
 #[must_use]
-#[allow(clippy::cast_possible_truncation, clippy::cast_sign_loss)]
-pub fn tile_index(lon: f64, lat: f64, zoom: u8) -> (u32, u32) {
-    assert!(zoom <= MAX_ZOOM, "zoom {zoom} must be <= {MAX_ZOOM}");
-    let tile = web_merc().tile(lon, lat, zoom).unwrap();
-    let max_value = (1_u64 << zoom) - 1;
-    (tile.x.min(max_value) as u32, tile.y.min(max_value) as u32)
+#[allow(clippy::cast_possible_truncation)]
+#[allow(clippy::cast_sign_loss)]
+pub fn tile_colrow(lng: f64, lat: f64, zoom: u8) -> (u32, u32) {
```
nyurik: lets keep this as `tile_index` to reduce the churn

martin-tile-utils/src/lib.rs
```diff
@@ -196,35 +188,46 @@ impl Display for TileInfo {
...
+    let col = (((x - (EARTH_CIRCUMFERENCE * -0.5)).abs() / tile_size) as u32).min((1 << zoom) - 1);
+    let row = ((((EARTH_CIRCUMFERENCE * 0.5) - y).abs() / tile_size) as u32).min((1 << zoom) - 1);
+    (col, row)
 }
 
 /// Convert min/max XYZ tile coordinates to a bounding box values.
 /// The result is `[min_lng, min_lat, max_lng, max_lat]`
 #[must_use]
-pub fn xyz_to_bbox(zoom: u8, min_x: u32, min_y: u32, max_x: u32, max_y: u32) -> [f64; 4] {
+pub fn xyz_to_bbox(zoom: u8, min_col: u32, min_row: u32, max_col: u32, max_row: u32) -> [f64; 4] {
```
nyurik: i'm a bit worried about calling anything `col` and `row` because mbtiles redefines the meaning of that to be "inverted Y".  Lets keep x & y as simpler concepts

martin-tile-utils/src/lib.rs
```diff
@@ -248,13 +251,31 @@ pub fn webmercator_to_wgs84(x: f64, y: f64) -> (f64, f64) {
...
 }
 
+/// transform WGS84 to WebMercator
+// from https://github.com/Esri/arcgis-osm-editor/blob/e4b9905c264aa22f8eeb657efd52b12cdebea69a/src/OSMWeb10_1/Utils/WebMercator.cs
+#[must_use]
+pub fn wgs84_to_webmercator(lon: f64, lat: f64) -> (f64, f64) {
+    (to_mercator_x(lon), to_mercator_y(lat))
+}
+
+fn to_mercator_y(lat: f64) -> f64 {
```
nyurik: we shouldn't create tiny individual functions unless they are used in multiple places

nyurik (commented):
The extra tests are great!  I think we keep using old names (index/x/y) as they are more expected.  I think this is almost ready to go. It is totally ok to not have a "perfect roundtrip" - as long as they are close enough, esp if we reduce the lat/lng by some tiny percentage, we are good to go

martin-tile-utils/src/lib.rs
nyurik: I think this code is actually from the OSM wiki that has all this math

nyurik (approved):
looks great, thanks!

PR #1105, opened 2023-12-25
Implement tile caching

nyurik:
Add a top level config parameter -- the size of cache memory (in MB) to use for caching tiles and PMT directories, defaulting to 512, and 0 to disable. This also removes the `pmtiles.dir_cache_size_mb` parameter (it will be ignored, but will give a warning)

```
cache_size_mb: 512
```

The new cache will contain all tiles as provided by the source.  So if PostgreSQL returns a non-compressed tile, the cache will contain the uncompressed variant, and will be compressed for each response. This will be fixed in the later releases.

Note that fonts and sprites are not cached at this time, and are still a TODO.


donnyv:
How does it know when to dump the cache for a table? If does dump the cache. Does it dump the whole cache or only features that have been updated?

nyurik:
@donnyv there is no cache flushing at the moment, because Martin assumes all data is static. We will need a way to flush cache going forward.

bernatfortet:
> @donnyv there is no cache flushing at the moment, because Martin assumes all data is static. We will need a way to flush cache going forward.

Hi! Thanks for all the work. Martin is awesome.
Were you able to implement cash flushing?

nyurik:
@bernatfortet no, not yet, mostly because we don't actually know what would be the trigger for the refresh - i.e. does the user want some magical header passed in to flush the cache? in that case this is a potential for DOS atack. If its some sort of a server-side command, we need to come up with an api and limit it to `localhost` only.

A much more flexible solution for you would be to setup a Varnish cache proxy - which has all the purging / tile to live / expiration capabilities, and you can even write a simple script that tells Varnish which items should be cached in what way. This way you would also disable Martin's own caching.

bernatfortet:
Gotcha I'll take a look. Thanks @nyurik!

PR #1124, opened 2024-01-09
Make the mbtiles and pmtiles deps optional.

jleedev:
This also makes the `mbtiles-cp` bin optional.

The binary is modestly smaller (by about 6%), but the target directory contains a lot less build products (by about 40%).


nyurik (approved):
Thanks!

jleedev:
I think this broke because sprites was indirectly depending on tokio having certain features enabled.

nyurik:
I'll be happy to merge the moment it passes CI.  Let me know if you need help

Issue #1102, opened 2023-12-25
Document using Martin as a AWS Lambda using PMTiles on S3, or another cloud provider


nyurik:
We urgently need anyone with the AWS Lambda experience to write a short documentation on how to use Martin with Amazon Lambda functions, or similar tech with other cloud providers.  The tiles data should be placed on an AWS bucket, with public access. Martin does not support S3 authentication yet, PRs welcome. Note that the bucket URL should not be publicized - rather it should simply be used by the Lambda function, which provides adequate protection from malicious users who may want to download large data directly, thus raising hosting costs.

The markdown doc file would go into https://github.com/maplibre/martin/tree/main/docs/src

nyurik:
There is now documentation page at https://maplibre.org/martin/run-with-lambda.html, thx to @jleedev - but it only applies to the v0.14 of Martin (not yet released).   I did a pass at rewording it a bit in https://github.com/maplibre/martin/commit/bdce8bb592d10d3b74563c6ac82cec005f8cae97, but it may need some more love:

* [ ] add images with AWS user interface to explain how to do all these steps
* [ ] elaborate why zip deployment should use application model
* [ ] ...?

jleedev:
Some context for the choice of how to talk to AWS (as a human): No real reason except that it's a useful way to ensure that the steps are reproducible. (And AWS has a million ways to do everything.)

Certainly possible to document the other method if that's simpler to follow, e.g: Create an empty directory, download the martin binary to `bin/martin`, write the appropriate script to `bootstrap`, zip the contents (_not_ inside a directory, e.g. `zip martin.zip bootstrap bin/martin`), upload the layer in the console, create the function in the console.

Which, perhaps, could be more instructive than "paste this yaml and a bunch of stuff happens", to explore all the objects and settings involved.

For the container deployment, actually, it's possible to use more SAM and not have to create an ECR by hand, which might be an improvement:

```yaml
AWSTemplateFormatVersion: 2010-09-09
Transform: 'AWS::Serverless-2016-10-31'
Resources:
  demotiles:
    Type: 'AWS::Serverless::Function'
    Properties:
      Architectures:
        - arm64
      PackageType: Image
      ImageConfig:
        Command:
          - '<tile url here>'
      FunctionUrlConfig:
        AuthType: NONE
    Metadata:
      Dockerfile: Dockerfile
      DockerContext: .
```

[...]

PR #1127, opened 2024-01-11
Add runtime for AWS Lambda

jleedev:
This adds the lambda-web crate to adapt the actix App to speak to Lambda by way of the lambda_runtime crate.

Briefly: AWS Lambda has native support for scripting languages to execute a function directly; compiled languages must embed a runtime to fetch incoming events from Lambda and post the responses. This detects the environment variables to start up in Lambda mode instead of the normal HTTP server, and is added as an optional feature.

Lambda has five (!) distinct ways of routing HTTP requests to a function; this supports some of them. (Specifically, the most obvious way to do this is with a Function URL, which is newest and simplest, and perhaps with CloudFront, which speaks to the Function URL and not Lambda directly.)

The error handling could probably be refined, I was just trying to get this to compile.

(Supported: API Gateway HTTP API with payload format version 2.0; API Gateway REST API; Lambda function URLs / Not supported: API Gateway HTTP API with payload format version 1.0; Application Load Balancer)

Necessary for #1102 to be able to run the released packages directly, and only having to configure the appropriate environment.

### TODOs

* [x] Update Book Documentation
* [ ] Update README
* [x] Add CI test(s)


martin/Cargo.toml
```diff
@@ -60,8 +60,9 @@ name = "bench"
 harness = false
 
 [features]
-default = ["fonts", "mbtiles", "pmtiles", "postgres", "sprites"]
+default = ["fonts", "lambda", "mbtiles", "pmtiles", "postgres", "sprites"]
```
nyurik, line 63: do we want to group all lambda supports into one group, or will we need to use different dependencies/code-paths for aws vs gcp vs azure?

martin/src/srv/server.rs
```diff
@@ -127,15 +127,32 @@ pub fn new_server(config: SrvConfig, state: ServerState) -> MartinResult<(Server
...
-    .shutdown_timeout(0)
-    .workers(worker_processes)
-    .run();
-
-    Ok((server, listen_addresses))
+    };
+
+    match () {
+        #[cfg(feature = "lambda")]
+        () if is_running_on_lambda() => {
```
nyurik: why not a simple block like this?  Why do you need a match statement?

```
#[cfg(feature = "lambda")]
if is_running_on_lambda() {
    ....
}
```

nyurik (commented):
looks great!  My biggest concern is how can we test this, especially as part of the CI?

martin/src/srv/server.rs
nyurik: P.S. we need to document that a new env var variable has magical properties per https://github.com/hanabu/lambda-web/blob/ddceea3b8ffcfae465fc8d0059aaa52a8ff94f8e/src/lib.rs#L34 

if `AWS_LAMBDA_RUNTIME_API` is defined, runs under AWS lambda runtime

martin/Cargo.toml
jleedev: Fortunately, only Lambda (to my knowledge) requires shenanigans like this, everything else supports regular HTTP servers.

martin/src/srv/server.rs
jleedev: I think I was frustrated that an if/else wouldn't work with the conditional compilation, and this is a way to express that only one of the code paths will run. But an if statement and a return would probably be clearer.

nyurik: I may refactor it later, no worries. I think we should focus on the main issue - Testing, and ideally - testing in CI.  We need a clear way to ensure that after some person adds some random code, we don't have to do too much work to make sure it continues to work in all expected scenarios.

jleedev:
> looks great! My biggest concern is how can we test this, especially as part of the CI?

Looks like AWS SAM (Serverless Application Model) is the harness to use: https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-cli-using-automated-tests.html

Hah, this was surprisingly painless:

```yaml
~/sam-app$ cat template.yaml
AWSTemplateFormatVersion: 2010-09-09
Transform: 'AWS::Serverless-2016-10-31'
Resources:
  MartinFunction:
    Type: 'AWS::Serverless::Function'
    Properties:
      PackageType: Image
      Architectures:
        - arm64
      ImageUri: martin
      ImageConfig:
        Command:
          - 'https://jleedev-planet.s3.us-east-2.amazonaws.com/demotiles.pmtiles'
      FunctionUrlConfig:
        AuthType: NONE
```

```bash
~/sam-app$ sam local generate-event apigateway http-api-proxy | jq '.rawPath="/catalog"|.requestContext.http.method="GET"' >event.json 
~/sam-app$ sam local invoke --event event.json 
[…]
{"body": "eyJ0aWxlcyI6eyJkZW1vdGlsZXMiOnsiY29udGVudF90eXBlIjoiYXBwbGljYXRpb24veC1wcm90b2J1ZiIsImNvbnRlbnRfZW5jb2RpbmciOiJnemlwIiwibmFtZSI6Im1hcGxpYnJlIiwiZGVzY3JpcHRpb24iOiIiLCJhdHRyaWJ1dGlvbiI6IlJlbmRlcmVkIHdpdGggPGEgaHJlZj1cImh0dHBzOi8vd3d3Lm1hcHRpbGVyLmNvbS9kZXNrdG9wL1wiPk1hcFRpbGVyIERlc2t0b3A8L2E+In19LCJzcHJpdGVzIjp7fSwiZm9udHMiOnt9fQ==", "cookies": [], "headers": {"content-type": "application/json", "vary": "Origin, Access-Control-Request-Method, Access-Control-Request-Headers"}, "isBase64Encoded": true, "statusCode": 200}
```

```bash
~/sam-app$ sam local start-lambda
~/sam-app$ AWS_ACCESS_KEY_ID=x AWS_SECRET_ACCESS_KEY=x AWS_REGION=us-east-1 aws --endpoint http://localhost:3001 lambda invoke --function-name MartinFunction --payload file://event.json --cli-binary-format raw-in-base64-out /dev/stdout 
{"body": "eyJ0aWxlcyI6eyJkZW1vdGlsZXMiOnsiY29udGVudF90eXBlIjoiYXBwbGljYXRpb24veC1wcm90b2J1ZiIsImNvbnRlbnRfZW5jb2RpbmciOiJnemlwIiwibmFtZSI6Im1hcGxpYnJlIiwiZGVzY3JpcHRpb24iOiIiLCJhdHRyaWJ1dGlvbiI6IlJlbmRlcmVkIHdpdGggPGEgaHJlZj1cImh0dHBzOi8vd3d3Lm1hcHRpbGVyLmNvbS9kZXNrdG9wL1wiPk1hcFRpbGVyIERlc2t0b3A8L2E+In19LCJzcHJpdGVzIjp7fSwiZm9udHMiOnt9fQ==", "cookies": [], "headers": {"content-type": "application/json", "vary": "Origin, Access-Control-Request-Method, Access-Control-Request-Headers"}, "isBase64Encoded": true, "statusCode": 200}{
    "StatusCode": 200
}

File: martin/src/bin/martin.rs [hunk 3]
```diff
@@ -9,7 +9,7 @@ use martin::{read_config, Config, MartinResult};
 
 const VERSION: &str = env!("CARGO_PKG_VERSION");
 
-async fn start(args: Args) -> MartinResult<Server> {
+async fn start(args: Args) -> MartinResult<impl Future<Output = MartinResult<()>>> {
```
nyurik: this looks ... scary. Could you add a comment explaining why you have future of a future here, or if we can somehow simplify this?  Basically here we have a desugared `Future<Result<Future<Result>>>>`, right?

martin/src/srv/server.rs
```diff
@@ -98,16 +100,14 @@ pub fn router(cfg: &mut web::ServiceConfig) {
     cfg.service(crate::srv::fonts::get_font);
 }
 
-/// Create a new initialized Actix `App` instance together with the listening address.
-pub fn new_server(config: SrvConfig, state: ServerState) -> MartinResult<(Server, String)> {
+/// Create a future for an Actix web server together with the listening address.
+pub fn new_server(
+    config: SrvConfig,
+    state: ServerState,
+) -> MartinResult<(LocalBoxFuture<'static, MartinResult<()>>, String)> {
```
nyurik: same thing as above - any way to simplify the result of this, or at least lots of comments?

nyurik (commented):
much simpler, thanks!  I added a few more comments, but this is looking better and better, almost ready to go.  I will add a few TODOs to the top comment.

nyurik:
@jleedev hi, let me know if you need any help with this PR, i really do want to get this merged in, but no pressure! :)

jleedev:
Current status:

Figured out how to build and test a Lambda function on GitHub actions. The `sam local invoke` command is straightforward but starts and stops the server each time, which would be suboptimal for running a full test suite. The `sam local start-lambda --warm-containers EAGER` is better, but requires actually using the AWS client to send the Invoke RPC. Rigging that up with bash and the aws cli is not extremely terrible.

Also requires having a (about ten line) yaml SAM template to actually run it, which is sort of useless boilerplate; the point of this change is so that the binary will "just work" when tossed into Lambda, and perhaps a sample template that someone can launch, which isn't really the same as a template whose only purpose is to run the integration tests.

At any rate, I propose adding a CI test that invokes the lambda once, to verify that it works in a minimal sense, and a followup can integrate the test suite fully. Translating the normal HTTP calls to Lambda invocations, etc.

I'll address the type comments above. If you could make the documentation changes that would be good I think.

martin/src/bin/martin.rs
jleedev: Exactly, the actix `Server` type was just being used as a future, and desugars to this. I note that the main function is already awkward with the `.await.unwrap().await.unwrap()`.

The only reason to do anything with the server besides awaiting it immediately would be to pause or stop it, and also to print out the listening address as we do here.

Defining a type alias for what the "Server" is would reduce the clutter in the other file, but this function can really be changed to return `MartinResult<()>`, since this is essentially the main.

nyurik:
Thanks, sounds good - a simple Lambda smoke test would be a great start that we can improve upon later.  Adding a doc page is easy - I created a simple placeholder https://github.com/jleedev/martin/pull/1  - but the actual "how to" is a mystery to me - i have never used lambdas.

jleedev:
Well, that worked in that SAM actually started martin with no arguments and thus failed during the Init phase. And then neglected to fail the CI.

nyurik:
I'll have to think of which CI step to move this to.  Most likely it will be part of the linux one, maybe even a separate test that simply downloads the linux binary from the temp storage, and runs it without postgres database - e.g. it will simply use a small test pmtiles.  Ideally, it would be great if we can simulate an S3 bucket that the lambda accesses - in that case we can place a sample pmtiles into it, and the lambda runner will run martin which will have just one CLI parameter - the URL of the s3 file.

nyurik:
@jleedev for CI, add something like this into ci.yml around line 291:

```yml
  test-aws-lambda:
    name: Test AWS Lambda
    runs-on: ubuntu-latest
    needs: [ build ]
    steps:
      - name: Checkout sources
        uses: actions/checkout@v4
      - name: Download build artifact build-x86_64-unknown-linux-gnu
        uses: actions/download-artifact@v3
        with:
          name: build-x86_64-unknown-linux-gnu
          path: target/
      - run: chmod +x target/martin
      - run: test/test-aws-lambda.sh
        env:
          MARTIN_BIN=target/martin
```

nyurik:
P.S. Actually an even better path would be to write a simple script similar to `tests/test.sh`, e.g.  `tests/test-aws-lambda.sh` -- which would have some of the bits and pieces of the test.sh script:

```bash
#!/usr/bin/env bash
set -euo pipefail

MARTIN_PORT="${MARTIN_PORT:-3111}"
MARTIN_URL="http://localhost:${MARTIN_PORT}"
MARTIN_ARGS="${MARTIN_ARGS:---listen-addresses localhost:${MARTIN_PORT}}"  # adjust this to use lambda?
MARTIN_BIN="${MARTIN_BIN:-target/debug/martin} ${MARTIN_ARGS}"

function wait_for {
  # copy this function from test.sh
}

function kill_process {
  # copy this function from test.sh
}

$MARTIN_BIN "${ARG[@]}"  &
MARTIN_PROC_ID=`jobs -p | tail -n 1`
trap "echo 'Stopping Martin server $MARTIN_PROC_ID...'; kill -9 $MARTIN_PROC_ID 2> /dev/null || true; echo 'Stopped Martin server $MARTIN_PROC_ID';" EXIT HUP INT TERM
wait_for $MARTIN_PROC_ID Martin "$MARTIN_URL/health"

# do whatever tests needed

kill_process $MARTIN_PROC_ID Martin
```

You may also add the logging parts to keep it even more consistent

nyurik:
@jleedev please run `just fmt-md` to fix some markdown issues. Thx!

jleedev:
Regarding this libc thing: I used the docker images that the workflow produced in my fork, and it ran successfully on AWS Lambda. The docker image uses musl.

The release tarball uses gnu. I just tried on CloudShell, and it fails in the same way as on Lambda:

```
$ curl -OL https://github.com/maplibre/martin/releases/download/v0.13.0/martin-x86_64-unknown-linux-gnu.tar.gz
$ tar -xzvf martin-x86_64-unknown-linux-gnu.tar.gz     
$ ./martin 
./martin: /lib64/libm.so.6: version `GLIBC_2.35' not found (required by ./martin)
$ cat /etc/os-release 
NAME="Amazon Linux"
VERSION="2023"
ID="amzn"
ID_LIKE="fedora"
VERSION_ID="2023"
PLATFORM_ID="platform:al2023"
PRETTY_NAME="Amazon Linux 2023"
```

So the Lambda documentation (and CI step) should specify to pull from the container image, not the release tarballs.

nyurik:
> So the Lambda documentation (and CI step) should specify to pull from the container image, not the release tarballs.

We have both `-gnu` and `-musl` in the releases. Do they both fail?  BTW, feel free to ping me on slack to discuss

jleedev:
Aha, I was confusing the "Artifacts" section of the run with the Release assets. So the CI must extract the cross-build zip artifact to get at the musl build, and the documentation can refer to the musl release (it already does).

.github/files/lambda-function/Makefile
```diff
@@ -0,0 +1,2 @@
+build-MartinFunction:
```
nyurik, line 1: where is this makefile being used?

.github/files/lambda-layer/bootstrap
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+set -eu
+exec martin -c "${_HANDLER}"
```
nyurik, line 3: will `martin` be available in the PATH?  I guess it is if it passes tests?

docs/src/run-with-lambda.md
```diff
@@ -0,0 +1,124 @@
...
+Download the binary and place it in your staging directory. The `bin` directory of your Layer will be added to the PATH.
+
+```bash
+mkdir -p martin_layer/src/bin/
+cd martin_layer
+curl -OL https://github.com/maplibre/martin/releases/download/VERSION_NUMBER_HERE/martin-aarch64-unknown-linux-musl.tar.gz
+tar -C src/bin/ -xzf martin-aarch64-unknown-linux-musl.tar.gz martin
+```
+
+Every zip-based Lambda function runs a file called `bootstrap`. `vim src/bootstrap`
```
nyurik: better use heredoc for this - easier, and does not require familiarity with `vim`.  Might also use `#!/usr/bin/env sh`, but not certain which shebang is better.

```
cat <<EOF > src/bootstrap
#!/bin/sh
set -eu
exec martin -c ${_HANDLER}.yaml
EOF
```

docs/src/run-with-lambda.md
```diff
@@ -0,0 +1,124 @@
...
+
+```bash
+#!/bin/sh
+exec martin -c ${_HANDLER}.yaml
+```
+
+Write the SAM template. `vim template.yaml`
```
nyurik: use heredoc as above

docs/src/run-with-lambda.md
```diff
@@ -0,0 +1,124 @@
...
+
+   ```yaml
+   pmtiles:
+     sources:
+       demotiles: <url to a pmtiles file>
+   ```
+
+2. Click Deploy, wait for the success banner, and visit your function URL.
+
+### Extra things
```
nyurik: "things" here looks weird - either `Extras` or `TODO` ?

```suggestion
### TODO
```

docs/src/env-vars.md
```diff
@@ -9,3 +9,4 @@ You can also configure Martin using environment variables, but only if the confi
 | `PGSSLCERT` <br/> `ssl_cert`             | `./postgresql.crt`                   | A file with a client SSL certificate. [docs](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-SSLCERT)                                                                             |
 | `PGSSLKEY` <br/> `ssl_key`               | `./postgresql.key`                   | A file with the key for the client SSL certificate. [docs](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-SSLKEY)                                                                |
 | `PGSSLROOTCERT` <br/> `ssl_root_cert`    | `./root.crt`                         | A file with trusted root certificate(s). The file should contain a sequence of PEM-formatted CA certificates. [docs](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-SSLROOTCERT) |
+| `AWS_LAMBDA_RUNTIME_API`                 |                                      | If defined, connect to AWS Lambda to handle requests. The regular HTTP server is not used.                                                                                                                         |
```
nyurik: This should probably link to the book's page on AWS lambda

docs/src/run-with-lambda.md
```diff
@@ -0,0 +1,124 @@
...
+
+1. Click “Create function”.
+2. Choose “Container image”.
+3. Put something in “Function name”. (Note: This is an internal identifier, not exposed in the function URL.)
+4. Click “Browse images”, and select your repository and the tag. (If you can’t find it, see if you’re in the same region?)
+5. Expand “Container image overrides”, and under CMD put the URL of a .pmtiles file.
+6. Set “Architecture” to arm64 to match the platform that we pulled. (Lambda has better ARM CPUs than x86.)
+7. Click “Create function”.
+8. Find the “Configuration” tab, select “Function URL”, “Create function URL”.
+9. Set “Auth type” to NONE. (Don’t enable CORS, as Martin already will, and you will have duplicate headers and your CORS will be broken.)
```
nyurik:
```suggestion
9. Set “Auth type” to `NONE`
   * Do not enable CORS. Martin already has CORS support, so it will create duplicate headers and break CORS.
```

docs/src/run-with-lambda.md
```diff
@@ -0,0 +1,124 @@
...
+It’s possible to deploy the entire codebase from the AWS console, but we will use Serverless Application Model. Our function will consist of a “Layer”, containing the Martin binary, and our function itself will contain the configuration in .yaml format.
+
+#### The layer
+
+Download the binary and place it in your staging directory. The `bin` directory of your Layer will be added to the PATH.
+
+```bash
+mkdir -p martin_layer/src/bin/
+cd martin_layer
+curl -OL https://github.com/maplibre/martin/releases/download/VERSION_NUMBER_HERE/martin-aarch64-unknown-linux-musl.tar.gz
```
nyurik: Use `latest` instead of requiring version number

```suggestion
curl -OL https://github.com/maplibre/martin/releases/latest/download/martin-aarch64-unknown-linux-musl.tar.gz
```

docs/src/run-with-lambda.md
```diff
@@ -0,0 +1,124 @@
...
+Outputs:
+  LayerArn:
+    Value: !Ref MartinLayer
+    Export:
+      Name: !Sub "${AWS::StackName}-LayerArn"
+```
+
+Run `sam deploy --guided`.
+
+1. Stack Name: [martin-layer]
```
nyurik: broken link?

docs/src/run-with-lambda.md
```diff
@@ -0,0 +1,124 @@
...
+1. Select “Functions”, “Create function”.
+2. Put something in “Function name”.
+3. Set “Runtime” to “Amazon Linux 2023”.
+4. Set “Architecture” to “arm64”.
+5. Under “Advanced settings”, choose “Enable function URL” with “Auth type” of “NONE”.
+6. Click “Create function”.
+
+Add your layer:
+
+1. Click “add a layer” (green banner at the top, or the very bottom).
```
nyurik, line 106: in cases like these, you may want to add screenshot images, but we can do this in a separate PR.

docs/src/run-with-lambda.md
```diff
@@ -0,0 +1,124 @@
...
+     sources:
+       demotiles: <url to a pmtiles file>
+   ```
+
+2. Click Deploy, wait for the success banner, and visit your function URL.
+
+### Extra things
+
+- Lambda has a default timeout of 3 seconds, and 128 MB of memory, maybe this is suboptimal.
+- Martin will cache tiles in memory, but this might be useless for an ephemeral server.
```
nyurik: I don't think this is useless - if martin binary will handle more than one request, it will be better to cache at least some things - for example, when using pmtiles, cache the directory structure.

martin/src/bin/martin.rs
```diff
@@ -35,20 +34,15 @@ async fn start(args: Args) -> MartinResult<Server> {
...
 async fn main() {
     let env = env_logger::Env::default().default_filter_or("martin=info");
     env_logger::Builder::from_env(env).init();
 
-    start(Args::parse())
-        .await
-        .unwrap_or_else(|e| on_error(e))
-        .await
-        .unwrap_or_else(|e| on_error(e));
+    start(Args::parse()).await.unwrap_or_else(|e| on_error(e));
```
nyurik: probably better to inline the on_error here. Also, I noticed the same issue in `martin-cp` if you want to inline it there at the same time.

```suggestion
    start(Args::parse()).await.unwrap_or_else(|e| {
        if log_enabled!(log::Level::Error) {
            error!("{e}");
        } else {
            eprintln!("{e}");
        }
        std::process::exit(1);
    });
```

tests/test-aws-lambda.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
```
nyurik, line 1: were you able to run this script locally?  If so, could you add a new recipe to the `justfile` too, showing how it can be run/tested locally?  thx

nyurik (commented):
awesome!!!  i think its almost done. I added a few nits that should be very easy to fix, and just to clarify for myself if they are needed and how they are being used.  Thanks again!

.github/files/lambda-function/Makefile
jleedev: Implicitly by `sam build` of the AWS::Serverless::Function. It _requires_ a Makefile. (Or like, something such as Java or JavaScript or Python which it knows how to build.)

(But AWS::Serverless::LayerVersion is happy to just zip up a directory. Which is a little bit silly since that's the complicated one and the function code is just one yaml, but oh well.)

See: https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/building-custom-runtimes.html and https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/building-rust.html

Note that they also have a `cargo lambda` utility which is in beta, but might be worth a look later.

.github/files/lambda-layer/bootstrap
jleedev: Yes, because it's under `bin/` in the zip file, and the layer extracts somewhere on Lambda's path.

docs/src/run-with-lambda.md
jleedev: Looks like AL2023's /bin/sh is "GNU bash, version 5.2.15(1)-release (x86_64-amazon-linux-gnu)", and they also have a merged / and /usr. So it should be fine.

I'm amused to learn that `#!/bin/sh` is actually not a POSIX compliant thing to do, as the path to `sh` cannot be assumed.

Anyway, `/bin/sh` causes Bash to start up in its posix mode, and refers to Dash on Debian and Ubuntu systems, but this small shell script shouldn't encounter any differences :)

tests/test-aws-lambda.sh
jleedev: Ooh, even better:

```bash
 ¶ git diff .github/files/lambda.yaml
@@ -17,3 +17,6 @@ Resources:
         - Ref: MartinLayer
       CodeUri: lambda-function/
       Handler: config.yaml
+      Events:
+        ApiEvent:
+          Type: HttpApi
 ¶ sam local start-api --warm-containers eager
```

```bash
curl http://127.0.0.1:3000
```

Function URL is the best choice for running on AWS, but for local emulation we can configure an API Gateway HTTP API, and then send it HTTP requests instead of having to use `aws lambda invoke` with specially crafted event objects. Good for a followup.

jleedev:
Made all the changes.

nyurik:
@jleedev there are a few minor comments in the diff view

jleedev:
Think I got them all now!

tests/test-aws-lambda.sh
nyurik: @jleedev would it make sense to create an issue to track it? Won't remember otherwise.

nyurik (approved):
awesome work, thank you!!!

We may want to create follow up issues to remind later of any further tasks, but this is totally up to you :)

Issue #1149, opened 2024-01-22
Support docker compose plugin to justfile


sharkAndshark:
The [Docker compose plugin](https://docs.docker.com/compose/install/linux/) way:
```shell
docker compose up
```

The [Docker compose standalone](https://docs.docker.com/compose/install/standalone/) way:
```shell
docker-compose up
```

Is there one way to support them both? I tried many in justfile but all failed.

nyurik:
I think we may need to test if the system has a `docker-compose`, and if it does not, use `docker compose` - something similar to the check we have in the `cargo-install` just command - but do it once when just starts up?

nyurik:
A temporary workaround could be to add an alias to bash

justb4:
This is a common problem, that will still remain some years, until `docker compose` is the only command.
We solved this in a workshop launch control Bash script like @nyurik indicated: sniff for the variant and set an `alias`. See [these lines](https://github.com/geopython/geopython-workshop/blob/master/workshop/geopython-workshop-ctl.sh#L14) i.e.:

```
# Sniff which Docker Compose variant is installed
# and set an alias.
# See https://github.com/geopython/geopython-workshop/issues/82
if command docker-compose --version &> /dev/null
then
  alias dockercompose='docker-compose'
  echo "Using docker-compose"
else
  if !command docker compose version &> /dev/null
  then
    echo "Neither docker-compose nor docker compose is available"
    echo "Check your Docker Installation"
    exit 1
  fi
  alias dockercompose='docker compose'  
  echo "Using docker compose"
fi

```
..and then use the alias `dockercompose`. Never heard of `just` (other than my first name..) but seems something to be added in the [justfile](https://github.com/maplibre/martin/blob/main/demo/justfile) as `bash` is presumed? Will PR...

justb4:
The "bash-script" approach in my comment above does not really work direct in a `justfile` because of the 'just-syntax'. But figured out an alternative, using standard var and -substitution, so an approach with oneliner:

```
dockercompose := `if docker-compose --version &> /dev/null; then echo "docker-compose"; else echo "docker compose"; fi`
.
.
up:
   {{ dockercompose }} up

up-build:
    {{ dockercompose }} up --build
.
.
```
 
There is a `justfile` in the repo root and the `demo` dir. Will go ahead with this approach.

PR #1153, opened 2024-01-25
Adapt justfiles for Docker Compose command

justb4:
This should solve #1149. See also [my comment there for the approach](https://github.com/maplibre/martin/issues/11). In short through a variable that is substituted. Note that this may not work if an older Docker version is installed without the Python `docker-compose` script...As both variants will not be present. Maybe add an extra test+warning msg for that case.

```
dockercompose := `if docker-compose --version &> /dev/null; then echo "docker-compose"; else echo "docker compose"; fi`
.
.
up:
   {{ dockercompose }} up

up-build:
    {{ dockercompose }} up --build
.
.
```

### DIFF
--- a/demo/justfile
+++ b/demo/justfile
@@ -2,6 +2,8 @@
 
 set shell := ["bash", "-c"]
 
+dockercompose := `if docker-compose --version &> /dev/null; then echo "docker-compose"; else echo "docker compose"; fi`
+
 @_default:
     just --list --unsorted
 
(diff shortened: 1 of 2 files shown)


sharkAndshark (approved):

sharkAndshark:
Thx @justb4  It works!  Now I can run recipes without docker-compose standalone.

nyurik (approved):
thx!

PR #1165, opened 2024-01-31
Update macos runner

### LINKED ISSUE #1162: Migrate MacOS CI to use new MacOS M1 runners
Per [blog post](https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/) the M1 MacOS runners are now available on GitHub CI, so we need to migrate to them, instead of the nasty hacks we have.  Help wanted :)

sharkAndshark:
Try to fix #1162

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -248,7 +248,7 @@ jobs:
           - target: debian-x86_64
             os: ubuntu-latest
           - target: x86_64-apple-darwin
-            os: macos-14  # M1 CPU
+            os: macos-12 # x64 CPU
           - target: x86_64-pc-windows-msvc
             os: windows-latest
             ext: '.exe'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -244,11 +244,11 @@ jobs:
       matrix:
         include:
           - target: aarch64-apple-darwin
-            os: macOS-latest
+            os: macos-14
           - target: debian-x86_64
             os: ubuntu-latest
           - target: x86_64-apple-darwin
-            os: macOS-latest
+            os: macos-14 
           - target: x86_64-pc-windows-msvc
             os: windows-latest
             ext: '.exe'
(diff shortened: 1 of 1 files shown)


nyurik (approved):
I updated the runners a bit.  Sadly, it seems it might be trickier to actually test nginx on M1 -- so up to you - we can merge this as is, and work on the nginx test on M1 separately, or do it all in this PR

sharkAndshark:
Thx @nyurik ! I will try Ngnix on my local M1 MacBook and update this PR then.

.github/workflows/ci.yml
```diff
@@ -248,7 +248,7 @@ jobs:
           - target: debian-x86_64
             os: ubuntu-latest
           - target: x86_64-apple-darwin
-            os: macos-12  # x64 CPU
+            os: macos-14  # M1 CPU
```
nyurik: You don't want M1 CPU here because we are building x86 target

sharkAndshark: Oh ok. I thought we can build x86 target on M1, can we?

sharkAndshark: I thought we could generate x86 and arm target both on M1, I'm not sure.

sharkAndshark: Oh I see I see.

sharkAndshark: Seems M1 could build and run(by Rosetta) `x86_64-apple-darwin`, but it's not the best choice. I've already switch back to macos-12 runner. Thx @nyurik

.github/workflows/ci.yml
```diff
@@ -320,6 +320,7 @@ jobs:
       - name: Start HTTP Server (MacOS, no Docker)
         if: runner.os == 'macos'
         run: |
+          mkdir -p /usr/local/etc/nginx
```
nyurik, line 323: i'm not sure we should create a directory that nginx doesn't know about.  We may have to just download nginx from somewhere...    anyway, i think it might not be worth our time at this stage - until they install nginx on -13 and -14 the same way they have it on -12 (latest).

sharkAndshark: We could do more when[ `macos-14` is stable, now is `beta`](https://docs.github.com/en/actions/using-jobs/choosing-the-runner-for-a-job#choosing-github-hosted-runners).  There are many issues bigger than #1162.

PR #1182, opened 2024-02-07
write! perf optimization - use write_str

nyurik:


sharkAndshark (approved):

sharkAndshark:
I remember we had a performance bench in CI before ( I‘m not sure. Maybe it's another repo), could we add it back?

nyurik:
Its a bit tricky - because performance requires a lot of specialized hardware configurations, and not really doable with Github runners.  So easier to just do it adhoc as needed.

Issue #1185, opened 2024-02-08
Add `--base-path` CLI option to override the URL path in the tilejson


nyurik:
As discussed in #1184, Martin should have an option to override the base path when its being hosted behind a proxy that cannot easily set `x-rewrite-url` header.

## CLI parameter

```shell
# Force Martin to ignore the x-rewrite-url header and set tilejson's tiles URL
# as   <protocol>://<host>/<source>/{z}/{x}/{y}
martin --base-path /

# Force it to be  <protocol>://<host>/foo/bar/<source>/{z}/{x}/{y}
martin --base-path /foo/bar

# This should work the same as above
martin --base-path /foo/bar/

# This should produce an error
# (must start with a slash, but could be relaxed in the future, e.g. can combine path with x-rewrite-uri?)
martin --base-path foo/bar
```

## Configuration Files

Add `base_path` to the root (in the `struct SrvConfig`)

```yaml
---
base_path: /foo/bar
```

## Code changes
- [ ] Add `base_path: Option<String>` to [SrvArgs](https://github.com/maplibre/martin/blob/79a89125edd6529177b00bca8d09fc9e9fb898a5/martin/src/args/srv.rs#L5) - similar to how `listen_addresses` is implemented
- [ ] Add `base_path` to [SrvConfig](https://github.com/maplibre/martin/blob/main/martin/src/srv/config.rs#L8)
- [ ] Similar to catalog, pass `Data::new(config)` to the `app_data` in [new_server](https://github.com/maplibre/martin/blob/79a89125edd6529177b00bca8d09fc9e9fb898a5/martin/src/srv/server.rs#L126)
- [ ] Modify [`get_source_info`](https://github.com/maplibre/martin/blob/79a89125edd6529177b00bca8d09fc9e9fb898a5/martin/src/srv/tiles_info.rs#L28) to additionally take `config: Data<SrvConfig>` and use it to override the `x-rewrite-url` handling.
- [ ] Make sure [`finalize`](https://github.com/maplibre/martin/blob/79a89125edd6529177b00bca8d09fc9e9fb898a5/martin/src/config.rs#L73) does all the needed validations of this value or it should return an error. You will need to add a new error to `MartinError` for this.

sharkAndshark:
Do we need to add it to config either?

nyurik:
@sharkAndshark I just updated this ticket with all the details needed

sutanAE:
> Do we need to add it to config either?

It would be nice to add this on config.

[...]

PR #1189, opened 2024-02-11
Add `--preferred-encoding (gzip|brotli)` to use when tile is not pre-encoded by source

### LINKED ISSUE #1178: Implement `--preferred-encoding (gzip|brotli)` configuration
Brotli encoding produces smaller content, but is slower to encode.  GZip is the opposite of that. Until v0.11, Martin used GZIP as the default, but later switched to Brotli, causing slower performance for some users, as reported in https://github.com/maplibre/martin/issues/1172

Per https://github.com/maplibre/martin/issues/1172#issuecomment-1926047552, we need a new configuration parameter in both CLI and in config.  CLI could use `--preferred-encoding (gzip|brotli)` (should also allow `"br"` for consistency with `Accept-Encoding` header, and config:

```
# root level
preferred_encoding: brotli
```

This setting will only work if the client provided [Accept-Encoding header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding) where `gzip` and `br` have the same top `q` value (the `q` value could be missing for both, which also makes it the same 1.0).  In theory, the client may prefer `identity`, which we also support.

sharkAndshark:
Try to fix #1178 

- [x] Add optional `--preferred-encoding` to cli, `br` `brotli` and `gzip` are allowed
- [x] Add optional `preferred-encoding` to config, `br` `brotli` and `gzip` are allowed
- [x] Add test
- [x] Update doc

### DIFF
--- a/docs/src/config-file.md
+++ b/docs/src/config-file.md
@@ -27,7 +27,7 @@ worker_processes: 8
 # Amount of memory (in MB) to use for caching tiles [default: 512, 0 to disable]
 cache_size_mb: 1024
 
-# Preferred tiles encoding, gzip or brotli, default brotili. You could also use br as a shortcut for brotli
+# If the client accepts multiple compression formats, and the tile source is not pre-compressed, which compression should be used. `gzip` is faster, but `brotli` is smaller, and may be faster with caching.  Defaults to brotli.
 preferred_encoding: gzip
 
 # Database configuration. This can also be a list of PG configs.
--- a/docs/src/config-file.md
+++ b/docs/src/config-file.md
@@ -27,6 +27,9 @@ worker_processes: 8
 # Amount of memory (in MB) to use for caching tiles [default: 512, 0 to disable]
 cache_size_mb: 1024
 
+# Preferred tiles encoding, gzip or brotli, default brotili. You could also use br as a shortcut for brotli
+preferred_encoding: gzip
+
 # Database configuration. This can also be a list of PG configs.
 postgres:
   # Database connection string. You can use env vars too, for example:
--- a/docs/src/run-with-cli.md
+++ b/docs/src/run-with-cli.md
@@ -32,7 +32,7 @@ Options:
           Number of web server workers
 
       --preferred-encoding <PREFERRED_ENCODING>
-          Preferred tiles encoding. gzip or brotli, default brotili. You could also use br as a shortcut for brotli
+          Martin server preferred tile encoding. If the client accepts multiple compression formats, and the tile source is not pre-compressed, which compression should be used. `gzip` is faster, but `brotli` is smaller, and may be faster with caching.  Defaults to brotli
           
           [possible values: brotli, gzip]
           
--- a/docs/src/run-with-cli.md
+++ b/docs/src/run-with-cli.md
@@ -31,6 +31,11 @@ Options:
   -W, --workers <WORKERS>
           Number of web server workers
 
+      --preferred-encoding <PREFERRED_ENCODING>
+          Preferred tiles encoding. gzip or brotli, default brotili. You could also use br as a shortcut for brotli
+          
+          [possible values: brotli, gzip]
+          
   -b, --auto-bounds <AUTO_BOUNDS>
           Specify how bounds should be computed for the spatial PG tables. [DEFAULT: quick]
 
--- a/martin/src/args/srv.rs
+++ b/martin/src/args/srv.rs
@@ -12,7 +12,7 @@ pub struct SrvArgs {
     /// Number of web server workers
     #[arg(short = 'W', long)]
     pub workers: Option<usize>,
-    /// Preferred tiles encoding. gzip or brotli, default brotili. You could also use br as a shortcut for brotli
+    /// Martin server preferred tile encoding. If the client accepts multiple compression formats, and the tile source is not pre-compressed, which compression should be used. `gzip` is faster, but `brotli` is smaller, and may be faster with caching.  Defaults to brotli.
     #[arg(long)]
     pub preferred_encoding: Option<PreferredEncoding>,
 }
--- a/martin/src/args/srv.rs
+++ b/martin/src/args/srv.rs
@@ -13,7 +13,7 @@ pub struct SrvArgs {
     #[arg(short = 'W', long)]
     pub workers: Option<usize>,
     /// Preferred tiles encoding. gzip or brotli, default brotili. You could also use br as a shortcut for brotli
-    #[arg(short, long)]
+    #[arg(long)]
     pub preferred_encoding: Option<PreferredEncoding>,
 }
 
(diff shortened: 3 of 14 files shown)


martin/src/args/srv.rs
```diff
@@ -10,6 +12,8 @@ pub struct SrvArgs {
     /// Number of web server workers
     #[arg(short = 'W', long)]
     pub workers: Option<usize>,
+    #[arg(help = "to do", short, long)]
+    pub preferred_encoding: Option<String>,
```
nyurik: Don't use String here. Create a new `enum PreferredEncoding` similar to `BoundsCalcType`.  Note that serde supports aliases (i.e. support both br and brotli in one value).

martin/src/args/srv.rs
```diff
@@ -10,6 +12,8 @@ pub struct SrvArgs {
     /// Number of web server workers
     #[arg(short = 'W', long)]
     pub workers: Option<usize>,
+    #[arg(help = "to do", short, long)]
```
nyurik: don't use `help`, use the `/// ...` right above this line

martin/src/srv/server.rs
```diff
@@ -115,7 +115,8 @@ pub fn new_server(config: SrvConfig, state: ServerState) -> MartinResult<(Server
 
         let app = App::new()
             .app_data(Data::new(state.tiles.clone()))
-            .app_data(Data::new(state.cache.clone()));
+            .app_data(Data::new(state.cache.clone()))
+            .app_data(Data::new(state.preferred_encoding));
```
nyurik: I think we should store `SrvConfig` object in the data.  Read the steps in https://github.com/maplibre/martin/issues/1185 -- here you have almost the same thing - a configuration value you can use in `get_tile` in `tiles.rs`

nyurik (commented):
good first step, left a few comments

martin/src/srv/tiles.rs
```diff
@@ -184,9 +185,30 @@ impl<'a> DynTileSource<'a> {
                 }
             }
             if tile.info.encoding == Encoding::Uncompressed {
+                let ordered_encodings = if let Some(prefered) = self.prefered_encodings {
+                    match prefered {
+                        PreferredEncoding::Brotli => [
```
nyurik: I am not too certain if the compiler can optimize these cases on the fly - maybe replace these with 2 statics, like we used to have with `SUPPORTED_ENCODINGS`:  

```rust
static PREFER_BROTLI_ENC: &[HeaderEnc] = &[ ... ];
static PREFER_GZIP_ENC: &[HeaderEnc] = &[ ... ];
```

and in this match statement just do this:

```rust
let ordered_encodings = match self.prefered_encodings {
  Some(PreferredEncoding::Gzip) => PREFER_GZIP_ENC,
  Some(PreferredEncoding::Brotli) | None => PREFER_BROTLI_ENC
};
```

nyurik (commented):
love it!  One small nit, and seems it is pretty good to go!

martin/src/srv/tiles.rs
```diff
@@ -93,7 +93,8 @@ impl<'a> DynTileSource<'a> {
             info,
             query_str,
             query_obj,
-            encodings,
+            accept_encodings: accpect_encodings,
```
nyurik: maybe better rename these into `accept_enc` and `preferred_enc` (notice you misspelled the var there... and you are not the first - similar fate with [Referer header](https://en.wikipedia.org/wiki/HTTP_referer)) :)

One uncertainty - is this a LIST of accepted encodings, or the encoding that has been accepted (chosen)?  The name is a bit confusing - might need to make it clearer.

thx!

sharkAndshark: I need to improve my English more XD 😂
Any better renaming suggestion? Like `accept_enc` => `browser_accept_enc`? I'm not sure...

martin/src/srv/tiles.rs
sharkAndshark: Thx!  Much better and efficiency.

martin/src/srv/tiles.rs
sharkAndshark: Or `browser_wanted_enc`?

docs/src/config-file.md
```diff
@@ -27,6 +27,9 @@ worker_processes: 8
 # Amount of memory (in MB) to use for caching tiles [default: 512, 0 to disable]
 cache_size_mb: 1024
 
+# Preferred tiles encoding, gzip or brotli, default brotili. You could also use br as a shortcut for brotli
```
```suggestion
# If the client accepts multiple compression formats, and the tile source is not pre-compressed, which compression should be used. `gzip` is faster, but `brotli` is smaller, and may be faster with caching.  Defaults to brotli.
```

docs/src/run-with-cli.md
```diff
@@ -31,6 +31,11 @@ Options:
   -W, --workers <WORKERS>
           Number of web server workers
 
+      --preferred-encoding <PREFERRED_ENCODING>
```
nyurik, line 34: similar as above?

martin/src/args/srv.rs
```diff
@@ -10,6 +12,19 @@ pub struct SrvArgs {
     /// Number of web server workers
     #[arg(short = 'W', long)]
     pub workers: Option<usize>,
+    /// Preferred tiles encoding. gzip or brotli, default brotili. You could also use br as a shortcut for brotli
```
nyurik: same

martin/tests/mb_server_test.rs
```diff
@@ -24,6 +25,12 @@ macro_rules! create_app {
                 ))
                 .app_data(actix_web::web::Data::new(::martin::NO_MAIN_CACHE))
                 .app_data(actix_web::web::Data::new(state.tiles))
+                .app_data(actix_web::web::Data::new(SrvConfig {
```
nyurik: for tests might as well use `SrvConfig::default()` here and below

nyurik (commented):
nah, not worth it, will fix naming if needed later, good enough.  I made a few more minor suggestions

martin/src/srv/tiles.rs
```diff
@@ -183,10 +196,17 @@ impl<'a> DynTileSource<'a> {
...
             if tile.info.encoding == Encoding::Uncompressed {
+                let ordered_encodings = match self.preferred_enc {
+                    Some(PreferredEncoding::Gzip) => PREFER_GZIP_ENC,
+                    Some(PreferredEncoding::Brotli) | None => PREFER_BROTLI_ENC,
+                };
+
                 // only apply compression if the content supports it
                 if let Some(HeaderEnc::Known(enc)) =
-                    accept_enc.negotiate(SUPPORTED_ENCODINGS.iter())
+                    // accept_enc.negotiate(SUPPORTED_ENCODINGS.iter())
```
nyurik:
```suggestion
```

nyurik (approved):
I'm not too happy with the description, but can't think think too clearly. Will try in the morning. In the mean time, there is one line that i think was left by an accident, but good to go otherwise, so remove it and feel free to merge :)

martin/src/srv/tiles.rs
sharkAndshark: Oops. Deleted.

sharkAndshark: "Preferred tile encoding when martin is negotiating with browser, only works if the tile itself is not pre-compressed. The default Brotli is much smaller(and may be faster with caching) but a little bit slower than gzip."
Well this version maybe much worse..

PR #1205, opened 2024-02-20
Add --base-path CLI option to override the URL path in the tilejson

### LINKED ISSUE #1185: Add `--base-path` CLI option to override the URL path in the tilejson

sharkAndshark:
Try to fix #1185 
- [x] Add --base-path to CLI
- [x] Add base_path to config file
- [x] Add tests
- [x] Update doc


martin/src/args/srv.rs
```diff
@@ -9,6 +9,8 @@ pub struct SrvArgs {
     pub keep_alive: Option<u64>,
     #[arg(help = format!("The socket address to bind. [DEFAULT: {}]", LISTEN_ADDRESSES_DEFAULT), short, long)]
     pub listen_addresses: Option<String>,
+    #[arg(help = format!("Base path of martin if it's behind a proxy server. It's useful when you couldn't set x-rewrite-url of proxy server"), long)]
```
nyurik:
```suggestion
    /// Set TileJSON URL path prefix. Must begin with a `/`. Examples: `/`, `/tiles`
    #[arg(long)]
```

martin/src/config.rs
```diff
@@ -71,6 +73,17 @@ impl Config {
         let mut res = UnrecognizedValues::new();
         copy_unrecognized_config(&mut res, "", &self.unrecognized);
 
+        if self
```
nyurik: parsing here should be the same as is done for `x-rewrite-url` -- probably we need a dedicated function for that, used from both places.  The error message should probably say something like `Base path must be a valid URL path, and must begin with a '/' symbol`.

nyurik (commented):
thx for working on this!

martin/src/srv/tiles_info.rs
```diff
@@ -26,18 +27,23 @@ async fn get_source_info(
...
-    let tiles_path = req
-        .headers()
-        .get("x-rewrite-url")
-        .and_then(|v| v.to_str().ok())
-        .and_then(|v| v.parse::<Uri>().ok())
-        .map_or_else(|| req.path().to_owned(), |v| v.path().to_owned());
-
+    // if base_path is set, use it as the tiles path
+    // or try to use x-rewrite-url header value as the tiles path
+    let tiles_path = if srv_config.base_path.is_some() {
```
nyurik: rule of thumb:  if you have `.unwrap()`, you are probably doing something bad (unless its a unit test)...  In this case - use `if let Some(base_path) = srv_config.base_path { ... } else { ... }`

martin/src/srv/tiles_info.rs
```diff
@@ -26,18 +27,23 @@ async fn get_source_info(
...
+        srv_config.base_path.clone().unwrap()
+    } else {
+        // Get `X-REWRITE-URL` header value, and extract its `path` component.
+        // If the header is not present or cannot be parsed as a URL, return the request path.
+        req.headers()
+            .get("x-rewrite-url")
+            .and_then(|v| v.to_str().ok())
+            .and_then(|v| v.parse::<Uri>().ok())
```
nyurik, line 40: shouldn't this code also use `parse_base_path` ?

martin/src/srv/tiles_info.rs
```diff
@@ -58,6 +53,29 @@ async fn get_source_info(
...
 }
 
+fn generate_tiles_base(
+    srv_config: &SrvConfig,
+    req: &HttpRequest,
+    path: &Path<SourceIDsRequest>,
+) -> String {
+    let mut result = None;
+    if let Some(path_from_config) = &srv_config.base_path {
+        let base = format!("{path_from_config}/{0}", &path.source_ids);
```
sharkAndshark: Should we allow the SrvConfig.base_path with a tail slash, like "/foo/bar/"?

martin/src/utils/utilities.rs
```diff
@@ -28,3 +31,32 @@ pub fn encode_brotli(data: &[u8]) -> Result<Vec<u8>, std::io::Error> {
...
+        assert_eq!("/", parse_base_path(&case1).unwrap());
+
+        let case2 = String::new();
+        assert!(parse_base_path(&case2).is_err());
+
+        let case3 = "/foo/bar".to_string();
+        assert_eq!("/foo/bar", parse_base_path(&case3).unwrap());
+
+        let case4 = "foo/bar".to_string();
+        assert!(parse_base_path(&case4).is_err());
```
sharkAndshark: @nyurik  
```rust
let case5 = "/foo/bar/".to_string();
assert!(parse_base_path(&case5).is_err());
// Should we remove the tail slash in parse_base_path or just return an error? 
```

sharkAndshark:
I will refactor these functions/tests later.

.github/workflows/ci.yml
```diff
@@ -393,6 +393,7 @@ jobs:
           "@
 
           # Replace the default Nginx configuration file
+          mkdir "C:\tools\nginx-1.25.3\conf\"
```
nyurik: we should not create this directory.  If it is missing, it means `nginx` is not properly installed. Perhaps the version changed? In that case, we need a way to find where the current nginx configuration is located. Regardless, this shouldn't be part of this PR - if our current CI fails, lets fix it there.

docs/src/run-with-cli.md
```diff
@@ -27,7 +27,9 @@ Options:
 
   -l, --listen-addresses <LISTEN_ADDRESSES>
           The socket address to bind. [DEFAULT: 0.0.0.0:3000]
-
+      --base-path <BASE_PATH>
+          Set TileJSON URL path prefix. Must begin with a `/`. Examples: `/`, `/tiles`
```
nyurik:
```suggestion
          Set TileJSON URL path prefix, ignoring X-Rewrite-URL header. Must begin with a `/`. Examples: `/`, `/tiles`
```

martin/src/args/srv.rs
```diff
@@ -9,6 +9,9 @@ pub struct SrvArgs {
     pub keep_alive: Option<u64>,
     #[arg(help = format!("The socket address to bind. [DEFAULT: {}]", LISTEN_ADDRESSES_DEFAULT), short, long)]
     pub listen_addresses: Option<String>,
+    /// Set TileJSON URL path prefix. Must begin with a `/`. Examples: `/`, `/tiles`
```
nyurik:
```suggestion
    /// Set TileJSON URL path prefix, ignoring X-Rewrite-URL header. Must begin with a `/`. Examples: `/`, `/tiles`
```

martin/src/srv/tiles_info.rs
```diff
@@ -9,7 +9,8 @@ use serde::Deserialize;
 use tilejson::{tilejson, TileJSON};
 
 use crate::source::{Source, TileSources};
-
+use crate::srv::SrvConfig;
+use crate::utils::parse_base_path;
```
nyurik: please run `just fmt2` - there should be an empty line after `use` section. I wonder why `cargo fmt` didn't do it?

martin/src/srv/tiles_info.rs
nyurik: That's a good question... `X-Rewrite-URL` and `--base-path` must use exactly the same rules - so if one allows something, so should the other. We may allow some "normalization" logic (for both cases) - e.g. if it has a trailing slash, we should internally convert it to a string without it.

martin/src/utils/utilities.rs
nyurik: see my answer above

.github/workflows/ci.yml
sharkAndshark: Yeah. We should fix it another PR.

martin/src/srv/tiles_info.rs
sharkAndshark: Trailing '/' would be removed in prase_base_path function now.

martin/src/srv/tiles_info.rs
sharkAndshark: Would run fmt2 after the nightly toolchain is installed(might take 30 minutes as my internet connection is poor...)

nyurik: hm, i think we should change fmt2 to use the stable version of cargo, but use the "unstable feature" of it.

martin/src/utils/utilities.rs
```diff
@@ -28,3 +32,39 @@ pub fn encode_brotli(data: &[u8]) -> Result<Vec<u8>, std::io::Error> {
     encoder.write_all(data)?;
     Ok(encoder.into_inner())
 }
+
+pub fn parse_base_path(path: &String) -> MartinResult<String> {
```
nyurik: several things:
* never pass `&String` as a parameter - use `&str` instead
* you don't need to clone anything here, except when returning an error
* use `.trim_end_matches('/')` to convert one `&str` into another `&str`

martin/src/srv/tiles_info.rs
```diff
@@ -58,6 +54,27 @@ async fn get_source_info(
     Ok(HttpResponse::Ok().json(merge_tilejson(&sources, tiles_url)))
 }
 
+fn generate_tiles_base(
```
nyurik: * if `srv_config.base_path` is set, you should not parse it on every request.  It should be in the right format, parsed during config parsing.
* I would inline `generate_tiles_base` here -- the whole `get_source_info` is about constructing `tiles_url` - it would be confusing to break it into two functions I think, and creates unneeded complexities with error handling.
* `generate_tiles_base` should not create Strings (clones), but instead `tiles_path` should be a `&str` that will be used in the `let path_and_query = ...`

martin/src/utils/utilities.rs
```diff
@@ -28,3 +32,39 @@ pub fn encode_brotli(data: &[u8]) -> Result<Vec<u8>, std::io::Error> {
...
+        return Err(BasePathError(path.to_string()));
+    }
+    if let Ok(uri) = path.parse::<Uri>() {
+        let mut result = uri.path();
+        while result.len() > 1 {
+            result = result.trim_end_matches('/');
+        }
+        return Ok(result.to_string());
+    }
+    Err(BasePathError(path.to_string()))
```
nyurik: wouldn't this be an infinite loop?

You don't need a loop - just do this:

```suggestion
    if let Ok(uri) = path.parse::<Uri>() {
      Ok(uri.path().trim_end_matches('/').to_string())
    } else {
      Err(BasePathError(path.to_string()))
    }
```

martin/src/srv/tiles_info.rs
```diff
@@ -26,17 +28,19 @@ async fn get_source_info(
...
-    // Get `X-REWRITE-URL` header value, and extract its `path` component.
-    // If the header is not present or cannot be parsed as a URL, return the request path.
-    let tiles_path = req
-        .headers()
-        .get("x-rewrite-url")
-        .and_then(|v| v.to_str().ok())
-        .and_then(|v| v.parse::<Uri>().ok())
-        .map_or_else(|| req.path().to_owned(), |v| v.path().to_owned());
+    let tiles_path = if let Some(path_from_config) = &srv_config.base_path {
+        format!("{path_from_config}/{0}", &path.source_ids)
```
nyurik:
```suggestion
    let tiles_path = if let Some(base_path) = &srv_config.base_path {
        format!("{base_path}/{}", path.source_ids)
```

nyurik (commented):
hmm, seems like you can't easily get rid of an extra string alloc here. Oh well.

martin/src/utils/utilities.rs
sharkAndshark: Seems for "//" `trim_end_matches` would return empty string `""`.. That would fail our test

nyurik: but if you leave it, it will break `format!("{base_path}/{}", path.source_ids)` -- because it would create two slashes

martin/src/utils/utilities.rs
```diff
@@ -28,3 +32,38 @@ pub fn encode_brotli(data: &[u8]) -> Result<Vec<u8>, std::io::Error> {
...
     Ok(encoder.into_inner())
 }
+
+pub fn parse_base_path(path: &str) -> MartinResult<String> {
+    if !path.starts_with('/') {
+        return Err(BasePathError(path.to_string()));
+    }
+    if let Ok(uri) = path.parse::<Uri>() {
+        let mut result = uri.path().to_string();
+        while result.chars().last().is_some_and(|v| v == '/') && result.len() > 1 {
```
sharkAndshark: Use pop() back as the trim_end_matches would return "' for "//", any  better way to use the trim methods?

nyurik: ok, lets keep it simple - use this method just for the user-supplied base_path parsing - in which case it should be empty if the user only passed `/`. And keep the original code for x-rewrite-url header.

I will look at it later, see if it can be optimized, but its not worth it now

sharkAndshark: Actually It's not bad to use pop() here?

sharkAndshark: But your suggestion is more simple

nyurik (approved):
thanks!!!

PR #1220, opened 2024-02-29
fix: Use ST_EstimatedExtent for quick bounds calc

### LINKED ISSUE #1206: For `--auto-bounds quick`, use `ST_EstimatedExtent` instead of `ST_Extent` for quicker results
Modify PG querying to use a quicker extent function when the bound check is set to `quick`

Implementation is easy: simply add another bool param to `calc_bounds`, and use a different function depending on that param

https://github.com/maplibre/martin/blob/e30467891df3b668d17a31e5a569f397e0e9b975/martin/src/pg/query_tables.rs#L214

sharkAndshark:
Try to fix #1206 
- [x] Use St_EstimatedExtent
- [x] Update test cases


martin/src/pg/query_tables.rs
```diff
@@ -217,10 +218,12 @@ async fn calc_bounds(
     table: &str,
     geometry_column: &str,
     srid: i32,
+    is_quick: bool,
 ) -> PgResult<Option<Bounds>> {
-    Ok(pool.get()
-        .await?
-        .query_one(&format!(
+    let sql = if is_quick {
+        format!("SELECT ST_EstimatedExtent('{schema}', '{table}', '{geometry_column}') as bounds")
```
nyurik: why in quotes?  these values are not strings, they are escaped PostgreSQL identifiers (schema/table/column names)

sharkAndshark: Per [postgis doc](https://postgis.net/docs/ST_EstimatedExtent.html), the arguments type of `ST_EstimatedExtent` is text.

`box2d ST_EstimatedExtent(text schema_name, text table_name, text geocolumn_name);`

nyurik: interesting, thx, did not realize that.  Should the parameters be pre-escaped though?  We would need to test this on the `MixedCase`.  Apparently there was a bug 10 years ago -- https://trac.osgeo.org/postgis/ticket/2834 that temporarily used the escaped version, but then it was reverted it seems

sharkAndshark: Made a little bit test.
```sql

SELECT ST_EstimatedExtent('public','Project_Copy1','geom')
-- BOX(399529.5 3384786.5,412743.5 3394297)
```

```sql
SELECT ST_EstimatedExtent('public','"Project_Copy1"','geom')
-- syntax error
```

I will add a test on `MixedCase` later.

nyurik:
I added a test for weird table names - https://github.com/maplibre/martin/pull/1222  -- please review, and maybe add a test for a similar function too, probably in the same namespace.

nyurik:
Also note this code: https://github.com/postgis/postgis/blob/bccab952f00696d5a48e9eb495e827825eda23a2/postgis/gserialized_estimate.c#L2315 - postgis adds double quotes around passed in schema & table names.  I really hope that function returns pre-escaped value without double quotes, or else it will fail spectacularly

martin/src/pg/query_tables.rs
```diff
@@ -217,10 +233,18 @@ async fn calc_bounds(
...
     srid: i32,
+    is_quick: bool,
 ) -> PgResult<Option<Bounds>> {
-    Ok(pool.get()
-        .await?
-        .query_one(&format!(
+    let sql = if is_quick {
+        let schema = escape_literal(schema);
+        let table = escape_literal(table);
+        let geometry_column = escape_literal(geometry_column);
```
sharkAndshark: @nyurik  any help on this? I think we should call `escape_literal` before `ST_EstimatedExtent`, but the tests still fails.
`[2024-03-13T04:33:08Z ERROR martin::pg::builder] Failed to create a source: Postgres error while querying table bounds: db error: ERROR: invalid name syntax`

sharkAndshark: And the SQL is : `SELECT ST_EstimatedExtent('"Quotes'' and Space.Dot.', '. Points" ''quote', '. "Geom"') as bounds`

nyurik: Sigh, one of those painful cases I guess - I created a bug https://trac.osgeo.org/postgis/ticket/5697 

So to move forward, I think we will have to do a workaround with the following logic:
* if schema, table, and geom column is "simple" (matches this regex for all 3 values:  `^[a-zA-Z_][a-zA-Z_0-9]*$`), pass these values "as is", but make sure to pass them as parameters (second param in `.query_one()`)
* if not, use the existing query

nyurik:
I have managed to get it to work in some cases, but still some issues remain.  The trick I think is to remove the `"` around schema and table, but keep the geometry column as a string value, so something like this, but there is still something wrong

```rust
    let params: Vec<&(dyn ToSql + Sync)>;
    let sql: String;

    // Both queries require identifier-escaped format for schema and table
    let schema = escape_identifier(schema);
    let table = escape_identifier(table);
    let mut schema: &str = schema.as_str();
    let mut table: &str = table.as_str();

    if is_quick {
        // Remove first and last quote (but only one on each side)
        schema = &schema[1..schema.len() - 1];
        table = &table[1..table.len() - 1];
        params = vec![&schema, &table, &geometry_column];
        sql = "SELECT ST_EstimatedExtent($1, $2, $3) as bounds".to_string();
    } else {
        params = vec![];
        let geometry_column = escape_identifier(geometry_column);
        sql = format!(
            r#"
WITH real_bounds AS (SELECT ST_SetSRID(ST_Extent({geometry_column}), {srid}) AS rb FROM {schema}.{table})
SELECT ST_Transform(
            CASE
                WHEN (SELECT ST_GeometryType(rb) FROM real_bounds LIMIT 1) = 'ST_Point'
                THEN ST_SetSRID(ST_Extent(ST_Expand({geometry_column}, 1)), {srid})
                ELSE (SELECT * FROM real_bounds)
            END,
            4326
        ) AS bounds
FROM {schema}.{table};
                "#
        );
    };
```

nyurik:
I got it to work in https://github.com/sharkAndshark/martin/pull/103/files -- once merged into your branch, it will appear here as well. Let me know what you think

CommanderStorm:
@nyurik @sharkAndshark
You both worked on this PR and I think you both found the current state good.
Am I misunderstanding something?
Is something still missing?

CommanderStorm (approved):
(From my POV, this looks good)

martin/src/pg/query_tables.rs
```diff
@@ -108,25 +109,37 @@ pub async fn table_to_query(
...
         match bounds_type {
             BoundsCalcType::Skip => {}
             BoundsCalcType::Calc => {
                 debug!("Computing {} table bounds for {id}", info.format_id());
-                info.bounds = calc_bounds(&pool, &schema, &table, &geometry_column, srid).await?;
+                info.bounds = calc_bounds(
+                    &pool,
+                    &info.schema,
+                    &info.table,
+                    &info.geometry_column,
```
nyurik: Might as well just pass a ref to `info`

```suggestion
                    &info
```

nyurik:
I need to dive into it a bit more, but from what i recall, nothing fundamentally wrong with this one

martin/src/pg/query_tables.rs
CommanderStorm: Well..
These are not identical and the reason for it being this way is a few lines further down:

image
image

=> since bounds was not borrowed above, this worked before the suggested change ^^

CommanderStorm: => reverted in https://github.com/maplibre/martin/pull/1220/commits/daae7f2023a76735205a2ae0c4f2544e9afdc01d

nyurik: @CommanderStorm see https://github.com/sharkAndshark/martin/pull/149

nyurik:
A small optimization for this PR is in https://github.com/sharkAndshark/martin/pull/149

PR #1253, opened 2024-03-15
Adjust readme and martin book

### LINKED ISSUE #1245: Reduce README, moving more things to the book
Currently the readme has far too much info, and some of it less or more descriptive than the book. We shouldn't maintain two copies of the same stuff - better move it to the book. For example, [install section](https://maplibre.org/martin/installation.html) of the book has fewer details than the readme.

sharkAndshark:
Try to fix #1245 
## Add or keep
- [ ] Add a `Get Started` to readme, it must be easy to take, like `curl -O ..` and then `martin http://xxx.demo.pmtiles` I guess? OR `docker run ..` with a small http://xxx.demo.pmtiles, do we have some host to serve a demo pmtiles file?
- [ ] Keep `License` in readme

## Move or remove from `readme.md`
- [ ] Move `Re-use Martin as a library` to book
- [ ] Move installation from readme to book
- [ ] Remove `Running Martin Service` 
- [ ] Remove `API` from readme


README.md
```diff
@@ -14,107 +14,6 @@ Additionally, there are [several tools](https://maplibre.org/martin/tools.html)
 
 See [Martin book](https://maplibre.org/martin/) for complete documentation.
 
-## Installation
```
nyurik, line 17: I think we should create some something like a list of important information, and point it to the book - e.g. installation should point directly to the installation section, etc.  Just the most important topics, not everything?

docs/src/Re-use-Martin-as-a-library.md
```diff
@@ -0,0 +1,9 @@
+# Re-use Martin as a library
```
nyurik: this should be added to the index.  Perhaps under `For developers` section?  There we should also add things like how to contribute to Martin

docs/src/installation.md
```diff
@@ -2,47 +2,73 @@
...
+### From Binary Distributions Manually
 
 You can download martin from [GitHub releases page](https://github.com/maplibre/martin/releases).
 
-| Platform | Downloads (latest)     |
-|----------|------------------------|
-| Linux    | [64-bit][rl-linux-tar] |
-| macOS    | [64-bit][rl-macos-tar] |
-| Windows  | [64-bit][rl-win64-zip] |
+| Platform | AMD-64                                                                                           | ARM-64                              |
```
nyurik: AMD-64 is a bit confusing (although technically correct). Perhaps use the `x64` is better, vs `arm-64`?

nyurik (commented):
thx for working on this!!

docs/src/Re-use-Martin-as-a-library.md
sharkAndshark: Oops. Forget to add it to the index.    
And any thounghts on the developers section? A new `For developers` section?  Perhaps place these things to the old `Development` section?

```markdown
- [Development](development.md)
 - [Prerequisites](Prerequisites.md) things about the `docker` `docker-compose` `just` `jq`  etc
 - [with VS Code](with-VS-Code.md) how to build and debug with vscode and the plugins needed(CODELLDB, rust analyzer.. etc)
 - [With Rust Rover](with-rust-rover.md) have no too much experience on this...
 - [Martin as a library](martin-as-a-library.md)
```

README.md
sharkAndshark: Like this?

## Helpful Martin docs
* [Installation](https://maplibre.org/martin/installation.html)
* Running with [CLI](https://maplibre.org/martin/run-with-cli.html) or [configuration file](https://maplibre.org/martin/config-file.html)
* [Usage and API](https://maplibre.org/martin/using.html)

nyurik: Yes, but without the header - just a few links.  We may also want to add a "quick start" - makes novice users go there first

docs/src/installation.md
sharkAndshark: Yup.  Updated.

README.md
```diff
@@ -14,110 +14,10 @@ Additionally, there are [several tools](https://maplibre.org/martin/tools.html)
...
-* **fonts** - enable font sources
-* **sprites** - enable sprite sources
-
-## Documentation
-
-See [Martin book](https://maplibre.org/martin/) for complete documentation.
+* [Installation](https://maplibre.org/martin/installation.html)
+* Running with [CLI](https://maplibre.org/martin/run-with-cli.html) or [configuration file](https://maplibre.org/martin/config-file.html)
+* [Usage and API](https://maplibre.org/martin/using.html)
+<!-- * [Quick Start](https://maplibre.org/martin/quick-Start.html) -->
```
sharkAndshark: To bypass the markdwon link check, this line is removed temporally. Will be added after this get merged.

nyurik:
will review shortly. One thing for sure - i don't think we want to use case-sensitive dir name like `Images` - use `img` or `images` instead

nyurik:
P.S. and same for the filename -- `Re-use-Martin-as-a-library.md` caps don't work well, esp when some people use windows (which is case insensitive) vs linux (which is case sensitive).

Perhaps a better name is `martin-as-library` or `using-martin-as-lib` or something like that?

nyurik:
Great work, thanks!!!  It gave me a good start to continue improving it. Especially big thx for the QGIS doc images, they make things much easier for the novice users.

PR #1260, opened 2024-03-17
Switch default encoding to gzip for now

nyurik:
GZip is significantly faster, so for now lets use that until brotli project gains some performance, and until we get proper after-compression caching.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2379,6 +2379,7 @@ dependencies = [
  "postgres-protocol",
  "pprof",
  "regex",
+ "rstest",
  "rustls 0.23.8",
  "rustls-native-certs",
  "rustls-pemfile",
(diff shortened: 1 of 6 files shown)


sharkAndshark:
https://github.com/maplibre/martin/blob/7107db5887f6233f3066077513978cbfe2757c46/martin/src/srv/tiles.rs#L200-L204

Seems this should be updated

sharkAndshark:
https://github.com/maplibre/martin/blob/7107db5887f6233f3066077513978cbfe2757c46/martin/src/srv/tiles.rs#L281-L297

And maybe more tests should be added or updated here.

nyurik:
Thx @sharkAndshark!  Turns out it was never properly working -- it was always using the list of accepted encoding and ignored the order of the supported encodings... So we may have to rework that code.  I re-worked the tests and removed the confusing default, but this is currently broken. Will try to get back to it soonish

sharkAndshark:
Hi @nyurik  I make a [PR](https://github.com/nyurik/martin/pull/6) to this, Could you give it a review?

sharkAndshark (approved):

PR #1283, opened 2024-03-30
Update documentation for running with Docker Compose

hiddewie:
Some small changes to make the Docker compose documentation up to date:
- Update dependency versions of PostGIS and Martin
- Use the bundled `docker compose` extension, not `docker-compose`. (see https://docs.docker.com/compose/migrate/#docker-compose-vs-docker-compose)
- Remove `version` from the Docker Compose file (see https://docs.docker.com/compose/compose-file/04-version-and-name/)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,5 +1,3 @@
-version: '3'
-
 services:
   fileserver:
     image: nginx:alpine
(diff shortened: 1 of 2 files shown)


nyurik (approved):
Thanks!

PR #1355, opened 2024-05-29
Fix preferred encoding computation

### LINKED ISSUE #1315: very slow queries through martin
Hi there, 

i am trying to find out why martin is so much slower then a direct query to the db. 
 
this: 
http://localhost:3000/query_tile/10/544/339 
takes >13s to get the tile from the server via martin.
the martin log looks like this: 

```
[2024-04-24T07:43:12Z DEBUG martin::pg::pg_source] SQL: SELECT "public"."query_tile"($1::integer, $2::integer, $3::integer) AS tile [10,544,339]
[2024-04-24T07:43:12Z DEBUG tokio_postgres::query] executing statement s7 with parameters: [10, 544, 339]
[2024-04-24T07:43:25Z INFO  actix_web::middleware::logger] 127.0.0.1 "GET /query_tile/10/544/339 HTTP/1.1" 200 440578 "-" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/124.0.0.0 Safari/537.36" 13.798128
```

using the function directly: 
`EXPLAIN (ANALYZE,VERBOSE) select query_tile(10,544,339)
`
results in: 
```
"Planning Time: 328.475 ms"
"Execution Time: 0.023 ms"

```
so there is a gap of 13s where something happens. 
i thought it might be the compression, but compressing the mvt tile from the db cannot take 13s. for comparison, the attached file takes 0.05s to compress using gzip. (is this comparable?)
```
time gzip -c 339 > 339.gz

real	0m0,056s
user	0m0,056s
sys	0m0,000s

```

how can i narrow down what the issue is?  i have attached the resulting  tile here. 

All rows have spacial indices and are clustered. 

[339.zip](https://github.com/maplibre/martin/files/15090583/339.zip)

PS.: this is the martin config (changing from brotli to gzip does not make a huge difference): 
``` yaml
keep_alive: 75
listen_addresses: "0.0.0.0:3000"
worker_processes: 8
cache_size_mb: 0
preferred_encoding: brotli
postgres:
  connection_string: "postgresql://postgres:postgrespw@localhost:5432/tilesDBtest"
  default_srid: 4326
  pool_size: 20
  max_feature_count: 100000
  auto_bounds: calc

  functions:
    query_tile:
      schema: public
      function: query_tile
      minzoom: 4
      maxzoom: 20
      bounds: [-180.0, -90.0, 180.0, 90.0]

```

nyurik:
* Decides which encoding to use depending on both `Accept-Encode` and the preferred encoding (chooses the preferred one in case brotli and gzip have the same q value)
* Uses `gzip` by default

This will fix #1315

### DIFF
--- a/martin/src/srv/tiles.rs
+++ b/martin/src/srv/tiles.rs
@@ -1,5 +1,6 @@
+use actix_http::header::Quality;
 use actix_http::ContentEncoding;
-use actix_web::error::{ErrorBadRequest, ErrorNotFound};
+use actix_web::error::{ErrorBadRequest, ErrorNotAcceptable, ErrorNotFound};
 use actix_web::http::header::{
     AcceptEncoding, Encoding as HeaderEnc, Preference, CONTENT_ENCODING,
 };
@@ -21,13 +22,7 @@ use crate::utils::{
 };
 use crate::{Tile, TileCoord, TileData};
 
-static PREFER_BROTLI_ENC: &[HeaderEnc] = &[
-    HeaderEnc::brotli(),
-    HeaderEnc::gzip(),
-    HeaderEnc::identity(),
-];
-
-static PREFER_GZIP_ENC: &[HeaderEnc] = &[
+static SUPPORTED_ENC: &[HeaderEnc] = &[
     HeaderEnc::gzip(),
     HeaderEnc::brotli(),
     HeaderEnc::identity(),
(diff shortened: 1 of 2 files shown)


sharkAndshark (approved):

danielrh:
I've measured a lot of brotli vs gzip comparisons, and have always found that with the appropriate settings brotli is faster to compress, decompress and results in smaller filesize at every comparable setting.

It sounds like something is amiss with some setting that would cause gzip to ever win in a fair comparison--it's just an older and less capable algorithm.
What was being measured here?

PR #1356, opened 2024-05-29
Fix day picker when clicking the first day of a range 

### LINKED ISSUE #1193: Website breaks when you click on Jan 1st in the demo
Clicking on Jan 1st in the demo at https://martin.maplibre.org/ breaks the website. The page turns completely blank.

Console error:

```
TypeError: this.props.range is undefined
```

Browser: Firefox 122 on Ubuntu 22.04.3

jtmiclat:
Filters were breaking when `addToRange` returned an undefined. [ This happens when the date you want to add is the start date of the range.](https://github.com/gpbl/react-day-picker/blob/9ad13dc72fff814dcf720a62f6e3b5ea38e8af6d/src/contexts/SelectRange/utils/addToRange.ts#L23-L24) Based on the `react-day-picker`, undefined range means neither from nor to where set. 

I did the following to fix this 
1. Set from and to to be null when range is undefined to not break destructuring
2. Add a condition to only update the tile url when all the needed query parameters are complete to prevent incomplete tile fetching 

https://github.com/maplibre/martin/assets/30991698/dfbf111a-c7e0-4a0f-97c8-098f0951e3c3

Closes #1193

### DIFF
--- a/demo/frontend/src/Components/Map/Filters/Filters.tsx
+++ b/demo/frontend/src/Components/Map/Filters/Filters.tsx
@@ -28,8 +28,11 @@ class Filters extends PureComponent<{visibleLayer, toggleLayer, range, hour, cha
 
   handleDayClick = (day) => {
     const range = addToRange(day, this.props.range);
-
-    this.props.changeFilter('range', range);
+    if (range) {
+      this.props.changeFilter('range', range);
+    } else {
+      this.props.changeFilter('range', { from: null, to: null });
+    }
   };
 
   changeTime = (e) => {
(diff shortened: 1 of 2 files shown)


nyurik (approved):
thanks!

nyurik:
@jtmiclat my sincere apologies - I didn't follow up on this and turns out CI does not auto-merge demo fixes. I just merged #1373 by @jokopriyono which I think fixes the same issue(?).  It is deployed now, and seems like its working properly, but please retest. Thank you for looking at it!

jtmiclat:
Closing the MR as it was addressed in #1373 !

PR #1373, opened 2024-06-12
fix: handle if selected date was undefined while double click on a sa…

jokopriyono:
Handle a negative case if the selected date was undefined while double-clicking on the same date in the calendar on the demo website.

Screenshot 2024-06-12 at 15 45 49


jokopriyono:
Please review this @nyurik

nyurik (approved):

nyurik:
thanks!

nyurik:
and deployed

PR #1387, opened 2024-06-28
fix curl url for linux mbtiles demo

tbo47:


sharkAndshark:
Thx!!!! @tbo47

sharkAndshark (approved):

sharkAndshark:
Maybe we should update the links of `windows` and `macOS` either?  
If not, the user would be redirected to the web interface of this MBTiles. Then they could click the `Download raw file` button. @nyurik

tbo47:
> Maybe we should update the links of `windows` and `macOS` either? If not, the user would be redirected to the web interface of this MBTiles. Then they could click the `Download raw file` button. @nyurik

yes, I didn't update the Windows and MacOs ones for this reason.

nyurik:
@tbo47 thx for the PR. Could you explain why the original was not working for you?  I tried it locally, and the original works just as well, and IMO is a better (`authoritative`) URL.

tbo47:
```bash
➜  /tmp curl -L -O https://github.com/maplibre/martin/blob/main/tests/fixtures/mbtiles/world_cities.mbtiles
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  280k    0  280k    0     0   270k      0 --:--:--  0:00:01 --:--:--  270k
➜  /tmp head world_cities.mbtiles
<!DOCTYPE html>
<html
  lang="en"
  
```
and 
```bash
➜  /tmp apt info curl
Package: curl
Version: 8.8.0-1~bpo12+1
```

I'm on debian stable with backport activated.

nyurik:
Ah, sorry, i should have actually looked at the result.   The `raw` button actually points to this URL: `https://github.com/maplibre/martin/raw/main/tests/fixtures/mbtiles/world_cities.mbtiles` -- I still think pointing to the github site is better than resolve the redirect and pointing to githubusercontent

docs/src/quick-start-linux.md
```diff
@@ -5,7 +5,7 @@ mkdir martin
 cd martin
 
 # Download some sample data
-curl -L -O https://github.com/maplibre/martin/blob/main/tests/fixtures/mbtiles/world_cities.mbtiles
+curl -L -O https://raw.githubusercontent.com/maplibre/martin/main/tests/fixtures/mbtiles/world_cities.mbtiles
```
nyurik:
```suggestion
curl -L -O https://github.com/maplibre/martin/raw/main/tests/fixtures/mbtiles/world_cities.mbtiles
```

nyurik (approved):

nyurik:
I also fixed the mac and windows download links because the link to the release code (i.e. zip file) starts downloading right away - so for consistency we will want the same thing.

nyurik:
And once again - huge thx for @tbo47 for reporting this!

PR #1394, opened 2024-07-01
Create run-with-apache.md

agenceKanvas:
I've created a page as simple as I could. Feel free to tell me if you want me to change anything.

### DIFF
--- a/docs/src/SUMMARY.md
+++ b/docs/src/SUMMARY.md
@@ -12,6 +12,7 @@
   - [Running with Docker](run-with-docker.md)
   - [Running with Docker Compose](run-with-docker-compose.md)
   - [Running with NGINX](run-with-nginx.md)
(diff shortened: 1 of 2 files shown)


nyurik:
@agenceKanvas would it be possible to remove any TLS-related code?  I think we should not confuse things with the https aspects of setting up a proxy - there are enough online guides for that.  Instead, lets focus just on the proxying to the backend part.   Thx!

agenceKanvas:
> @agenceKanvas would it be possible to remove any TLS-related code? I think we should not confuse things with the https aspects of setting up a proxy - there are enough online guides for that. Instead, lets focus just on the proxying to the backend part. Thx!

Of course. I let it in the file because it already was in mine but if you think that it's not helpful, let's remove it

nyurik (approved):
thanks!

PR #1397, opened 2024-07-03
add HEALTHCHECK instruction to the Dockerfile

caspg:
Since Martin gives us `/health` endpoint we could use it by default in the Dockerfile.

I've added extra info to the doc about https://github.com/willfarrell/docker-autoheal but not sure if that's ok. I can remove it.


nyurik (approved):
lgtm, thx!

PR #1420, opened 2024-07-14
chore(docs): tried to document the `--base-path` option better

### LINKED ISSUE #1415: Usage of --base-path CLI option
I use Martin 0.14.2 with the CLI. I want to expose my API on 0.0.0.0:3000 with the prefix path /tileserver because I have a gateway and I want to expose several resources on my gateway, including the Martin tileserver connected to my database.

I launched the server using the command `martin --base-path /tileserver`, but it doesn't work as I understand it:

* The catalog is still reachable from 0.0.0.0:3000/catalog (no problem with that).
* Any TileJSON URL for a given tile source or function layer is reachable from 0.0.0.0:3000/my_super_source, even though the documentation explicitly mentions that the --base-path option will "Set TileJSON URL path prefix".
* What's more, the tiles key for a given source or function layer indicates that my --base-path is taken into account:

```
{
  "tilejson": "xxx",
  "tiles": [
    "https://MY_DNS/tileserver/my_super_source/{z}/{x}/{y}"
  ],
  ...
}
```

When I request https://MY_DNS/tileserver/my_super_source/{z}/{x}/{y}, I get a 404 response, but when I request https://MY_DNS/tileserver/my_super_source/{z}/{x}/{y}, the server responds with a 20X.

To achieve what I want, should I use the --base-path option or use an Nginx server?

CommanderStorm:
As noted, the option could likely be more explicite in what it does

Resolves https://github.com/maplibre/martin/issues/1415

### DIFF
--- a/docs/src/config-file.md
+++ b/docs/src/config-file.md
@@ -26,7 +26,10 @@ keep_alive: 75
 # The socket address to bind [default: 0.0.0.0:3000]
 listen_addresses: '0.0.0.0:3000'
 
-# Set TileJSON URL path prefix, ignoring X-Rewrite-URL header. Must begin with a `/`
+# Set TileJSON URL path prefix. This overides the default of respecting the X-Rewrite-URL header.
+# Only modifies the JSON (TileJSON) returned, martins' API-URLs remain unchanged. If you need to rewrite URLs, please use a reverse proxy.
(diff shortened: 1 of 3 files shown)


nyurik (approved):

nyurik:
thx!

PR #1427, opened 2024-07-23
feat:added a postgres version check

### LINKED ISSUE #1229: db error: ERROR: column ix.indnkeyatts does not exist
Dear amazing Martin team,

failed start martin. with error message:

```
[2024-03-05T02:24:14Z INFO  martin] Starting Martin v0.13.0
[2024-03-05T02:24:14Z INFO  martin] Config file is not specified, auto-detecting sources
[2024-03-05T02:24:14Z INFO  martin::config] Initializing main cache with maximum size 536870912B
[2024-03-05T02:24:14Z INFO  martin::pg::pool] Connecting with SSL support: Config { user: Some("postgres"), password: Some(_), dbname: Some("petadb"), options: None, application_name: None, ssl_mode: Prefer, host: [Tcp("XX.XX.XX.XX")], hostaddr: [], port: [5432], connect_timeout: None, tcp_user_timeout: None, keepalives: true, keepalives_idle: 7200s, keepalives_interval: None, keepalives_retries: None, target_session_attrs: Any, channel_binding: Prefer }
[2024-03-05T02:24:14Z ERROR martin] Postgres error while querying available tables: db error: ERROR: column ix.indnkeyatts does not exist
    HINT: Perhaps you meant to reference the column "ix.indnatts".
```

my postgres version 9.6.11

### LINKED ISSUE #1230: Detect old postgresql versions and warn user
As asked in #1229, postgres 9.6 had different schema discovery structure, and yet we do not tell users that their postgresql db is too old and there might be an issue.

What we need is:
* [ ] Query Postgres version (similar to how we currently detect Postgis version)
  * There are [two ways](https://stackoverflow.com/questions/13733719/which-version-of-postgresql-am-i-running):  `select version()` good for printing as part of the `[info]` text, and `SHOW server_version` (or `SELECT current_setting('server_version')` ?) for a semver variant that can be used for testing.
* [ ] Report postgres version string to the user
* [ ] If postgres version is older than whichever version we currently support (some global const), give a warning, and note that that postgres is no longer maintained

CommanderStorm:
- [x] adds a postgres version check
- [x] added a testcase (using testcontainers) that the version is parsed correctly

Resolves #1230 (I think at this point the PR can be considered unasigned again, feel free to reject otherwise. Don't want to step on peoples toes ^^)
Resolves #1229 (by using the minumum supported postgres version for postgis)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -581,6 +581,56 @@ dependencies = [
 ]
 
 [[package]]
+name = "bollard"
+version = "0.17.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d41711ad46fda47cd701f6908e59d1bd6b9a2b7464c0d0aeab95c6d37096ff8a"
+dependencies = [
+ "base64 0.22.1",
+ "bollard-stubs",
+ "bytes",
+ "futures-core",
+ "futures-util",
+ "hex",
+ "home",
+ "http 1.1.0",
+ "http-body-util",
+ "hyper 1.5.0",
+ "hyper-named-pipe",
+ "hyper-rustls",
+ "hyper-util",
+ "hyperlocal",
+ "log",
+ "pin-project-lite",
+ "rustls",
+ "rustls-native-certs 0.7.3",
+ "rustls-pemfile",
+ "rustls-pki-types",
+ "serde",
+ "serde_derive",
+ "serde_json",
+ "serde_repr",
+ "serde_urlencoded",
+ "thiserror",
+ "tokio",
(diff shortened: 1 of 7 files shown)


nyurik:
thanks!  test containers is an interesting feature, perhaps should use it in some other places too.

nyurik (approved):

CommanderStorm:
Bit of an overlooked downside:
since it requires docker, the CI runners for Mac/windows would need to install this manually, but otherwise a really great tool for testing, can only recommend..

PR #1428, opened 2024-07-23
Update deps, fix new sqlx usage

### LINKED ISSUE #1475: RUSTSEC-2024-0363: Binary Protocol Misinterpretation caused by Truncating or Overflowing Casts
> Binary Protocol Misinterpretation caused by Truncating or Overflowing Casts

| Details             |                                                |
| ------------------- | ---------------------------------------------- |
| Package             | `sqlx`                      |
| Version             | `0.7.4`                   |
| URL                 | [https://github.com/launchbadge/sqlx/issues/3440](https://github.com/launchbadge/sqlx/issues/3440) |
| Date                | 2024-08-15                         |
| Patched versions    | `>=0.8.1`                  |

The following presentation at this year&#39;s DEF CON was brought to our attention on the SQLx Discord:

&gt; SQL Injection isn&#39;t Dead: Smuggling Queries at the Protocol Level  
&gt; &lt;http://web.archive.org/web/20240812130923/https://media.defcon.org/DEF%20CON%2032/DEF%20CON%2032%20presentations/DEF%20CON%2032%20-%20Paul%20Gerste%20-%20SQL%20Injection%20Isn&#39;t%20Dead%20Smuggling%20Queries%20at%20the%20Protocol%20Level.pdf&gt;  
&gt; (Archive link for posterity.)

Essentially, encoding a value larger than 4GiB can cause the length prefix in the protocol to overflow, 
causing the server to interpret the rest of the string as binary protocol commands or other data.

It appears SQLx _does_ perform truncating casts in a way that could be problematic, 
for example: &lt;https://github.com/launchbadge/sqlx/blob/6f2905695b9606b5f51b40ce10af63ac9e696bb8/sqlx-postgres/src/arguments.rs#L163&gt;

This code has existed essentially since the beginning, 
so it is reasonable to assume that all published versions `&lt;= 0.8.0` are affected.

## Mitigation

As always, you should make sure your application is validating untrustworthy user input. 
Reject any input over 4 GiB, or any input that could _encode_ to a string longer than 4 GiB. 
Dynamically built queries are also potentially problematic if it pushes the message size over this 4 GiB bound.

[`Encode::size_hint()`](https://docs.rs/sqlx/latest/sqlx/trait.Encode.html#method.size_hint) 
can be used for sanity checks, but do not assume that the size returned is accurate. 
For example, the `Json&lt;T&gt;` and `Text&lt;T&gt;` adapters have no reasonable way to predict or estimate the final encoded size, 
so they just return `size_of::&lt;T&gt;()` instead.

For web application backends, consider adding some middleware that limits the size of request bodies by default.

## Resolution

Work has started on a branch to add `#[deny]` directives for the following Clippy lints:

* [`cast_possible_truncation`](https://rust-lang.github.io/rust-clippy/master/#/cast_possible_truncation)
* [`cast_possible_wrap`](https://rust-lang.github.io/rust-clippy/master/#/cast_possible_wrap)
* [`cast_sign_loss`](https://rust-lang.github.io/rust-clippy/master/#/cast_sign_loss)

and to manually audit the code that they flag.

A fix is expected to be included in the `0.8.1` release (still WIP as of writing).

See [advisory page](https://rustsec.org/advisories/RUSTSEC-2024-0363.html) for additional details.

nyurik:
Resolves https://github.com/maplibre/martin/issues/1475

One possible concern is that i32 is now used as i64 with sqlx. This **might** impact `tile_xxh3_64_hash` value validity, but I don't see conclusive evidence of that. Note that the tests were incorrectly dumped as i32, thus had to be updated, but it does not seem to affect the actual storage or validation...


mbtiles/src/queries.rs
```diff
@@ -409,8 +394,10 @@ FROM tiles;"
     .fetch_one(conn)
     .await?;
 
-    let min_zoom = validate_zoom(info.min_zoom, "zoom_level")?;
-    let max_zoom = validate_zoom(info.max_zoom, "zoom_level")?;
+    #[allow(clippy::cast_possible_truncation)]
+    let min_zoom = validate_zoom(info.min_zoom.map(|v| v as i32), "zoom_level")?;
+    #[allow(clippy::cast_possible_truncation)]
+    let max_zoom = validate_zoom(info.max_zoom.map(|v| v as i32), "zoom_level")?;
```
CommanderStorm: I don't think the truncation here is the best option.

`fn validate_zoom` does check if the value can be casted to `u8` via `u8::try_from`

Changing to `fn validate_zoom(zoom: Option<i64>, zoom_name: &'static str) -> MbtResult<Option<u8>>` seems like a better option

```suggestion
    let min_zoom = validate_zoom(info.min_zoom, "zoom_level")?;
    let max_zoom = validate_zoom(info.max_zoom, "zoom_level")?;
```

CommanderStorm (approved):
Nice, thanks for debugging the last part ❤️

PR #1444, opened 2024-07-28
Improved glyph url handling: accept arbitrary suffixes after {end} #1443

### LINKED ISSUE #1443: Improved handling of glyph urls as commonly used in styles
When trying to make the font service working, I stumbled across a small problem that is easy to fix, but not so easy to find out why it didn't work (at least for me).

I'm using a tilemaker mbtiles file together with publicly available openmaptiles style, for example, https://github.com/openmaptiles/osm-bright-gl-style/blob/master/style.json. Practically all openmapstyle styles listed on https://openmaptiles.org/styles/, have one thing in common in how they specify the glph url. They do it as something like 
```json
{
  "glyphs": "https://api.maptiler.com/fonts/{fontstack}/{range}.pbf?key={key}"
}
```
  
When I first tried to make Martin work as glyph server, I adapted the url to
```json
{
  "glyphs": "font/{fontstack}/{range}.pbf"
}
```

This didn't work. Martin successfully scanned the font directory that I have provided:

Configured font Open Sans Regular with 883 glyphs (0020-FFFD) from `fonts/open-sans/OpenSans-Regular.ttf`

But when I tried to access the service with [`http://localhost:3000/font/Open%20Sans%20Regular/0-255.pbf`](http://localhost:3000/font/Open%20Sans%20Regular/0-255.pbf) as I have watched maplibre request the glyphs, I would get a response `can not parse "255.pbf" to a u32`.

After enabling the debug logs, I got 
```raw
2024-07-28T10:59:28Z DEBUG actix_web::types::path] Failed during Path extractor deserialization. Request path: "/font/Open%20Sans%20Regular/0-255.pbf"
[2024-07-28T10:59:28Z DEBUG actix_web::middleware::logger] Error in response: Error("can not parse \"255.pbf\" to a u32")
```

After some tinkering, I found out that [`http://localhost:3000/font/Open%20Sans%20Regular/0-255`](http://localhost:3000/font/Open%20Sans%20Regular/0-255) works, but [`http://localhost:3000/font/Open%20Sans%20Regular/0-255.pbf`](http://localhost:3000/font/Open%20Sans%20Regular/0-255.pbf) doesn't. After changing the glyph urls in the styles to the following everything started to work perfectly

```diff
{
-  "glyphs": "font/{fontstack}/{range}.pbf"
+  "glyphs": "font/{fontstack}/{range}"
}
```

Ideally there would be either a pointer in the documentation how the glyph urls need to look like or the server would automatically detect such pbf urls and handle them correctly.

Thanks for this great server. I really enjoy tinkering with it!

bzeiss:
I have attempted to fix #1443 for myself just to challenge myself to write some Rust. I'm a total Rust newbie so please review with care.

This pull requests attempts to solve the problem that many styles out there add suffixes like '.pbf' to the glyph url at the end (see #1443 for examples). The change adds some parsing code to FontRequest that allows arbitrary suffixes after {{end}} and only returns the the leading digits of {{end}} as number for get_font_range. Everything after that is discarded. 

I hope that this way, the font server "just works" for most people and don't hit the issue that I did.

### DIFF
--- a/martin/src/srv/fonts.rs
+++ b/martin/src/srv/fonts.rs
@@ -1,5 +1,5 @@
-use std::string::ToString;
 use std::str::FromStr;
+use std::string::ToString;
 
 use actix_web::error::{ErrorBadRequest, ErrorNotFound};
 use actix_web::web::{Data, Path};
@@ -41,9 +41,7 @@ impl FontRequest {
 )]
 #[allow(clippy::unused_async)]
 async fn get_font(path: Path<FontRequest>, fonts: Data<FontSources>) -> ActixResult<HttpResponse> {
-    let (start, end) = path.parse().map_err(|e| {
-        ErrorBadRequest(e.to_string())
-    })?;
+    let (start, end) = path.parse().map_err(|e| ErrorBadRequest(e.to_string()))?;
 
     let data = fonts
         .get_font_range(&path.fontstack, start, end)
--- a/martin/src/srv/fonts.rs
+++ b/martin/src/srv/fonts.rs
@@ -25,7 +25,7 @@ impl FontRequest {
     }
 
     fn parse_leading_digits(input: &str) -> Result<u32, &'static str> {
-        let digits: String = input.chars().take_while(|c| c.is_digit(10)).collect();
+        let digits: String = input.chars().take_while(|c| c.is_ascii_digit()).collect();
         if digits.is_empty() {
             Err("No leading digits found")
         } else {
--- a/martin/src/srv/fonts.rs
+++ b/martin/src/srv/fonts.rs
@@ -25,7 +25,7 @@ impl FontRequest {
     }
 
     fn parse_leading_digits(input: &str) -> Result<u32, &'static str> {
-        let digits: String = input.chars().take_while(|c| c.is_ascii_digit()).collect();
+        let digits: String = input.chars().take_while(char::is_ascii_digit).collect();
         if digits.is_empty() {
             Err("No leading digits found")
         } else {
--- a/martin/src/srv/fonts.rs
+++ b/martin/src/srv/fonts.rs
@@ -1,4 +1,5 @@
 use std::string::ToString;
+use std::str::FromStr;
 
 use actix_web::error::{ErrorBadRequest, ErrorNotFound};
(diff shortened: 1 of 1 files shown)


martin/src/srv/fonts.rs
```diff
@@ -11,19 +12,39 @@ use crate::srv::server::map_internal_error;
...
+            Err("No leading digits found")
+        } else {
+            digits.parse::<u32>().map_err(|_| "Failed to parse number")
+        }
+    }
 }
 
 #[route(
-    "/font/{fontstack}/{start}-{end}",
+    "/font/{fontstack}/{start}-{end}*",
```
CommanderStorm, line 38: the only reason you are adding the above code is to ignore possible trailing exstentsions, right?

Why not restrict this to a pattern like below (and adding it to `FontRequest`) which should™️ work as far as I read the docs [here](https://actix.rs/docs/url-dispatch/).

Relevant part:
> Replacement markers can optionally specify a regular expression which will be used to decide whether a path segment should match the marker. To specify that a replacement marker should match only a specific set of characters as defined by a regular expression, you must use a slightly extended form of replacement marker syntax. Within braces, the replacement marker name must be followed by a colon, then directly thereafter, the regular expression. The default regular expression associated with a replacement marker `[^/]+` matches one or more characters which are not a slash. For example, under the hood, the replacement marker {foo} can more verbosely be spelled as `{foo:[^/]+}`. You can change this to be an arbitrary regular expression to match an arbitrary sequence of characters, such as `{foo:\d+}` to match only digits.

I have not tested that it actually works.

```suggestion
    "/font/{fontstack}/{start}-{end}{extension:[^\d]*}",
```

Could you add a testcase to ensure that this works and that there are no regressions?
Actix-webs' testing infratstructure is documented [here](https://actix.rs/docs/testing)

CommanderStorm (changes requested):
Hi, I have left a different approach below which I think would be simpler

(disclaimer: not a maintainer ^^)

martin/src/srv/fonts.rs
bendecaf: This looks like a good suggestion. I will try this out. Thanks for looking at the change.

PR #1491, opened 2024-08-29
docs: better differentiation between `image_id` and `sprite_id`s

CommanderStorm:
made sure that the differences between image and sprite ids is clearer

The rationale is that I got stuck here for a bit and got quite confused.
After reading the sourcecode, this was much clear.

I am a bit too explicite on purpose, feel free to (ask to) revert/reduce/rewrite as you see fit. Docs are never written alone ^^


nyurik (approved):
thanks!

PR #1492, opened 2024-08-30
feat: add `sdf` sprites via additional apis

### LINKED ISSUE #1075: Support SDF sprites
In some cases, images should be encoded using SDF first.  The @flother's [spreet crate](https://github.com/flother/spreet) already supports it with https://github.com/flother/spreet/issues/58, so it should be fairly easy (?) to use that.

## Overview
Similar to the regular spritesheets, given a directory with png/svg/... images, converts all images in that directory and subdirectories to SDF images before packaging them in a spritesheet.  If needed SDF and non-SDF spritesheets can be combined by the client, e.g. `GET /sprites/icons,sdf_icons.png` (or `.json|@2x.png|@2x.json`.  No new API endpoints are introduced.

## CLI Configuration
```bash
martin --sdf-sprites <directory_or_image>
```

## YAML Configuration

```yaml
# Sprite configuration
sprites:
  # This controls how all all images will be processed. false by default
  # This can be overridden only by setting   sources.<sprite_id>.sdf
  default_sdf: false | true | pretransformed
  paths:
    # all SVG files in this dir will be published as a "my_images" sprite source
    - /path/to/my_images
  sources:
    # SVG images in this directory will be published as a "my_sprites" sprite source
    my_sprites: /path/to/some_dir
    # SDF SVG images in this directory will be published as a "my_sprites" sprite source
    my_sprites:
      path: /path/to/some_sdf_dir
      # This tells Martin to handle images in some_sdf_dir as SDFs
      sdf: false | true | pretransformed
```

CommanderStorm:
Resolves https://github.com/maplibre/martin/issues/1075

I am not quite right if I have made everything correctly.
The project has lots of places to configure things and it was not always clear where I need an Option.

~Another thing which is likely wrong:
I'd assume as far as I read the blessing that should require new files...
Something is really odd here => need to debug that before public review makes sense~
I fucked up the testcase a bit. This is now fixed


docs/src/sources-sprites.md
```diff
@@ -54,6 +54,13 @@ configuration to the config file.
 martin --sprite /path/to/sprite_a --sprite /path/to/other/sprite_b
 ```
 
+To use generate Signed Distance Fields (SDFs) images instead, the `--make_sdf` Option can be added.
+This makes images be handled as a SDFs allowing their color to be set at runtime in the map redering engines.
```
CommanderStorm:
```suggestion
SDF-Images allow their color to be set at runtime in the map redering engine.
```

docs/src/sources-sprites.md
```diff
@@ -68,4 +75,8 @@ sprites:
   sources:
     # SVG images in this directory will be published as a "my_sprites" sprite source
     my_sprites: /path/to/some_dir
+  # This tells Martin to handle images in directories as Signed Distance Fields (SDFs)
+  # Images are handled as a signed-distance field (SDF) allow their color to be set at runtime in the map redering enines.
```
CommanderStorm:
```suggestion
  # SDF-Images allow their color to be set at runtime in the map redering engine.
```

nyurik:
Thanks, this is an awesome feature to have.  One thing I kept thinking while reviewing:  would it be possible to NOT have any of the associated config values, and instead simply allow both sdf and non-sdf usage?  It's not very difficult to run through the `spreet` crate twice and generate two spritesheets, and this would simplify usage -- either use `/sprite` or `/sdf` (or some other prefix)?

TBD:
* does it makes any business sense to do this?
* are there are any cases where sdf would fail to generate, while regular sprites would work?
* do we ever want to support non-svg files for sprites, and how would that work?

CommanderStorm:
Honestly:
I have not see a difference between the pngs and sdf images, except for the `sdf: true` metadata.

I need to look into if this is possible, but I think we could just activate this automatically if users are using monochrome icons.

Using a different url is fine too.
You likely don't want to mix the monochrome, but color-configurable sdf images and the regularly colored pngs, but that solution achieves the goal.
What do you think about `/sprites/sdf/..` instead? They are still sprites, so having that under this url is likely better.

Will have to test if this would be a breaking change though ^^
=> Will rework this today


Answering your questions:

- > does it makes any business sense to do this?
 
 Yes, thinking about it, I can see a reason why one would want both (multi colored and dynamically colored icons on the same map)
- > are there are any cases where sdf would fail to generate, while regular sprites would work?
  
  I have not come across cases where this, *fails* to generate, but sdfs don't support multiple colors
  => the output here is black, unless colored in the map
- > do we ever want to support non-svg files for sprites, and how would that work?
  
  generating sdf from pngs is not possible (we are missing the distances) => users would need to supply sdf+png images to have both functionality.
  If we have such images, we should likely only support adding pngs to the pngs, not sdfs. 
  Reasoning: for an sdf to exist, you would need to have a svg file (or other vector format) at some point (unless I am missing something)
  => just add the svg instead.

In any case, I need to make this more clear in the docs, why one would choose the one or the other image format.

hiddewie:
Hi 👋🏻, chipping into this nice PR because it would be very helpful for improving the symbol rendering one of my projects using the MapLibre stack (https://github.com/hiddewie/OpenRailwayMap-vector).

>  I have not see a difference between the pngs and sdf images, except for the sdf: true metadata.

Testing the output from the `spreet` commandline shows a difference for me when rendering the same directory

```
spreet symbols/general sprites
```
gives me
sprites

while 

```
spreet symbols/general sprites --sdf
```
gives me
sprites

(colorless and with the distance field visible).

So the Martin SDF sprites should output something similar when using the spreet library. 

>  One thing I kept thinking while reviewing: would it be possible to NOT have any of the associated config values, and instead simply allow both sdf and non-sdf usage?

It would be useful for me to have both normal and SDF sprites output and available for map rendering. The first thing I can think of is rendering the same symbol twice, one SDF with the icon outline, and the colorful icon on top to automatically generate a (fixed color) halo around a symbol.

And the other reason mentioned above, to render both colored and monochrome variants of the same image on the map is also useful.

---

The combination of a colorful image together with the SDF data would be ideal, but I don't think that is supported in a single sprite image by any of the libraries or SDF sprite consumers.

Maybe there could be a way to change the consuming map rendering libraries to accept two paths of the same sprite resource, one for a colorful image, and one for the SDF image, and the rendering library could combine the information to render colorful icons together with an icon halo.

CommanderStorm:
Yea.. I think this might be a bug in spreet..

if I do
```
SpritesheetBuilder::new()
    .make_sdf()
    .sprites(sprites.into_iter().collect());
    .generate()
```

<details><summary>I get this (click to expand)</summary>
<p>

image

</p>
</details>

CommanderStorm:
> I have not see a difference between the pngs and sdf images, except for the sdf: true metadata.

That was user error. Thanks for noticing, that feature would have not worked.

I have raised a PR documenting the behaviour upstream https://github.com/flother/spreet/pull/85 and fixed the bug in my code ^^

martin/src/sprites/mod.rs
```diff
@@ -186,14 +189,19 @@ async fn parse_sprite(
...
+        Sprite::new(tree, pixel_ratio).ok_or_else(|| SpriteInstError(path.clone()))?
+    };
 
     Ok((name, sprite))
 }
 
 pub async fn get_spritesheet(
     sources: impl Iterator<Item = &SpriteSource>,
     pixel_ratio: u8,
+    as_sdf: bool,
```
CommanderStorm, line 203: Attention: There are 2-3 semver-breaking changes here.

Without reworking the API completely, I don't see how this can be avoided.

=> @nyurik would a major version bump be okay?

martin/src/sprites/mod.rs
```diff
@@ -186,14 +187,19 @@ async fn parse_sprite(
     let tree = Tree::from_data(&buffer, &Options::default())
         .map_err(|e| SpriteParsingError(e, path.clone()))?;
 
-    let sprite = Sprite::new(tree, pixel_ratio).ok_or_else(|| SpriteInstError(path.clone()))?;
+    let sprite = if as_sdf {
+        Sprite::new_sdf(tree, pixel_ratio).ok_or_else(|| SpriteInstError(path.clone()))?
```
CommanderStorm: unsure if I should add another error variant..
Looking at spreet, the failiour cases are similar-ish.

=> accuracy vs semver-breakage
=> your call @nyurik

docs/src/sources-sprites.md
```diff
@@ -40,6 +41,19 @@ the PNG, there is a high DPI version available at `/sprite/<sprite_id>@2x.json`.
 }
 ```
 
+##### Coloring at runtime via Signed Distance Fields (SDFs)
```
CommanderStorm, line 44: > I need to make this more clear in the docs, why one would choose the one or the other image format

@hiddewie (hope you don't mind the ping too much)
Could you have a look at the docs if there is something that I missed?

hiddewie: This looks great, thanks!

martin/src/sprites/mod.rs
nyurik: this is fine, no need for a new error i think.  You don't need to duplicate error creation though - both `new` and `new_sdf` return `Option<Self>` - so you may want to consolidate error handling

martin/src/sprites/mod.rs
nyurik: we are totally ok to keep bumping major for martin - no biggie

nyurik (commented):
looks better with each pass!

WRT url - i'm still a bit on the fence on the API...  `/sprite/sdf/...` does make sense, but then you have to handle if the user gives a sprite source named `sdf` (which probably is not that unusual)...  On the other hand, a source called `/sprite_sdf/...` or `/sdf/...` or `/sdf_sprite/...` keeps naming collision resolution in one place (we already do this for sources), and keeps overall pattern consistent (no URL sub-branches).  I would prefer to keep API flatter, just not sure of the naming.

martin/src/sprites/mod.rs
```diff
@@ -186,14 +187,19 @@ async fn parse_sprite(
...
+    } else {
+        Sprite::new(tree, pixel_ratio).ok_or_else(|| SpriteInstError(path.clone()))?
+    };
```
CommanderStorm: The reason why I left this duplicated is that I don't think that the alternatives below are better ^^

here are the alternatives:

```suggestion
    let new_sprite = if as_sdf { Sprite::new_sdf } else { Sprite::new };
    let sprite = new_sprite(tree, pixel_ratio).ok_or_else(|| SpriteInstError(path.clone()))?;
```
```suggestion
    let sprite = if as_sdf {
        Sprite::new_sdf(tree, pixel_ratio)
    } else {
        Sprite::new(tree, pixel_ratio)
    }
    .ok_or_else(|| SpriteInstError(path.clone()))?;
```

```suggestion
    let sprite = if as_sdf {
        Sprite::new_sdf(tree, pixel_ratio)
    } else {
        Sprite::new(tree, pixel_ratio)
    };
    let sprite = sprite.ok_or_else(|| SpriteInstError(path.clone()))?;
```

martin/src/sprites/mod.rs
CommanderStorm: continued in https://github.com/maplibre/martin/pull/1492/files#r1809425239

CommanderStorm:
Have changed it to `/sdf_sprite/..` as far as the choices prefered by you, that is the best imo.

The reason why I introduced it as `/sprite/sdf/..` without any backwards compatibiltiy considerations is the following:

> The default regular expression associated with a replacement marker `[^/]+` matches one or more characters which are not a slash. For example, under the hood, the replacement marker `{foo}` can more verbosely be spelled as `{foo:[^/]+}.` 

https://actix.rs/docs/url-dispatch/#generating-resource-urls

=> even if a sprite was previously named `"sdf"`, it could not match the new routes

martin/src/sprites/mod.rs
nyurik: thx, slightly shorter now.  I wasn't even thinking of the first option, nice, but too magical :)  I generally prefer consolidated error handling when possible, but not too crazy (whatever that even means )

nyurik:
Thx  for changing - i do think it looks a bit better now, but if anyone has other thoughts, please chime in.  @CommanderStorm could you add some tests to the [test.sh](https://github.com/maplibre/martin/blob/6f201fcf69322272bb3681d6874839431954e264/tests/test.sh#L353)?  Make sure to run `just bless` afterwards.

CommanderStorm:
> could you add some tests to the [test.sh](https://github.com/maplibre/martin/blob/6f201fcf69322272bb3681d6874839431954e264/tests/test.sh#L353)?

Shure thing, have added it

nyurik (approved):
awesome work!!!

hiddewie:
Great work @CommanderStorm, this gives nice results on my maps!

CommanderStorm:
We finally also had the time to finish the refactoring.
A bit of eye candy:

image

PR #1501, opened 2024-09-05
Use layer id if it is provided

### LINKED ISSUE #1472: Explicit layer_id ignored in tile.json
https://github.com/maplibre/martin/blob/c000fd950151e119ae51e5b7777380e89485f641/martin/src/pg/config_table.rs#L81-L106

Shouldn't the “layer_id” (if available) be used in line 96 and not the “source_id”?

I have explicitly set a “layer_id” in my config.yaml, which is also correctly present in the tiles itself. But in the tile.json is the “source_id”.

There are therefore contradictory entries.

MapLibre displays the layer, but at the same time complains that the layer does not exist.

image
image
image

sharkAndshark:
Try to fix #1472

### DIFF
--- a/martin/src/pg/config_table.rs
+++ b/martin/src/pg/config_table.rs
@@ -83,7 +83,7 @@ impl PgInfo for TableInfo {
         format!("{}.{}.{}", self.schema, self.table, self.geometry_column)
     }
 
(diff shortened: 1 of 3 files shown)


nyurik:
thanks!

PR #1507, opened 2024-09-09
Use comment on view to generate TileJson

### LINKED ISSUE #1494: Comment as TileJSON for Database View is not working
Hey, i am trying to customize the TileJSON in order to have some additional parameters in it with my custom text.

I am running a 

- martin tiles server v0.14.2 on a ubuntu VM
- "PostgreSQL 16.3 on x86_64-pc-linux-gnu, compiled by gcc (GCC) 7.3.1 20180712 (Red Hat 7.3.1-12), 64-bit"

and using View as the tile source.

In this views I have tried to make change via the function as described in the [docs](https://maplibre.org/martin/sources-pg-tables.html)
without success.

Additionally, I have tried just to put a simple JSON formatted text into the View description e.g.:

```
{
    "source_name": "Some source name",
    "source_description": "Some source description"
}
```

without a effect.

This applies for both cases, "auto_publish" and specific configuration of the tile source via "tables" section of the config.yaml

In all cases just the "standard" generated TileJSON is displayed without any information from the view description.
There are also any errors displayed, when I am setting `RUST_LOG=debug`.

Do have any advise?

sharkAndshark:
Try to fix #1494 
- [x] Update [query_available_tables.sql](https://github.com/maplibre/martin/pull/1507/files#diff-449496b294ee2413fc0d5e4290688a50f008e07fece5a613da1c2417548a06bf)
- [x] Add and update test

### DIFF
--- a/martin/src/pg/scripts/query_available_tables.sql
+++ b/martin/src/pg/scripts/query_available_tables.sql
@@ -55,9 +55,10 @@ WITH
         SELECT
             pg_namespace.nspname AS schema_name,
             relname AS table_name,
-            CAST(obj_description(relfilenode, 'pg_class') AS VARCHAR) AS description
+            pg_description.description AS description
         FROM pg_class
             JOIN pg_namespace ON pg_class.relnamespace = pg_namespace.oid
+            LEFT JOIN pg_description ON pg_class.oid = pg_description.objoid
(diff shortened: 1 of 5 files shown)


nyurik (approved):
thanks!

Issue #1539, opened 2024-10-09
Add an ability to wait for the PostgreSQL to become available before starting


nyurik:

teyotan:
May I try working on this?

nyurik:
sure thing

mzeedijk:
Not sure if this helps but I added a `After` and a `ExecStartPost` line for this to the systemd service service file. For me it did the trick:

```
[Unit]
Description = A blazing fast map tile server which can generate and serve vector tiles on the fly from large PostGIS databases, PMTile, and MBTile files
After = postgresql.service

[Service]
ExecStart = /usr/bin/martin --config /usr/local/etc/martin/config.yaml
ExecStartPost = /usr/bin/timeout 30 sh -c 'while ! ss -H -t -l -n sport = :5432 | grep -q "^LISTEN.*:5432"; do sleep 1; done'
Restart=always

[Install]
WantedBy = multi-user.target
```

CommanderStorm:
Naiively retriying would mean that we don't give a status what we are doing for a long time.

If the database is dead / incorrect config string,  some users might not want to wait 30s, without giving an indication of why it is taking so long.
For this, we should log the fact that we are trying to re-establish the connection.

The requirements for this issue to be a success are:
* we give clear indication of the status, but without flooding the output, and only if it is taking more than 2s (?) from the start of the program
* we offer a clear way to bypass or configure this with a cli + config params, notifying of this method on the first printed warning

### Configuration ideas
Add `pg-connection-retries (<number>|infinite)` as a CLI parameter, as well as in the config block

PR #1545, opened 2024-10-21
Retry connections to PostgreSQL before starting martin

teyotan:
Trying to solve (#1539)

There are still a lot of improvements that can be done, such as making the retry params configurable and adding test for unreachable db instance. But would love to get feedbacks first before I proceed on those.


nyurik:
Thanks, looks good!  We should discuss the configuration for this too as part of this PR. Jitter is a good idea, thx.

At some point we should also test/evaluate what happens when PG restarts (drops connections, switches to another primary, etc) while Martin is running - in theory PG pool should simply reconnect, but I have never tested that. Probably good for another PR.

martin/src/pg/config.rs
```diff
@@ -114,7 +116,18 @@ impl PgConfig {
     }
 
     pub async fn resolve(&mut self, id_resolver: IdResolver) -> MartinResult<TileInfoSources> {
-        let pg = PgBuilder::new(self, id_resolver).await?;
+        // Retry strategy: Fixed 5 seconds interval backoff with jitter (random variation)
+        let retry_strategy = FixedInterval::from_millis(5000)
+            .map(jitter) // Add random jitter to avoid "thundering herd" problem
```
CommanderStorm: I think this might be a bit unexpected for users:

This strategy might not wait at all.
=> all 3 request may be either a total of 15s or 0s.

https://docs.rs/tokio-retry/latest/src/tokio_retry/strategy/jitter.rs.html#3-5

The amount to which jitter affects the `FixedInterval` should likely not be 0%-100%

nyurik: Good catch, thx, I agree - we should probably do something like +/- 20% - does jitter support that?

CommanderStorm: We can just have our own function which does what we need..
The one-liner is not that complicated 😁

Also we might want to lower the retry time and increase the retry amount. As far as I remember it (might be misremembering) Postgres takes a hot few seconds (somewhere between 10 and 20) to start up

teyotan: Sorry for the late reply, suddenly got a lot on my plate. I'll continue working on this around Tuesday or Wednesday.

There are multiple parameters that we can configure:
1. Retry strategy. Tokio_retry has [3 built in strategy](https://docs.rs/tokio-retry/latest/tokio_retry/strategy/index.html), do we want it to be configurable or do we just choose one? If we do choose one, exponential or fibonacci back off might be a better choice than fixed interval.
2. the interval (and factor if we use exponential/fibonacci). Choosing the default value for this one is a bit tricky. In my limited experience, CMIIW, cold starting a postgresql server does not take very long, <1s. It's another story if you also start a server or a container. I think a good option right now is to have retry amount of 5 and the sum of wait duration across these 5 retry is 60 seconds.
3. Jitter. Yeah I didn't realize jitter just multiply the duration by 0-100%, my bad (first time working with rust, im guilty of skipping some part that I don't understand lol). We can do jitter by 80%-120%. Do we want jitter to be configurable (enable/disable and jitter range)?

CommanderStorm (approved):
I moved a bit of code around and think this works better now.

Yes, the 20s total time for PG to accept our connection is a bit much, but as long as this saves someone from debugging something really stupid like adding a bit too long running startup sql..

nyurik:
The code looks good, but I do have a big concern:

if the database is dead / incorrect config string,  we now will wait for up to 30s, without giving an indication of why it is taking so long, and the fact that we are trying to re-establish the connection?

We cannot merge this until:
* we give clear indication of the status, but without flooding the output, and only if it is taking more than 2s (?) from the start of the program
* we offer a clear way to bypass or configure this with a cli + config params, notifying of this method on the first printed warning

### Configuration ideas
Add `pg-connection-retries (<number>|infinite)` as a CLI parameter, as well as in the config block

nyurik (changes requested):
The code looks good, but I do have a big concern:

if the database is dead / incorrect config string,  we now will wait for up to 30s, without giving an indication of why it is taking so long, and the fact that we are trying to re-establish the connection?

We cannot merge this until:
* we give clear indication of the status, but without flooding the output, and only if it is taking more than 2s (?) from the start of the program
* we offer a clear way to bypass or configure this with a cli + config params, notifying of this method on the first printed warning

### Configuration ideas
Add `pg-connection-retries (<number>|infinite)` as a CLI parameter, as well as in the config block

CommanderStorm:
Thanks for the contribution! I believe Yuri's direction on this aligns better with our goals, especially in making things fully transparent to the user.

Since there hasn’t been much activity here for a while, I’ll go ahead and close this PR for now. If I’ve misunderstood the status or you’d like to continue working on it, feel free to reopen or ping us—we would be delighed to reopen the PR!

Issue #1542, opened 2024-10-18
Fail to start if PG pool size is set to `<= 0`


nyurik:
Martin should not allow PostgreSQL pool size to be set to less than 1

sikehish:
I'm looking into this :)

PR #1546, opened 2024-10-21
Added validate function to PgConfig and invoked it in finalize and pool initialization

Saru2003:
This PR addresses issue [#1542](https://github.com/maplibre/martin/issues/1542), which states that the PostgreSQL pool size should not be allowed to be set to less than 1.

Changes:
1. Introduced a validate method in the PgConfig struct to check the following:

- The pool_size must be greater than or equal to 1.
- A connection string must be provided.

2. Updated the finalize and resolve methods to call the validate method. This ensures that the configuration is validated before proceeding with further processing.

3. Updated pool.rs, ensures the validation process is integrated when creating a new pool by calling the validate method from the PgConfig instance.

Please review my implementation and provide feedback.


Saru2003:
@nyurik Kindly refer to the latest commit I have made.

martin/src/pg/config.rs
```diff
@@ -110,10 +128,12 @@ impl PgConfig {
             self.auto_publish = OptBoolObj::Bool(true);
         }
 
+        self.validate()?;
         Ok(res)
     }
 
     pub async fn resolve(&mut self, id_resolver: IdResolver) -> MartinResult<TileInfoSources> {
+        self.validate()?;
```
nyurik: why do we need to validate it in so many places - here, in the pool, and in finalize?  I don't remember for sure which codepath always executes, but we should probably limit it to just one. Probably finalize? Please check. Thx!

martin/src/pg/errors.rs
```diff
@@ -69,4 +69,7 @@ pub enum PgError {
 
     #[error(r#"Unable to get tile {2:#} with {:?} params from {1}: {0}"#, query_to_json(.3.as_ref()))]
     GetTileWithQueryError(#[source] TokioPgError, String, TileCoord, Option<UrlQuery>),
+
+    #[error("Validation error: {0}")]
+    ValidationError(String),
```
nyurik: Lets call this `ConfigError`. Also, at least for now, all errors are `&'static str` - so we don't need to call `.into()` on them to convert str into an allocated String.

nyurik (commented):
a few minor things, and this will be ready to go! :)

nyurik (approved):
thanks!!

Saru2003:
@nyurik Thanks for merging! Could you please add the "hacktoberfest-accepted" label? Thanks for considering!

CommanderStorm:
@Saru2003 adding labels is only nessesary if a PR is not accepted via an approving review or merged.

nyurik:
i'm a bit confused - what labeling are you talking about?  We only add labels relevant to the Martin project, e.g. which subsystem the PR applies to, or things like "help needed", etc.

PR #1559, opened 2024-10-30
Do not update tiles with same hash

pka:


mbtiles/src/mbtiles.rs
```diff
@@ -216,7 +216,7 @@ impl Mbtiles {
                 ),
                 Some(format!(
                     "
-    INSERT {on_duplicate} INTO images (tile_id, tile_data)
+    INSERT OR IGNORE INTO images (tile_id, tile_data)
```
CommanderStorm, line 219: This PR is missing a lot of context. Describing the changes you want/need to make goes a long way..

- What was your original issue? 
  Could you attach a testcase to avoid regressions?
- Why does changing just this insert statement to [`CopyDuplicateMode::Ignore`](https://github.com/maplibre/martin/blob/88aa4f93dbb5f8d80530da589ceb1cf692ddf592/mbtiles/src/copier.rs#L33) solve this? Why should this behave differently for `images`?

nyurik:
@pka hi, i'm not sure what you tried to do here - in the copying, it is already possible to set how override will be done with a CLI flag

PR #1585, opened 2024-11-16
feat!: `DashMap` based catalog

CommanderStorm:
This PR changes the internal datastructure to be `Dashmap<..>` as suggested in https://github.com/maplibre/martin/pull/1179#pullrequestreview-2149936800
I thought that the data/configuration/.. split might be more managble in terms of reviwing and shipping this.

Sadly two changes did cascade a bit:
- `&dyn Source` -> `Box<dyn Source>`
- `DashMap` does not guarantee sorting of keys => I need to manually do so

I had some free time and this seemed like a nice issue to progress on, I hope you (@sharkAndshark) don't mind me stealing this part ^^


CommanderStorm:
Here is a benchmark of the versions "requesting" tiles (i.e. without the network/file io overhead)

| branch | features | image |
|--------|--------|--------|
| `master` | - | image |
| `dashmap-based-catalog` | `inline` | image | 
| `dashmap-based-catalog` | `rayon` | image |
| `dashmap-based-catalog` | `rayon`<br>`inline` | image | 


=> it is a significant performance drop, but 60ns to enable updating is likely an okay tradeoff. What do you people think?

sharkAndshark:
I'm really happy to see you pick this up again. It's a looooooong requested feature. And Martin needs more  contributor. Go ahead! : D @CommanderStorm

sharkAndshark:
Consider making the use of `dashmap` and `/refresh API` optional (with a default)? @nyurik This would help users like me who don't mind rebooting Martin to reflect database updates.

CommanderStorm:
> Consider making `dashmap` optional

No, splitting the implementation on one of the core datastructures is a great way to get bugs and maintainability issues.
The 60ns performance drop should not matter to you (remember that 60ns = 0.000060ms)

`/refresh` is not included in this PR, but will be disablable in the config as far as I read the spec

sharkAndshark:
60ns means nothing to me. I'm totally Ok the performance part.
So yes let's use dashmap and only make /refresh optional(in the next PRs):D

martin/Cargo.toml
```diff
@@ -113,6 +113,7 @@ tilejson.workspace = true
 tokio = { workspace = true, features = ["io-std"] }
 tokio-postgres-rustls = { workspace = true, optional = true }
 url.workspace = true
+dashmap.workspace = true
```
nyurik: please sort this list

martin/src/fonts/mod.rs
```diff
@@ -301,44 +300,43 @@ fn parse_font(
             .replace_all(name.as_str(), " ")
             .to_string();
 
-        match fonts.entry(name) {
-            Entry::Occupied(v) => {
+        if let Some(v) = fonts.get(&name) {
```
nyurik: dashmap support the entry api - why did you switch to the get+insert, with a potential for a race condition?

martin/src/bin/martin-cp.rs
```diff
@@ -392,7 +392,7 @@ fn parse_encoding(encoding: &str) -> MartinCpResult<AcceptEncoding> {
 async fn init_schema(
     mbt: &Mbtiles,
     conn: &mut SqliteConnection,
-    sources: &[&dyn Source],
+    sources: &[Box<dyn Source>],
```
nyurik: I'm still not certain why you need a boxed type here - what caused it to change?  Also, you might want to use `TileInfoSource` for consistency

martin/src/sprites/mod.rs
```diff
@@ -130,17 +129,12 @@ impl SpriteSources {
         let disp_path = path.display();
         if path.is_file() {
             warn!("Ignoring non-directory sprite source {id} from {disp_path}");
+        } else if let Some(v) = self.0.get(&id) {
```
nyurik: same here - any reason to avoid entry api?

martin/src/sprites/mod.rs
```diff
@@ -198,18 +195,18 @@ async fn parse_sprite(
 }
 
 pub async fn get_spritesheet(
-    sources: impl Iterator<Item = &SpriteSource>,
+    paths: impl Iterator<Item = &PathBuf>,
```
nyurik: is this relevant to this PR?  I am usually ok to mix in a bunch of unrelated changes into smaller / linter / cleanup PRs, but in case of a the complex PRs that already do a lot of fundamental system changes, it is better to move unrelated stuff to a separate PR

nyurik (commented):
great work!!  Seriously impressive :)

A few things:
* could you explain a bit why you need boxed source? Can we work around it?
* please move the sorting output into a separate PR - this way test changes do not get in the way of the review
* this also goes for anything unrelated to this pr - switching to dashmap is a complex PR - so the fewer changes we have in it, the easier it is to spot issues

plus a few minor inlines

martin/src/source.rs
```diff
@@ -37,16 +38,17 @@ impl TileSources {
...
-    pub fn get_source(&self, id: &str) -> actix_web::Result<&dyn Source> {
+    pub fn get_source(&self, id: &str) -> actix_web::Result<TileInfoSource> {
         Ok(self
             .0
             .get(id)
             .ok_or_else(|| ErrorNotFound(format!("Source {id} does not exist")))?
-            .as_ref())
+            .value()
+            .clone())
     }
```
CommanderStorm, line 52: Lets provide the rationale why the `Box` change was done:

The change to the boxed variant comes from here and cascades pretty far sadly.
Because accessing a value returns `Ref<..>` => a temporary variable

image

martin/src/fonts/mod.rs
CommanderStorm: I somehow thought that the entry api did not exist.
Seems baffeling in hindsight 🤷🏻‍♂️

CommanderStorm:
@nyurik
might it be possible to re-review this PR?

This would unblock further work in re-integrating the watch-mode..

nyurik:
Sure thing, will do tonight

nyurik (approved):
Actually i don't see any issues, thanks!  Could you re-run the perf tests just to make sure we are still ok with all the changes?  feel free to merge when ready

CommanderStorm:
> Could you re-run the perf tests just to make sure we are still ok with all the changes? feel free to merge when ready

Shure, have done so.
I am seeing the same difference

PR #1587, opened 2024-11-17
chore(deps): bump from `macos-14` to `macos-latest`

CommanderStorm:
not related to the actual macos CI failour, but we might want to bump the secondary macos test up to `-latest`

On the related note:
I could not get to the bottom to the actual `macos-13` issue. Would it be possible to just drop it and replace it with `macos-latest`?


nyurik (approved):
yeah, macos + postgis has been a massive pain to test - my brew-based github action keeps going up and down. Really annoying.

PR #1599, opened 2024-11-28
Fix clippy lints 1.83, bump deps

nyurik:


martin/src/pg/utils.rs
```diff
@@ -13,6 +13,7 @@ use tokio::time::timeout;
 use crate::source::UrlQuery;
 
 #[cfg(test)]
+#[expect(clippy::ref_option)]
```
CommanderStorm: CI is failing due to it not knowing this lint.

image

=> would it be an option to postpone this until CI is updated?

```suggestion
```

nyurik: Yep, totally not in a hurry:)

nyurik: Ps ironically, I'm the one who created that lint :)

CommanderStorm (approved):
LGTM

PR #1612, opened 2024-12-13
added `/font/%7Bfontstack%7D/%7Brange%7D`

### LINKED ISSUE #1610: Support Fonts List
When using with Maputnik the Glyph url from Martin (https://martin./font/{fontstack}/{range}), Maputnik tries to get the list of available font names by the url:
https://martin./font/%7Bfontstack%7D/%7Brange%7D.

That would be great to have Matrtin provide that list.

CommanderStorm:
This is a pretty simple API-addition enabling the previews in maputnik.

Resolves #1610

### DIFF
--- a/docs/src/sources-fonts.md
+++ b/docs/src/sources-fonts.md
@@ -59,6 +59,18 @@ curl http://127.0.0.1:3000/catalog
 }
 ```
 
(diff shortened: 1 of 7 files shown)


nyurik:
lets first figure out if we need this or not in the #1610

CommanderStorm:
lets fix this upstream instead, agree

PR #1614, opened 2024-12-15
test if moving the CI to aarch64 fixes the homebrew issues

CommanderStorm:


CommanderStorm:
(it is not the case 😢)

PR #1616, opened 2024-12-15
tests: migrate to insta binary snapshots

CommanderStorm:
This PR migrates us to insta's binary snapshots.
These snapshots were added via
- https://github.com/maplibre/martin/pull/1563
- https://github.com/mitsuhiko/insta/issues/610

<details><summary>previous blockers</summary>
<p>

<del>Currently, this PR is not mergable as the binary snapshots don't actually work. I encountered these two bugs:</del>
- [x] <del>the binary snapshots get created as `FILENAME.snap.new.png`. This naming is inconsistent with how other snapshots are created.</del>
- [x] <del>Re-running the tests creates different metadata files which contain different data:</del>
  
  ```diff
  ---
  source: martin/src/sprites/mod.rs
  - assertion_line: 273
  expression: png
  - extension: png
  - snapshot_kind: binary
  ---
  +
  
  ```
  
  <del>
  Yes, I looked if I am using the same versions of insta, I am. `cargo test` and `cargo insta` still provide really weird results
  </del>
<del>
=> to be mergable, this PR would need work upstream.
I am going to try to write this up during the christmas holidays and provide a fix upstream.</del>

</p>
</details>


nyurik:
Note that image (binary) snapshots might not be ... stable, i.e. SDF->PNG, especially when multiple images are packed as a single sprite might produce slightly different results on different platforms.  So in short - this might cause more confusion than its worth. Maybe.  I am not saying lets not do it, i'm saying we may not get what we want in some cases - so we should check multiple platforms on this on

CommanderStorm:
I don't quite see where platform issues might come into image encoding/optimisation.
They *SHOULD* encode the same way, right?

The conclusion does not change though.
This is blocked upstream first before we can even consider running it through our CI.

nyurik:
well, technically any hashmap-based implementation is already randomized - so even that may produce different results for image packing. Plus any kind of SVG->PNG encoding might in theory differ depending on arm64 vs x64 (in theory)...

nyurik:
insta finally fixed the issue - no longer adding the type of the snapshot

CommanderStorm:
seems like the issues from https://github.com/maplibre/martin/pull/1616/#issue-2740859494 have indeed been fixed 🎉 (even without me investigating 😅)

=> lets see what CI thinks of this change

CommanderStorm:
> check multiple platforms on this on

I don't currently have an arm PC.
Does one of you two by happenstance have one?

Otherwise, I can likely borrow one of my colleges macbooks

nyurik (approved):
looks great, and it seems like it doesn't have any issues (yet?), so I guess we can merge

PR #1623, opened 2024-12-16
[Enh] Updating Group dependency for GitHub Actions and Cargo

### LINKED ISSUE #1622: Modify dependabot settings to suggest multiple changes together
https://github.com/maplibre/martin/blob/main/.github/dependabot.yml needs to be modified to stop suggesting individual dependency bumps - I think it now supports batching it

nakul-py:
### What does this PR do?

Updating the `.github/dependabot.yml` file to group dependency updates for:

- GitHub Actions: All updates will now be grouped under a single PR using the `all-actions` group.
- Cargo: Updates will be batched into a single PR using the `all-cargo` group.

### Which issue does this PR Fixes?
- Fixes #1622


.github/dependabot.yml
```diff
@@ -5,9 +5,17 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+    groups:
+      all-actions:
+        patterns:
+          - "*"
```
sharkAndshark:
```suggestion
    groups:
      all-actions-version-updates:
        applies-to: version-updates
        patterns:
          - "*"
      all-actions-security-updates:
        applies-to: security-updates
        patterns:
          - "*"
```

.github/dependabot.yml
```diff
@@ -5,9 +5,17 @@ updates:
...
   - package-ecosystem: cargo
     directory: "/"
     schedule:
       interval: daily
       time: "02:00"
     open-pull-requests-limit: 10
+    groups:
+      all-cargo:
+        patterns:
+          - "*"
```
sharkAndshark:
```suggestion
    groups:
      all-cargo-version-updates:
        applies-to: version-updates
        patterns:
          - "*"
      all-cargo-security-updates:
        applies-to: security-updates
        patterns:
          - "*"
```

sharkAndshark (commented):
Thx!!  @nakul-py Seems we must made separated groups. See the [doc](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups)

> You cannot apply a single grouping set of rules to both version updates and security updates. Instead, if you want to group both version updates and security updates using the same criteria, you must define two, separately named, grouping sets of rules. To do this, you can copy the group configuration block for the ecosystem and directory, and name each set of rules differently.

.github/dependabot.yml
```diff
@@ -5,9 +5,27 @@ updates:
...
     open-pull-requests-limit: 10
+    groups:
+      all-cargo-version-updates:
+        applies-to: version-updates
+        patterns:
+          - "*"
+      all-cargo-security-updates:
+        applies-to: security-updates
+        patterns:
+          - "*"
```
CommanderStorm: small formatting fix

```suggestion
          - "*"

```

CommanderStorm (approved):
> Seems we must made separated groups

So stupid that we need to do so.
Otherwise looks good to me

PR #1637, opened 2024-12-30
Make tile service  GoogleMaps-Compatible when the cog file is GoogleMapsCompatible

sharkAndshark:
GDAL provides an interesting option when creating COG, [TILING_SCHEME=[CUSTOM​/​GoogleMapsCompatible​/​...]](https://gdal.org/en/stable/drivers/raster/cog.html#reprojection-related-creation-options)

The goal of this PR is to allow frontend could use  the googlemaps-compatible cog without extra steps(like make a custom tile grid), as google tilling schema seems is the most widely used schema, almost every JS lib support it( even only support it.)

LIke if we hanve a tiff file which is from zoom 0 - zoom 5 and it's googlemaps-compatible.

So there is a mapping from tiff-internal-xyz to google-map-xyz, I'm wondering if we could take this mapping and pretending that it's a tile source which use google tilling schema.

A sample like below: 

|tiff ifd number|zoom level |zoom level in google maps|
|----|----|----|
|0|5|14|
|1|4|13|
|2|3|12|
|3|2|11|
|....|....|....|

## Update the tilejson
To implement this, we could update the tilejson 
```json
{
    ...,
    // tell user the zoom level is  from 9-14 instead of 0-5 
   "min_zoom": 9, 
   "max_zoom": 14,
    ...
}
```
## Convert the user requested zxy internally
To implement this, we could transform the resquested xyz to the actuall xyz of tiff tile
|user requested|the actual zxy internally|
|----|----|
|`/14/0/0`|`/5/4/3`
|....|....|


sharkAndshark:
Still a concept proofing draft.

martin/src/cog/mod.rs
```diff
@@ -375,3 +399,64 @@ fn get_images_ifd(decoder: &mut Decoder<File>) -> Vec<usize> {
     }
     res
 }
+
+fn get_tilling_schema(
+    decoder: &mut Decoder<File>,
+) -> Result<(Option<String>, Option<u8>), CogError> {
+    let gdal_metadata = decoder
+        .get_tag_ascii_string(Tag::Unknown(42112))
+        .map_err(|e| CogError::TagsNotFound(e, vec![42112], 0, PathBuf::new()))?;
```
sharkAndshark: Just return None. It's normal a tiff file have no tag 42112

martin/src/cog/mod.rs
```diff
@@ -75,9 +79,14 @@ impl Source for CogSource {
             Decoder::new(tif_file).map_err(|e| CogError::InvalidTiffFile(e, self.path.clone()))?;
         decoder = decoder.with_limits(tiff::decoder::Limits::unlimited());
 
-        let ifd = self.meta.zoom_and_ifd.get(&(xyz.z)).ok_or_else(|| {
+        let actual_zoom = if let Some(google_zoom) = self.meta.google_compatible_max_zoom {
```
sharkAndshark: should calculate the actual_x and actual_y either.

sharkAndshark:
More tests should be added and a lot of cleanup is needed.

PR #1639, opened 2025-01-02
Cleanup and bless

sharkAndshark:
- [x] Add some comments
- [x] Add a warn when ignoring a mask subfile
- [x] Fix failed test(args/root.rs)
- [x] Move some verifications to a standalone `verify_requirments` function
- [x] Remove unused errors

### DIFF
--- a/martin/src/cog/mod.rs
+++ b/martin/src/cog/mod.rs
@@ -244,7 +244,7 @@ impl SourceConfigExtras for CogConfig {
         false
     }
 }
-fn verify(decoder: &mut Decoder<File>, path: &Path) -> Result<(), CogError> {
+fn verify_requirments(decoder: &mut Decoder<File>, path: &Path) -> Result<(), CogError> {
     let chunk_type = decoder.get_chunk_type();
 
     if chunk_type != ChunkType::Tile {
@@ -296,7 +296,7 @@ fn get_meta(path: &PathBuf) -> Result<Meta, FileError> {
         .map_err(|e| CogError::InvalidTiffFile(e, path.clone()))?
         .with_limits(tiff::decoder::Limits::unlimited());
 
-    verify(&mut decoder, path)?;
+    verify_requirments(&mut decoder, path)?;
     let mut zoom_and_ifd: HashMap<u8, usize> = HashMap::new();
     let mut zoom_and_tile_across_down: HashMap<u8, (u32, u32)> = HashMap::new();
 
--- a/martin/src/cog/mod.rs
+++ b/martin/src/cog/mod.rs
@@ -324,6 +324,10 @@ fn get_meta(path: &PathBuf) -> Result<Meta, FileError> {
         zoom_and_tile_across_down.insert(zoom, (tiles_across, tiles_down));
     }
 
+    if images_ifd.is_empty() {
+        Err(CogError::NoImagesFound(path.clone()))?;
+    }
+
     Ok(Meta {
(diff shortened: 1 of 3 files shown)


martin/src/cog/mod.rs
```diff
@@ -237,67 +244,78 @@ impl SourceConfigExtras for CogConfig {
...
+
+    let images_ifd = get_images_ifd(&mut decoder, path);
+
+    for (idx, image_ifd) in images_ifd.iter().enumerate() {
         decoder
             .seek_to_image(*image_ifd)
             .map_err(|e| CogError::IfdSeekFailed(e, *image_ifd, path.clone()))?;
 
-        let zoom = u8::try_from(images_ifd.len() - (image_ifd + 1))
+        let zoom = u8::try_from(images_ifd.len() - (idx + 1))
```
sharkAndshark, line 318: The result of `images_ifd.len() - (image_ifd + 1)` would be wrong when there are masks.  Updated this to `images_ifd.len() - (idx + 1)`

martin/src/cog/mod.rs
```diff
@@ -306,18 +324,9 @@ fn get_meta(path: &PathBuf) -> Result<Meta, FileError> {
...
-
-    let max_zoom = zoom_and_ifd
-        .keys()
-        .max()
-        .ok_or_else(|| CogError::NoImagesFound(path.clone()))?;
     Ok(Meta {
-        min_zoom: *min_zoom,
-        max_zoom: *max_zoom,
+        min_zoom: 0,
+        max_zoom: images_ifd.len() as u8 - 1,
```
CommanderStorm, line 333: Is `NoImagesFound` not an error that can happen?

If `images_ifd.len()==0` this seems like buggy logic.

martin/src/cog/mod.rs
```diff
@@ -306,18 +324,9 @@ fn get_meta(path: &PathBuf) -> Result<Meta, FileError> {
...
-        .ok_or_else(|| CogError::NoImagesFound(path.clone()))?;
-
-    let max_zoom = zoom_and_ifd
-        .keys()
-        .max()
-        .ok_or_else(|| CogError::NoImagesFound(path.clone()))?;
     Ok(Meta {
-        min_zoom: *min_zoom,
-        max_zoom: *max_zoom,
+        min_zoom: 0,
```
CommanderStorm, line 332: Just that this is intentional:
COGs always include data for the zoom level `0`?

CommanderStorm (approved):
Looks good to merge as far as I am concearned.
Lets just make sure that the metadata changes are correct. ^^

martin/src/cog/mod.rs
sharkAndshark: Oops I made a mistake here.  Thx!!  :D

martin/src/cog/mod.rs
sharkAndshark: Add a is_empty() check now.

CommanderStorm: `is_empty` does not seem relevant to me.

Lets rephrase the question:
Do COGs always include data for the zoom level `0`?

(if yes: lets merge this PR)

sharkAndshark: A valid COG must have at least one image.

> A COG file SHALL have at least one IFD with a NewSubfileType tag whose bit 0 is not set, or without a NewSubfileType tag (full-resolution image data).

Issue #1632, opened 2024-12-26
Releasing `v0.15.0`


CommanderStorm:
Hi @nyurik @sharkAndshark 
Quick check in if the current state would be a good place to do a release or if something has to be finished first (we have not done one in some time => lets not leave too many features "in limbo").

I went through the commits and came up with the following. What do you think? ^^
The list is ordered somewhat by how important I think they are => please feel free to reorder

```
[![Book](https://img.shields.io/badge/docs-Book-informational)](https://maplibre.org/martin)

## Features

* Add basic `cog` (Cloud Optimized GeoTIFF)  support by @sharkAndshark in https://github.com/maplibre/martin/pull/1590 and https://github.com/maplibre/martin/pull/1635
* Add `sdf` sprites which can be colored by the map dynamically via additional apis by @CommanderStorm in https://github.com/maplibre/martin/pull/1492
* Add support for `geography` columns by @rshk in https://github.com/maplibre/martin/pull/1627
* Add basic web UI (not currently usable) d by @paigewilliams in https://github.com/maplibre/martin/pull/1395
* Add `HEALTHCHECK` instruction to the Dockerfile by @caspg in https://github.com/maplibre/martin/pull/1397
* Use comment on view to generate TileJson by @sharkAndshark in https://github.com/maplibre/martin/pull/1507
* Add a postgres version check to not run on versions of postgis that we don't support by @CommanderStorm in https://github.com/maplibre/martin/pull/1427
* Use layer id if it is provided by @sharkAndshark in https://github.com/maplibre/martin/pull/1501
* Added validate function to PgConfig and invoked it in finalize and pool initialization by @Saru2003 in https://github.com/maplibre/martin/pull/1546

## Refactorings and enhancements
* Add port fwd test by @nyurik in https://github.com/maplibre/martin/pull/1411
* Minor lints by @nyurik in https://github.com/maplibre/martin/pull/1412
* Update tests, connection limit by @nyurik in https://github.com/maplibre/martin/pull/1530
* Cleanup `test.sh`, fix json rounding issue by @nyurik in https://github.com/maplibre/martin/pull/1543
* unified the types to be based on `TileInfoSource` by @CommanderStorm in https://github.com/maplibre/martin/pull/1611
* Group dependency updates for GitHub Actions and Cargo by @nakul-py in https://github.com/maplibre/martin/pull/1623
* Temporary disable macos-13 testcase by @CommanderStorm in https://github.com/maplibre/martin/pull/1626

[...]

PR #1641, opened 2025-01-03
chore: dependency bump and re-bless

CommanderStorm:
This PR re-blesses the output and does https://github.com/maplibre/martin/issues/1632#issuecomment-2562152935


martin/src/args/root.rs
```diff
@@ -296,11 +296,13 @@ mod tests {
         let mut config = Config::default();
         let err = args.merge_into_config(&mut config, &env);
         assert!(err.is_ok());
-        assert_yaml_snapshot!(config, @r#"
+        assert_yaml_snapshot!(config, @r###"
```
nyurik: your local snap needs updating - it shouldn't generate three `#` symbols

CommanderStorm:
To be transparent what is happening in the git history:
I am having trouble reproducing the build faliour in CI locally.
This is because I am not running node 16 (as CI does). Because `vite@v6` is using `structuredClone`, `node>=17` would be required.

[Bumping the node version](https://github.com/maplibre/martin/pull/1641/commits/0c38ea22fb6767fc5272dc25d1ceb6e153263eb9) in CI is not quite working:
image

=> currently trying out how other node versions are/are not able to be installed ^^

A patch would be to [revert to `vite@v5`](https://github.com/maplibre/martin/pull/1641/commits/1360fcb58fd5008eee8f66bf40b1579b28d91323)

CommanderStorm:
Okay. `vite@v5` and `node@17` seem to be the most recent that I can push this to without totally changing the way we write our docker files.

TBH, we should likely move the building of the binary into docker anyway, to eliminate `libc`-issues with `node>=18`.
That is a different PR though.

nyurik (approved):
lgtm, thx

PR #1665, opened 2025-01-26
fix Changelog.md being chosen as a source for releases

CommanderStorm:
both `body_part` and `draft` override the current text.
That is really annoying and must be a bug, but I don't think we need this functionality (do we?)

I don't know if we want `generate_release_notes`. I have added it, but we might not want this..

(and I should have used my fork, sorry)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -653,9 +653,8 @@ jobs:
         if: startsWith(github.ref, 'refs/tags/')
         uses: softprops/action-gh-release@v2
         with:
-          draft: true
           files: 'target/files/*'
-          body_path: CHANGELOG.md
+          generate_release_notes: true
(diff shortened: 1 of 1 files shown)


nyurik:
tbh, i never used the auto-release tools (I would love to, just haven't had the brain capacity to go into it).  I'm ok with whatever method tbh. Perhaps lets do v15 by hand, and merge this after?

nyurik (approved):

CommanderStorm:
> Perhaps lets do v15 by hand, and merge this after?

I don't think that will be possible as the action overrides the content of the release, and is triggered on-release.
Which makes it back to a draft with the wrong content.

(I honestly have no clue what I and you are doing differently that you are not triggering this part and I don't want to debug it ^^)

PR #1676, opened 2025-02-02
Add custom tile schema to the TileJSON of cog

sharkAndshark:


martin/src/cog/mod.rs
```diff
@@ -39,11 +39,20 @@ impl CogSource {
...
-        let tilejson = tilejson! {
+        let mut tilejson = tilejson! {
             tiles: vec![],
             minzoom: meta.min_zoom,
             maxzoom: meta.max_zoom
         };
+
+        // Add COG-specific metadata to tilejson.other
+        let mut cog_info = serde_json::Map::new();
+        // Add extent bounds
```
CommanderStorm: FYI: tilejson has the bounds parameter which looks extremely similar.. Would this be more appropriate?

image

sharkAndshark: Hi @CommanderStorm ^^
The purpose for this PR is that COG could has tile schema totally different with 3857.  And even it's 3857 compatible the user still need to set a custom tile schema (same with 3857) to load it if without the support from #1637 
So I plan to add a `custom_tile_schema`(a better name required, let's discuss) to `tileJSON`.


## The field `.bounds` woulbe be filled or not in this PR?

Depends on whether the SRS of the COG file could be easily transform to 4326 as per the spec of `tilejson` it must [be represented in WGS 84]( https://github.com/mapbox/tilejson-spec/tree/master/3.0.0#35-bounds). I would only fill this field if it's 4326 already or 3857. It's the simplest case.

* If it's not 4326/3857, `bounds` would be left default(-180, -85.05112877980659, 180, 85.0511287798066)
* If it's 4326/3857, `bounds` would be filled

## The external extent field should be named  `.custom_tile_schema.bounds`  or`.custom_tile_schema.extent`?

It's a good question. Let's discuss. ^^

sharkAndshark:
A nested `custom_tile_schema` might be a little bit ugly. Another option is we could add the required fields for custom tile schema `tile size` `extent`(might reuse the `bounds` field , TBD) `origin` `resolutions` directly to tile json with a not nested way. But seems this could break the compatibility of tile json spec in the future I guess.
```json
{
    minzoom: 4,
    maxzoom: 14,
    origin: [0, 0],
    resolutions: [ 1.1, 2.2, 3.3, ....],
    tilesize: 256,
    ...
}
```

sharkAndshark:
1. Make services google compatible if it's google compatible. We should do this I guess. I will finish it [there](https://github.com/maplibre/martin/pull/1637)
2. Add custom tile schema fields nested.
3. Add custom tile schema fields flat.


## The nested

```
{
    ...,
    custom_tile_schema:{
           extent: [],
           tilesize: 256,
           resolutions: [],
           origin: [ ],
       },
    ....
}
```

## The flat

```
{
    minzoom: 4,
    maxzoom: 14,
    origin: [0, 0],
    resolutions: [ 1.1, 2.2, 3.3, ....],
    tilesize: 256,
    ...
}

```
flat or nested, I'm a little prefer flat now.

martin/src/cog/mod.rs
CommanderStorm: If it is in a differen projection, lets not use the `bounds` name.
Some rendering engines might use the field to not request tiles ^^

sharkAndshark:
CC @nyurik

sharkAndshark:
It's still an early draft. We could decide flat or not after all fields added.

nyurik:
If we want to modify tilejson, we may need to step back for a bit and come up with a "standard-compatible" way to:
* support non-web-mercator projections
* support COG data type -- currently there is no [official](https://github.com/mapbox/tilejson-spec/tree/master/3.0.0) `type` field. We may need to establish a MIME-based type field. See also https://github.com/opengeospatial/CloudOptimizedGeoTIFF/issues/1
* [schema](https://github.com/mapbox/tilejson-spec/tree/master/3.0.0#315-scheme) is already documented as either `tms` or `xyz` (default).  If we need a different tile addressing schema, this is the place to do it.

martin/src/cog/source.rs
```diff
@@ -459,4 +714,128 @@ mod tests {
...
+
+    #[test]
+    fn can_get_model_infos() {
+        let path = PathBuf::from("../tests/fixtures/cog/rgb_u8.tif");
+        let tif_file = File::open(&path).unwrap();
+        let mut decoder = Decoder::new(tif_file).unwrap();
+
+        let (pixel_scale, tie_points, transformation) = super::get_model_infos(&mut decoder, &path);
+
+        assert_yaml_snapshot!(pixel_scale, @r###"
```
nyurik, line 764: please update your `cargo insta` version

sharkAndshark: thx!

sharkAndshark:
I believe it's almost there but got really busy these days, would be back next week

nyurik:
wow, quiet a lot of work went into this one!  @sharkAndshark what did you think about my last  [comment](https://github.com/maplibre/martin/pull/1676#issuecomment-2634866823)?

sharkAndshark:
> If we want to modify tilejson, we may need to step back for a bit and come up with a "standard-compatible" way to:
> 
> * support non-web-mercator projections
> * support COG data type -- currently there is no [official](https://github.com/mapbox/tilejson-spec/tree/master/3.0.0) `type` field. We may need to establish a MIME-based type field. See also [ Media (MIME) type for COG opengeospatial/CloudOptimizedGeoTIFF#1](https://github.com/opengeospatial/CloudOptimizedGeoTIFF/issues/1)
> * [schema](https://github.com/mapbox/tilejson-spec/tree/master/3.0.0#315-scheme) is already documented as either `tms` or `xyz` (default).  If we need a different tile addressing schema, this is the place to do it.


## 

## Schema

No new tile schema added in this PR.  The y-axis for tile index is from top to down in COG.

I would say custom `tile grid` added instead of  custom `tile schema` added to make the goal of this PR clear.

## MIME

I was not aware of this. But seems we no worry about it? The format output tile of COG is `png`

## TileJSON compatible

I prefer add all these fields nested in a `custom_grid` field to not break the compatiblity. Not sure if it's this good. How do you think? @nyurik

tests/expected/auto/rgb_u8.json
```diff
@@ -1,4 +1,28 @@
 {
+  "custom_grid": {
```
sharkAndshark, line 2: Actually I'm not sure shoud we added them here( It's a little bit ugly..)

Maybe add these fileds to the json of `/catalog`?

sharkAndshark: `curl http://127.0.0.1:3000/catalog`
```json
{
   "fonts": {...},
    "spirtes": {...},
    "tiles": {....},
    "tileGrid": {
         "rgb_u8": {....}
    }
}
```

sharkAndshark: Maybe we need to figure out a general way to support tile grid not 3857 in Martin.

sharkAndshark:
If it's difficult to make a decision about where to place these fields maybe it's not bad to make a CLI to output them(or just print them in console log). The compatibility would be greatly honored.

And maybe discuss more about support for tile grid not Web Mercator in another issue. It could be related this PR I guess.

docs/src/sources-cog-files.md
```diff
@@ -57,6 +57,104 @@ cog:
...
+var source = new XYZ({
+  url: "http://10.1.155.35:3000/rgb_u8/{z}/{x}/{y}", tileGrid: custom_grid
+});
+
+const map = new Map({
+  target: 'map',
+  layers: [
+    new TileLayer({
+      source: new XYZ({
+        url: 'https://api.maptiler.com/maps/basic/{z}/{x}/{y}.png?key=vgdy5zgzM4IIqQCC6SOn',
```
nyurik: probably shouldn't have a key here, especially since it is not related to COG

nyurik (commented):
I keep thinking how we can stabilize the custom grid thing - perhaps collaborate with the OpenLayers on how to parametrize it?  You already have most of the needed values there I think

docs/src/sources-cog-files.md
sharkAndshark: Oops. I should remove this.

sharkAndshark:
I suggest to remove the doc about cog before we figure out how to support non Web Mercator tile grid and how to tell user the tile grid  of cog. 
As user could not display tiles correctly without these fileds, keeping this doc would be confusing. 

I'm sorry for this. But definitely I would keep working.  @nyurik

nyurik:
Thank you so much for your hard work on this!  I am sure we can figure it out - clearly not an easy thing to do right, and even harder to change once we release it.  Luckly, we are not the first to tackle this problem, so should use the knowledge gained by others

sharkAndshark:
> Thank you so much for your hard work on this!  I am sure we can figure it out - clearly not an easy thing to do right, and even harder to change once we release it.  Luckly, we are not the first to tackle this problem, so should use the knowledge gained by others

Yea! Let's do it! ^^

nyurik:
BTW, #611 was trying to add non-web-mercator projection - we may want to consolidate these efforts

CommanderStorm:
I am not sure what the status of this PR is, but the following PR seems like solving a very similar thing
- https://github.com/maplibre/martin/pull/2510

Lets track this there

PR #1679, opened 2025-02-03
Upgrade to postgis:16-3.5 in test

sharkAndshark:
Related to #1651 

- [x] Upgrade our postgis version to 16-3.5
- [x] Update mbtiles fixtures


CommanderStorm:
Could we add a warning like in
- https://github.com/maplibre/martin/pull/1427

Something like
> In versions prior to postgis 3.5 some geometry was hidden on some zoom levels. If You encounter this bug, please consider updating your postgis installation. For further details please refer to ISSUE

sharkAndshark:
Let's add the warning another PR to make each PR dedicated and small enough to gain a quick review, how do you think? @CommanderStorm ^^

CommanderStorm (approved):
Fair enough, LGTM

PR #1681, opened 2025-02-06
add a version warning for Postgis<3.5

CommanderStorm:
could not go to sleep, so PR instead.. insomnia is a bitch ^^

This adds a version warning for postgis<3.5

Yes, it might be a bit too early too tell, but all signs point to older versions being buggy.
Lets keep https://github.com/maplibre/martin/issues/1651 open to keep tabs on weather the upstream issue changes.

### DIFF
--- a/martin/src/pg/pool.rs
+++ b/martin/src/pg/pool.rs
@@ -50,15 +50,19 @@ impl PgPool {
         if pg_ver < MINIMUM_POSTGRES_VERSION {
             return Err(PostgresqlTooOld(pg_ver, MINIMUM_POSTGRES_VERSION));
         }
-        if pg_ver < RECOMMENDED_POSTGRES_VERSION {
-            warn!(
-                "PostgreSQL {pg_ver} is older than the recommended {RECOMMENDED_POSTGRES_VERSION}."
-            );
-        }
 
         let postgis_ver = get_postgis_version(&conn).await?;
         if postgis_ver < MINIMUM_POSTGIS_VERSION {
             return Err(PostgisTooOld(postgis_ver, MINIMUM_POSTGIS_VERSION));
+        }
+
+        // In the warning cases below, we could technically run.
+        // This is not ideal for reasons explained in the warnings
+
+        if pg_ver < RECOMMENDED_POSTGRES_VERSION {
+            warn!(
+                "PostgreSQL {pg_ver} is older than the recommended minimum {RECOMMENDED_POSTGRES_VERSION}."
+            );
         }
         let margin_not_supported = postgis_ver < ST_TILE_ENVELOPE_POSTGIS_VERSION;
         if margin_not_supported {
--- a/martin/src/pg/pool.rs
+++ b/martin/src/pg/pool.rs
@@ -33,7 +33,7 @@ pub struct PgPool {
     /// Indicates if `ST_TileEnvelope` supports the margin parameter.
     ///
     /// `true` if running postgis >= 3.1
-    /// This being `False` indicates that tiles may be cut off at the edges.
+    /// This being `false` indicates that tiles may be cut off at the edges.
     supports_tile_margin: bool,
 }
 
--- a/martin/src/pg/pool.rs
+++ b/martin/src/pg/pool.rs
@@ -70,7 +70,7 @@ impl PgPool {
             warn!("PostGIS {postgis_ver} is older than {ST_TILE_ENVELOPE_POSTGIS_VERSION}. Margin parameter in ST_TileEnvelope is not supported, so tiles may be cut off at the edges.");
         }
         if postgis_ver < MISSING_GEOM_FIXED_POSTGIS_VERSION {
-            warn!("PostGIS {postgis_ver} is older than the recommended minimum {MISSING_GEOM_FIXED_POSTGIS_VERSION}. In prior versions, some geometry may be hidden on some zoom levels. If You encounter this bug, please consider updating your postgis installation. For further details please refer to https://github.com/maplibre/martin/issues/1651#issuecomment-2628674788");
+            warn!("PostGIS {postgis_ver} is older than the recommended minimum {MISSING_GEOM_FIXED_POSTGIS_VERSION}. In the used version, some geometry may be hidden on some zoom levels. If You encounter this bug, please consider updating your postgis installation. For further details please refer to https://github.com/maplibre/martin/issues/1651#issuecomment-2628674788");
         }
 
         info!("Connected to PostgreSQL {pg_ver} / PostGIS {postgis_ver} for source {id}");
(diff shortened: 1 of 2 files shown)


martin/src/pg/pool.rs
```diff
@@ -42,21 +45,26 @@ impl PgPool {
...
+        if pg_ver < MINIMUM_POSTGRES_VERSION {
+            return Err(PostgresqlTooOld(pg_ver, MINIMUM_POSTGRES_VERSION));
         }
-        if pg_ver < RECOMMENDED_POSTGRES_VER {
-            warn!("PostgreSQL {pg_ver} is older than the recommended {RECOMMENDED_POSTGRES_VER}.");
+        if pg_ver < RECOMMENDED_POSTGRES_VERSION {
+            warn!("PostgreSQL {pg_ver} is older than the recommended {RECOMMENDED_POSTGRES_VERSION}.");
         }
 
         let postgis_ver = get_postgis_version(&conn).await?;
```
nyurik, line 55: lets move error cases before the warning cases. This way we won't print a warning and then error out

martin/src/pg/pool.rs
```diff
@@ -42,21 +45,26 @@ impl PgPool {
...
         }
-        let margin = postgis_ver >= RECOMMENDED_POSTGIS_VER;
-        if !margin {
-            warn!("PostGIS {postgis_ver} is older than the recommended {RECOMMENDED_POSTGIS_VER}. Margin parameter in ST_TileEnvelope is not supported, so tiles may be cut off at the edges.");
+        let margin_not_supported = postgis_ver < ST_TILE_ENVELOPE_POSTGIS_VERSION;
+        if margin_not_supported {
+            warn!("PostGIS {postgis_ver} is older than {ST_TILE_ENVELOPE_POSTGIS_VERSION}. Margin parameter in ST_TileEnvelope is not supported, so tiles may be cut off at the edges.");
         }
+        let tiles_could_be_missing_geometry = postgis_ver < MISSING_GEOM_FIXED_POSTGIS_VERSION;
+        if !tiles_could_be_missing_geometry {
```
nyurik: i think you meant this without the `!`
```suggestion
        if tiles_could_be_missing_geometry {
```

martin/src/pg/pool.rs
```diff
@@ -30,8 +30,10 @@ const RECOMMENDED_POSTGRES_VERSION: Version = Version::new(12, 0, 0);
 pub struct PgPool {
     id: String,
     pool: Pool,
-    // When true, we can use margin parameter in ST_TileEnvelope
-    margin: bool,
+    /// Indicates if the margin parameter in `ST_TileEnvelope` is supported.
+    ///
+    /// This being `False` indicates that tiles may be cut off at the edges.
+    st_envelope_margin_is_supported: bool,
```
nyurik: perhaps `use_st_envelope_margin`?  A bit shorter and avoids a passive.

nyurik: maybe even remove the `st`? not certain

CommanderStorm: That is a bit cryptic. 🤔

I like your version from code below is better:
`supports_tile_margin`

martin/src/pg/pool.rs
```diff
@@ -42,25 +48,39 @@ impl PgPool {
...
+        }
+        let margin_not_supported = postgis_ver < ST_TILE_ENVELOPE_POSTGIS_VERSION;
+        if margin_not_supported {
+            warn!("PostGIS {postgis_ver} is older than {ST_TILE_ENVELOPE_POSTGIS_VERSION}. Margin parameter in ST_TileEnvelope is not supported, so tiles may be cut off at the edges.");
         }
-        let margin = postgis_ver >= RECOMMENDED_POSTGIS_VER;
-        if !margin {
-            warn!("PostGIS {postgis_ver} is older than the recommended {RECOMMENDED_POSTGIS_VER}. Margin parameter in ST_TileEnvelope is not supported, so tiles may be cut off at the edges.");
+        let tiles_could_be_missing_geometry = postgis_ver < MISSING_GEOM_FIXED_POSTGIS_VERSION;
+        if tiles_could_be_missing_geometry {
```
nyurik: might as well inline this for simplicity. The warning text is self explanatory enough
```suggestion
        if postgis_ver < MISSING_GEOM_FIXED_POSTGIS_VERSION {
```

martin/src/pg/pool.rs
```diff
@@ -42,25 +48,39 @@ impl PgPool {
...
+
+        // In the warning cases below, we could technically run.
+        // This is not ideal for reasons explained in the warnings
+
+        if pg_ver < RECOMMENDED_POSTGRES_VERSION {
+            warn!(
+                "PostgreSQL {pg_ver} is older than the recommended minimum {RECOMMENDED_POSTGRES_VERSION}."
+            );
+        }
+        let margin_not_supported = postgis_ver < ST_TILE_ENVELOPE_POSTGIS_VERSION;
```
nyurik: for clarity, lets also use a positive value here - otherwise below you end up with a confusing double negative `!margin_not_supported`. Also means you can use var as is when creating an instance.

```suggestion
        let supports_tile_margin = postgis_ver >= ST_TILE_ENVELOPE_POSTGIS_VERSION;
```

martin/src/pg/pool.rs
nyurik: yep, i like this name too

martin/src/pg/pool.rs
```diff
@@ -15,20 +15,26 @@ pub const POOL_SIZE_DEFAULT: usize = 20;
...
 
 #[derive(Clone, Debug)]
 pub struct PgPool {
     id: String,
     pool: Pool,
-    // When true, we can use margin parameter in ST_TileEnvelope
-    margin: bool,
+    /// Indicates if `ST_TileEnvelope` supports the margin parameter.
+    ///
+    /// `True` if running postgis >= 3.1
```
nyurik: if keyword, should not be capitalized.

```suggestion
    /// `true` if running postgis >= 3.1
```

martin/src/pg/pool.rs
```diff
@@ -15,20 +15,26 @@ pub const POOL_SIZE_DEFAULT: usize = 20;
...
+    /// This being `False` indicates that tiles may be cut off at the edges.
```
nyurik:
```suggestion
    /// This being `false` indicates that tiles may be cut off at the edges.
```

sharkAndshark:
Seems new warning needs to be removed either in `test.sh`

https://github.com/maplibre/martin/blob/e024ff00405dbf9602004e76eba21ad45d062331/tests/test.sh#L186-L189

CommanderStorm:
What a messy PR.. Sorry about this.. Next time no PRs at 3:30AM..

martin/src/pg/pool.rs
```diff
@@ -108,9 +127,13 @@ impl PgPool {
         self.id.as_str()
     }
 
+    /// Indicates if `ST_TileEnvelope` supports the margin parameter.
+    ///
+    /// `True` if running postgis >= `3.1`
```
sharkAndshark:
```suggestion
    /// `true` if running postgis >= `3.1`
```

sharkAndshark:
I got insomnia either last night.. loll

nyurik (approved):
thanks!

PR #1686, opened 2025-02-11
fix: make docker-`HEALTHCHECK` only query ipv4 by default 

### LINKED ISSUE #1674: Docker `HEALTHCHECK` fails due to trying to access Martin at IPv6
Hi, with the new version we got a problem, where HEALTHCHECK instruction (introduced in #1397) fails due to trying to access Martin at IPv6 (localhost is translated to [::1]), while Martin is listening on IPv4 in the container.

Tried to run the healthcheck command inside the running container and found out the following:

This fails, makes the container status unhealthy and prevents the container to be recognized by our reverse proxy (Traefik), thus not being online:
```sh
/ # wget --spider http://localhost:3000/health
Connecting to localhost:3000 ([::1]:3000)
wget: can't connect to remote host: Connection refused
```

This works okay:
```sh
/ # wget --spider http://127.0.0.1:3000/health
Connecting to 127.0.0.1:3000 (127.0.0.1:3000)
remote file exists
```

CommanderStorm:
This PR changes the docker-`HEALTHCHECK` only query ipv4 by default.

Thanks @SmallhillCZ for the report ❤️
Resolves: #1674

### DIFF
--- a/.github/files/multi-platform.Dockerfile
+++ b/.github/files/multi-platform.Dockerfile
@@ -4,5 +4,5 @@ ARG TARGETPLATFORM
 LABEL org.opencontainers.image.description="Blazing fast and lightweight tile server with PostGIS, MBTiles, and PMTiles support"
 COPY target_releases/$TARGETPLATFORM/* /usr/local/bin
 
-HEALTHCHECK CMD wget --spider http://localhost:3000/health || exit 1
+HEALTHCHECK CMD wget --spider http://127.0.0.1:3000/health || exit 1
 ENTRYPOINT ["/usr/local/bin/martin"]


sharkAndshark (approved):

sharkAndshark:
lgtm!  : )

nyurik:
looks good, thanks.  I wonder if we can make Martin listen on both IPv4 and IPv6 at the same time?

nyurik (approved):

sharkAndshark:
That would be cool if martin support IPv6( Or maybe we have supportd it already with the actix?)

nyurik:
Wouldn't IPv6 be like a one-liner to add?  Probably a separate PR though

PR #1689, opened 2025-02-15
feat(mbtiles) - Add validation to martin .mbtiles #741

### LINKED ISSUE #741: Add more validations to .mbtiles detection
As part of [mbtiles type detection](https://github.com/maplibre/martin/blob/8e7bba3535ae4d5617bbda58f05d80ac80af6de3/martin-mbtiles/src/mbtiles.rs#L279C18-L279C29), we should catch some common issues:
* [ ] detect if `tiles` (simple) or `map` (deduplicated) tables have an index by z,x,y columns
* [ ] ???

In case of a problem, use a `warn!` macro to print a warning, unless we are certain the issue makes it impossible to use a file, in which case it should be appropriate error (add it to the error enum).

A bigger discussion is if mbtiles should stop using logging, and return some `Vec<MbtilesWarning>` if non-critical issues are detected.

Auspicus:
closes #741


martin/src/mbtiles/mod.rs
```diff
@@ -31,7 +55,31 @@ impl ConfigExtras for MbtConfig {
...
+                        validation_error.to_string(),
+                    ));
+                }
+                OnInvalid::Ignore => {
+                    return Err(FileError::IgnoreOnInvalid(
+                        path,
+                        validation_error.to_string(),
+                    ));
+                }
+                OnInvalid::Warn => {
```
Auspicus: We might want to do further processing of the `validation_error` to determine if this is a critical issue or not.

martin/src/mbtiles/mod.rs
```diff
@@ -149,8 +203,10 @@ mod tests {
                 pm-src3: https://example.org/file3.ext
                 pm-src4:
                   path: https://example.org/file4.ext
+            validate: thorough
```
nyurik: minor nit: probably best to put the simple `key: value` things before the nested ones - i.e. move these two above `paths:`

mbtiles/src/errors.rs
```diff
@@ -57,7 +57,7 @@ pub enum MbtError {
     #[error("The destination file {0} is not empty. Some operations like creating a diff file require the destination file to be non-existent or empty.")]
     NonEmptyTargetFile(PathBuf),
 
-    #[error("The file {0} does not have the required uniqueness constraint")]
+    #[error("The file {0} does not have the required uniqueness constraint. Try adding a unique index on the `map` or `tiles` table.")]
```
nyurik: Perhaps link to https://maplibre.org/martin/mbtiles-schema.html - and also update that page to include some information about performance?  Note that `tiles` is the only official requirement of an mbtiles, but it could be either a table or a view - and it would be difficult to explain all this in a few lines.

martin/src/mbtiles/mod.rs
nyurik: I keep wondering if "fast" and "skip" are actually the same thing. We should always do some basic validation - like making sure the `tiles` table exists with the right columns - or else we can't even use it. The only difference is in what we do afterwards - ignore the source or continue with it...  So perhaps:

* validation level is `fast` vs `thorough` (no skip)
* severity would be `good`, `warn`, `error`
* and the action could be `abort` or `ignore source` for `error`, and `abort` or `ignore source` or `warn` on `warning`?
* we may even introduce `warnings-as-errors` parameter? (slowly building a compiler here :) )

martin/src/file_config.rs
```diff
@@ -264,7 +270,11 @@ async fn resolve_int<T: SourceConfigExtras>(
                 let dup = if dup { "duplicate " } else { "" };
                 let id = idr.resolve(&id, url.to_string());
                 configs.insert(id.clone(), source);
-                results.push(cfg.custom.new_sources_url(id.clone(), url.clone()).await?);
+                match cfg.custom.new_sources_url(id.clone(), url.clone()).await {
```
nyurik: i'm not a big fan of all this dup code for each call to `new_sources` - i wonder if we can do this logic elsewhere, and here keep the simple `?` operator as before?

nyurik (commented):
I really like where you are going with this!  I left a few minor comments inline.

A few overall thoughts:
* we may have more than one type of "local source validation" (pmtiles, sprites, fonts, cog files, ...) - so we should keep CLI parameters simple and global (not mbtiles specific)
* all configuration sources may introduce some format-specific validation steps, but we should try to keep them all consistent from configuration file perspective
* we should have multi-level config: CLI / config file root / per source type / specific path/file
* lets cleanup error handling dup code somehow
* as i wrote more in depth in a comment, we may want to have `warning-as-error` non-default mode

overall, awesome work!  I think it is pretty close to the working state. One last thing - usually you do not want to submit PRs from the main branch of your fork, but that's really a minor thing :)

nyurik:
hi @Auspicus, let me know if you need any help moving it forward, seems like a really good feature to have in the longer term

Auspicus:
Hey @nyurik! Thanks for checking in.

I'm working on this in the interest of using Martin as a replacement for MapTiler Server for my day job. Just had to pause for a bit to tackle some other things that have come up but should be able to get back to this next week, hopefully.

Cheers!

nyurik:
Thx!  You might be interested in my work on https://github.com/nyurik/maplibre-native-rs - ability to do server-side rendering... still some time away, but already shows promising results

sharkAndshark:
> Thx!  You might be interested in my work on https://github.com/nyurik/maplibre-native-rs - ability to do server-side rendering... still some time away, but already shows promising results

Wow. Really excited for this! Sadly I know nothing about cpp ( need some cpp knowledge to get involved in I guess?)

martin/src/args/root.rs
```diff
@@ -115,11 +115,6 @@ impl Args {
             config.pmtiles = parse_file_args(&mut cli_strings, &["pmtiles"], true);
         }
 
-        #[cfg(feature = "mbtiles")]
```
Auspicus: need to add back

martin/src/config.rs
```diff
@@ -209,7 +210,35 @@ impl Config {
...
 
-        Ok(TileSources::new(try_join_all(sources).await?))
+        let resolved_sources = try_join_all(sources)
+            .await?
+            .into_iter()
+            .flatten()
+            .collect::<TileInfoSources>();
+
+        let mut sources_to_prune: Vec<usize> = vec![];
+        for (idx, source) in resolved_sources.iter().enumerate() {
```
Auspicus, line 220: need to do parallel validation rather than this.

martin/src/mbtiles/mod.rs
```diff
@@ -127,14 +175,18 @@ mod tests {
...
-    use crate::file_config::{FileConfigEnum, FileConfigSource, FileConfigSrc};
+    use crate::file_config::{FileConfigEnum, FileConfigSource, FileConfigSrc, OnInvalid};
     use crate::mbtiles::MbtConfig;
 
     #[test]
     fn parse() {
-        let cfg = serde_yaml::from_str::<FileConfigEnum<MbtConfig>>(indoc! {"
+        let cfg: FileConfigEnum<MbtConfig> =
+            serde_yaml::from_str::<FileConfigEnum<MbtConfig>>(indoc! {"
+            validate: thorough
```
Auspicus, line 188: might want to make consistent with `validation_level` name

Auspicus:
Hey @nyurik I've adjusted this one a bit to be more in line with what you were talking about:
- removed `skip` validation level, always apply at least `fast`
- moved `validation_level` and `on_invalid` to global arguments / config that can apply to any tile source (could be extended to fonts, sprites and styles) not sure if server args are the best place for them but I didn't see any other existing structs that made sense to put it in
- moved the validation check to a higher level so that it can be applied to all tile sources although it is disabled by default for any tile source other than mbtiles
- `validation_level` and `on_invalid` can still be specified at the tile source level and will override global default

I don't think it's 100% over the line yet but before I polish this off I'd like to know if this general direction makes sense to you. Let me know.

Cheers! 🍻

CommanderStorm:
It has been a while and master and this branch has significantly diverged.
I think it is best to close this PR and copy any changes over to a new one rather than rebasing due to the amount of conflicts.

Sorry that this happenend and thanks for looking into this.
I would still love to have this feature (more validation is always good)!

Auspicus:
I agree @CommanderStorm. I'll get around to this again sometime soon.

Issue #1694, opened 2025-02-20
Tiles get lost


panterlo:
The video shows the issue (similar with fill, line etc). Source is Postgres.

https://github.com/user-attachments/assets/b689b4bc-441d-452f-8fb5-af81a86d0bc0

If I export to an MBTiles file with martin-cp same table and repoint MapLibre source to the MBTile served through Martin everything works but for some reason the source Postgres table get's this wierd issue.

CommanderStorm:
Hmmm..
Are you maybe using an old postgres version?

- https://github.com/maplibre/martin/issues/1651

panterlo:
I am using 
```
POSTGIS="3.5.2 dea6d0a" [EXTENSION] PGSQL="170" GEOS="3.12.1-CAPI-1.18.1" PROJ="9.4.0 NETWORK_ENABLED=OFF URL_ENDPOINT=https://cdn.proj.org USER_WRITABLE_DIRECTORY=/tmp/proj DATABASE_PATH=/usr/share/proj/proj.db" (compiled against PROJ 9.4.0) LIBXML="2.9.14" LIBJSON="0.17" LIBPROTOBUF="1.4.1" WAGYU="0.5.0 (Internal)"
```

CommanderStorm:
Hmm.. so likely a differnt issue.

The line seems pretty thin.. could this be a styling/rendering issue?
Unlikely though given that martin-cp does solve your issue.

Is the data you are using maybe publicly avaliable somewhere? (can I debug on the same data?)

CommanderStorm:
And are there any clues in the network console/ other indicators

panterlo:
Nothing in the console, same also applies for fill / polygons.

https://github.com/user-attachments/assets/1e49cc0f-5275-46db-82e7-62c25a2464c6

 Let me publish our server so you can try out...

CommanderStorm:
> Let me publish our server

I trust you that this is happening. Publishing said server is not something I nessesarily need ^^

question is what data you are running (maybe OSM?) and whether I can run a debug build agains said data.

CommanderStorm:
Do you maybe have `max_feature_count` configured?

panterlo:
oh... terrible.... yeah... I just copied the example and it had  max_feature_count: 1000

panterlo:
```
postgres:
  - connection_string: 'postgresql://postgres:XXX@localhost:5432/lm_top10'
    default_srid: 4326
    pool_size: 20
[...]

PR #1695, opened 2025-02-20
Max feature count docs

CommanderStorm:
This PR tries to make the logic behind `max_feature_count` a bit clearer and adds a short example when/why one would activate this

Related #1694

### DIFF
--- a/docs/src/config-file.md
+++ b/docs/src/config-file.md
@@ -70,7 +70,7 @@ postgres:
   # This feature allows to put a maximum latency bound on tiles with extreme amount of detail at the cost of not returning all data.
   # It is sensible to set this limit if you have user generated/untrusted geodata, e.g. a lot of data points at [Null Island](https://en.wikipedia.org/wiki/Null_Island).
   # `null` (the default): Unlimited
-  max_feature_count: null
+  max_feature_count: 100_000
 
   # Control the automatic generation of bounds for spatial tables [default: quick]
   # 'calc' - compute table geometry bounds on startup.
--- a/docs/src/config-file.md
+++ b/docs/src/config-file.md
@@ -64,8 +64,13 @@ postgres:
   # Maximum Postgres connections pool size [default: 20]
   pool_size: 20
 
-  # Limit the number of table geo features included in a tile. Unlimited by default.
-  max_feature_count: 1000
+  # Limit the number of table geo features included in a tile.
+  #
+  # If the number of features is higher than this, they will not be included in the tile and thus look "cut off"/incomplete.
+  # This feature allows to put a maximum latency bound on tiles with extreme amount of detail at the cost of not returning all data.
+  # It is sensible to have such a limit in place if you have user generated/untrusted geodata. (Think: lots of datapoints near null island)
+  # `null` (the default): Unlimited
+  max_feature_count: null
 
   # Control the automatic generation of bounds for spatial tables [default: quick]
   # 'calc' - compute table geometry bounds on startup.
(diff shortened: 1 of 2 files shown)


docs/src/config-file.md
```diff
@@ -64,8 +64,13 @@ postgres:
...
 
-  # Limit the number of table geo features included in a tile. Unlimited by default.
-  max_feature_count: 1000
+  # Limit the number of table geo features included in a tile.
+  #
+  # If the number of features is higher than this, they will not be included in the tile and thus look "cut off"/incomplete.
+  # This feature allows to put a maximum latency bound on tiles with extreme amount of detail at the cost of not returning all data.
+  # It is sensible to have such a limit in place if you have user generated/untrusted geodata. (Think: lots of datapoints near null island)
+  # `null` (the default): Unlimited
+  max_feature_count: null
```
nyurik: Let's keep a real number here - `null` looks really weird tbh.    You can say unlimited if not set or set to `~` (null)?

docs/src/config-file.md
```diff
@@ -64,8 +64,13 @@ postgres:
   # Maximum Postgres connections pool size [default: 20]
   pool_size: 20
 
-  # Limit the number of table geo features included in a tile. Unlimited by default.
-  max_feature_count: 1000
+  # Limit the number of table geo features included in a tile.
```
```suggestion
  # Limit the number of geo features per tile.
```

docs/src/config-file.md
```diff
@@ -64,8 +64,13 @@ postgres:
...
+  #
+  # If the number of features is higher than this, they will not be included in the tile and thus look "cut off"/incomplete.
```
```suggestion
  # If the source table has more features than set here, they will not be included in the tile and the result will look "cut off"/incomplete.
```

docs/src/config-file.md
```diff
@@ -64,8 +64,13 @@ postgres:
...
+  # This feature allows to put a maximum latency bound on tiles with extreme amount of detail at the cost of not returning all data.
+  # It is sensible to have such a limit in place if you have user generated/untrusted geodata. (Think: lots of datapoints near null island)
```
```suggestion
  # It is sensible to set this limit if you have user generated/untrusted geodata, e.g. a lot of data points at  [Null Island](https://en.wikipedia.org/wiki/Null_Island).
```

nyurik (approved):
i did a bit of a brushing up of the text.  Please review and merge when you are happy with it.

docs/src/config-file.md
CommanderStorm: lets not have double spaces, otherwise quite reasonable

```suggestion
  # It is sensible to set this limit if you have user generated/untrusted geodata, e.g. a lot of data points at [Null Island](https://en.wikipedia.org/wiki/Null_Island).
```

docs/src/config-file.md
CommanderStorm: Well.. `~` in yaml is the same as `null`, and likely the part of the spec which users are not quite know.

What about setting this to a really high value instead?

```suggestion
  max_feature_count: 100_000
```

alternatively, we could change to a "float" (and truncate to an integer) and allow the following:

```suggestion
  max_feature_count: inf
```

nyurik: nah, lets not introduce float, as that would be REALLY confusing.  A large number may work.

CommanderStorm:
Btw:
Should we remove these lines? (currently @nyurik will need to press the `Merge without waiting for requirements to be met (bypass rules)` button...)

https://github.com/maplibre/martin/blob/895d825ce1f36ed739ac0e10043318e2cd75406c/.github/workflows/ci.yml#L8-L11

nyurik:
we might have to come up with some "success rules" that would be able to bypass unneeded files if they haven't changed... not sure if there is any good way to do that, may need to ask AI :)

nyurik (approved):

PR #1668, opened 2025-01-27
Fix martin-ui crate packaging

nyurik:
STATUS:  this PR can use some help from other folks - it deals with NPM integration, which turns out is not working as intended... :(

The martin-ui dir must be part of the martin package, otherwise it does not get included in the `cargo publish -p martin`.  We could also do it with a symlink from inside the `/martin` dir, but that gets a bit confusing - easier to just keep ui inside martin itself, as it is not used anywhere outside of that crate.

In order to prevent these bugs in the future, we might want to run `cargo +nightly publish --workspace -Z package-workspace --dry-run` as part of some CI -- this will take into account that some crates depend on the other, and that some/all of them have not been published yet

### TODO
* [ ] modify `martin/build.rs` to install node files inside the standard build dir.  Current dry run shows this error:

```
error: failed to verify package tarball

Caused by:
  Source directory was modified by build.rs during cargo publish. Build scripts should not modify anything outside of OUT_DIR.
  Added: <...>/martin/target/package/martin-0.15.0/martin-ui/dist
  	<...>/martin/target/package/martin-0.15.0/martin-ui/dist/_
  	<...>/martin/target/package/martin-0.15.0/martin-ui/dist/_/assets
  	<...>/martin/target/package/martin-0.15.0/martin-ui/dist/_/assets/favicon.ico
  	<...>/martin/target/package/martin-0.15.0/martin-ui/dist/_/assets/index-Cu-91x5d.js
  	<...>/martin/target/package/martin-0.15.0/martin-ui/dist/_/assets/index-DZmjP79K.css
  	<...>/martin/target/package/martin-0.15.0/martin-ui/dist/_/assets/logo-BlOKpv3a.png
  	<...>/martin/target/package/martin-0.15.0/martin-ui/dist/index.html
  	<...>/martin/target/package/martin-0.15.0/martin-ui/node_modules
  	           (and many files in this dir)

  To proceed despite this, pass the `--no-verify` flag.
```

I am not sure we should use `--no-verify`

fixes #1667

PR #1699, opened 2025-02-23
Martin-ui packaging issue

### LINKED ISSUE #1667: `cargo publish -p martin` does not work due to Node.js package issue
Running `cargo publish -p martin --dry-run` generates this error, which means I cannot publish v0.15 martin. The other two packages have been published ok.

```
led to run custom build command for `martin v0.15.0 (/home/nyurik/dev/rust/martin/target/package/martin-0.15.0)`

Caused by:
  process didn't exit successfully: `.../martin/target/package/martin-0.15.0/target/debug/build/martin-c5f8c0cc6225768e/build-script-build` (exit status: 1)
  --- stderr
  Cannot execute npm install: Os { code: 2, kind: NotFound, message: "No such file or directory" }
  Error: Os { code: 2, kind: NotFound, message: "No such file or directory" }
warning: build failed, waiting for other jobs to finish...
error: failed to verify package tarball
```

cc: @CommanderStorm @sharkAndshark

nyurik:
Seems we still have a packaging issue - npm saves files outside the build dir.

```
error: failed to verify package tarball

Caused by:
  Source directory was modified by build.rs during cargo publish. Build scripts should not modify anything outside of OUT_DIR.
  Added: .../martin/target/package/martin-0.15.0/martin-ui/dist
  	.../martin/target/package/martin-0.15.0/martin-ui/dist/_
  	.../martin/target/package/martin-0.15.0/martin-ui/dist/_/assets
        ...
```

Fixes #1667
See prior partial fix in #1668


CommanderStorm:
Let me prephase this: What I have just pushed is a hooorible hack.

What I just pushed works locally (both publish and running).. wish me luck

image

CommanderStorm:
I will not continue working on this today, seems like CI is really against this change..
I need fresh eyes for this => will continue sometime in the week/on the weekend.

If someone wants to look into this or has better oppinions, I would love feedback.

Somewhere I apparently made a change in the code I pushed during final cleanup (=before my first commit here) 😞

martin/build.rs
```diff
@@ -1,13 +1,87 @@
...
+        }
+    }
+    Ok(())
+}
+
 fn main() -> std::io::Result<()> {
     #[cfg(feature = "webui")]
     {
-        static_files::NpmBuild::new("martin-ui")
+        // rust requires that all changes are done in OUT_DIR.
```
nyurik: I wonder if we should simply create a symlink in the outdir - no need to copy anything?

CommanderStorm: Sounds good, but as far as my logic goes, that would require basically the same code..

- if you mean placing a symlink in `OUT_DIR`:
  The directory is dynamic based on compilation target (`package`,`debug`,..) => I don't think hardcoding them is a good way to go.
  These directories are also not documented => we would be setting ourselves up for breaking changes..
- if you mean generating a symlink in the `build.rs` file:
  `dist` and `node_modules` may not live outside of the `OUT_DIR`, if compiled by the `build.rs` file as far as I understand. => code would basically be identical. Maybe 1-2 lines simpler, but not much..

Or do you mean something different?

What might work (thinking about it: I have not tried this part) is deleting `dist` and `node_modules` after the `build.rs` file..
This might fuck devs over in development though..
(think: start `martin-ui`, then server. `node_modules` is now gone)

CommanderStorm: Thinking about it:
Generating symlinks is a bit simpler code.
No recursion or dir-deletion required.. => more like 4-5 lines

Will migrate to that once I have the time.

nyurik: I looked at all of these usages at github search, but the [results](https://github.com/search?q=%22NpmBuild%3A%3Anew%22+out_dir+path%3A**%2Fbuild.rs&type=code) are not too many.  Mostly they modify `.target(...)` to use the `OUT_DIR`, but that would probably only affect the `dist` dir, not the location of `node_modules`.  I posted an [issue](https://github.com/static-files-rs/static-files/issues/11), and also asked on stackoveflow... will see

CommanderStorm: > Generating symlinks is a bit simpler code

Turns out it is not.
A few parts in the js building part don't support symlinks and die pretty quickly...
image
image

Since the logic would need to account for this anyhow, I think this is not simpler anymore.
The risk of not insiginificant debugging without actual benefit is quite high..

CommanderStorm (approved):
Okay.. this works now.
It is still a hoooorible hack, but at least it woks

nyurik:
thx, i refactored the build script a bit further, trying to simplify it (not much luck), plus made sure not to use results - with build, its ok to crash asap as it gives us good debugging context.  When we build with the npm packaging command, where does the generated file end up in?  I only see `dist/index.html`.

If everything is ok, lets merge

CommanderStorm:
> When we build with the npm packaging command, where does the generated file end up in? I only see `dist/index.html`.

what do you mean? Not quite sure what you are asking
It is in the `OUT_DIR`, more precicely most of from the node build generated files will end up in the `_` folder.
image

martin/build.rs
```diff
@@ -1,13 +1,115 @@
...
+    });
+    let excludes = exclude_dirs.iter().map(|v| src.join(v)).collect::<Vec<_>>();
+
+    let mut it = walkdir::WalkDir::new(src).into_iter();
+    loop {
+        let entry = match it.next() {
+            None => break,
+            Some(Err(err)) => panic!("failed to read directory entry: {err}"),
+            Some(Ok(entry)) => entry,
+        };
```
CommanderStorm: more of a nit: I would find while-let to be more readable than loop-break

```suggestion
    while let Some(entry) = it.next() {
        let entry = match entry {
            Ok(entry) => entry,
            Err(err) => panic!("failed to read directory entry: {err}"),
        };
```

martin/build.rs
```diff
@@ -1,13 +1,115 @@
...
+            entry
+                .path()
+                .strip_prefix(src)
+                .expect("path is not a prefix of the source directory"),
+        );
+
+        // Tell Cargo to monitor all root level entries for changes
+        if dst_path.components().count() == 1 {
+            println!("cargo:rerun-if-changed={}", entry.path().display());
+        }
```
CommanderStorm: Is this not already covered by `static_files::NpmBuild::change_detection` below, or am I missing something?

As far as I read this, we can keep either, but lets not keep duplicate code..

martin/build.rs
```diff
@@ -1,13 +1,115 @@
...
+    let _ = fs::remove_dir_all(dst); // ignore if dir does not exist
+    fs::create_dir_all(dst).unwrap_or_else(|e| {
+        panic!(
+            "failed to create destination directory {}: {e}",
+            dst.display()
+        )
+    });
+    let excludes = exclude_dirs.iter().map(|v| src.join(v)).collect::<Vec<_>>();
+
+    let mut it = walkdir::WalkDir::new(src).into_iter();
```
CommanderStorm: > Crate walkdir provides an efficient and cross platform implementation of recursive directory traversal. ***Several options are exposed to control iteration, such as whether to follow symbolic links (default off)***, limit the maximum number of simultaneous open file descriptors and the ability to efficiently skip descending into directories

Something to be aware of => no directories which are symlinks

We could disable this footgun.. What do you think?

```suggestion
    let mut it = walkdir::WalkDir::new(src).follow_links(true).into_iter();
```

CommanderStorm (commented):
Looks good to me
Could be merged as is, asssuming `rerun-if-changed` is correct ^^

martin/build.rs
nyurik: good catch! I forgot about `while let` (copied `walkdir` example for skipping dirs). Going further, I think we should use `if let Err(...) = .. {panic}` too

martin/build.rs
nyurik: good catch!

martin/build.rs
nyurik: you mean just in case we start using symlinks?  Sure

nyurik:
thanks!  Turns out I shouldn't code at night - didn't notice the `_` dir. Thank you for good suggestions, will merge shortly.

martin/build.rs
```diff
@@ -1,13 +1,107 @@
...
+        .build()
+        .expect("failed to build webui npm dir");
+
+    let target_to_keep = martin_ui_dir.join("dist");
+    assert!(
+        !target_to_keep.exists() || target_to_keep.is_dir(),
+        "the martin-ui/dist must either not exist or have been produced by previous builds"
+    );
+
+    // TODO: we may need to move index.html one level down per change_detection() docs
```
CommanderStorm, line 98: Could you clairify what you mean here?
When I rebuild, this should be reproducible enough to depend on the input files and not on the output file..

As far as I see it, this is a non-issue..

```suggestion
```

CommanderStorm (approved):
LGTM
I think we can remove the TODO though

nyurik:
Thanks @CommanderStorm for tackling this one!  Messy, but it seems like it works now :)

This is why i put that todo (which we might want to remove later):

https://github.com/static-files-rs/static-files/blob/bac6352faddffdc36d96cac04b4115dfd6f2f35e/src/mods/npm_build.rs#L81-L93

PR #1700, opened 2025-02-23
Minor cleanup, dep bump, rm pre-push hook

nyurik:
* get rid of the husky pre-commit hook - it became much more of a nuisance than help.
* Add pre-commit CI to automatically do code formatting, just like we do in maplibre-native and some other projects
* remove unused dependencies
* bump some dependencies


Cargo.toml
```diff
@@ -89,7 +87,7 @@ subst = { version = "0.3", features = ["yaml"] }
 testcontainers-modules = { version = "0.11.6", features = ["postgres"] }
 thiserror = "2"
 tiff = "0.9.1"
-tile-grid = "0.6"
+# tile-grid = "0.6"
```
CommanderStorm: lets remove entirely instead of commenting out.

If intentional, a comment for future devs would be helpful

```suggestion
```

CommanderStorm (approved):
LGTM

Cargo.toml
nyurik: oops, mised that, thx

PR #1703, opened 2025-02-24
Minor pre-commit-config change

nyurik:


.pre-commit-config.yaml
```diff
@@ -11,14 +11,14 @@ repos:
...
-      - id: trailing-whitespace
-      - id: end-of-file-fixer
-      - id: mixed-line-ending
-        args:
-          - --fix=lf
+      - id: check-added-large-files
       - id: check-toml
       - id: check-yaml
-      - id: check-added-large-files
+        args: [ --allow-multiple-documents, --unsafe ]
```
CommanderStorm: Why this change?

> `--unsafe`
> Instead of loading the files, simply parse them for syntax. A syntax-only check enables extensions and unsafe constructs which would otherwise be forbidden. Using this option removes all guarantees of portability to other yaml implementations. Implies `--allow-multiple-documents`

.pre-commit-config.yaml
```diff
@@ -11,14 +11,14 @@ repos:
...
     rev: v5.0.0
     hooks:
-      - id: trailing-whitespace
-      - id: end-of-file-fixer
-      - id: mixed-line-ending
-        args:
-          - --fix=lf
+      - id: check-added-large-files
       - id: check-toml
       - id: check-yaml
```
CommanderStorm, line 18: lets also check json files while we are at it

```suggestion
      - id: check-yaml
      - id: check-json
        exclude: '.+/tsconfig.json'
```

.pre-commit-config.yaml
```diff
@@ -11,14 +11,14 @@ repos:
...
-      - id: check-added-large-files
+        args: [ --allow-multiple-documents, --unsafe ]
+      - id: end-of-file-fixer
+      - id: mixed-line-ending
+        args: [ --fix=lf ]
+      - id: trailing-whitespace
```
CommanderStorm, line 23: There are a few more that might be usefull imo.
What do you think?

```suggestion
      - id: trailing-whitespace
      - id: check-executables-have-shebangs
      - id: check-shebang-scripts-are-executable
        exclude: '.+\.rs' # would be triggered by #![doc = include_str!("../README.md")]
      - id: check-symlinks
      - id: destroyed-symlinks
```

CommanderStorm (approved):

.pre-commit-config.yaml
nyurik: copy paste, i guess we don't need it, thx

PR #1706, opened 2025-02-25
docs: change default `max_feature_count` documentation to null

CommanderStorm:
I think `null` is less confusing for users than `~`..
Feel free to overrule ^^

Maybe adding an enum and having `unlimited|null|usize` be the options would be less confusing?

If we were to accept breakage (I don't think we should), `unlimited|NonZeroUsize` would likely be a better choice.

### DIFF
--- a/docs/src/config-file.md
+++ b/docs/src/config-file.md
@@ -69,8 +69,7 @@ postgres:
   # If the source table has more features than set here, they will not be included in the tile and the result will look "cut off"/incomplete.
   # This feature allows to put a maximum latency bound on tiles with extreme amount of detail at the cost of not returning all data.
   # It is sensible to set this limit if you have user generated/untrusted geodata, e.g. a lot of data points at [Null Island](https://en.wikipedia.org/wiki/Null_Island).
-  # `null` (the default): Unlimited
-  max_feature_count: 100_000
+  max_feature_count: max_feature_count: null # null=unlimited (default), but can also be a positive integer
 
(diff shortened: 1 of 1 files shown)


docs/src/config-file.md
```diff
@@ -69,8 +69,7 @@ postgres:
   # If the source table has more features than set here, they will not be included in the tile and the result will look "cut off"/incomplete.
   # This feature allows to put a maximum latency bound on tiles with extreme amount of detail at the cost of not returning all data.
   # It is sensible to set this limit if you have user generated/untrusted geodata, e.g. a lot of data points at [Null Island](https://en.wikipedia.org/wiki/Null_Island).
-  # `null` (the default): Unlimited
-  max_feature_count: 100_000
+  max_feature_count: max_feature_count: null # null=unlimited (default), but can also be a positive integer
```
```suggestion
  max_feature_count: null   # either a positive integer, or null=unlimited (default)
```

nyurik (approved):
sounds good, just fix the dup

docs/src/config-file.md
CommanderStorm: Shure, but lets not have so many spaces ^^

```suggestion
  max_feature_count: null # either a positive integer, or null=unlimited (default)
```

CommanderStorm:
(and needs CI-overruling again ^^)

nyurik (approved):

nyurik:
thanks!

Issue #578, opened 2023-02-17
Improve packaging for various platforms


nyurik:
Our CI currently only publishes Docker images and adds binary artifacts to the release, e.g. [here](https://github.com/maplibre/martin/releases/tag/v0.7.0) (at the bottom). We need to greatly simplify re-use of Martin:

- [x] debian packages
- [x] homebrew (used to work before?)
- [x] `cargo install` from crates.io - Martin at the moment uses internal non-published mbtiles crate - we need to either publish them if reasonable, or move the code into a sub-module
- [ ] windows install (is this still a thing?)
- [ ] packaging for other platforms, and publish it to relevant package repositories

Useful links:
* [tutorial](https://rust-cli.github.io/book/tutorial/packaging.html#getting-your-app-into-package-repositories)
* (?) https://github.com/axodotdev/cargo-dist/
* https://github.com/sharkdp/bat/blob/master/.github/workflows/CICD.yml -- this is how `bat` creates debian packages in CI
* use https://repology.org/ to track packaging on different platforms -- currently only found [this one](https://repology.org/project/martin/versions)
  * https://github.com/NixOS/nixpkgs/blob/release-21.11/pkgs/servers/martin/default.nix - someone's maintaining (outdated) Urbica version here

ankane:
Just fyi, Martin is now available in homebrew-core: https://github.com/Homebrew/homebrew-core/pull/206380

nyurik:
nice, thanks @ankane !

mr-sven:
Hi, the build matrix should be modified. The Debian package could not be installed on current Debian Stable 12.9 because of wrong libc dependency:
```
dpkg: dependency problems prevent configuration of martin:
 martin depends on libc6 (>= 2.38); however:
  Version of libc6:amd64 on system is 2.36-9+deb12u9.
```

mr-sven:
Found a solution in another repo, you have to hardwire the Ubuntu Image to `ubuntu-22.04`:
https://github.com/maplibre/martin/blob/d55d7bc96f4c76513059760b711ce3013b1a6175/.github/workflows/ci.yml#L264
```diff
--             os: ubuntu-latest
++             os: ubuntu-22.04
```
This will change to glibc 2.35, so debian 12 glibc 2.36 will work.

nyurik:
Thx @mr-sven , want to submit a pr?

mr-sven:
Tested the build artifact, it could now be installed and run on Debian 12 :thumbsup:

[...]

PR #1710, opened 2025-03-01
fix: debian 12 libc compatibility

mr-sven:
Should fix debian libc problem mentioned in #578

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -275,6 +275,8 @@ jobs:
       - name: Rust Versions
         run: rustc --version && cargo --version
       - uses: Swatinem/rust-cache@v2
+        with:
+          key: ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


mr-sven:
Looks like the rust cache uses wrong source.

.github/workflows/ci.yml
```diff
@@ -261,7 +261,7 @@ jobs:
           - target: aarch64-apple-darwin
             os: macos-latest  # M-series CPU
           - target: debian-x86_64
-            os: ubuntu-latest
+            os: ubuntu-22.04
```
```suggestion
            os: ubuntu-22.04 # downgraded to have lower libc support. Debian-12 (stable at the time of writing) does not support the same version ubuntu does.
            # TODO: update to newer ubuntu version again once debian 13 launches in fall of 2025
```

CommanderStorm (approved):

PR #1711, opened 2025-03-02
feat: style serving

### LINKED ISSUE #1607: Serve styles
Is it possible to configure support for serving static files or Mapbox-style JSON files using a configuration file?

```yaml
static:
 - /path/to/static_dir
```
or

```yaml
styles:
 - /path/to/styles_dir
```

Additionally, when serving Mapbox-style JSON files specifically, could the system automatically add related fonts, sprites, and sources to the file?

CommanderStorm:
This PR adds style serving to the features we support.
Resolves https://github.com/maplibre/martin/issues/1607

I am not quite certain if `sources` is something that works for styles.. The key to the dict not playing any role is somewhat counterintuitive.


docs/src/config-file.md
```diff
@@ -230,4 +230,18 @@ fonts:
   # A list of *.otf, *.ttf, and *.ttc font files and dirs to search recursively.
   - /path/to/font/file.ttf
   - /path/to/font_dir
+
+# We support serving styles
+# In the future, we plan on adding support for rendering too
+styles:
+   paths:
+     # all files in this dir will be published as a "styles_dir_FILENAME" style source
```
nyurik: this is not very clear, perhaps something like this?  Note that I don't think we want to keep dir name in the style... or do we?   We may even introduce some setting like `id_format` (similar to `source_id_format` for postgres), where we could have `"{parent2}_{parent1}_{filename}"` e.g. to preserve two levels of the dir as a possible configuration for this... But this might be an overkill.

As a separate thought - a style usually has fonts+sprites.  Currently all of them are independent global assets. In theory, it might make sense for a style designer to somehow keep one style dir with all the needed stuff in it... Or am I overthinking it, and it is better to have independent assets and styles?

```suggestion
     # each json file in this dir will publish as style
     # The name of the file will be used as the style name
```

martin/src/args/root.rs
```diff
@@ -303,6 +308,7 @@ mod tests {
           - "../tests/fixtures/cog"
           - "../tests/fixtures/cog/rgba_u8_nodata.tiff"
           - "../tests/fixtures/cog/rgba_u8.tif"
+        styles: "../tests/fixtures/cog"
```
nyurik: cog?

martin/src/config.rs
```diff
@@ -17,8 +17,10 @@ use crate::file_config::FileConfigEnum;
 use crate::fonts::FontSources;
 use crate::source::{TileInfoSources, TileSources};
 #[cfg(feature = "sprites")]
-use crate::sprites::{SpriteConfig, SpriteSources};
+use crate::sprites::SpriteSources;
```
nyurik: why?

docs/src/config-file.md
CommanderStorm: > it might make sense for a style designer to somehow keep one style dir with all the needed stuff in it... Or am I overthinking it, and it is better to have independent assets and styles?

I toyed around with parsing the files for "is this actually json", but wanted to get an "intial draft" out the door that is not too weird compared to the other things..
If I just parse if it is json, this would behave more like the sprite part.

Sprites currently look at all the files which are not hidden and have the `.svg` ending..

Will look into adding further validation on which to skip importing if that is fine with you ^^

nyurik: yes, that's actually another consideration i haven't mentioned - that a style file should really be "style" rather than some random file that just happened to have `.json` extension, but I agree - lets do validation later on.  I was more thinking in terms of if we want to have any support for "entire style in a dir" model, that includes stylesheet, fonts, and sprites all together.  I do suspect that it won't be very valuable because many things like highway shields are shared across multiple styles

martin/src/args/root.rs
CommanderStorm: Yes, a folder..
See above: ´pmtiles´/`mbtiles` are configured the same way..

maybe targeting the folder one further up is a better idea

Also: should `styles` use the same mechanism as the tile-formats? I am a bit torn

=> the relevant logic is `config.styles = parse_file_args(&mut cli_strings, &["json"], false);`

martin/src/config.rs
CommanderStorm: purely aestetic..
all the other config imports are done this way:

image

nyurik: ah, ok.  When evaluating complex PRs, these things sometime get in a way, thus confusion

martin/src/args/root.rs
nyurik: right, but what does `cog` have to do with the new styling feature you are adding?

CommanderStorm: Nothing, that is just the name of the folder the testcase uses.
I have split the testcase in two to prevent this confusion in the future.

I have removed the "big folder" support => being able to run `martin ./big_folder_with_styles_and_stuff` from this PR.
This needs better testing anyway and is a good fire-break.

nyurik:
@CommanderStorm one more thing - please use your fork for all contributions - https://gist.github.com/nyurik/4e299ad832fd2dd43d2b27191ed3ec30

CommanderStorm:
Oops sorry. Had to setup my environment on a new PC and forgot about that automation.
Could you add a ruleset to prevent this?

image

nyurik:
I'm not sure it will be possible because we need dependabot and precommit ci to create branches

CommanderStorm:
good point, but those can just be added to the bypass list 
image

nyurik:
ah, thx, i wasn't aware of that feature. Created.

CommanderStorm:
To be fair:
Rulesets replacing branch protection rules is pretty new.. https://github.blog/news-insights/product-news/github-repository-rules-are-now-generally-available/

docs/src/config-file.md
```diff
@@ -230,4 +230,19 @@ fonts:
   # A list of *.otf, *.ttf, and *.ttc font files and dirs to search recursively.
   - /path/to/font/file.ttf
   - /path/to/font_dir
+
+# We support serving styles
+# In the future, we plan on adding support for rendering too
+styles:
```
CommanderStorm, line 236: TODO:
Discuss how we want to add rendering-related config options here, to make sure that this does not feel awkward/..

docs/src/using.md
```diff
@@ -69,7 +70,12 @@ curl localhost:3000/catalog | jq
       "end": 65533
     },
     ...
-  }
+  },
+  "styles": {
+    "americana": {
```
CommanderStorm: TODO(decide) usually paths on the local file system do contain `.json` => possibly remove just this ending or all file suffixes /..

docs/src/using.md
```diff
@@ -69,7 +70,12 @@ curl localhost:3000/catalog | jq
...
+      "path": "path/to/americana.json",
```
CommanderStorm: Note: schema has to be able to fit rendering.
I think current schema can, but please double check my reasoning here.
Depending on how we want to configure this, different catalog design might be nessesary.

I am thinking about parameters like:
- allow rendering file format xyz
- allow rendering

=> things which likely will have to be configured at the styles level.

Unsure if more options are nessesary to show..

martin/src/srv/styles.rs
```diff
@@ -0,0 +1,39 @@
...
+    style_id: String,
+}
+
+#[route(
+    "/style/{style_id}",
+    method = "GET",
+    wrap = "middleware::Compress::default()"
+)]
+#[allow(clippy::unused_async)]
+async fn get_style_json(
```
CommanderStorm: currently, this does not ensure that the style is valid json or compress the whitespace (other than general compression).

Unsure if adding is better to streaming the file out via `actix_files`

martin/src/config.rs
```diff
@@ -67,7 +71,11 @@ pub struct Config {
 
     #[cfg(feature = "sprites")]
     #[serde(default, skip_serializing_if = "FileConfigEnum::is_none")]
-    pub sprites: FileConfigEnum<SpriteConfig>,
+    pub sprites: FileConfigEnum<crate::sprites::SpriteConfig>,
+
+    #[cfg(feature = "styles")]
+    #[serde(default, skip_serializing_if = "FileConfigEnum::is_none")]
+    pub styles: FileConfigEnum<crate::styles::StyleConfig>,
```
nyurik, line 71: perhaps move the `StyleConfig` into the `use` statement at the top since it is also feature-gated?  Same for `SpriteConfig`?

nyurik (commented):
thanks, looks good! Minor nit in the code. Also, I don't think we want such huge style files as test fixtures. Perhaps delete most of the rules from them, only leaving the relevant parts that we may want to parse?  Less is usually more when doing tests, as it helps clarity (i'm guilty of doing it wrong myself, but i try)

martin/src/srv/styles.rs
nyurik: Compacting json on the fyl sounds like a good idea, especially if we enable caching (separate PR please, caching is messy). Streaming might be a good option - I wonder if we can just hand a stream to actix?

docs/src/using.md
nyurik: removing `.json` is probably enough... what other extensions could it have?

docs/src/config-file.md
nyurik: I think this should work (per style options are probably overkill, so likely will just keep the top level)

```yaml
styles:
  render: true
  render_formats: [ 'png', 'jpg' ]
  render_dpi: [ 1, 2 ]
  paths:
    # each json file in this dir will publish as style
    # The name of the file will be used as the style name
    - /path/to/styles_dir
    # will be published as maplibre_style
    - /path/to/maplibre_style.json
    - path: /path/to/style.json
      render: true
      render_formats: [ 'png', 'jpg' ]
      render_dpi: [ 1, 2 ]
  sources:
    # will be published as style1_FILENAME
    style1: /path/to/styles_dir2
    # will be published as style2
    style2: /path/to/americana.json
    # will be published as style3
    style3:
      path: /path/to/americana.json
      render: true
      render_formats: [ 'png', 'jpg' ]
      render_dpi: [ 1, 2 ]

```

martin/src/config.rs
CommanderStorm: but that would make it inconsistent with `postgres`, `pmtiles`, `mbtiles` and `cog`..
Feel free to overrule ^^

martin/src/srv/styles.rs
CommanderStorm: Not sure if caching is something that we need..
local file reads are cached by the os anyway

=> I would not expect much more performance

A bit, but not much..

docs/src/using.md
CommanderStorm: no idea what other exstenstions, have removed just `.json` (and spaces)

martin/src/srv/styles.rs
CommanderStorm: Caching for this seems a bit overengineered..
Lets remember that if we access the disk often, the file system caches this too.
=> I would not expect much more performance. A bit, but not much..

Caching the compressed output would be fair gain though..

Streaming was a nice idea, but does not work given that `middleware::Compress::default` does need to buffer.
I have pushed this and etags or last_modified header handling to a future PR.

docs/src/config-file.md
```diff
@@ -230,4 +230,24 @@ fonts:
...
+   sources:
+     # each json file in this directory will publish as style
+     # The filename will be used as the styles' id
+     #
+     # source_name1 will be logged and can be useful for debugging
+     source_name1: /path/to/styles_dir1
+     # will be published as style2
+     #
+     # source_name2 will be logged and can be useful for debugging
+     source_name2: /path/to/style2.json
```
CommanderStorm: This part feels really akward to configure. Unsure if we should remove it for styles

nyurik:
@CommanderStorm do note that in the case of a failure, all generated test files are saved as a build artifact in the build pipeline. You can easily download and compare them locally

docs/src/config-file.md
nyurik: perhaps.  It all depends on the "typical" use case. Would it be enough to point to a file? Point to a directory with multiple styles? Do multiple styles thing actually exists, or is it always a style by itself in a directory, together with additional assets like fonts/sprites?

CommanderStorm:
Where? I don't see a button for this..

(For this one I could also have looked at the file itsself ^^)
image

CommanderStorm:
Ah sorry, nevermind.
Found it

image

nyurik:
It might make sense to add a note to our dev guide that in case of a failure, all logs are saved

nyurik:
@CommanderStorm hi, what's the status of this?

CommanderStorm:
@nyurik I think everything is tested, documented & stuff.
-> Ready to review/merge

docs/src/config-file.md
```diff
@@ -230,4 +230,24 @@ fonts:
...
```
nyurik: @CommanderStorm I would propose to change the behavior of the parser to this logic.  This will allow us to configure directories that also contain additional assets (icons/images/fonts).  I don't think `sources` section should ignore the IDs as it does now.

```suggestion
   paths:
     # publish all *.json files in this directory
     # The name of the file will be used as the style name
     - /path/to/styles_dir
     # publish a single file - here `maplibre_style` will be the style name 
     - /path/to/maplibre_style.json
   sources:
     # publish first JSON file found in this directory as `style_name1`
     # If more than one JSON file is found, Martin will print a warning
     style_name1: /path/to/styles_dir1
     #  Publish specific file as `style_name2` style source
     style_name2: /path/to/style2.json
```

martin/src/styles/mod.rs
```diff
@@ -0,0 +1,319 @@
...
+        entries
+    }
+
+    /// Adds all the contained files in the given file/directory as style sources.
+    fn add_sources(&mut self, id: &str, base_path: &PathBuf) {
+        match list_contained_files(base_path) {
+            Ok(contained_paths) => {
+                for child_path in contained_paths {
+                    let Some(name) = child_path.file_name() else {
+                        // FIXME: we should never panic in production code
```
nyurik: we should never crash on loading - asserts are evil

martin/src/styles/mod.rs
```diff
@@ -106,6 +106,7 @@ impl StyleSources {
             Ok(contained_paths) => {
                 for child_path in contained_paths {
                     let Some(name) = child_path.file_name() else {
+                        // FIXME: we should never panic in production code
```
CommanderStorm, line 109: this assertion is here, as it logically cannot happen.

The base path is checked if it is a valid file name one level up (where the id is extracted).

docs/src/config-file.md
CommanderStorm: lets remove the `source_name1` vs `source_name2` confusion factor. Looks too much like `source.1` for tiles.

```suggestion
   paths:
     # publish all *.json files in this directory
     # The name of the file will be used as the style name
     - /path/to/styles_dir
     # publish a single file - here `maplibre_style` will be the style name 
     - /path/to/maplibre_style.json
   sources:
     # publish first JSON file found in this directory as `some_style_name`
     # If more than one JSON file is found, Martin will print a warning
     some_style_name: /path/to/styles_dir
     #  Publish specific file as `style_name2` style source
     other_style_name: /path/to/other_style.json
```

CommanderStorm: @nyurik
Am fine with said change.

Though, what will happen if said warning happens => `some_style_name.1`, like for tiles?

Also, likey with an sorting. => alphanumeric or [natural](https://en.wikipedia.org/wiki/Natural_sort_order)

> If more than one JSON file is found, Martin will print a warning

martin/src/styles/mod.rs
CommanderStorm: this assertion is here, as it logically cannot happen.

The base path is checked if it is a valid file name one level up (where the id is extracted).

docs/src/config-file.md
nyurik: I guess we can just skip dirs with multiple json files for the sources section? With a warning. Btw, we really need treat warnings as errors option, but that's separate feature

martin/src/styles/mod.rs
CommanderStorm: wait no: actually you are entirely right:
if I `foo: bar.json` in the `sources`-part, the ID is not drawn from the parents filename

martin/src/styles/mod.rs
```diff
@@ -0,0 +1,319 @@
...
+    fn add_sources(&mut self, id: &str, base_path: &PathBuf) {
+        match list_contained_files(base_path) {
+            Ok(contained_paths) => {
+                for child_path in contained_paths {
+                    let Some(name) = child_path.file_name() else {
+                        // FIXME: we should never panic in production code
+                        assert!(
+                            !base_path.is_file(),
+                            "base_path cannot be a file without name as otherwise the id would not exist"
+                        );
```
CommanderStorm:
```suggestion
                        // child_path is a file and may be equal to base_path
```

docs/src/config-file.md
CommanderStorm: So the following?
=> drawing a bit of attention to the different "ability" of `paths` vs `sources`

```suggestion
   paths:
     # publish all *.json files in this directory
     # The name of the file will be used as the style name
     - /path/to/styles_dir
     # publish a single file - here `maplibre_style` will be the style name 
     - /path/to/maplibre_style.json
   sources:
     # publish the JSON file found in this directory as `some_style_name`
     # If more than one JSON file is found, Martin will print a warning and ignore the folder
     # If you require serving multiple styles in a folder, we support either naming them explicitliy or publishing the path
     some_style_name: /path/to/styles_dir
     #  Publish a specific style as `other_style_name` style source
     other_style_name: /path/to/other_style.json
```

docs/src/sources-styles.md
```diff
@@ -0,0 +1,25 @@
...
+Use `/catalog` API to see all the `<style_id>`s.
+
+### Map Style
+
+Use the `/style/<style_id>` API to get `<style_id>` JSON content.
+
+Changes or removals of styles are reflected immediately, but additions are not.
+A restart of Martin is required to see new styles.
+
+### Add server-side raster tile rendering
```
CommanderStorm: Noticed that my gramar is off here

```suggestion
Use the `/catalog` API to see all the `<style_id>`s.

### Map Style

Use the `/style/<style_id>` API to get a `<style_id>`'s JSON content.

Changes or removals of styles are reflected immediately, but additions are not.
A restart of Martin is required to see new styles.

### Server-side raster tile rendering
```

martin/src/styles/mod.rs
nyurik: @CommanderStorm I don't think this is the case - `id` is being ignored in the `fn add_sources` except to report warnings. Instead, it should be used when appropriate.

nyurik:
@CommanderStorm so what's the next steps here?  My thinking:
* fix it so that `sources: { src: path, src2: path }` would use the `src` and `src2` keys as the ids
* modify so that if the path is a directory, and has more than one json file, the whole directory is ignored (with a warning). Only applies to `sources`, not to `paths` section.
* decide how we want to deal with the other style assets like fonts and sprites. Both are already supported, but we need to understand how we want to tie them all together. This item is not a requirement for merging, but it would help us understand if we are on the right path.

CommanderStorm:
Yes, I will do the style-change that on the weekend or possibly next week today was quite buisy with other work.

About the fonts/sprites aspect: idk, we serve all 3. Unsure how you would like to tie them togehter. Likely something like #1757 ?
Think this is farther-future => would not worry about this regarding styles directly.

nyurik:
I think we are almost there... one thought that kept bugging me is the whole "support directories in sources" aspect... I think you were right that we should not even allow dirs there, e.g. `sources: { foo: some_file.yml }` -- should be the only one allowed.  No directories there.  We would still allow directories in the `paths` section.

nyurik (approved):
Awesome!!!

PR #1713, opened 2025-03-02
chore: add missing cfg feature gates

CommanderStorm:
These feature gates were accidentally forgotten.
I noticed them during working on #1711

Unsure if there is a better way to find them rather than just noticing them


nyurik (approved):
thx!

PR #1717, opened 2025-03-03
Fix tilejson scheme

### LINKED ISSUE #1716: MapLibre-GL-JS problem with MBTile source
At the moment I face a problem using MapLibre GL JS and an MBTile source.
The tilejson responds with `scheme: tms` what is correct I've read because mbtiles are always `tms`?!?!
It looks like MapLibre GL JS is taking care of that and requests a inverted y axis, but in the mbtiles mod the y axis is inverted again: 
https://github.com/maplibre/martin/blob/00757ee708e18a654e03e78ba843d106c2593010/mbtiles/src/mbtiles.rs#L144 
So I'm getting no tiles for the specified region. If I point maplibe to a modified tilejson with `scheme: xyz` it will work.

mr-sven:
This fix #1716

### DIFF
--- a/mbtiles/src/metadata.rs
+++ b/mbtiles/src/metadata.rs
@@ -138,7 +138,8 @@ impl Mbtiles {
                     "agg_tiles_hash" => agg_tiles_hash = Some(value),
                     "scheme" => {
                         if value != "tms" {
-                            warn!("File {file} has an unexpected metadata value {name}='{value}'. Only 'tms' is supported. Ignoring.")
+                            let file = &self.filename();
+                            warn!("File {file} has an unexpected metadata value {name}='{value}'. Only 'tms' is supported. Ignoring.");
                         }
                     }
                     _ => {
--- a/mbtiles/src/metadata.rs
+++ b/mbtiles/src/metadata.rs
@@ -111,10 +111,7 @@ impl Mbtiles {
         let query = query!("SELECT name, value FROM metadata WHERE value IS NOT ''");
         let mut rows = query.fetch(&mut *conn);
 
-        let mut tj = tilejson! {
-            tiles: vec![],
-            scheme: "xyz".to_string()
-        };
+        let mut tj = tilejson! { tiles: vec![] };
         let mut layer_type: Option<String> = None;
         let mut json: Option<JSONValue> = None;
         let mut agg_tiles_hash: Option<String> = None;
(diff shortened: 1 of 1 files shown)


mbtiles/src/metadata.rs
```diff
@@ -111,7 +111,10 @@ impl Mbtiles {
         let query = query!("SELECT name, value FROM metadata WHERE value IS NOT ''");
         let mut rows = query.fetch(&mut *conn);
 
-        let mut tj = tilejson! { tiles: vec![] };
+        let mut tj = tilejson! {
+            tiles: vec![],
+            scheme: "xyz".to_string()
```
nyurik: I don't think we need this - the default of `xyz` is sufficient

mbtiles/src/metadata.rs
```diff
@@ -136,6 +139,7 @@ impl Mbtiles {
                         tj.other.insert(name, Value::String(value));
                     }
                     "agg_tiles_hash" => agg_tiles_hash = Some(value),
+                    "scheme" => (), // ignore scheme because it defaults to xyz
```
nyurik: I see the issue - so basically we treat `schema` as a custom field, and because your mbtiles just happen to have it set, I copy it into the output as is... I see how this could be a problem.

We should not silently ignore it though, so perhaps something like this?  (haven't tested, might use some wordsmithing)

```suggestion
                    "scheme" => {
                        if value != "tms" {
                            warn!("File {file} has an unexpected metadata value schema='{value}'. Only 'tms' is supported. Ignoring.")
                        }
                    }
```

mbtiles/src/metadata.rs
```diff
@@ -136,6 +136,11 @@ impl Mbtiles {
                         tj.other.insert(name, Value::String(value));
                     }
                     "agg_tiles_hash" => agg_tiles_hash = Some(value),
+                    "scheme" => {
+                        if value != "tms" {
+                            warn!("File {file} has an unexpected metadata value schema='{value}'. Only 'tms' is supported. Ignoring.")
```
nyurik: Typo, lets use a var instead

```suggestion
                            warn!("File {file} has an unexpected metadata value {name}='{value}'. Only 'tms' is supported. Ignoring.")
```

nyurik (approved):

PR #1698, opened 2025-02-22
2024 rust edition

CommanderStorm:
This PR upgrades (`cargo fix --eddition`, then fix what does not make sense) our crates to the 2024 edition.

While quite early, this enables things such as the v3 resolver, which might be helpfull to prevent MRSV breakages in the future by a MSRV breakage now.

Most of the changes are from rustfmt => going commit-by-commit is likely a good approach

PR #1718, opened 2025-03-03
Revert "2024 rust edition"

CommanderStorm:
Reverts maplibre/martin#1698

I have no idea why this was automerged.. that was a mistake

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,12 +1,12 @@
 [workspace]
-resolver = "3"
+resolver = "2"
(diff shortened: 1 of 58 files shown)


nyurik (approved):

nyurik:
Usually I have a "finishing workflow" - something that does not actually do any testing, but depends on all the critical CI jobs that must pass. This way branch protection rule can depend on just that one job, which in turn depends on all the critical ones we expect to work. No idea if something changed though, may need to investigate

PR #1719, opened 2025-03-03
2024 rust edition

CommanderStorm:
Re-Reverts maplibre/martin#1718

This PR upgrades (`cargo fix --eddition`, then fix what does not make sense) our crates to the 2024 edition.

While quite early, this enables things such as the v3 resolver, which might be helpfull to prevent MRSV breakages in the future by a MSRV breakage now.

Most of the changes are from rustfmt => going commit-by-commit is likely a good approach

Depends on:
- rust 1.85 being avaliable in githubs CI
  https://github.com/actions/runner-images/issues/11637
  - [x] Windows Runner rollout
  - [x] Unix Runner rollout
  - [x] Mac Runner rollout
- rust 1.85 being avaliable in pre-commit.ci
  - [x] https://github.com/pre-commit-ci/runner-image/pull/294
  - [x] pre-commit CI runner rollout
- [x] #1752


nyurik:
I think precommit has updated their ubuntu images, so this might be ready to go

CommanderStorm:
> I think precommit has updated their ubuntu images

Apparently rollout for both GH and pre-commits' CI is slower than thought.
We are patient ^^

martin/src/lib.rs
```diff
@@ -45,8 +45,10 @@ pub use crate::args::Env;
 mod test_readme {
     macro_rules! external_doc_test {
         ($x:expr) => {
-            #[doc = $x]
-            extern "C" {}
+            unsafe {
+                #[doc = $x]
+                extern "C" {}
+            }
```
nyurik: wouldn't this work?

```suggestion
            #[doc = $x]
            unsafe extern "C" {}
```

martin/src/args/environment.rs
```diff
@@ -14,16 +14,16 @@ pub trait Env<'a>: VariableMap<'a> {
...
         match self.var_os(key) {
-            Some(s) => {
-                match s.into_string() {
-                    Ok(v) => Some(v),
-                    Err(v) => {
-                        let v = v.to_string_lossy();
-                        warn!("Environment variable {key} has invalid unicode. Lossy representation: {v}");
-                        None
-                    }
+            Some(s) => match s.into_string() {
```
nyurik, line 17: **might** be easier with `.map_err(...).ok()` ?

martin/src/lib.rs
CommanderStorm: Lets try it

martin/src/args/environment.rs
CommanderStorm: maybe, but it would introduce a clone due to how the borrow checker currently works..
(happy to be proven wrong, could not get it to work out)

Lets keep it as is for now

.github/workflows/grcov.yml
```diff
@@ -51,7 +51,7 @@ jobs:
       - name: Install nightly toolchain
         uses: dtolnay/rust-toolchain@master
         with:
-          toolchain: nightly-2024-10-01
+          toolchain: nightly-2024-03-15
```
CommanderStorm: oops, small typo ^^

```suggestion
          toolchain: nightly-2025-03-15
```

CommanderStorm:
I think this should be everything needed for upgrading to the 2024 edition ^^

nyurik (approved):
nice!

PR #1725, opened 2025-03-07
Add "extract" subcommand to mbtiles tool

NicoManderySWM:
This PR adds a "extract" subcommand to the `mbtiles` tool, to write the tiles contained in a mbtiles archive to the filesystem or to a object store. 

While I am not sure if this feature is still in scope for the `mbtiles` tool, we where in need of a feature like this to switch our rendering process to `martin` while still being able to deploy the rendered tiles directly to a object store. Maybe the naming of this subcommand is not perfect as it does a bit more than just extracting, but due to lack of a better naming idea I went with this name for now ;)

The implementation is based on the [`object_store` crate ](https://docs.rs/object_store) of the Apache Arrow project and brings support for Amazon S3 (and compatible alternatives), Azure BlobStore and Google Cloud Store. Local file access is also done through the abstraction of this crate.

As configuration options for the various object store types differs, the CLI interface of the new subcommand allows passing options to the `object_store` crate using key-value pairs. I tried to document this in the commands help texts and to provide links to upstream documentation.

When the object-store type supports Content-Type and Content-Encoding metadata, these are set accordingly to the contents of the mbtiles archive.

The key / filename of the single tiles is specified using a string template to provide a maximum of flexibility for the user.


nyurik:
Thanks! I'll look at it in the morning, but this is definitely something we should support...  but "how" is slightly nuanced. 

There are several use cases, each somewhat different:
* copy a lot of tiles from one backend to another. For this, we have 'martin-cp' - I think it is much better suited for various bulk operations, while supporting all backends and merging. At the moment, it only supports mbtiles as the destination, but can be extended
* debugging - dump a single tile from mbtiles, possibly also decoding it - this is more in line with mbtiles util
*  Perhaps a new debugging interface or a cli util to get a tile from any backend... this is less certain, but it could be good to support all

CommanderStorm:
> to write the tiles contained in a mbtiles archive to a object store.

Given that object storage is charging you partially by the objects amount, this might get pretty expensive. 2.5ct/month per 1k objects is not insignificant..
=> for a better cost estimate see https://tools.geofabrik.de/calc/#type=geofabrik_standard

PMTiles are possibly a better option for your usecase, if you only need static tiles.
`martin-cp` currently cannot do this step though..
As Yuri hinted, a PR here would likely be better for your budget.

NicoManderySWM:
I placed this in the `mbtiles` tool as this functionality only operates on pre-rendered mbtiles archives. As I understood `martin-cp` operates on a martin configuration file.

PMTiles is certainly a better way to deploy static tiles, but the billing argument can be less severe depending on the usecase - for example when tiles are only rendered for smaller regions (like our original usecase), or when non-cloud services are used. ... and yes, PMTiles is the better solution to deploy to the big cloud providers ;)

nyurik:
`martin-cp` should appear very similar to `martin` itself as far as CLI goes -- see [docs](https://maplibre.org/martin/martin-cp.html). In other words, `martin-cp` adds two required params: `--source` and `--output-file` -- the rest are the same params as in martin, like `postgres://...` would auto-add PG sources, and `foo.mbtiles` would add an mbtiles source.  If you want an ability to extract all the tiles from mbtiles, we **may** want something like this:

```
martin-cp  \
           --output-dir /path/to/dir     \  # <-- instead of --output-file
           --mbtiles-type normalized     \  # optional
           "--bbox=-180,-90,180,90"      \  # optional
           --min-zoom 0                  \ 
           --max-zoom 10                 \
           --source source_name          \  # <-- same name as the mbtiles file next line
           /path/to/source_name.mbtiles
```

This will keep the overall structure the same, give ability to combine multiple sources as neeeded, and will not introduce any mbtiles-specific interface

CommanderStorm:
As indicated above:
In `martin-cp` this functionality would be in-scope. An object store being a destination sounds resonable.
However, in `mbtiles`, it is kind of akward, at least how it is currently implemented.

I am going to close the PR as I don't think this PR has momentum to migrate the code to integrating the functionality into `martin-cp`.
We can obviously reopen if this changes.

Thanks for the effort you put into this even if this did not lead to a merge.

PR #1730, opened 2025-03-08
Remove cog doc

sharkAndshark:
To load tiles from COG the fields in  #1676  is required. To make the users not confusing, I suggested to remove the doc of cog before
* find generally way to support non web mercator tile grid in martin
* find the way to tell user the fields in #1676 

Let's see cog as a nightly feature I will keep working on this.

And I've pinned the #343 , get more inspiration form other repos and talk more there!

@nyurik


nyurik (approved):

nyurik:
it seems this PR somehow broke the summary list - only showing items after the comment. Probably a bug in the rendering engine, but :shrug:. I will delete it.

image

PR #1734, opened 2025-03-08
chore: add additional metadata to `Cargo.toml`

CommanderStorm:
During #1720 I noticed that these fields are missing
This PR adds them for completeness ^^


nyurik:
Isn't Readme field autofilled if Readme exists? I'll need to double check

CommanderStorm:
image

True, that is the default

nyurik (approved):
lgtm

PR #1736, opened 2025-03-09
chore: improve CLI line breaks and markdown-handling

CommanderStorm:
During #1720 I noticed that there are a few places in the CLI where a few more line breaks and handling markdown better could help readability.

Yes, this adds the `unstable-markdown` feature of clap, but given that removal would only impact the CLI-Formatting, I think here the risk is okay.
Tracking issue:
- https://github.com/clap-rs/clap/issues/5900

Hidden rationale:
I have one place in #1720 where making it readable without markdown formatting is nearly impossible


nyurik (approved):
thanks!

PR #1752, opened 2025-03-15
tests: migrate from gcov to llvm-cov

CommanderStorm:
gcov is not supported on rust 2024 and the `-Zprofile` option was dropped on nightly.

This migrates to the `llvm-cov` instrumentation

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -52,15 +52,12 @@ jobs:
       - name: Install cargo-llvm-cov
         uses: taiki-e/install-action@cargo-llvm-cov
       - name: Generate code coverage
-        run: cargo llvm-cov --all-features --workspace --lcov --output-path lcov.info
+        run: cargo llvm-cov --all-features --workspace --codecov --output-path codecov.info
         env:
-          CARGO_INCREMENTAL: '0'
-          RUSTFLAGS: '-Ccodegen-units=1 -Cinline-threshold=0 -Clink-dead-code -Coverflow-checks=off -Cpanic=abort'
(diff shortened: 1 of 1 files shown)


nyurik (approved):
thanks so much for working on this!  I can see it being posted in https://app.codecov.io/gh/maplibre/martin/tree/CommanderStorm%3Acoverage-reports

CommanderStorm:
Seems to have worked: https://app.codecov.io/github/maplibre/martin/commit/4a1e00349b3166c940da6a4702770099999843a3

nyurik:
my main concern with the code coverage is that for some reason all our tests are not combined into a single coverage report - i.e. we have unit tests, integration tests, and automated tests with the `test.sh` script - and ideally we should see the coverage report for all of them combined before being uploaded to the codecov

nyurik:
i wonder why codecov no longer posts directly to this PR's comments

CommanderStorm:
> my main concern with the code coverage is that for some reason all our tests are not combined into a single coverage report - i.e. we have unit tests, integration tests, and automated tests with the test.sh script - and ideally we should see the coverage report for all of them combined before being uploaded to the codecov

Lets test external test coverage in a different PR
=> https://github.com/taiki-e/cargo-llvm-cov?tab=readme-ov-file#get-coverage-of-external-tests

> why codecov no longer posts directly to this PR's comments

For this PR, I would not expect a change in coverage => there should not be a comment:
image

nyurik:
oh, thx, i wasn't aware of the additional coverage commands. This would be so awesome to include here (I'll gladly copy it to all my other projects too :) )

PR #1753, opened 2025-03-15
Include full test into the coverage

CommanderStorm:
This is a test to see if multiple runs in the grcov github workflow would increase the coverage. So far seems like only the last one is used

Codecov page: https://app.codecov.io/gh/maplibre/martin/tree/CommanderStorm%2Fmartin%3Asplit-codecov

Resolves #541


nyurik:
I think this approach might be good for martin:

https://github.com/nyurik/sqlite-hashes/pull/72/files

nyurik:
This PR is even cleaner I think:  https://github.com/nyurik/sqlite-compressions/pull/84/files

The main idea is that this way we can test coverage locally (shows an html report), and the same `just` command is used in both ci and locally

CommanderStorm:
@nyurik bit of a gocha:
image

CommanderStorm:
Some things about the codecov-data are a bit sus thb..
I don't think that "it" checks that the log output is asserted for example..

image

CommanderStorm:
Otherwise, is this something that we can be proud enough to put in the readme as a badge?

image

nyurik:
Is this better than what we had? Does it actually combine multiple runs?

CommanderStorm:
compared to main, yes.
image

Lets see what happens if I toggle the unit tests (the only part which could be overridden..)

CommanderStorm:
I noticed that above, I accidentally had unittests enabled, which does not quite work without serious hackery currently.
Here are the versions without doing that:

- integration + unit tests: https://app.codecov.io/gh/maplibre/martin/commit/511bdf521c559969ddb573a06e7099e29f138c30
-  integration tests: https://app.codecov.io/gh/maplibre/martin/commit/d917993db83c7bb06935d9f9085cbe1351bc0c77

(Once CI is done)

nyurik:
I'm looking from a phone, but I don't see a comment from the cov site - especially if the coverage has changed, it should have posted some comment about it, shouldn't it?

CommanderStorm:
Yea, seems weird..
Something is definitively fucked there.

Maybe a permission issue?
=> if we are using a user account, we maybe forgot to update the token
=> if we are using the [integration](https://docs.codecov.com/docs/team-bot), is there something in the github admin settings?

CommanderStorm:
image

Okay, the unit tests defitively get looked at ^^

justfile
```diff
@@ -181,6 +181,24 @@ test-int: clean-test install-sqlx
         fi
     fi
 
+# Generate code coverage report
+coverage *ARGS="--open": clean start (cargo-install "cargo-llvm-cov")
+    #!/usr/bin/env bash
+    set -euo pipefail
+    if ! rustup component list | grep llvm-tools-preview > /dev/null; then \
```
nyurik: why do you need to manually install `llvm-tools-preview`?  Shouldn't the cargo-llvm-cov automatically do that?

justfile
```diff
@@ -181,6 +181,24 @@ test-int: clean-test install-sqlx
         fi
     fi
 
+# Generate code coverage report
+coverage *ARGS="--open": clean start (cargo-install "cargo-llvm-cov")
```
nyurik: please keep these in the same place in the just file - makes it much easier to review the changes vs cleanups

.github/workflows/coverage.yml
```diff
@@ -20,44 +20,20 @@ jobs:
     runs-on: ubuntu-latest
     env:
       CARGO_TERM_COLOR: always
-      # PG_* variables are used by psql
```
nyurik, line 23: why remove the service here? How is this related to the test coverage?

CommanderStorm: The just based coverage tests already do this.
No need to do this in the GitHub yaml format too.

Can re-add if you prefer

justfile
CommanderStorm: Yes and no.
They would, but since we execute this the first time in the env-variable step it is not printed to the screen that the user needs to press y..

nyurik (approved):
looks great!  Thanks!!

PR #1759, opened 2025-03-22
Do not publish Cargo.lock to creates.io

nyurik:


sharkAndshark (approved):

CommanderStorm (approved):
so this effectively auto-update transitive dependencys.
Am a bit torn.
- on the pro side: we don't need to do a release if something is found there
- on the con side: one of these could break us and we might not know...

=> Lets try this and see where it leads

nyurik:
Heh, I am not certain about my own PR here tbh.
* `cargo install` uses it only if `--locked` flag is passed - see [docs](https://doc.rust-lang.org/cargo/commands/cargo-install.html#dealing-with-the-lockfile)
* The libs.rs owner's [dashboard](https://lib.rs/~nyurik/dash) shows a weird warning, which might actually be ok to ignore?
image

CommanderStorm:
I think at this point we can close the pr. We can obviously reopen if you want to merge this

PR #1774, opened 2025-04-07
chore: add SQLFluff to pre-commit hooks

IvanPiatnishin:
This adds SQLFluff with the PostgreSQL dialect to enforce SQL formatting rules via pre-commit.


.pre-commit-config.yaml
```diff
@@ -16,15 +15,15 @@ repos:
       - id: check-json
         exclude: '.+/tsconfig.json'
       - id: check-shebang-scripts-are-executable
-        exclude: '.+\.rs' # would be triggered by #![some_attribute]
+        exclude: '.+\.rs'
```
CommanderStorm: please re-add these comments. I think they add nice context

CommanderStorm:
the added CI-step is failing though. This needs to be fixed for mergability ^^
https://results.pre-commit.ci/run/github/105363726/1744040351.aYINARI8SCavxUA6wdemiQ

.pre-commit-config.yaml
IvanPiatnishin: sure, done

CommanderStorm: I think you forgot to push ^^

.pre-commit-config.yaml
```diff
@@ -36,16 +35,9 @@ repos:
...
-  #   rev: 3.3.1
-  #   hooks:
-  #     - id: sqlfluff-fix
-  #       name: sqlfluff-fix
-  #       args: [ "--dialect", "postgres" ]
-  #       files: '^.*\.sql$'
-
-# TODO: Add NPM hooks for formatting and linting
+  - repo: https://github.com/sqlfluff/sqlfluff
+    rev: 3.0.7
```
CommanderStorm: Copilot is right (thanks @sharkAndshark for thinking of this)
We should use latest

```suggestion
    rev: 3.3.1
```

.pre-commit-config.yaml
IvanPiatnishin: I forgot, I’ll do it a bit later, sorry:)

IvanPiatnishin: done!

.pre-commit-config.yaml
```diff
@@ -1,7 +1,6 @@
 # See https://pre-commit.com for more information
 # See https://pre-commit.com/hooks.html for more hooks
 
-# exclusions should be separated with a pipe (|) character and a newline
```
CommanderStorm: I also meant this with comments.

FYI: the formatting in CI is failing

CommanderStorm (changes requested):

.pre-commit-config.yaml
IvanPiatnishin: i added, sorry again

CommanderStorm: no worries ^^

sharkAndshark:
@IvanPiatnishin You may want to run commands to fix the formatting errors.

```bash
sqlfluff fix --dialect postgres demo/db/initdb/01.sql
sqlfluff fix --dialect postgres demo/db/initdb/03.sql
sqlfluff fix --dialect postgres martin/src/pg/scripts/query_available_function.sql
```

CommanderStorm:
There are a few things where you will need to look into the formatting. CI is still failing.
https://results.pre-commit.ci/run/github/105363726/1744269374.byjF0W4QT7WxnLg-J1GjkQ and https://github.com/maplibre/martin/actions/runs/14374564930/job/40304086214?pr=1774

sharkAndshark:
Or just run this to fix all `.sql` files
```bash
sqlfluff fix --dialect postgres
```

nyurik:
Note that ideally this should not just verify the style, but just like cargo fmt it should push changes to the pr

nyurik:
Note that this PR will not auto-run workflows because @IvanPiatnishin is a new github contributor. A workaround for this may be to create a PR in their own fork -- https://github.com/IvanPiatnishin/martin/compare/IvanPiatnishin:main...IvanPiatnishin:setup-sqlfluff-formatting?expand=1

nyurik:
On a more substantive notes - sqlfluff currently catches these two lints:

```
== [tests/fixtures/functions/function_special_characters.sql] FAIL
L:   1 | P:  32 | RF05 | Do not use special characters in identifiers.
                       | [references.special_chars]
== [martin/src/pg/scripts/query_available_tables.sql] FAIL
L:  11 | P:  39 | RF04 | Keywords should not be used as identifiers.
```

Both are actually OK for us because that's the whole idea of those tests - to see if our code will work correctly when the user does something weird.  So we should disable these specific lints in the config somewhere

CommanderStorm:
I changed a few things about the setup in https://github.com/maplibre/martin/pull/1774/commits/be894254fb564a4b6c3fa5e13ad3c5900588db2f
- enabled fixing
- re-added an removed comment
- changed which rules we ignore (see below for rationale)

I think `RF02` (Unqualified reference), `RF04` (Keywords should not be used as identifiers) and `RF05` (special characters in identifiers) they are interesting to keep as indicating possible correctness issues.
Not strong indicators, but a smell.

I would like them to be ignored instead:
https://docs.sqlfluff.com/en/3.2.3/configuration/ignoring_configuration.html

`LT05` (line lenght) is mostly triggering on comments. => lets ignore this one

nyurik (approved):
looks good!  It seems we might need to remove PG 11 support as it is no longer installing? (it should probably be a separate PR to keep things separate)

nyurik:
Is it possible to ignore certain rules just for a single file, or maybe even for a specific line in a file?  We could use something like that for keywords and special character tests.

sharkAndshark:
Seems sqlfluff support [In-File Configuration Directives](https://docs.sqlfluff.com/en/stable/configuration/setting_configuration.html#in-file-configuration-directives), we can do that

> to ignore certain rules just for a single file, or maybe even for a specific line in a file

And certain line could be configured by [this](https://docs.sqlfluff.com/en/stable/configuration/ignoring_configuration.html)

```sql
-- Ignore all errors
SeLeCt  1 from tBl ;    -- noqa

-- Ignore rule CP02 & rule CP03
SeLeCt  1 from tBl ;    -- noqa: CP02,CP03

-- Ignore all parsing errors
SeLeCt from tBl ;       -- noqa: PRS

```

CommanderStorm:
> Is it possible to ignore certain rules just for a single file

Yes, see for example this file. We already do so:
https://github.com/maplibre/martin/blob/be894254fb564a4b6c3fa5e13ad3c5900588db2f/tests/fixtures/functions/function_special_characters.sql

CommanderStorm (approved):
Forgot to ACK this PR

PR #1776, opened 2025-04-07
docs: add styling documentation by fixing wrong linkage

CommanderStorm:
Bit akward and really hard to catch in review.
I never read the mdbook => never knew that if I don't add files to `SUMMARY.md`

I also moved the `sources-sprites.png` to the images directory as semi-related


nyurik (commented):
thx!

nyurik (approved):

PR #1785, opened 2025-04-12
Remove postgis 11-3.0 test and fix gdalbin installation

sharkAndshark:


.github/workflows/ci.yml
```diff
@@ -454,9 +454,6 @@ jobs:
       matrix:
         include:
           # These must match the versions of postgres used in the docker-compose.yml
-          - img_ver: 11-3.0-alpine
-            args: postgres
-            sslmode: disable
```
CommanderStorm: I might be missing something, but I don't think we need to drop this.
https://github.com/maplibre/martin/pull/1774 only had problems with gdal as far as I can see. 🤔

CommanderStorm (approved):
LGTM, assuming https://github.com/maplibre/martin/pull/1785#discussion_r2040624877 has a reason ^^

.github/workflows/ci.yml
sharkAndshark: Yes.It failed because gdal installation. And Maybe it's time to drop the test as v11 is too old? @nyurik

.github/workflows/ci.yml
```diff
@@ -454,9 +454,6 @@ jobs:
       matrix:
         include:
           # These must match the versions of postgres used in the docker-compose.yml
```
CommanderStorm: Lets split these apart. the CI issue is getting annoying

```suggestion
          # These must match the versions of postgres used in the docker-compose.yml
          - img_ver: 11-3.0-alpine
            args: postgres
            sslmode: disable
```

.github/workflows/ci.yml
nyurik: yeah, i think v11 and even v12 might be ok to be dropped

Issue #580, opened 2023-02-20
Add support for etag header


nyurik:
CDNs rely heavily on etag (usually a hash) to identify if the tile has changed or not. In some cases, the tile backend like .mbtiles is already storing tiles with some hash. For example, .mbtiles frequently uses two tables to remove duplicate tiles - one stores `z, x, y, hash` columns, and another table - `hash, data blob`.

So there are multiple steps:
- [x] add automatic (fallback) generation of an etag for all tiles https://github.com/maplibre/martin/pull/1834
- [x] add automatic generation of an etag for all stable non-tile endpoints https://github.com/maplibre/martin/pull/1836
- [x] make mbtiles backend recognize when the `.mbtiles` file has two frequently used tables as described above, and use the hash value as etag https://github.com/maplibre/martin/pull/1787
- [ ] allow individual backends like mbtiles to supply a pre-generated etag
- [ ] allow PostgreSQL functions to return rows with two columns, treating the second column as the key. OpenMapTiles already generates these types of functions.
- [ ] support etags for `/catalog`

dericke:
The documentation at https://maplibre.org/martin/sources-pg-functions.html reads, to me, like this feature is already implemented. I landed here while troubleshooting why I wasn't seeing ETags on my tiles. If Martin isn't able to serve ETags yet, perhaps that should be clarified in the docs?

CommanderStorm:
I don't think we actually do, but we would love a PR ^^

Said part of the docs is talking primarily about how we recognise functions.
Quite misleading, true.

nyurik:
Thx for handling it @CommanderStorm !

CommanderStorm:
> support etags for `/catalog`

This is acutally not as simple as it looks.
When I changed to `HashMaps` for storage of the non-internal catalog, this also lost us the ability to hash this effectively.
And since the output is not stable, bytewise changes are also out of the question.

Likely this will be simpler to do after #288

nyurik:
Why can't we make the output stable, e.g. do a recursive key sort?

nyurik:
P.S.  initially i was actually thinking that the catalog should not be a hashmap, but rather a `Vec<(String, SomeValue)>` -- this way the order will always be guaranteed, and easier to regenerate/sort.

CommanderStorm:
> Why can't we make the output stable, e.g. do a recursive key sort

If we want to go this route, we should store our public catalog in `BTreeMap`'s.
[...]

PR #1786, opened 2025-04-12
docs: add more docstrings to `mbtiles`

CommanderStorm:
I did a bit of investigation what changes would be nessesary to support https://github.com/maplibre/martin/issues/580 and these are the changes where I worked myself through the functions in said crate.

Most are not related to the actual investigation, don't infer too much from which functions I looked at.

### DIFF
--- a/mbtiles/src/mbtiles.rs
+++ b/mbtiles/src/mbtiles.rs
@@ -131,6 +131,9 @@ impl Mbtiles {
         Ok(())
     }
 
+    /// Get a tile from the database
+    ///
+    /// See [`MBTiles::get_tile_and_hash`] if you need the hash too
     pub async fn get_tile<T>(
         &self,
         conn: &mut T,
--- a/mbtiles/src/mbtiles.rs
+++ b/mbtiles/src/mbtiles.rs
@@ -132,8 +132,6 @@ impl Mbtiles {
     }
 
     /// Get a tile from the database
-    ///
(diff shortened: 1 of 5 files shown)


mbtiles/src/mbtiles.rs
```diff
@@ -131,6 +131,9 @@ impl Mbtiles {
         Ok(())
     }
 
+    /// Get a tile from the database
+    ///
+    /// See [`MBTiles::get_tile_and_hash`] if you need the hash too
```
CommanderStorm:
```suggestion
```

mbtiles/src/pool.rs
```diff
@@ -18,11 +18,17 @@ impl MbtilesPool {
...
+    ///
+    /// See [`Metadata`] for more information.
     pub async fn get_metadata(&self) -> MbtResult<Metadata> {
         let mut conn = self.pool.acquire().await?;
         self.mbtiles.get_metadata(&mut *conn).await
     }
 
+    /// Get a tile from the pool
+    ///
+    /// See [`MbtilesPool::get_tile_and_hash`] if you need the tiles' hash for cases where it exists.
```
CommanderStorm:
```suggestion
```

CommanderStorm (commented):
xD
well, they don't exist (yet)

nyurik (approved):
thanks!

PR #1787, opened 2025-04-12
feat: `mbtiles` hash extraction support

CommanderStorm:
(PR is based on #1786 => merge said PR first before looking at this one.)

This PR adds support to the `mbtiles` crate to hand out which hash corresponds to a tile.

The flat case might be debatable: Here I chose to md5 hash it to make upstream code more readable. The alternative would be to in this case return `None` for the hash.

This looks more intimidating than it actully is. Most of the lines are unit-tests.


mbtiles/src/mbtiles.rs
```diff
@@ -153,6 +155,62 @@ impl Mbtiles {
...
+        &self,
+        conn: &mut SqliteConnection,
+        mbt_type: MbtType,
+        z: u8,
+        x: u32,
+        y: u32,
+    ) -> MbtResult<Option<(Vec<u8>, String)>> {
+        if mbt_type == MbtType::Flat {
+            // we need to register the md5 functions here as the connection might change
+            let mut handle_lock = conn.lock_handle().await?;
```
nyurik: I think we should do this as part of pool creation/mbtiles opening, not on tile retrival.  Moreover, i think this only needs to be done once per file, not once per pool connection

CommanderStorm: That was my initial approach as well.

Thing is, I did not find an option to register functions when a new "thing" in sqlx's Pool gets added. I can change the first "thing", but not subsequent "spans".
Luckyly my tests fail on the naive approach.

We could `load_extension("name")`, but this requires https://github.com/nyurik/sqlite-hashes to be installed on a users device. Not ideal.

CommanderStorm: the reason why I am using the dedicated hashes on the lowest level (instead of the file-level agg hash) is that `ETAG` seem to require that they represent the resource
=> 1/0/1 and 1/0/0 being `ETAG=FOOBAR` would not be allowed.

Thinking about it: I could hash the tile inline without loading the sqlite extension.

nyurik: I think a better path would be to add generic hashing support, and not neseserily the MD5 (which is not that great for many reasons). This way if a tile source has a hash, use that - which would work on all tile sources. If not, compute it on the fly.  I think [xxhash](https://crates.io/crates/xxhash-rust) is the better choice - e.g. the XXH128 is the same size as MD5 IIRC, but much faster to compute.

CommanderStorm: When you say make it generic do you mean:
- please use xxhash here
- change hash to an option, return None if not present => hash one level up

nyurik: The later

mbtiles/src/pool.rs
```diff
@@ -26,9 +37,303 @@ impl MbtilesPool {
...
+                      - "Washington, D.C."
+                      - Ürümqi
+                      - Ōsaka
+                count: 68
+                geometry: Point
+                layer: cities
+        agg_tiles_hash: D4E1030D57751A0B45A28A71267E46B8
+        "#);
+        assert_eq!(pool.detect_type().await.unwrap(), MbtType::FlatWithHash);
+        let t1 = pool.get_tile(0, 0, 0).await.unwrap().unwrap();
```
CommanderStorm: no clue why all of the sudden getting a tile for this fails.
Will look at this tomorrow/in the evening.

Weird

CommanderStorm: Reason was that 0,0,0 never existed in said mbtiles archive.

nyurik:
we really ought to just generate our own test files - perhaps have a short SQLite script that fake-creates a few mbtiles with well understood properties.  This way instead of checking in binary blobs, we can check in and generate these files on the fiy, and have an easy to inspect SQL that shows what they have.

mbtiles/src/pool.rs
```diff
@@ -26,9 +37,302 @@ impl MbtilesPool {
...
+        tilejson:
+          tilejson: 3.0.0
+          tiles: []
+          vector_layers:
+            - id: cities
+              fields:
+                name: String
+              description: ""
+              maxzoom: 6
+              minzoom: 0
```
CommanderStorm, line 205: Agree that generating test files might uncover bugs.
Here we are handling this too simplistic for example.

We are trusting that the metadata is actually valid.

nyurik: yep,  and actually generating sqlite files should be very simple - we can just use the `.dump` command in sqlite to export the content of all existing files, delete the originals, clean up the generated dumps, remove any unrelated stuff from them, and use them to generate .mbtiles files during testing - e.g. with some extra `init-mbtiles` just command to set them up in some build dir.

nyurik (commented):
this looks good, but I don't want to merge something that is not being used, i.e. dead code -- can you add usage of the new fn to martin itself?

CommanderStorm:
This is not actually dead code... mbtiles is a libary

The actual PR implementing ETAGs is a bit messy and needed another refactoring round before being an PR.
I would like to not merge them, to keep this reviewable.

We can leave it open and merge once I get to finishing that work.

nyurik (approved):
fair enough, lets merge, and thanks!

PR #1801, opened 2025-04-22
docs: add more meat to the development section

CommanderStorm:
This PR tries to add more meat to the devlopment seciton

Looks like much changed, actually is not.

Here is how the two `details` blocks look rendered
screencapture-file-home-frank-dev-maplibre-martin-target-book-development-html-2025-04-22-02_21_26

Content wise, some of the explainations are inspired by http://gist.github.com/nyurik/4e299ad832fd2dd43d2b27191ed3ec30/

PR #1806, opened 2025-04-25
hack: remove the PR file changed checks for CI

CommanderStorm:
Yes, a bit hacky and not the proper solution, but gets this bug silenced.
I don't like that CI always fails on docs PRs.
F.ex https://github.com/maplibre/martin/pull/1801 could be merged, but needs admin approval.

A proper solution would be to skip all except `CI FInished` if just docs were changed.


nyurik (approved):
sigh, i guess optimizations are just not worth the trouble of force override... i have seen some workflows that somehow iterate over all workflows and check if any failed... TBD

PR #1813, opened 2025-04-30
Add `just fmt-sql` to just recipe

### LINKED ISSUE #1778: Add sqlfluff fix to just fmt recipe
Currently `just fmt` only format rust code. It would be better if format `.sql` files either?

Call `sqlfluff fix` in `just fmt` command. 

* [just doc](https://github.com/casey/just)
* [sqlfluff doc](https://docs.sqlfluff.com/en/stable/)
* [the just file martin using](https://github.com/maplibre/martin/blob/6cb8fc4f3e7a88d3b772f391c1a2ceb5dd0707be/justfile)

sharkAndshark:
Fix #1778

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -43,6 +43,7 @@ repos:
         # AL07: Avoid table aliases in from clauses and join conditions <- we sometimes do this for long names
         # LT05: Line is too long <- not useful. mostly triggers for comments
         # LT12: Files must end with a single trailing newline <- already done by trailing-whitespace
+        # If you update this rule, please also update the rules in justfile, the fmt-sql recipe
         args: [--dialect=postgres, "--exclude-rules=AL07,LT05,LT12"]
         files: \.sql$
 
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -43,6 +43,7 @@ repos:
         # AL07: Avoid table aliases in from clauses and join conditions <- we sometimes do this for long names
         # LT05: Line is too long <- not useful. mostly triggers for comments
         # LT12: Files must end with a single trailing newline <- already done by trailing-whitespace
+        # If you update this rule, please also update the rules in justfile, the fmt-sql recipe
```
sharkAndshark, line 46: Or maybe use a same config file to avoid this

CommanderStorm (approved):

.pre-commit-config.yaml
CommanderStorm: That would only work if pre-commit were not pretty stubborn in not using it.

sharkAndshark: Guess it's enough with this comment.

sharkAndshark: We will know it's not enough when we konw loll

Issue #931, opened 2023-10-10
Enabling CORS via Martin


nouryf:
Hi,

I'm trying to prevent access to resources on a website (from any other not supposed to) by using CORS on Nginx. I did setup a reverse proxy to direct tiles requests to Martin (localhost:3000). On Nginx I tried :
add_header "Access-Control-Allow-Origin"  "https://www.myWebsite.com";

Looking at the response header I don't see that. So I went and edited Martin to enable it there by modifying "[server.rs](https://github.com/maplibre/martin/blob/main/martin/src/srv/server.rs)":
**_line 16:_**

```rust
use actix_web::{
  http, middleware, route, web, App, HttpMessage, HttpRequest, HttpResponse, HttpServer, Responder,
   Result,
};
```

**line 422:**

```rust
let server = HttpServer::new(move || {
let cors_middleware = Cors::default()
  .allowed_origin("https://www.myWebsite.com")
  .allowed_origin_fn(|origin, _req_head| {
     origin.as_bytes().ends_with(b".myWebsite.com")
  })
  .allowed_methods(vec!["GET"])
 .allowed_headers(vec![http::header::AUTHORIZATION, http::header::ACCEPT])
  .allowed_header(http::header::CONTENT_TYPE)
  .max_age(3600);
```

I rebuilt Martin and started it. Then I looked at the response header and I'm not seeing the allow origin there still:

```http
HTTP/1.1 200 OK
Server: nginx/1.18.0 (Ubuntu)
Date: Tue, 10 Oct 2023 10:49:33 GMT
Content-Type: image/jpeg
Content-Length: 19397
Connection: keep-alive
vary: Origin, Access-Control-Request-Method, Access-Control-Request-Headers
X-Cache-Status: MISS
```

My ngnix config for the reverse proxy is:
```nginx
proxy_set_header        X-Rewrite-URL $uri;
proxy_set_header        X-Forwarded-Host $host:$server_port;
proxy_set_header        X-Forwarded-Proto $scheme;
proxy_redirect          off;

proxy_connect_timeout   5m;
proxy_send_timeout      5m;
proxy_read_timeout      5m;
send_timeout            5m;

proxy_cache             backend_cache;
proxy_cache_lock        on;
proxy_cache_revalidate  on;
proxy_cache_valid       200 204 302 1d;
proxy_cache_valid       404 1m;
proxy_cache_use_stale   error timeout http_500 http_502 http_503 http_504;
[...]

PR #1815, opened 2025-05-01
feat(martin-server):  add support for basic CORS configuration 

NINNiT:
As mentioned in  #931, I've tried to come up with a basic implementation/spec which allows configuration of CORS.. 

**Before**:
Static CORS configuration in `server.rs`. 
```rust
let cors_middleware = Cors::default()
         .allow_any_origin()
         .allowed_methods(vec!["GET"]);
```
This translates to the following logic:
- If an incoming request has an `Origin` Header, the value of this Headers is used to set `Access-Control-Allow-Origin` (not `*`)
- If an incoming request has no `Origin` Header, `Access-Control-Allow-Origin` is not set, however `vary` headers are still being sent

**After**:
- Same default behavior as before (CORS enabled, all origins allowed, only `GET` methods allowed). There should be no breaking changes for existing deployments.
- Allows configuration of the following properties:

```yaml
# CORS Configuration
# Can also be disabled via `cors: false`
cors:
  # Sets the `Access-Control-Allow-Origin` header [default: *]
  # '*' will use the requests `ORIGIN` header
  origin:
    - https://example.org
  # Sets `Access-Control-Max-Age` Header. [default: null]
  # null means not setting the header for preflight requests
  max_age: 3600
```
- Conditionally wraps the actix app in the CORS middleware, if the latter is enabled
```rust
   app.wrap(Condition::new(
            cors_middleware.is_some(),
            cors_middleware.unwrap_or_default(),
        ))
```
- Adds tests for CORS

- Config is nested under `SrvConfig` . I think this is fitting, as it's not a source. 

Let me know if you have any suggestions, changes that should be applied or if you'd like to take this over.  :crab: 
@CommanderStorm im gonna ping you here, hope that's okay


docs/src/config-file.md
```diff
@@ -44,6 +44,17 @@ preferred_encoding: gzip
...
+# CORS configuration
+cors:
+  # enable/disable CORS [default: true]
+  enable: true
+  # sets the Access-Control-Allow-Origin header [default: *]
+  # '*' will use the requests ORIGIN header
+  origin:
+    - https://example.org
+  # sets Access-Control-Max-Age Header. Remove to use default. [default: None]
+  max_age: 3600
```
CommanderStorm: Mostly, this is great work ❤️ 

I don't quite like the `enable: true` part of the config.
If CORS is entirely unconfigured, the rest of the options don't quite make sense.

What about below?

```suggestion
# CORS configuration
#
# Can also be disabled via `cors: false`
cors:
  # Sets the Access-Control-Allow-Origin header [default: *]
  # '*' will use the requests ORIGIN header
  origin:
    - https://example.org
  # Sets Access-Control-Max-Age Header. [default: null]
  # null means not seting the header
  max_age: 3600
```

(And changing the `CorsConfig` to be an enum and with `#[serde(transparent)]` on the variant with configuration)

.gitignore
```diff
@@ -23,3 +23,4 @@ tmp/
 ####
 #### Above content must match .dockerignore ####
 ####
+.aider*
```
nyurik: please move this above the comment, and also add it to the .dockerignore

docs/src/config-file.md
```diff
@@ -44,6 +44,17 @@ preferred_encoding: gzip
 # Enable or disable Martin web UI. At the moment, only allows `enable-for-all` which enables the web UI for all connections. This may be undesirable in a production environment. [default: disable]
 web_ui: disable
 
+# CORS Configuration
+# Can also be disabled via `cors: false`
```
nyurik: need to document what does `true` means, and what happens if not set at all (default).  Note that default here and in sub-items is a bit confusing, i.e. does `cors: {}` equal to `cors: ~` ?

docs/src/config-file.md
```diff
@@ -44,6 +44,17 @@ preferred_encoding: gzip
...
+cors:
+  # Sets the `Access-Control-Allow-Origin` header [default: *]
```
nyurik, line 52: just to double check -- does the default of `*` mean exactly the same as the current default behavior - i.e. allow all?  I.e. currently it "will use the requests `ORIGIN` header" ?

NINNiT: yep, "will use the requests ORIGIN header", is the current default .

[docs.rs](https://docs.rs/actix-cors/latest/actix_cors/struct.Cors.html#method.allowed_origin) 
`send_wildcard()` is disabled by default in `actix_cors`. 

> When all origins are allowed and send_wildcard is set, * will be sent in the Access-Control-Allow-Origin response header. If send_wildcard is not set, the client’s Origin request header will be echoed back in the Access-Control-Allow-Origin response header.

docs/src/config-file.md
NINNiT: good point - i'll expand on that and double check what happens. might even add some tests
never seen  `~` or `{}` in the wild though.

NINNiT: Right now, both `cors: {}` and `cors: ~` seem to enable CORS using our defaults.

NINNiT: As it stands right now, it'll always fall back to the defaults. 

- no `cors` option at all -> Enable using defaults
- `cors: true` -> Enable using defaults
- only configuring a subset and omitting certain properties-> use defaults for properties that are not explicitly set. 

 this...
```yaml
cors:
  max_age: 3600
```
would translate to this:
```yaml
cors:
  origin: ["*"]
  max_age: 3600
```
The only way to disable CORS right now, is to explicitly set `cors: false`. 

I guess `cors: ~` should also disable it? What would you prefer

docs/src/config-file.md
NINNiT: To expand on this a bit, this is the current CORS initialization on `main`:
```rust
    let cors_middleware = Cors::default()
            .allow_any_origin()
            .allowed_methods(vec!["GET"]);
```

- `default()`: sets restrictive defaults
*No* allowed origins, methods, request headers or exposed headers. Credentials
not supported. No max age (will use browser's default).
- `allow_any_origin()`: allows any origin, either by returning `*` in the header or by just returning the requests `ORIGIN` (this is the default in actix_cors as `send_wildcard()` is NOT enabled). .
- `allowed_methods(vec!["GET"])` sets `Access-Control-Allow-Methods`, only used in OPTIONS/preflight requests. 

This PR keeps the exact same defaults for now, but allows disabling CORS altogether and dynamically setting `ACCESS_CONTROL_ALLOW_ORIGIN` and `ACCESS_CONTROL_MAX_AGE`. 

If the defaults were to change at some point (this would be a breaking change), we could think about making it a bit more restrictive by default.

Maybe enable CORS by default, but require users to set specific origins. This would also ensure that an informed decision is made when opting out of "security" by allowing any origin or disabling it.

docs/src/config-file.md
nyurik: No, null is the same as skipped, so it should be the default

NINNiT:
I'm wondering if we actually want to fall back to defaults when CORS is explicitly configured.  
For example, when only a subset of properties is specified (e.g., just `max_age`), should we implicitly apply defaults to the omitted properties?

This:

```yaml
cors:
  max_age: 3600
```

would currently be interpreted as:

```yaml
cors:
  origin: ["*"] # default value
  max_age: 3600
```

Is that the intended behavior, or should partial configuration require more explicit values to avoid unintended/hidden defaults?

nyurik:
@NINNiT that's actually a good point - we want to avoid ppl making accidental mistakes, so yes, I agree - if they specify cors config, we should require it to have explicit origin.  That said, we obviously should print a helpful error message explaining what happens and what the fix would be

martin/src/srv/server.rs
```diff
@@ -165,7 +165,8 @@ pub fn new_server(config: SrvConfig, state: ServerState) -> MartinResult<(Server
             .clone()
             .cors
             .unwrap_or_default()
-            .make_cors_middleware();
+            .make_cors_middleware()
+            .expect("CORS configuration is invalid");
```
CommanderStorm: We likely want to bubble this up via `?` instead.
Dependeing on the way the borrow checker allows this, some cloning might be required.

```suggestion
            .make_cors_middleware()?;
```

martin/src/srv/cors.rs
```diff
@@ -30,12 +39,38 @@ impl Default for CorsProperties {
     }
 }
 
+impl CorsProperties {
+    pub fn validate(&self) -> Result<(), CorsError> {
+        if self.origin.is_empty() {
+            Err(CorsError::PropertyValidationError(
+                "When configuring CORS properties, 'origin' must be explicitly specified"
+                    .to_string(),
+            ))
```
CommanderStorm:
```suggestion
            Err(CorsError::NoOriginsConfigured)
```

martin/src/srv/cors.rs
```diff
@@ -1,6 +1,15 @@
 use actix_http::Method;
+use log::info;
 use serde::{Deserialize, Serialize};
 
+use crate::{MartinError, MartinResult};
+
+#[derive(thiserror::Error, Debug)]
+pub enum CorsError {
+    #[error("Invalid CORS configuration")]
+    PropertyValidationError(String),
```
CommanderStorm: If we only have one error, then keeping it overly generic as an `PropertyValidationError(String)` is not quite nessesary.

```suggestion
    #[error("At least one 'origin' must be specified in the 'cors' configuration")]
    NoOriginsConfigured,
```

martin/src/srv/cors.rs
```diff
@@ -134,4 +169,108 @@ mod tests {
...
+        let validation_result = properties.validate();
+        assert!(validation_result.is_err());
+
+        // Check that the error is the right type
+        match validation_result.unwrap_err() {
+            CorsError::PropertyValidationError(msg) => {
+                assert!(msg.contains("origin"));
+            }
+        }
+    }
```
CommanderStorm: We can cut out a lot of slack in these testcases.

Given the amount of comments, I somewhat assume that these are AI generated => please give it the task to be concice 😉

```suggestion
    #[test]
    fn test_cors_validation_error_empty_origin() {
        let properties = CorsProperties {
            origin: vec![],
            max_age: Some(3600),
        };
        assert_eq!(properties.validate(), Err(CorsError::NoOriginsConfigured));
    }
```

martin/src/srv/cors.rs
```diff
@@ -86,7 +121,7 @@ mod tests {
     #[test]
     fn test_cors_middleware_disabled() {
         let config = CorsConfig::SimpleFlag(false);
-        assert!(config.make_cors_middleware().is_none());
+        assert!(config.make_cors_middleware().is_ok_and(|x| x.is_none()));
```
CommanderStorm:
```suggestion
        assert_eq!(config.make_cors_middleware(), Ok(None));
```

CommanderStorm (commented):
I just noticed that this is still a draft. Unsure if you want feedback.
If you don't yet want feedback, feel free to ignore.

If there is anything where I can help you push this over the line, feel free to ask. I think this PR is basically done.

martin/src/srv/server.rs
NINNiT: Yep, those commits are still WIP - had to commit to be able to continue working on this on my notebook the next few days. Feedback is always welcome though! :heart: 

errors/logs aren't even printed right now, i think due to them being thrown within the actix worker threads.  i think we might have to do this outside of `factory` and clone it into the closure.

NINNiT:
> I just noticed that this is still a draft. Unsure if you want feedback.
> If you don't yet want feedback, feel free to ignore.
> 
> If there is anything where I can help you push this over the line, feel free to ask. I think this PR is basically done.

Feel free to clean/finish this up! 
Otherwise i'll do it later today/tomorrow. Would be nice if this made it into v0.17.

.dockerignore
```diff
@@ -32,6 +32,7 @@ tmp/
 **/node_modules/
 **/dist/
 
+.aider*
```
nyurik: please sort this line together with the ones above, both here and in the gitignore

martin/src/srv/cors.rs
```diff
@@ -0,0 +1,233 @@
...
+            assert_eq!(properties.max_age, None);
+        } else {
+            panic!("Expected Properties variant for default config");
+        }
+    }
+
+    #[test]
+    fn test_cors_middleware_disabled() {
+        let config = CorsConfig::SimpleFlag(false);
+        assert_eq!(config.make_cors_middleware(), Ok(None));
```
NINNiT: @CommanderStorm 
Some of your test changes aren't compiling on my end (stable toolchain, v1.87.0)

```
error[E0369]: binary operation `==` cannot be applied to type `std::result::Result<std::option::Option<Cors>, utils::error::MartinError>`
   --> martin/src/srv/cors.rs:121:9
    |
121 |         assert_eq!(config.make_cors_middleware(), Ok(None));
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |         |
    |         std::result::Result<std::option::Option<Cors>, utils::error::MartinError>
    |         std::result::Result<std::option::Option<Cors>, utils::error::MartinError>
    
    note: an implementation of `PartialEq` might be missing for `utils::error::MartinError`
   --> martin/src/utils/error.rs:30:1
    |
30  | pub enum MartinError {
    | ^^^^^^^^^^^^^^^^^^^^ must implement `PartialEq`
```
Looks like we'll have to implement `PartialEq` for `MartinError`
or assert  `Ok` and `Option` separately (the latter after unwrapping)

docs/src/config-file.md
```diff
@@ -44,6 +44,18 @@ preferred_encoding: gzip
 # Enable or disable Martin web UI. At the moment, only allows `enable-for-all` which enables the web UI for all connections. This may be undesirable in a production environment. [default: disable]
 web_ui: disable
 
+# CORS Configuration
+# Defaults to `cors: true`, which allows all origins.
```
CommanderStorm, line 49: Formatting nit

```suggestion
#
# Defaults to `cors: true`, which allows all origins.
```

docs/src/config-file.md
```diff
@@ -44,6 +44,18 @@ preferred_encoding: gzip
...
+# Can be completely disabled via `cors: false`
```
```suggestion
# Sending/acting on cors headers be completely disabled via `cors: false`
```

martin/src/srv/cors.rs
```diff
@@ -0,0 +1,236 @@
...
+                info!("CORS is disabled");
+                Ok(None)
+            }
+            CorsConfig::SimpleFlag(true) => {
+                // log out the default properties / serializte to string
+                info!(
+                    "CORS is enabled with default properties: {:?}",
+                    CorsProperties::default()
+                );
+                Ok(Some(Self::create_cors(&CorsProperties::default())))
```
CommanderStorm:
```suggestion
                let properties = CorsProperties::default();
                info!("Enabled CORS with defaults: {properties:?}");
                Ok(Some(Self::create_cors(&properties)))
```

martin/src/srv/cors.rs
```diff
@@ -0,0 +1,236 @@
...
+
+        if let CorsConfig::Properties(settings) = config {
+            assert_eq!(settings.origin, vec!["https://example.org".to_string()]);
+            assert_eq!(settings.max_age, Some(3600));
+        } else {
+            panic!("Expected Settings variant for detailed config");
+        }
+
+        let config: CorsConfig = serde_yaml::from_str("false").unwrap();
+        assert!(matches!(config, CorsConfig::SimpleFlag(false)));
```
CommanderStorm:
```suggestion
        assert_eq!(config, CorsConfig::SimpleFlag(false));
```

martin/src/srv/cors.rs
```diff
@@ -0,0 +1,236 @@
...
+
+        let config: CorsConfig = serde_yaml::from_str("true").unwrap();
+        assert!(matches!(config, CorsConfig::SimpleFlag(true)));
```
CommanderStorm:
```suggestion
        assert_eq!(config, CorsConfig::SimpleFlag(true));
```

martin/src/srv/server.rs
CommanderStorm: I need to check if logs are printed.
Might need to move this up a bit.

CommanderStorm (approved):
Good Job.
I have included a few nits and will finish up the rest when I have more stable Internet.

Sorry for the long march this feature has been and thanks for completing it.
We really appreciate the effort ♥️

martin/src/srv/server.rs
NINNiT: I think separating the validation from the middleware creation is the way to go here.

`actix_cors::Cors` is a tad picky with what it's allowing us to do. 

E.g when moving the whole part outside of the `factory` closure, we get the following error:
```
   │ │ │   ├╴  `std::rc::Rc<actix_cors::inner::Inner>` cannot be sent between threads safely
   │ │ │   │    within `{closure@martin/src/srv/server.rs:169:19: 169:26}`, the trait `std::marker::Send` is not implemented for `std::rc::Rc<actix_cors::inner::Inner>` rustc (E0277) [205, 42]
   │ │ │   ├╴  the trait bound `std::option::Option<actix_cors::Cors>: std::clone::Clone` is not satisfied
   │ │ │   │    required for `std::option::Option<actix_cors::Cors>` to implement `std::clone::Clone` rustc (E0277) [205, 42]
```
So:
Validation -> outside
Middleware Creation -> inside

I'll push that change soon, you can always change it tho  if you think there's a better way.

NINNiT:
If we are through with all the changes, I'd like to test this one last time in the real world.
Just to be sure, before this makes it into the release  :laughing:

CommanderStorm:
> I need to check if logs are printed.
Might need to move this up a bit.

I also need to check this part ^^

martin/src/srv/cors.rs
```diff
@@ -0,0 +1,243 @@
+use actix_http::Method;
+use log::info;
+use serde::{Deserialize, Serialize};
+
+use crate::{MartinError, MartinResult};
+
+#[derive(thiserror::Error, Debug)]
+pub enum CorsError {
+    #[error("At least one 'origin' must be specified in the 'cors' configuration")]
```
nyurik, line 9: we should probably mention the `"*"` as the common value? 

```suggestion
    #[error("At least one 'origin' must be specified in the 'cors' configuration, e.g. `origin: *`")]
```

martin/src/srv/server.rs
NINNiT: Change is pushed via 64d1133415e13a8880142f557eb3d1c9be24f403

This _should_ work and bubble up / print the errors/logs correctly.

I kept the double validation (once outside the closure, once within via `make_cors_middleware()`) -  just to make it a bit safer. This would at least panic early, if someone deletes the outside validation method by mistake.

martin/src/srv/cors.rs
NINNiT: I think `*` is pretty well known and would pop up in all CORS docs. 
It's also something that should generally be avoided from a security standpoint, so not sure if we should point users to that direction...
If they opt-in into configuring martin's CORS with their custom rules,  they might already have specific origins in mind.

nyurik (approved):

nyurik:
Thank you!!!

PR #1823, opened 2025-05-09
feat(martin-cp): Allow --source to not be required if only one source is available #1809

### LINKED ISSUE #1809: Allow `--source` to not be required for `martin-cp` if only one source is available
Currently, [martin-cp](https://maplibre.org/martin/martin-cp.html) requires the `--source` parameter. This is ok when martin would discover multiple tile sources, e.g. multiple tables in a PostgreSQL DB, but if the source is given as a single mbtiles, it becomes redundant. We need to allow the `--source` CLI parameter to be optional - unless more than one tile source is discovered. In that case, we should error out while providing a helpful message with the list of all configured sources.

sdf-jkl:
As mentioned in #1809 I was working on making the `--source` env var optional.
First I changed the `pub struct CopyArgs {`
`pub source: String,` to   
    `pub source: Option<String>,`
    
Then I created a match pattern in the `async fn start(mut copy_args: CopierArgs) -> MartinCpResult<()>`  <- (had to make copy_args mutable.)
 to check the `sources` length and base the actions on it.
```
    let source_count = sources.tiles.len();

    match source_count {
        0 => return Err(MartinCpError::NoSources),

        1 => {
            if copy_args.copy.source.is_none() {
                // Get the only source name from the DashMap
                if let Some(entry) = sources.tiles.iter().next() {
                    let only_source_name = entry.key().clone();
                    info!("Only one source detected: {}", &only_source_name);
                    copy_args.copy.source = Some(only_source_name)
                }
            }
        }

        _ => {
            if copy_args.copy.source.is_none() {
                let available_sources = sources.tiles.source_names().join(", ");
                return Err(MartinCpError::MultipleSources(available_sources));
            }
        }
    }
```
I had to create some `TileSources` methods to get the sources outside of struct, since it's a DashMap inside of a tuple and it's not `pub`:
```
    pub fn len(&self) -> usize {
        self.0.len()
    }

    pub fn iter(&self) -> dashmap::iter::Iter<'_, String, TileInfoSource> {
        self.0.iter()
    }

    pub fn source_names(&self) -> Vec<String> {
        self.0.iter()
        .map(|entry| entry.key().clone())
        .collect()
    }
```

I have also created two more `MartinCPError`'s:
```

    #[error("No sources found")]
    NoSources,
    #[error("More than one source found, please specify source using --source. \nAvailable sources: {0}")]
    MultipleSources(String),
```

fixes #1809


martin/src/bin/martin-cp.rs
```diff
@@ -154,6 +154,30 @@ async fn start(copy_args: CopierArgs) -> MartinCpResult<()> {
         info!("Use --save-config to save or print configuration.");
     }
 
+    let source_count = sources.tiles.len();
```
nyurik: i would simplify this match a bit, something like this:

```rust
if sources.tiles.is_empty() {
  return Err(MartinCpError::NoSources);
}

if copy_args.copy.source.is_none() {
  if sources.tiles.len() != 1 {
    return Err(...);
  }
  copy_args.copy.source = Some(only_source)
}
```

Also, I would move this logic into `run_tile_copy`

martin/src/bin/martin-cp.rs
```diff
@@ -273,13 +303,18 @@ fn iterate_tiles(tiles: Vec<TileRect>) -> impl Iterator<Item = TileCoord> {
     })
 }
 
-async fn run_tile_copy(args: CopyArgs, state: ServerState) -> MartinCpResult<()> {
+async fn run_tile_copy(mut args: CopyArgs, state: ServerState) -> MartinCpResult<()> {
+    let source = args
```
nyurik: the logic to decide witch source to use should all be done here - this way you avoid a potential panic of passing in a None instead of a value.  I.e., we shouldn't rely on panics (run time checks) if we can do the same thing at the compile time (by passing in a real value rather than an Option that will never be None)

nyurik (commented):
thanks for working on this!  It will need a few minor changes, but the path seems to be ok

CommanderStorm:
I think this would be nice to add to the integration tests.
This way, we guarantee, that this feature does not break by accident in the future.

Could you add a testcase around here
https://github.com/maplibre/martin/blob/528994fca6355c89027bac1b6059de5a3b6b283d/tests/test.sh#L476

martin/src/bin/martin-cp.rs
sdf-jkl: I've moved the logic into `run_tile_copy` and moving on to the next comment.

sdf-jkl:
> I think this would be nice to add to the integration tests. This way, we guarantee, that this feature does not break by accident in the future.
> 
> Could you add a testcase around here
> 
> https://github.com/maplibre/martin/blob/528994fca6355c89027bac1b6059de5a3b6b283d/tests/test.sh#L476

I'll work on it, thanks

martin/src/bin/martin-cp.rs
sdf-jkl: I've moved the logic into the `run_til_copy` function but did not fully remove the option part.
Can you please check the newest commit file changes?

martin/src/bin/martin-cp.rs
```diff
@@ -273,13 +279,26 @@ fn iterate_tiles(tiles: Vec<TileRect>) -> impl Iterator<Item = TileCoord> {
...
-async fn run_tile_copy(args: CopyArgs, state: ServerState) -> MartinCpResult<()> {
+async fn run_tile_copy(mut args: CopyArgs, state: ServerState) -> MartinCpResult<()> {
     let output_file = &args.output_file;
     let concurrency = args.concurrency.unwrap_or(1);
 
+    if state.tiles.is_empty() {
+        return Err(MartinCpError::NoSources);
+    }
+
+    if args.source.is_none() {
```
nyurik: don't modify `args` - because you would still have the same minor issue of unwrapping it later.  Just create a new var, and assign it either the value inside `args.source`, e.g. using this pattern -- notice that there are no unwraps, no are there any creation of combined string when we don't need it (wasteful, even if it is not critical performance here)

```rust
let source = if let Some(source) = args.source {
  source
} else {
  let sources = state.tiles.source_names();
  if let Some(source) = sources.first() {
    if sources.len() > 1 {
      return <error>;
    }
    source
  } else {
    return <noSources error>
  }
}
```

sdf-jkl: Thanks, I was having trouble coming up with a solution yesterday.

I've implemented your logic into `run_tile_copy`. Made the `args` unmutable again and removed the `unwrap()`

nyurik: np, there are a lot of rusty tricks like this - little brain teasers :)   See also this pattern for checking none vs 1 vs many:

https://github.com/maplibre/martin/blob/ef2af962bfeda52911e6158f7a339ede71b6f557/martin/src/utils/cfg_containers.rs#L46-L57

sdf-jkl: Now once I add the integration tests this PR should be good and the issue can be closed?

nyurik: the issue will close automatically as soon as this PR is merged (github recognizes "fixes" and similar keywords - I added it to the bottom of the top comment)

martin/src/source.rs
```diff
@@ -42,6 +42,18 @@ impl TileSources {
...
     }
 
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+
+    pub fn len(&self) -> usize {
+        self.0.len()
+    }
+
```
nyurik: not needed

```suggestion
```

martin/src/bin/martin-cp.rs
```diff
@@ -277,9 +283,23 @@ async fn run_tile_copy(args: CopyArgs, state: ServerState) -> MartinCpResult<()>
     let output_file = &args.output_file;
     let concurrency = args.concurrency.unwrap_or(1);
 
+    let source = if let Some(source) = &args.source {
+        source.to_string()
```
nyurik: this is totally fine here (low usage), but in high usage case, you would want to change this to `Cow::Borrowed(source.as_str())` -- this way you can have a var that can be either an owned or a borrowed value. The line below would than become `source.to_string().into()` or `Cow::Owned(source.to_string())` or similar.

nyurik (commented):
great!  Now all we need is a test, and we can merge :)

nyurik:
oh, and don't forget to make CI happy :)

sdf-jkl:
I'm trying to figure out how to proceed with testing. Specifically multiple scenarios part.

### Questions:
1. **Should I test all three scenarios (0 tables, 1 table, multiple tables) with different number of sources in the database?**
2. **If so, what would be the best way to implement it?**

### My current thoughts:
 1. I looked through the `tests/test.sh` and it looks like the database is already setup before the testing begins.
 2. I found the `tests/fixtures/` and `initdb.sh` inside, which is later run in `.github/workflows/ci.yml`.

My idea was to create 3 different database instances for this specific test. 
Then we can run something like:
```
martin-cp "source --" postgresql://postgres@localhost:5432/0_tables

martin-cp "source --" postgresql://postgres@localhost:5432/1_table

martin-cp "source --" postgresql://postgres@localhost:5432/2_tables
```

Another idea was to create functions that would change the database contents during the test phase.
Running sql queries on the database to test each scenario.

Would this approach work, or there is a better way to test it?

CommanderStorm:
> My idea was to create 3 different database instances for this specific test

That sounds quite reasonable - though a lot of work.
I would understand if you'd say that you don't have the time/energy for this.

sdf-jkl:
> > My idea was to create 3 different database instances for this specific test
> 
> That sounds quite reasonable - though a lot of work. I would understand if you'd say that you don't have the time/energy for this.

I was talking to @nyurik and he suggested a solution. It is to test copying one `.mbtiles` file into another. This way we avoid creating new databases and test the no `--source` scenario.

nyurik:
see if you can refactor `run_tile_copy` to be shorter and use some functions, but it is not critical, and if code gets less readable, just add a `#[expect(clippy::...)]` to that function

nyurik:
@sdf-jkl let me know if you need help with this

sdf-jkl:
> @sdf-jkl Let me know if you need help with this.

Sorry for the delay - last week was hectic and I was also sick.

I ran into some trouble setting up the local environment for testing and figuring out how to break up the `run_tile_copy` function, since Clippy flags it for being over 100 lines.

Also, since I haven’t gotten the local test environment working yet, I was a bit stuck on `tests.sh`. Running the full test suite on every GitHub commit takes a lot of time, which slowed things down.

I’ve been trying to figure things out on my own to avoid being a bother, since you’ve already helped a lot - I really appreciate it.

nyurik:
totally ok!  I was just wondering if you ran into any blockers.  What issues did you have to run tests locally? I would love to streamline this process as much as possible

sdf-jkl:
I was wondering if I should set things up in a way that allows me to run the workflows locally without relying on GitHub Actions.

I tried running `just coverage --codecov --output-path target/codecov.info` to test things locally, but I kept running into issues.

I also considered simply running `tests.sh`, but that would require setting up a local database instance beforehand in the correct way to run all the tests. (Which I assume is part of a different workflow, to set everything up for the github actions.)

nyurik:
i would recommend simply running `just test`, `just clippy`, and for the most comprehensive - `just ci-test`.  You don't want to run `setup.sh` by hand.  What errors were you getting?

sdf-jkl:
I think I just overcomplicated everything and didn't figure out it was as simple as `just test` etc. 

I will run the tests and write back if I have any problems. 

Thanks for your help!

sdf-jkl:
The code coverage check didn't pass because there are no `tests/expected/martin-cp/` files for the `no-source` test.

~~I believe I was supposed to run `just bless` after the error when comparing the output to the expected files. However, I'm using an ARM64 machine and had trouble getting `sqldiff` to work, so I couldn't proceed to that part of the test. As a result, I wasn't able to update the expected files folder.~~

I simply had to update my computer and `sqldiff` is working now.

The tests went through without errors on my machine and I realized it's because it skips the integration testing part since it's for linux only.
This part in the `justfile`:
```# Run integration tests
test-int: clean-test install-sqlx
    #!/usr/bin/env bash
    set -euo pipefail
    tests/test.sh
    if [ "{{os()}}" != "linux" ]; then
        echo "** Integration tests are only supported on Linux"
        echo "** Skipping diffing with the expected output"
 ...
```
@nyurik How should I generate and add the `expected/martin-cp/` files for the `no-source` test without being able to run the tests?

nyurik:
i just pushed the expected files (please double check).  In some distant future, we should make it run on all platforms... or at least have an easy way to do it with docker

nyurik (approved):
well done, thanks!!

PR #1825, opened 2025-05-12
feat: remove wastefull `Dashmap` from the catalog

CommanderStorm:
Change the wastefull dashmap::Dashmap from the catalog to `std::collections::HashMap`.
We are only using it to return values, so concurency here is just waste.


nyurik (approved):
thanks!

PR #1832, opened 2025-05-17
feat(docker): add labels, pin alpine tag to the latest version

NINNiT:
Just a few minimal `Dockerfile` improvements.

---

Adds the following standardized `Dockerfile` Labels: 
```dockerfile
LABEL org.opencontainers.image.source="https://github.com/maplibre/martin" 
LABEL org.opencontainers.image.licenses="Apache-2.0 OR MIT"
```

I've noticed that some dependency upgrade tools (`renovate`, `dependabot`, ...) can't fetch the martin changelog when creating Bump MRs.  `org.opencontainers.image.source` should alleviate that..
Providing the licenses also doesn't hurt. 

---

Furthermore I've pinned the tag to `latest`, which makes it a bit cleaner. (same behavior as not pinning the tag at all).
Pinning to a fixed version would be better - but without some sort of automation, keeping the tag up-to-date can be a hassle. 

Seems like there are no problems regarding CI image caching however, as `martin` is already using the newest alpine version:

```bash
docker run --entrypoint /bin/sh ghcr.io/maplibre/martin:v0.16.0 -c "cat /etc/os-release""
```
```
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.21.3
PRETTY_NAME="Alpine Linux v3.21"
HOME_URL="https://alpinelinux.org/"
BUG_REPORT_URL="https://gitlab.alpinelinux.org/alpine/aports/-/issues"
```

### DIFF
--- a/.github/files/multi-platform.Dockerfile
+++ b/.github/files/multi-platform.Dockerfile
@@ -7,6 +7,7 @@ LABEL org.opencontainers.image.source="https://github.com/maplibre/martin"
 LABEL org.opencontainers.image.licenses="Apache-2.0 OR MIT"
 LABEL org.opencontainers.image.documentation="https://maplibre.org/martin/"
 LABEL org.opencontainers.image.vendor="maplibre"
+LABEL org.opencontainers.image.authors="Yuri Astrakhan, Stepan Kuzmin and MapLibre contributors"
 
 COPY target_releases/$TARGETPLATFORM/* /usr/local/bin
 
--- a/.github/files/multi-platform.Dockerfile
+++ b/.github/files/multi-platform.Dockerfile
@@ -5,6 +5,8 @@ ARG TARGETPLATFORM
 LABEL org.opencontainers.image.description="Blazing fast and lightweight tile server with PostGIS, MBTiles, and PMTiles support"
 LABEL org.opencontainers.image.source="https://github.com/maplibre/martin"
 LABEL org.opencontainers.image.licenses="Apache-2.0 OR MIT"
+LABEL org.opencontainers.image.documentation="https://maplibre.org/martin/"
+LABEL org.opencontainers.image.vendor="maplibre"
 
 COPY target_releases/$TARGETPLATFORM/* /usr/local/bin
 
--- a/.github/files/multi-platform.Dockerfile
+++ b/.github/files/multi-platform.Dockerfile
@@ -3,7 +3,7 @@ FROM alpine:latest
 ARG TARGETPLATFORM
 
 LABEL org.opencontainers.image.description="Blazing fast and lightweight tile server with PostGIS, MBTiles, and PMTiles support"
-LABEL org.opencontainers.image.source="https://github.com/maplibre/martin" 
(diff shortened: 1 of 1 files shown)


.github/files/multi-platform.Dockerfile
```diff
@@ -1,7 +1,11 @@
-FROM alpine
+FROM alpine:latest
```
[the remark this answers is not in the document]
CommanderStorm:
```suggestion
FROM alpine:3.21
```

.github/files/multi-platform.Dockerfile
```diff
@@ -1,7 +1,11 @@
-FROM alpine
+FROM alpine:3.21
```
CommanderStorm, line 1: we have automatic dependency updates on
=> we get notified if something here is breaking => pinning to minor is imo fine for this one ^^

CommanderStorm (approved):
Thanks for the improvements

.github/files/multi-platform.Dockerfile
```diff
@@ -1,7 +1,11 @@
...
+
 ARG TARGETPLATFORM
 
 LABEL org.opencontainers.image.description="Blazing fast and lightweight tile server with PostGIS, MBTiles, and PMTiles support"
+LABEL org.opencontainers.image.source="https://github.com/maplibre/martin"
+LABEL org.opencontainers.image.licenses="Apache-2.0 OR MIT"
+
```
CommanderStorm, line 11: We can actually add a bit more static metadata:

```suggestion
LABEL org.opencontainers.image.documentation="https://maplibre.org/martin/"
LABEL org.opencontainers.image.vendor="maplibre"

```

.github/files/multi-platform.Dockerfile
```diff
@@ -1,7 +1,13 @@
-FROM alpine
+FROM alpine:3.21
+
 ARG TARGETPLATFORM
 
 LABEL org.opencontainers.image.description="Blazing fast and lightweight tile server with PostGIS, MBTiles, and PMTiles support"
+LABEL org.opencontainers.image.source="https://github.com/maplibre/martin"
+LABEL org.opencontainers.image.licenses="Apache-2.0 OR MIT"
+LABEL org.opencontainers.image.documentation="https://maplibre.org/martin/"
+LABEL org.opencontainers.image.vendor="maplibre"
```
CommanderStorm, line 9: actually, we can also add this label

```suggestion
LABEL org.opencontainers.image.vendor="maplibre"
LABEL org.opencontainers.image.authors="Yuri Astrakhan, Stepan Kuzmin and MapLibre contributors"
```

CommanderStorm:
Actually, I am not sure if we use these labels or completely ignore them:
https://github.com/maplibre/martin/blob/c833034e0c79bcd1e69ea50bda31133e9aee2b2c/.github/workflows/ci.yml#L233-L247

Will have to investigate

CommanderStorm:
(simplest way to investigate is merging and looking at the [`:main` image](https://github.com/maplibre/martin/pkgs/container/martin/413998570?tag=main))

PR #1835, opened 2025-05-19
chore: add more doc-comments

CommanderStorm:
During working through the pg-sourcecode I noticed that these functions did not have doc comments. Now they have.


martin/src/source.rs
```diff
@@ -103,10 +103,15 @@ impl TileSources {
...
+    /// ID under which this [`Source`] is identified if accessed externally
     fn get_id(&self) -> &str;
 
+    /// TileJSON of this [`Source`]
+    ///
+    /// Will be communicated verbatim to the outside to give rendering engines information about the source's contents such as zoom levels, center points, ...
     fn get_tilejson(&self) -> &TileJSON;
 
+    /// Information for serving the source such as which Mime-type to apply or how compression should work
     fn get_tile_info(&self) -> TileInfo;
```
sharkAndshark, line 115: I thought they were self-documenting?

nyurik: i think it is better to be explicit in docs, thx @CommanderStorm!  this may only be a concern when implementing a trait for a specific struct - in that case it would be mostly dups

nyurik (approved):
lgtm, thx!

PR #1834, opened 2025-05-19
feat: basic etag handling for tile serving

CommanderStorm:
This PR adds basic etag handling to the tile serving.

I am currently using an okay hash function for this.
There [are better ones like `gxhash`](https://github.com/ogxd/gxhash?tab=readme-ov-file#benchmarks), but for compatability with the `mbtiles`-crate, using `xxhash` is likely also fine. Also, `xxhash` does not require SIMD.

<details><summary>benchmarks (click to expand)</summary>
<p>

<table>
<tr>
<td>this branch</td>
<td>main</td>
</tr>
<tr>
<td>

```raw
oha --latency-correction -z 5s --no-tui http://localhost:3000/function_zxy_query/18/235085/122323 > /dev/null
oha --latency-correction -z 60s         http://localhost:3000/function_zxy_query/18/235085/122323
Summary:
  Success rate: 100.00%
  Total:        60.0012 secs
  Slowest:      0.0102 secs
  Fastest:      0.0000 secs
  Average:      0.0003 secs
  Requests/sec: 165899.5459

  Total data:   1.39 GiB
  Size/request: 150 B
  Size/sec:     23.73 MiB

Response time histogram:
  0.000 [1]       |
  0.001 [9880415] |■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■
  0.002 [71916]   |
  0.003 [1465]    |
  0.004 [255]     |
  0.005 [70]      |
  0.006 [19]      |
  0.007 [9]       |
  0.008 [1]       |
  0.009 [8]       |
  0.010 [9]       |

Response time distribution:
  10.00% in 0.0001 secs
  25.00% in 0.0002 secs
  50.00% in 0.0003 secs
  75.00% in 0.0004 secs
  90.00% in 0.0005 secs
  95.00% in 0.0007 secs
  99.00% in 0.0010 secs
  99.90% in 0.0015 secs
  99.99% in 0.0023 secs


Details (average, fastest, slowest):
  DNS+dialup:   0.0013 secs, 0.0004 secs, 0.0025 secs
  DNS-lookup:   0.0001 secs, 0.0000 secs, 0.0020 secs

Status code distribution:
  [200] 9954168 responses

Error distribution:
  [4] aborted due to deadline
oha --latency-correction -z 5s --no-tui http://localhost:3000/png/0/0/0 > /dev/null
oha --latency-correction -z 60s         http://localhost:3000/png/0/0/0
Summary:
  Success rate: 100.00%
[...]

PR #1836, opened 2025-05-19
feat: add simpler etag handling to all non-critical endpoints

CommanderStorm:
This PR adds etag handling via an actix wrap instruction to all "non-tile" endpoints.
I ommitted some endpoints which I suspect are not called more than once such as `/` if the webui is not compiled in.

The reason for not being smarter for some is that honestly, the impact of this is less than I expected based on benchmarks in #1834 and that `actix-middleware-etag` uses the same hash algorithm.

<del>
I have not addded tests for this as frankly, this does not seem like a feature that needs deep tests.
</del>
By request, this PR now includes all stemi-static headers.


martin/src/srv/server.rs
```diff
@@ -105,7 +105,8 @@ async fn get_health() -> impl Responder {
     "/catalog",
     method = "GET",
     method = "HEAD",
-    wrap = "middleware::Compress::default()"
+    wrap = "middleware::Compress::default()",
+    wrap = "actix_middleware_etag::Etag::default()"
```
nyurik: please add a `use actix_middleware_etag::Etag;` everywhere, and also a trailing comma for all - to avoid churn

CommanderStorm: > also a trailing comma for all - to avoid churn

tell that to `cargo fmt`.
I don't see an option for this here too: https://rust-lang.github.io/rustfmt/

CommanderStorm:
I don't super like the importing part as this makes `headers::ETag` and `actix_middleware_etag::Etag` indistinguisable, but am ok with doing it like this.

martin/src/srv/fonts.rs
```diff
@@ -18,7 +19,8 @@ struct FontRequest {
 #[route(
     "/font/{fontstack}/{start}-{end}",
     method = "GET",
-    wrap = "middleware::Compress::default()"
+    wrap = "middleware::Compress::default()",
+    wrap = "Etag"
```
nyurik: a bit weird here - the above wrap is using `::default()`, but this one does not?  Would it make sense to add both `Compress` and `Etag` to the `use` statements at the top, and have a compact `wrap = "Compress", wrap = "Etag"` ?

nyurik (approved):
there is a minor nit, and feel free to merge after addressing it

martin/src/srv/fonts.rs
sharkAndshark: And what's the order here?  
> 
```rust
let app = App::new()
    .wrap(MiddlewareA::default())
    .wrap(MiddlewareB::default())
    .wrap(MiddlewareC::default())
    .service(service);
```
> The request first gets processed by the middleware specified last - MiddlewareC. So the chain is C -> B -> A

1 or 2?

1. Etag -> Compress
2. Compress -> Etag

Note: The Etag must be loaded as the last middleware.

sharkAndshark:
> I have not addded tests for this as frankly, this does not seem like a feature that needs deep tests

I think a really simple test for this is still necessary. Maybe next PR.

martin/src/srv/fonts.rs
sharkAndshark: A second thought: If the ordering is 1(Etag -> Compress),Comparing the etag before compressing seems Ok but seems a little bit tricky?

CommanderStorm: No clue how this is handled on the way out of the request.
Correctly, ertags should be checked first on the way out, then compression as that enables the fast-path there.

The docs on arctic side are a bit unclear on this.

Will have to benchmark which one is faster

CommanderStorm: > above wrap is using ::default(), but this one does not

Yes, clippy complained that I am using default on a unit struct. Compression does have settings => committing this does not work.

CommanderStorm:
> I think a really simple test for this is still necessary.

I added a testcase for this. Whas what I added what you thought of?

martin/src/srv/fonts.rs
CommanderStorm: Have resorted the wraps as suggested.
Benchmarking shows a barely measurable improvement. Compressing so small thigns is inherently fast.

For future reference:

image

.tmp
sharkAndshark, line 1: Seems we need to remove this empty file

tests/test.sh
```diff
@@ -94,15 +94,15 @@ test_jsn() {
 
   echo "Testing $(basename "$FILENAME") from $URL"
   # jq before 1.6 had a different float->int behavior, so trying to make it consistent in all
-  $CURL "$URL" | jq --sort-keys -e 'walk(if type == "number" then .+0.0 else . end)' > "$FILENAME"
+  $CURL  --dump-header  "$FILENAME.headers" "$URL" | jq --sort-keys -e 'walk(if type == "number" then .+0.0 else . end)' > "$FILENAME"
```
sharkAndshark, line 97: Maybe do a test like we did under the `martin/tests`, and sure, it's good to test it in `test.sh` either. 

But dump the header for each request it's not a good idea I guess...

nyurik: at first I also thought headers might be too flaky, but than it seems headers only have the relatively stable things... with the possible exception of the length? i.e. in theory we could have a slightly different compression between different brotli implementation versions?  So lets add it, and if we ever run into issues with the test flakiness, we could  filter the header files to remove unstable headers?

tests/test.sh
```diff
@@ -613,4 +613,13 @@ fi
 
 rm -rf "$TEST_TEMP_DIR"
 
+# now we need to strip the date header as it is undeterministic
+find -type f -name "*.headers" -exec sed --regexp-extended --in-place "s/date: .+//" {} \;
```
nyurik: ah, silly me, this step here is why you only have the stable things. My only minor gripe is that we have to use `find` here. Instead, lets create a dedicated `curl_with_headers $URL $FILE` function that downloads URL into a FILE, and also saves FILE.headers and removes unneeded stuff from it.  This way if there is an error, the header files will be in the right state right away, and whoever is debugging it won't get confused why expected and resulting files are different

nyurik: P.S. For this line, you can do fixes after the fact, i.e. instead of this:

```
$CURL  --dump-header  "$FILENAME.headers" "$URL" | jq --sort-keys -e 'walk(if type == "number" then .+0.0 else . end)' > "$FILENAME"
```

do this:

```
curl_with_headers "$URL" "$FILENAME"
cat "$FILENAME" jq --sort-keys -e 'walk(if type == "number" then .+0.0 else . end)' > "${FILENAME}__tmp"
mv "$FILENAME__tmp" "$FILENAME"
```

CommanderStorm: I think `curl_with_headers` is a bit too non-obvious in what it is doing.

I changed it to the following instead.

```sh
$CURL --dump-header  "$FILENAME.headers" "$URL" > "$FILENAME"
clean_headers_dump "$FILENAME.headers"
```

tests/expected/configured/pmt2_0_0_0.png.txt
```diff
@@ -1 +1 @@
-tests/output/configured/pmt2_0_0_0.png: RIFF (little-endian) data, Web/P image
+tests/output/configured/pmt2_0_0_0.png: RIFF (little-endian) data, Web/P image, with alpha, 511+1x511+1
```
CommanderStorm: my local environment and CI disagree if this file have alpha or not.

No clue why, likely a newer version of this detector.
In any case an unrelated bug.

```suggestion
tests/output/configured/pmt2_0_0_0.png: RIFF (little-endian) data, Web/P image
```

tests/test.sh
CommanderStorm: responding to this explicitely:
- If length changes, then the etag changes as well.
- If encoding changes, then the etag and length changes as well.

We cannot have both tests that
- ensure that the etag does not change and
- not check that the etag changes.

But since we have PRs for this kind of dependency changes, this should be fine.

This is not like the build in main being broken by new cippy lints every few months.

nyurik (approved):
i think this is good to go, thanks!

CommanderStorm:
Nice

PR #1837, opened 2025-05-19
Update run-with-docker-compose.md to support styles serving

Andreal2000:
I have lost a couple hours this morning trying to serve my styles, after tracing your pull request where you add this features i find out that i was using the wrong version of martin (copied from this file of the docs).

Maybe you could add a conf file validation when the server start, to check if there are some incompatible values.

### DIFF
--- a/docs/src/run-with-docker-compose.md
+++ b/docs/src/run-with-docker-compose.md
@@ -6,7 +6,7 @@ file as a reference
 ```yml
 services:
   martin:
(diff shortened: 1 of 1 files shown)


CommanderStorm (approved):
Thanks a ton and sorry 🥵

CommanderStorm:
> Maybe you could add a conf file validation when the server start, to check if there are some incompatible values

I thought we had this - apparently this is currently broken and does not send a warning 👀

PR #1814, opened 2025-04-30
Behave like a 3857 tile service when google compatible flag found in COG file

sharkAndshark:
Not an actual PR but a concept prove(Currently it's really messy). But if it works, it might be split to many PRs to make review easy.
Trying to fix #1820

PR #1840, opened 2025-05-20
Add get origin func to COG

### LINKED ISSUE #1820: Make COG web friendly
[WebMercatorQuad](https://docs.ogc.org/is/17-083r2/17-083r2.html#72) is the most popular tile grid. Almost all clients (maplibre, OL, Mapbox, etc) take it as default.

It would be better if martin could support it like:

## Configuration

```yml
cog:
  # default false. 
  # When enabled, martin would take the requested tile index(/{z}/{x}/{y}) as tile index of webmercatorquad, and martin will finish the clipping and warping internally.
  # It only works when the projection of COG file it's 3857
  web_friendly: true
```

## Tilejson

Martin would create a mapping between the inner custom tilegrid of cog file and the WebMercatorQuad.

The min/max zoom  would be updated then.

`/cog_source`

** **

```jsonc
{
//  "maxzoom": 3, Before web_friendly enabled
//  "minzoom": 0,  Before web_friendly enabled
 "maxzoom": 14,   // after web_friendly enabled
 "minzoom":  11,  // after web_friendly enabled
  "tilejson": "3.0.0",
  "tiles": [
    "http://localhost:3111/cog_source/{z}/{x}/{y}"
  ]
}
```

No any tilejson Compatibility got broken here and the most popular projection and tilegrid got supported then. 

[Titiler support WebMercatorQuad for COG either and even more custom tilegrid can be added](https://developmentseed.org/titiler/user_guide/tile_matrix_sets/), and its enpoint is flexible like `/cog/tiles/{tileMatrixSetId}/{z}/{x}/{y}[@{scale}x][.{format}]`, see https://developmentseed.org/titiler/endpoints/cog/#api

sharkAndshark:
It's the 2nd PR of #1814  to fix #1820 

A func to get the origin coord from COG is added. It's highly inspired by [geotiff.js](https://github.com/geotiffjs/geotiff.js) and the content of [B.6 of geotiff spec](https://docs.ogc.org/is/19-008r4/19-008r4.html#_geotiff_tags_for_coordinate_transformations)


martin/src/cog/source.rs
```diff
@@ -23,6 +23,7 @@ struct Meta {
     min_zoom: u8,
     max_zoom: u8,
     model: ModelInfo,
+    origin: [f64; 3],
```
CommanderStorm, line 27: Could you add doc comments explaining what the origin is?
This would be very handy as origin as 3*f64 is not quite self-explanatory.

martin/src/cog/errors.rs
```diff
@@ -52,4 +52,9 @@ pub enum CogError {
         "The pixel size of the image {0} is not squared, the x_scale is {1}, the y_scale is {2}"
     )]
     NonSquaredImage(PathBuf, f64, f64),
+
+    #[error(
+        "Get origin failed for {0}: the length of ModelTiepointTag should be >= 6, or the length of ModelTransformationTag should be >= 12"
+    )]
+    GetOriginFailed(PathBuf),
```
CommanderStorm, line 60: This requiring another server major bump is another case where decoupling martin and it's libs semver would be handy.

Regarding this: all error types should likely be marked as no eyhausive.

martin/src/cog/errors.rs
```diff
@@ -52,4 +52,9 @@ pub enum CogError {
         "The pixel size of the image {0} is not squared, the x_scale is {1}, the y_scale is {2}"
     )]
     NonSquaredImage(PathBuf, f64, f64),
+
+    #[error(
+        "Get origin failed for {0}: the length of ModelTiepointTag should be >= 6, or the length of ModelTransformationTag should be >= 12"
```
CommanderStorm: Would this be correct?

```suggestion
        "Calculating the tile origin failed for {0}: the length of ModelTiepointTag should be >= 6, or the length of ModelTransformationTag should be >= 12"
```

martin/src/cog/source.rs
```diff
@@ -521,4 +560,34 @@ mod tests {
...
+        None,Some(vec![
+            0.0, 100.0, 0.0, 400_000.0, 100.0, 0.0, 0.0, 500_000.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0,
+            0.0, 1.0,
+        ]),
+        [400_000.0, 500_000.0, 0.0]
+    )]
+    fn can_get_origin(
+        #[case] tie_point: Option<Vec<f64>>,
+        #[case] matrix: Option<Vec<f64>>,
+        #[case] expected: [f64; 3],
```
CommanderStorm: Not nesssesary, but I think adding a test cases for the error path would be very nice ^^

```suggestion
        #[case] expected: Option<[f64; 3]>,
```

CommanderStorm (approved):
If you add some documentation in the form if doc comments what an origin is, I am fine with this.
I honestly don't have a good understanding what an origin is.

martin/src/cog/source.rs
sharkAndshark: Hmm. I will try.  Actually it's the coords: raster_space_to_model_space(0,0,0), I'm not good at this, maybe you have better one? 
I add this to:
```rust
    // The geo coords of pixel(0, 0, 0) ordering in [x, y, z]
    origin: [f64; 3],
```

martin/src/cog/errors.rs
sharkAndshark: let's bump the version when we finish this feature

PR #1845, opened 2025-05-22
Add get_extent to COG

### LINKED ISSUE #1820: Make COG web friendly

sharkAndshark:
It's the 3rd PR of #1814  to fix #1820 

A func to get the extent from COG is added. It's highly inspired by [geotiff.js](https://github.com/geotiffjs/geotiff.js) and the content of [B.6 of geotiff spec](https://docs.ogc.org/is/19-008r4/19-008r4.html#_geotiff_tags_for_coordinate_transformations)


martin/src/cog/source.rs
```diff
@@ -456,6 +480,76 @@ fn get_origin(
...
+            if matrix[1] == 0.0 && matrix[4] == 0.0 {
+                Ok([matrix[0], matrix[5], matrix[10]])
+            } else {
+                let x_res = (matrix[0] * matrix[0]) + (matrix[4] * matrix[4]);
+                let y_res = ((matrix[1] * matrix[1]) + (matrix[5] * matrix[5])).sqrt() * -1.0;
+                let z_res = matrix[10];
+                Ok([x_res, y_res, z_res])
+            }
+        }
+        (None, None) => Err(CogError::GetFullResolutionFailed(path.to_path_buf())),
```
sharkAndshark: More comments would be added here

martin/src/cog/source.rs
```diff
@@ -456,6 +480,76 @@ fn get_origin(
...
+    path: &Path,
+) -> Result<[f64; 3], CogError> {
+    match (pixel_scale, transformation) {
+        // ModelPixelScaleTag = (ScaleX, ScaleY, ScaleZ)
+        (Some(scale), _) => Ok([scale[0], scale[1], scale[2]]),
+        (_, Some(matrix)) => {
+            if matrix[1] == 0.0 && matrix[4] == 0.0 {
+                Ok([matrix[0], matrix[5], matrix[10]])
+            } else {
+                let x_res = (matrix[0] * matrix[0]) + (matrix[4] * matrix[4]);
```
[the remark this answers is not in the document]
sharkAndshark: thx! AI! I didn't noticed that. I would review and test this part more carefully.

martin/src/cog/source.rs
```diff
@@ -456,6 +480,76 @@ fn get_origin(
...
+    (full_width, full_height): (f64, f64),
+) -> [f64; 4] {
+    if let Some(matrix) = transformation {
+        let corners = [
+            [0, 0],
+            [0, full_height_pixel],
+            [full_width_pixel, 0],
+            [full_width_pixel, full_height_pixel],
+        ];
+        let transformed = corners.map(|pixel| {
```
CommanderStorm: the min/max code below is a tad bit weird.

I think just iterating over transformed and using the min/max functions instead of below would be much simpler.

martin/src/cog/source.rs
```diff
@@ -456,6 +480,82 @@ fn get_origin(
...
+            x_res = x_res.copysign(matrix[0]);
+            let mut y_res =
+                (matrix[1] * matrix[1] + matrix[5] * matrix[5] + matrix[9] * matrix[9]).sqrt();
+            println!("martix[5]: {}", matrix[5]);
+            // A positive y_res indicates that model space Y cordinates decrease as raster space J indices increase. This is the standard vertical relationship between raster space and model space
+            y_res = y_res.copysign(-matrix[5]);
+            let mut z_res =
+                (matrix[2] * matrix[2] + matrix[6] * matrix[6] + matrix[10] * matrix[10]).sqrt();
+            z_res = z_res.copysign(matrix[10]);
+            Ok([x_res, y_res, z_res])
```
sharkAndshark, line 503: ## How to calculate them based on the transformation

Define 3 functions "raster2model(x,y,z) based on the transformation matrix, which return the coords in model space.

|||
|----|----|
|raster2model_x|`I*a + J*b + K*c + 1*d`|
|raster2model_y|`I*e + J*f + K*g + 1*h`|
|raster2model_z|`I*i + J*j + K*k + 1*l`|

```raw
coords =     matrix  * coords
|- -|     |-       -|  |- -|
| X |     | a b c d |  | I |
| | |     |         |  |   |
| Y |     | e f g h |  | J |
|   |  =  |         |  |   |
| Z |     | i j k l |  | K |
| | |     |         |  |   |
| 1 |     | m n o p |  | 1 |
|- -|     |-       -|  |- -|
```

## x_res

Its' the length from [0,0,0] to [1,0,0].

x_res = sqrt(x^2+y^2+z^2)

1. x == raster2model_x(1,0,0) - raster2model_x(0,0,0) => a
2. y == raster2model_y(1,0,0) - raster2model_y(0,0,0) => e 
3. z == raster2model_z(1,0,0) - raster2model_z(0,0,0) => i


x_res = sqrt(a^2+e^2+i^2)

**But the sign?**

> A positive ScaleX in the ModelPixelScaleTag SHALL indicate that model space X coordinates increase as raster space I indices increase. This is the standard horizontal relationship between raster space and model space. 

So if a < 0 then x_res *= -1.0;

## y_res

Its' the length from [0,0,0] to [0,1,0].

y_res = sqrt(x^2+y^2+z^2)

1. x == raster2model_x(0,1,0) - raster2model_x(0,0,0) => b
2. y == raster2model_y(0,1,0) - raster2model_y(0,0,0) => f 
3. z == raster2model_z(0,1,0) - raster2model_z(0,0,0) => j

y_res = sqrt(b^2+f^2+j^2)

**But the sign?**

> A positive ScaleY in the ModelPixelScaleTag SHALL indicate that model space Y coordinates decrease as raster space J indices increase. This is the standard vertical relationship between raster space and model space

So if f > 0 then y_res *= -1.0;

## z_res

Its' the length from [0,0,0] to [0,0,1].

z_res = sqrt(x^2+y^2+z^2)

1. x == raster2model_x(0,0,1) - raster2model_x(0,0,0) => c
2. y == raster2model_y(0,0,1) - raster2model_y(0,0,0) => g 
3. z == raster2model_z(0,0,1) - raster2model_z(0,0,0) => k


z_res = sqrt(c^2+g^2+k^2)

**But the sign?**

> The ScaleZ is primarily used to map the pixel value of a digital elevation model into the correct Z-scale (in other words a Z-Scaling factor).

I'm not very sure this part as the spec not make it very clear. But I suppose **model space Z coordinates increase as raster space K indices increase**.

So if k < 0 then z_res *= -1.0;

sharkAndshark: guys I need to add more tests to this PR. But almost every geotiff file use `pixelscale` and `tiepoints` to define their raster to model. I need made many by myself. I think this is import to test this carefully. I need more time here.

martin/src/cog/source.rs
```diff
@@ -456,6 +480,79 @@ fn get_origin(
...
+        let max_y = transformed
+            .iter()
+            .map(|(_, y)| *y)
+            .max_by(f64::total_cmp)
+            .expect("corners has >1 element => max exists");
+        let min_y = transformed
+            .iter()
+            .map(|(_, y)| *y)
+            .min_by(f64::total_cmp)
+            .expect("corners has >1 element => min exists");
```
CommanderStorm: @sharkAndshark what do you think about this instead? (I pushed some code changing this to `max_by`)

martin/src/cog/source.rs
```diff
@@ -456,6 +480,79 @@ fn get_origin(
...
+
+fn raster2model(i: u32, j: u32, matrix: &[f64]) -> (f64, f64) {
+    let i = f64::from(i);
+    let j = f64::from(j);
+    let x = matrix[3] + (matrix[0] * i) + (matrix[1] * j);
+    let y = matrix[7] + (matrix[4] * i) + (matrix[5] * j);
+    (x, y)
+}
+
+fn get_extent(
```
CommanderStorm, line 528:
```suggestion
/// Computes an `[min_x, min_y, max_x, max_y]` tuple and returns the geographic extent (bounding box) of the given spatial layer. 
///
/// Applies a transformation matrix to corner pixels if provided;
/// otherwise, computes extent from origin and raster size in model units.
///
/// # Parameters
/// - `origin`: Top-left origin `[x, y, z]` in model space.
/// - `transformation`: Optional pixel-to-model transformation matrix.
/// - `full_width_pixel`, `full_height_pixel`: Raster dimensions in pixels.
/// - `full_width`, `full_height`: Raster size in model units.
fn get_extent(
```

CommanderStorm (approved):
from my side, we could btw ship it.

martin/src/cog/source.rs
sharkAndshark: thx! I've made small updates based on your suggestion

martin/src/cog/source.rs
CommanderStorm: @sharkAndshark
Except adding a doc comment for this method (and maybe the other two also), I am fine with this.
Does not have to be such a long doc comment

```suggestion
/// Computes an `[min_x, min_y, max_x, max_y]` tuple and returns the geographic extent (bounding box) of the given spatial layer. 
///
/// Applies a transformation matrix to corner pixels if provided;
/// otherwise, computes extent from origin and raster size in model units.
fn get_extent(
```

sharkAndshark:
thx! @CommanderStorm

Issue #1841, opened 2025-05-20
Listing all tiles in an Mbtiles file


Murtaught:
Hello!

I'm currently working a tile downloading program, and I want to use `mbtiles` crate for reading a writing `.mbtiles` files in addition to individual tiles stored on the filesystem in `{z}/{x}/{y}.{ext}`-structure.

As far as I know, `mbtiles` crate is the most feature-complete one for this task in the Rust ecosystem as of now. Thank you for publishing it!

I've noticed there is currently no way to get the full list of tiles stored in a particular file. I can read the tile with `Mbtiles::get_tile()` method if I know its coordinates, but if I want to know what tiles are available, then there is no way for doing that. Am I correct? Maybe I'm missing something?

I have [a draft of an implementation](https://github.com/Murtaught/martin/commit/fb9a075b24a2e1b9272ea6a45162af01f54cc32e) of this feature in my fork of the project.
Would you be interested in a PR?
Should I change anything? (Beyond writing a proper test for `Mbtiles::all_tile_coords()`)

I've tested it with a file full of raster tiles, but if I correctly understand how `.mbtiles` are structured, it should also work fine with vector data.

CommanderStorm:
Listing the avaliable tiles would imo be a fair addition and I really like your impl.

Though, we would need a quick testcase to ensure that we don't accidentaly regress on this (should be fairly simple though - calling this method on one of the files in `tests/fixtures` and asserting the results shoudl be enogh).

CommanderStorm:
Let us know if there is something missing in said crate.
We currently have only optimised it to what we need as a tile server (or the functionality here: https://maplibre.org/martin/mbtiles.html)

nyurik:
I agree, we should have a way to iterate over all available tiles. Moreover, I think we need two modes - with and without the tile data (i.e. coordinates only).

Murtaught:
@CommanderStorm  @nyurik Thank you both for very quick response!

> Listing the avaliable tiles would imo be a fair addition and I really like your impl.
> Though, we would need a quick testcase to ensure that we don't accidentaly regress on this

Great, I'll figure out the tests and prepare a PR. Should not take more than a few days.

> Let us know if there is something missing in said crate.

Sure. I'm figuring this out myself as I go. 
[...]

PR #1850, opened 2025-05-24
feat: add `Mbtiles::contains`

CommanderStorm:
a bit of the pool testcase changes are a bit off-topic from the other changes there.
Have moved them to another PR => Should be reviewed after
- https://github.com/maplibre/martin/pull/1851

Otherwise should be rather self-explainatory.

CC @Murtaught does this improve the usecase you mentioned in https://github.com/maplibre/martin/issues/1841 ?

### DIFF
--- a/mbtiles/src/mbtiles.rs
+++ b/mbtiles/src/mbtiles.rs
@@ -345,16 +345,15 @@ impl Mbtiles {
         let table = match mbt_type {
             MbtType::Flat => "tiles",
             MbtType::FlatWithHash => "tiles_with_hash",
-            MbtType::Normalized { hash_view: _ } => "map",
+            MbtType::Normalized { .. } => "map",
         };
         let sql = format!(
             "SELECT 1 from {table} where zoom_level = ? AND tile_column = ? AND tile_row = ?"
         );
-        let y = invert_y_value(z, y);
         let row = query(&sql)
             .bind(z)
             .bind(x)
-            .bind(y)
+            .bind(invert_y_value(z, y))
             .fetch_optional(conn)
             .await?;
         Ok(row.is_some())
--- a/mbtiles/src/mbtiles.rs
+++ b/mbtiles/src/mbtiles.rs
@@ -319,21 +319,9 @@ impl Mbtiles {
 
     /// Check if a tile exists in the database.
     ///
-    /// This is a slight optimisation over [`Mbtiles::get_tile_and_hash`] or [`Mbtiles::get_tile`] and does only look up IF the tile exists, not the tile data.
-    ///
-    /// <div class="warning">
-    ///
-    /// **Note:**
-    /// You usually (unless the second query is unnecessary) want to retrieve the tile data if it exists instead.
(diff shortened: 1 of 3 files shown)


Murtaught:
This looks great!

I wanted to send another PR for this tomorrow, but you beat me to it. And you did a better job with doc comments, I really like them 👍🏻

mbtiles/src/mbtiles.rs
```diff
@@ -317,6 +317,49 @@ impl Mbtiles {
...
+    /// You usually (unless the second query is unnecessary) want to retrieve the tile data if it exists instead.
+    /// Sqlite already checks for a tiles existence and retrieving it if it exists in [`Mbtiles::get_tile_and_hash`] or [`Mbtiles::get_tile`].
+    ///
+    /// </div>
+    ///
+    /// *Tipp:*
+    /// If you need even more performance in this specific operation, consider scanning over the entire table once via [`Mbtiles::stream_coords`] and building a bloom filter (f.ex. [`fastbloom`](https://docs.rs/fastbloom/)) as a fast path.
+    /// This only works if
+    /// - the database is static or
+    /// - you maintain the bloom filter on inserts and deletes to the database.
```
nyurik: The simpler the better. I don't think we need to go into the bloom filter here, regardless of how cool that thing is.  I.e. less is more approach :)

```suggestion
    /// Check if a tile exists in the database.
    ///
    /// This method is slightly faster than [`Mbtiles::get_tile_and_hash`] and [`Mbtiles::get_tile`]
    /// because it only checks if the tile exists but does not retrieve tile data.
    /// Most of the time you would want to use the other two functions.
```

nyurik: P.S. similar change for the other new method

mbtiles/src/pool.rs
```diff
@@ -68,19 +68,43 @@ impl MbtilesPool {
...
+    /// You usually (unless the second query is unnecessary) want to retrieve the tile data if it exists instead.
+    /// Sqlite already checks for a tiles existence and retrieving it if it exists in [`MbtilesPool::get_tile_and_hash`] or [`MbtilesPool::get_tile`].
+    ///
+    /// </div>
+    ///
+    /// *Tipp:*
+    /// If you need even more performance in this specific operation, consider scanning over the entire table once via [`Mbtiles::stream_coords`] and building a bloom filter (f.ex. [`fastbloom`](https://docs.rs/fastbloom/)) as a fast path.
+    /// This only works if
+    /// - the database is static or
+    /// - you maintain the bloom filter on inserts and deletes to the database.
```
CommanderStorm:
```suggestion
    /// Check if a tile exists in the database.
    ///
    /// This method is slightly faster than [`Mbtiles::get_tile_and_hash`] and [`Mbtiles::get_tile`]
    /// because it only checks if the tile exists but does not retrieve tile data.
    /// Most of the time you would want to use the other two functions.
```

nyurik (approved):
thanks!

PR #1853, opened 2025-05-26
Ignore `DANGER_ACCEPT_INVALID_CERTS` env var

nyurik:
Remove any special handling of the obsolete `DANGER_ACCEPT_INVALID_CERTS` env var. It was used a while ago, but I don't think we have any usage of it now.


CommanderStorm (approved):
Good call

PR #1854, opened 2025-05-26
improve fmt and related just commands

nyurik:
* consolidated `just fmt` will format using nightly (preferred) or regular `cargo fmt`. This is identical to what we had before as `just fmt2`
* remove unused rustfmt.toml
* reformatted everything


CommanderStorm (approved):
Nice, thanks

PR #1856, opened 2025-05-26
fix(perf): clippy issues in newer rust releases

### LINKED ISSUE #1855: Optimize error enum size
The new clippy version reports an issue with our large errors - see [`result_large_err` docs](https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err).  The error exceeds 400 bytes -- and from the lints' explanation:

>  A Result is at least as large as the Err-variant. While we expect that variant to be seldom used, the compiler needs to reserve and move that much memory every single time. Furthermore, errors are often simply passed up the call-stack, making use of the ?-operator and its type-conversion mechanics. If the Err-variant further up the call-stack stores the Err-variant in question (as library code often does), it itself needs to be at least as large, propagating the problem.

I think one of the biggest culprit is the PMTiles error, but others might contribute as well.  cc: @lseelenbinder

CommanderStorm:
This PR updates `pmtiles` which reduces this specific error variant.
Some clippy lints regarding docs were also fixed

Resolves #1855

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -84,6 +84,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "actix-middleware-etag"
+version = "0.4.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bf947271526f363aba604e2a735fb2792f63768c6ebdd06a30bf6df74a5ba06c"
+dependencies = [
+ "actix-service",
+ "actix-web",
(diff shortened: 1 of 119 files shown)


CommanderStorm:
Needs https://github.com/stadiamaps/pmtiles-rs/pull/53 merged and `pmtiles=0.13` released to be mergable.

nyurik (approved):
looks good, thx

Issue #1857, opened 2025-05-26
martin=0.17.0 release


CommanderStorm:
this issue is for discussing what needs to be done for a release.

> [!NOTE]
> This is not the v1.0 release-talk. For this, we have https://github.com/maplibre/martin/discussions/1805

> [!TIP]
> as always, the `:master`-tag is avaliable if somebody needs for example the s3-changes currently
>
> This may be interesting for @Youssef-Harby 
> Thanks for your work on making this happen ❤

# Outstanding issues/PRs

Not nessesary, but would be nice:
- to properly anoucne this as a feature ^^
  https://github.com/maplibre/martin/pull/1836
- seems to be basically done, with minor follow ups still needed
  https://github.com/maplibre/martin/pull/1815

Must-have:
- https://github.com/maplibre/martin/pull/1856
- https://github.com/maplibre/martin/pull/1848
- https://github.com/maplibre/martin/pull/1859

# Changelog

url to see what is not included: https://github.com/maplibre/martin/compare/506627d63f85a352d325fe219b8c4597d9f00aa1...HEAD

```
## What's Changed

### using pmtiles via S3

We now support storing your pmtiles in S3 or S3-compatible blob stores and serving them.
For further information on this feature, please [see our documentation](https://maplibre.org/martin/sources-files.html#pmtiles-s3).

> [!NOTE]
> There is work outstanding upstream to extend PMtiles support to more than blob stores other than S3.
> If you would like to contribute on this, please see https://github.com/stadiamaps/pmtiles-rs/issues/42

Done by @alamminsalo and @Youssef-Harby in https://github.com/maplibre/martin/pull/1477

### `ETag` header handling

> [!IMPORTANT]
> An `ETag` offers a way for the client to validate that the tile is still the same.
> In prior versions, we did not have `ETag`s.
> In theory there could be a difference in behavior with regards to tile refreshes, how CDNs/Reverse proxys cache content or how you do accounting if you are a SAAS.

This optimisation enables
- CDN-style caching of requests and
- may reduce network bandwith and time spend downloading tiles a bit

It is particularly noticable in usecases with slow networking.

What we did:
We added [`ETag`-header](https://developer.mozilla.org/de/docs/Web/HTTP/Reference/Headers/ETag) handling to martin.
If your device caches tiles, it will attach an `If-None-Match`-header with the expected `ETag`.
[...]

PR #1859, opened 2025-05-27
chore: update transitive dependencys

CommanderStorm:
This PR runs the standard dependency update run before
- #1857


nyurik:
Maybe lets do rust deps separate from npm?  Not sure why things are failing

nyurik (approved):

PR #1867, opened 2025-06-02
A few more justfile cleanups

nyurik:
Nothing big, but added a check to ensure that the repo is not dirty after tests, and minor nits

### DIFF
--- a/justfile
+++ b/justfile
@@ -1,25 +1,39 @@
 #!/usr/bin/env just --justfile
 
-set shell := ["bash", "-c"]
+set shell := ['bash', '-c']
 
-#export DATABASE_URL="postgres://postgres:postgres@localhost:5411/db"
+
+main_crate := 'martin'
+
+#export DATABASE_URL='postgres://postgres:postgres@localhost:5411/db'
 
 # Set additional database connection parameters, e.g.   just  PGPARAMS='keepalives=0&keepalives_idle=15'  psql
-PGPARAMS := ""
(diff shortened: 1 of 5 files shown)


justfile
```diff
@@ -1,26 +1,40 @@
...
+# if running in CI, treat warnings as errors by setting RUSTFLAGS and RUSTDOCFLAGS to '-D warnings' unless they are already set
+# Use `CI=true just ci-test` to run the same tests as in GitHub CI.
+# Use `just env-info` to see the current values of RUSTFLAGS and RUSTDOCFLAGS
+ci_mode := if env('CI', '') != '' {'1'} else {''}
+export RUSTFLAGS := env('RUSTFLAGS', if ci_mode == '1' {'-D warnings'} else {''})
+export RUSTDOCFLAGS := env('RUSTDOCFLAGS', if ci_mode == '1' {'-D warnings'} else {''})
+export RUST_BACKTRACE := env('RUST_BACKTRACE', if ci_mode == '1' {'1'} else {''})
+
+alias _down := stop
+alias _stop-db := stop
```
CommanderStorm: Could you add a comment above what this does? 🤔
Assigning twice seems a bit weird.

CommanderStorm (approved):
No complaints 😉

justfile
nyurik: these were simply moved from below.  I guess we should just delete them

nyurik:
```suggestion
```

Issue #1846, opened 2025-05-22
Generate test mbtiles on the fly rather than storing them


nyurik:
We currently have a number of .mbtiles files checked in for various testing. Some of them are also used for documentation / quickstart guides.  I think we should rework this for security and readabilty reasons, with the following actionables:

* [x] Create several schema initialization .sql files, e.g. in `mbtiles/sql/` dir, add `init-flat.sql`, `init-flat-with-hash.sql`, etc.
  * [x] Use content of these files directly in the [docs](https://github.com/maplibre/martin/blob/428265025e0c76ba4a1919e0f9cbe7618137a4d1/docs/src/mbtiles-schema.md?plain=1#L10-L17), rather than copy/paste it. See [`{{#include file.rs}}`](https://rust-lang.github.io/mdBook/format/mdbook.html?highlight=include#including-files) syntax
  * [ ] Use these files in the [test.sh](https://github.com/maplibre/martin/blob/main/tests/test.sh) before populating them with data
  * [x] Use these files in the unit tests
  * [x] Update docs to use these files with the `sqlite` command to create a blank .mbtiles file
* [ ] Export all existing .mbtiles test files as .sql files, e.g. by using sqlite3's [`.dump`](https://www.sqlitetutorial.net/sqlite-dump/) command.
  * [ ] Remove any schema creation - and reuse the above ones.  We may need to include the schema name in the file name to ensure the right schema is used... unless we can somehow transclude one file inside the other -- i am not sure if `.read filename.sql` can be used from another .sql file?
  * [ ] For image content, each tile content should be stored as an individual image file in a subdir - this way they can also be easily inspected. I think the `readfile('...')` sqlite function can be used to get the file content as part of the insert statement.
  * [ ] Same goes for the .mvt / .mlt content, except that sadly we have no way to store it in a viewable format - until someday we may have some simple json-mvt-mlt debug converter, or we could use ogr2ogr or another utility to store geojson - but this is not great because the output is less guaranteed.

sdf-jkl:
Can I work on this?

sdf-jkl:
With the first PR we were able to tick some of the boxes:

* [x] Create several schema initialization .sql files, e.g. in `mbtiles/sql/` dir, add `init-flat.sql`, `init-flat-with-hash.sql`, etc.
[...]

PR #1868, opened 2025-06-02
test(mbtiles): Generate mbtiles dynamically from SQL files to increase debuggability and transparency

sdf-jkl:
Working on this:

* [x] Create several schema initialization .sql files, e.g. in `mbtiles/sql/` dir, add `init-flat.sql`, `init-flat-with-hash.sql`, etc.
   * [x] Use content of these files directly in the [docs](https://github.com/maplibre/martin/blob/428265025e0c76ba4a1919e0f9cbe7618137a4d1/docs/src/mbtiles-schema.md?plain=1#L10-L17), rather than copy/paste it. See [`{{#include file.rs}}`](https://rust-lang.github.io/mdBook/format/mdbook.html?highlight=include#including-files)
   * [x] Use these files in the [test.sh](https://github.com/maplibre/martin/blob/main/tests/test.sh) before populating them with data 
   -- Couldn't find where the mbtiles are being updated in tests.sh
   * [x] Use these files in the unit tests
   * [x] Update docs to use these files with the `sqlite` command to create a blank .mbtiles file
* [x] Export all existing .mbtiles test files as .sql files, e.g. by using sqlite3's [`.dump`](https://www.sqlitetutorial.net/sqlite-dump/) command. 
   * [x] Remove any schema creation - and reuse the above ones. We may need to include the schema name in the file name to ensure the right schema is used... unless we can somehow transclude one file inside the other 
   -- i am not sure if `.read filename.sql` can be used from another .sql file?
   * [ ] For image content, each tile content should be stored as an individual image file in a subdir - this way they can also be easily inspected. I think the `readfile('...')` sqlite function can be used to get the file content as part of the insert statement.
   * [ ] Same goes for the .mvt / .mlt content, except that sadly we have no way to store it in a viewable format - until someday we may have some simple json-mvt-mlt debug converter, or we could use ogr2ogr or another utility to store geojson - but this is not great because the output is less guaranteed.
   
Related #1846


sdf-jkl:
@nyurik 
I tested including `.read` in .sql files and it worked.

I created these test .sql files:
`init.sql`
```sql
CREATE TABLE meow (
    text TEXT
);
```
`read.sql`
```sql
.read init.sql

INSERT INTO meow VALUES ('meow');
```
`rightaway.sql`
```sql
CREATE TABLE meow (
    text TEXT
);

INSERT INTO meow VALUES ('meow');
```
Created .sqlite files to check their difference and there was no output:
```bash
sqlite3 read.sqlite < read.sql
sqlite3 rightaway.sqlite < rightaway.sql
sqldiff read.sqlite rightaway.sqlite
```
So we should be good with using` .read` inside .sql files.

nyurik:
awesome work!!  I noticed that all .sql files have tons of individual inserts.  Is there a sqlite command to dump the entire table as a single insert?  Something like this. We could of course modify the sql files by hand later, but if sqlite .dump can do it automatically, it will save some time.

```sql
INSERT INTO 'tablename' ('column1', 'column2') VALUES
    ('data1', 'data2'),
    ('data1', 'data2'),
    ('data1', 'data2'),
    ('data1', 'data2');
```

sdf-jkl:
Unfortunately, there isn't one. Will have to do that manually (a small script should do the work)

tests/fixtures/mbtiles/uncompressed_mvt.sql
```diff
@@ -0,0 +1,22 @@
...
+ANALYZE sqlite_schema;
+INSERT INTO sqlite_stat1 VALUES('tiles','tile_index','196 28 3 1');
+INSERT INTO sqlite_stat1 VALUES('metadata','name','11 1');
```
nyurik: you don't need stats - its an internal sqlite state that should not be copied. Also we probably don't need `analyze sqlite_schema`

nyurik:
i don't think you want to do a script for this :)   A simple search/replace, or a multi-cursor editor would do all this in a few minutes :)

sdf-jkl:
Should we create a new just recipe for creating .mbtiles from .sql files?

We can integrate it in `just test`: 
```diff
-test: start (test-cargo "--all-targets") test-doc test-int
+test: start prepare-fixtures (test-cargo "--all-targets") test-doc test-int
```

nyurik:
I think it would make sense to have a recipe for mbtiles generation, yes.  Just make sure it gets created in some directory in the `.gitignore` list (or add a new entry there)....  I will have to think of a good dir structure. 

One hope I had though was to avoid using the actual disk files in the unit tests. Instead, it would be good to use in-memory sqlite dbs, populate them with data, and pass them to martin components.   See https://github.com/maplibre/martin/blob/a1250b947ad25b9cee176554894e21cefbde211a/mbtiles/tests/copy.rs#L135 for an example how in-memory sqlite db can be created, and then opened as if it was a regular file.   This method might not work from the start - we may rely on temp files at first.

sdf-jkl:
I was thinking about generating them in the same folders the original mbtiles were and the new sql files are now ( tests/fixtures/{files, mbtiles}
And adding *.mbtiles to `.gitignore`

nyurik:
that could be an interim step, agree

nyurik:
the problem though is that users won't be able to run `cargo test` without doing an additional step first -- that's why I was hoping we won't rely on the files

sdf-jkl:
> I think it would make sense to have a recipe for mbtiles generation, yes. Just make sure it gets created in some directory in the `.gitignore` list (or add a new entry there).... I will have to think of a good dir structure.
> 
> One hope I had though was to avoid using the actual disk files in the unit tests. Instead, it would be good to use in-memory sqlite dbs, populate them with data, and pass them to martin components. See
> 
> https://github.com/maplibre/martin/blob/a1250b947ad25b9cee176554894e21cefbde211a/mbtiles/tests/copy.rs#L135
> 
> for an example how in-memory sqlite db can be created, and then opened as if it was a regular file. This method might not work from the start - we may rely on temp files at first.

I'll look into this more, before I resolve to use files.

sdf-jkl:
Sorry for the delay in getting back to this.
From what I understand, the process for the unit tests should be:
1. Create the in-memory SQLite databases.
2. Use `sqlx` to populate them by reading the fixture files.
3. Access these databases within the unit tests.

I was wondering whether the goal is to initialize the `.mbtiles` databases once and reuse them across all unit tests, or if we should initialize the required `.mbtiles` databases fresh within each unit test?

nyurik:
> 1. Create the in-memory SQLite databases.
> 2. Use `sqlx` to populate them by reading the fixture files.
> 3. Access these databases within the unit tests.
> 
> I was wondering whether the goal is to initialize the `.mbtiles` databases once and reuse them across all unit tests, or if we should initialize the required `.mbtiles` databases fresh within each unit test?

* Per test is much better -- each test should run independently, possibly even at the same time, so each in-memory file should either be anonymous `:memory:`, or when this is not possible (e.g. when a filename must be passed as input), use some temporary filename that is unique per test (it is still in memory, but uses some extra `?` params like some other tests already)

* For (2 - populating data) - you can just read the whole script file and pass it to sqlite:

```rust
let script = fs::read_to_string("path/to/something.sql").unwrap();
sqlx::raw_sql(script).execute(&mut conn).await.unwrap();
```

sdf-jkl:
I updated the `summary::tests::summary` unit test and it seems like the in-memory dbs do not have `file_size`. 
The `page_count` was also different.

```
────────────┬───────────────────────────────────────────────────────────────────
    0       │-file_size: 49152
          0 │+file_size: ~
    1     1 │ mbt_type: Flat
    2     2 │ page_size: 4096
    3       │-page_count: 12
          3 │+page_count: 11
    4     4 │ tile_count: 196
    5     5 │ min_tile_size: 64
    6     6 │ max_tile_size: 1107
    7     7 │ avg_tile_size: 96.2295918367347
────────────┴───────────────────────────────────────────────────────────────────
Stopped on the first failure. Run `cargo insta test` to run all snapshots.
```

mbtiles/src/summary.rs
```diff
@@ -222,8 +222,10 @@ mod tests {
 
     #[actix_rt::test]
     async fn summary() -> MbtResult<()> {
-        let mbt = Mbtiles::new("../tests/fixtures/mbtiles/world_cities.mbtiles")?;
+        let mbt = Mbtiles::new("file:world_cities?mode=memory&cache=shared")?;
```
nyurik: You don't need a reusable (shared) connection here i don't think.  You would only need it when you create the (in-memory) file with `...?mode=memory&cache=shared`, and later pass the same identical string to some API to work on that "file".

```suggestion
        let mbt = Mbtiles::new(":memory:")?;
```

nyurik (commented):
good progress, thanks! I think this approach is workable, and moreover, we may want to greatly reduce the size of those .sql files by removing things we don't care about (i.e. we do not need to store detalied `tilestats` and possibly other things for at least some of the files.  The rule of thumb:  remove anything that is not tested for, but keep anything related to the core functionality of the .mbtiles files.

nyurik:
I think its ok to modify the test with the new page_count and remove file_size - not that critical

sdf-jkl:
I was working on removing the schema creation from `invalid-tile-format.sql` and ran into an issue initializing a`.mbtiles` database using `init-normalized.sql`.
It includes an extra column in the `CREATE TABLE` statement:
```sql
CREATE TABLE map (
   zoom_level INTEGER,
   tile_column INTEGER,
   tile_row INTEGER,
   tile_id TEXT,
   grid_id TEXT --   <-----This one
);
```
I was not sure if that's intentional(to create an invalid schema for testing), but I noticed that the `invalid-tile-format.mbtiles` file is not used anywhere in the project repository.
I also found two additional unused files in the same directory:
- `tests/fixtures/files/invalid.sql` 
-  `tests/fixtures/files/tmp.json`

nyurik:
technically grid is part of the [spec](https://github.com/mapbox/mbtiles-spec/blob/master/1.3/spec.md#grids), but i haven't seen anyone ever using it.

sdf-jkl:
I actually found that all `.mbtiles` files using the `normalized` schema include an extra field(`grid_id`) in the `map` table, which is not part of the `normalized` schema documented in docs.

What the `normalized` schema in the docs looks like:
```sql
CREATE TABLE map (
    zoom_level  INTEGER,
    tile_column INTEGER,
    tile_row    INTEGER,
    tile_id     TEXT);
```
What the schema from sqlite `.dump` actually look like:
```sql
CREATE TABLE map (
   zoom_level INTEGER,
   tile_column INTEGER,
   tile_row INTEGER,
   tile_id TEXT,
   grid_id TEXT -- <---- This one
);
```
Is the `grid_id` field missing from the documented schema?

nyurik:
`grid_id` is NOT part of the spec. Neither is the `map` table.  See [spec](https://github.com/mapbox/mbtiles-spec/blob/master/1.3/spec.md#grids) only requires the `tiles` table or view, and an optional `grid` table/view.  So we can safely get rid of it.

sdf-jkl:
I made one of the tests pass the unit testing, but only after excluding the `.read init-flat.sql` part from `world_cities.sql`.

It seems that `.read` inside a `.sql` file works only with the `sqlite3` cli, but not with `sqlx`.

docs/src/mbtiles-meta.md
```diff
@@ -7,6 +7,9 @@ number of tiles per zoom level, the size of the smallest and largest tiles, and
 level. The command will also print the bounding box of the covered area per zoom level.
 
 ```bash
+sqlite3 tests/fixtures/mbtiles/world_cities.mbtiles < tests/fixtures/mbtiles/world_cities.sql
+mbtiles summary tests/fixtures/mbtiles/world_cities.mbtiles
```
nyurik:
```suggestion
# Create a sample .mbtiles file in some dir
sqlite3 target/world_cities.mbtiles < tests/fixtures/mbtiles/world_cities.sql
# Get mbtiles summary
mbtiles summary target/world_cities.mbtiles
```

sdf-jkl:
I'm working on making unit tests work and one of the files - `mbtiles\src\copier.rs` has a comment:
```
// TODO: Most of these tests are duplicating the tests from tests/mbtiles.rs, and should be cleaned up/removed.
```
But we don't have a `tests/mbtiles.rs` file in the repo. Is this comment about the `test.sh` or unit tests somewhere else?

mbtiles/src/metadata.rs
```diff
@@ -267,17 +274,20 @@ mod tests {
         assert_eq!(tj.name.unwrap(), "Geography Class");
         assert_eq!(
             tj.template.unwrap(),
-            "{{#__location__}}{{/__location__}}{{#__teaser__}}<div style=\"text-align:center;\">\n\n<img src=\"data:image/png;base64,{{flag_png}}\" style=\"-moz-box-shadow:0px 1px 3px #222;-webkit-box-shadow:0px 1px 5px #222;box-shadow:0px 1px 3px #222;\"><br>\n<strong>{{admin}}</strong>\n\n</div>{{/__teaser__}}{{#__full__}}{{/__full__}}"
+            "{{#__location__}}{{/__location__}}{{#__teaser__}}<div style=\"text-align:center;\">\\n\\n<img src=\"data:image/png;base64,{{flag_png}}\" style=\"-moz-box-shadow:0px 1px 3px #222;-webkit-box-shadow:0px 1px 5px #222;box-shadow:0px 1px 3px #222;\"><br>\\n<strong>{{admin}}</strong>\\n\\n</div>{{/__teaser__}}{{#__full__}}{{/__full__}}"
```
nyurik: this is strange - the newline `\n` values should not have their `\` escaped

nyurik:
> I'm working on making unit tests work and one of the files - `mbtiles\src\copier.rs` has a comment:
> 
> ```
> // TODO: Most of these tests are duplicating the tests from tests/mbtiles.rs, and should be cleaned up/removed.
> ```
> 
> But we don't have a `tests/mbtiles.rs` file in the repo. Is this comment about the `test.sh` or unit tests somewhere else?

tbh, not sure, it's been a while ago... maybe it has already been cleaned up? :)

mbtiles/src/metadata.rs
sdf-jkl: The `.sql` files that I created don't have them:
```
('bounds','-180,-85.0511,180,85.0511'),
('minzoom','0'),
('maxzoom','1'),
('legend','<div style="text-align:center;">\n\n<div style="font:12pt/16pt Georgia,serif;">Geography Class</div>\n<div ...
                                             ^here                                                             ^and here
```
but for some reason after the files are generated, `mbtiles` reads the metadata with extra `\`.
I did the change to please `assert_eq!`

sdf-jkl:
It is likely, I will remove the comment and update the tests.

mbtiles/src/metadata.rs
nyurik: yes, this is a bug. Use this:

```sql
('legend','<div style="text-align:center;">' || x'0A0A' || '<div style="font:12pt/16pt Georgia,serif;">Geography Class</div>' || x'0A' || '<div ...
```

mbtiles/src/pool.rs
```diff
@@ -158,22 +174,28 @@ mod tests {
...
     #[tokio::test]
-    async fn test_metadata_normalized() {
-        let pool = MbtilesPool::open_readonly(
-            "../tests/fixtures/mbtiles/geography-class-png-no-bounds.mbtiles",
-        )
-        .await
-        .unwrap();
+    async fn test_metadata_normalized() -> MbtResult<()> {
+        let mbt = Mbtiles::new("file:test_metadata_normalized?mode=memory&cache=shared")?;
+        let mut conn = mbt.open().await?;
```
nyurik: i think it's easier for tests to simply use `.unwrap()`, rather than use `?` and have a return type.  A test should just fail asap

sdf-jkl: Makes sense, I guess I should change that for all the unit tests.

sdf-jkl:
What should be the next step once I finish working with unit tests and integration tests?

nyurik:
once this PR passes tests, I will review it once again and will most likely merge. It is big enough as is, so any subsequent changes should probably go in other PRs

nyurik:
Note that pre-commit CI is failing because it thinks the new .sql files are for PostgreSQL:
https://github.com/maplibre/martin/blob/c43a3de50055c58cc931f0417fd1270ef83b51c2/.pre-commit-config.yaml#L48-L54

We will need to use filters to specify which files should use postgres, and which - sqlite

mbtiles/src/patcher.rs
```diff
@@ -169,42 +169,59 @@ mod tests {
...
+        .unwrap();
 
         // Apply patch to the src data in in-memory DB
-        let patch_file = PathBuf::from("../tests/fixtures/mbtiles/world_cities_diff.mbtiles");
-        apply_patch(src, patch_file, true).await?;
+        let patch_file = PathBuf::from("file:flat_patch_file_mem?mode=memory&cache=shared");
+        let mbt = Mbtiles::new(&patch_file).unwrap();
+        let mut conn = mbt.open().await.unwrap();
+        let script =
+            std::fs::read_to_string("../tests/fixtures/mbtiles/world_cities_diff.sql").unwrap();
```
nyurik: i would just use `include_str!("...")` and elsewhere - it's better for a test to focus on the testing, not to any extras like file IO

.pre-commit-config.yaml
```diff
@@ -45,12 +45,20 @@ repos:
...
         # LT12: Files must end with a single trailing newline <- already done by trailing-whitespace
         # If you update this rule, please also update the rules in justfile, the fmt-sql recipe
         args: [--dialect=postgres, "--exclude-rules=AL07,LT05,LT12"]
         files: \.sql$
+        exclude: ^tests/fixtures/(mbtiles|files)/.*\.sql$
+
+
+        # 2. SQLite dialect for .sql files in specific directories
+      - id: sqlfluff-fix
+        args: [--dialect=sqlite, "--exclude-rules=LT05,LT12"]
```
nyurik: I think we don't want to exclude https://docs.sqlfluff.com/en/stable/reference/rules.html#rule-LT12

nyurik (commented):
looks almost ready!  One thing - please look over all the `open(...)` and see if you can use the `:memory:` instead of the `file:...?mode=memory`.  I am sure some require named filenames, but if any do not, would be good to get rid of them.  Please mark as "ready for review" when ... "ready for review" :D

sdf-jkl:
Will do, but I still have to figure CORS unit tests.

martin/tests/cors_test.rs
```diff
@@ -48,7 +48,7 @@ async fn test_cors_explicit_disabled() {
         cors: false
         mbtiles:
           sources:
-            test: ../tests/fixtures/mbtiles/world_cities.mbtiles
+            test: ../tests/fixtures/mbtiles/world_cities.sql
```
nyurik: this will not work.  You need to create the in-memory file and use it. Something like this:

```rust
let file = PathBuf::from("file:test_cors_explicit_disable?mode=memory&cache=shared");
let mbt = Mbtiles::new(&file).unwrap();
let mut conn = mbt.open().await.unwrap();
let script = include_str!("../../tests/fixtures/mbtiles/world_cities.sql");
sqlx::raw_sql(script).execute(&mut conn).await.unwrap();

let app = create_app!(formatdoc! {"
    cors: false
    mbtiles:
      sources:
        test: {}
", file.display()});
```

Note that we should have a helper function to create these files and reuse the same function everywhere, e.g. `let file = temp_named_mbtiles!(test_cors_explicit_disable, "../../tests/fixtures/mbtiles/world_cities.sql");` -- this is a macro that takes `ident` as the first param (should match current function name), and a literal (string) that gets used inside the macro in `include_str!(...)`. The result would be a string that can be used in a test.  For anonymous we could also have a `let conn = anonymous_mbtiles!("../../tests/fixtures/mbtiles/world_cities.sql");` -- returns sql connection, but creates `:memory:` file

nyurik: P.S. Macros might be a bit overwhelming to work with at first, so feel free to do it as regular functions, and I can fix it later

sdf-jkl: That's great, thanks!
I was having trouble figuring out how to create an in-memory database file that also has a usable path - something that can be accessed not just by sqlx.

nyurik: The whole idea to use `file:test_cors_explicit_disable?mode=memory&cache=shared` instead of `:memory:` is so that multiple sqlite connections can be created to the same (in-memory) file.  The only thing to note is that the first connection to the `mode=memory&cache=shared` should stay open while you open a second connection to it - otherwise sqlite will delete it from memory

sdf-jkl: Shouldn't it be the case since the tests run in async?

nyurik: the in-memory file should only exist as long as the test itself runs.  Other tests should never use the same path. So from a single test's perspective, it all runs in a single thread. Async in theory shouldn't have that much effect on that - so as long as the state machine knows that the initial connection should stay around until the end of the test, you can re-open the same path as many times as needed within the same test

mbtiles/src/metadata.rs
```diff
@@ -231,33 +232,43 @@ impl Mbtiles {
...
 }
 
+pub async fn anonymous_mbtiles(script: &str) -> (Mbtiles, SqliteConnection) {
+    let mbt = Mbtiles::new(":memory:").unwrap();
+    let mut conn = mbt.open().await.unwrap();
+    sqlx::raw_sql(script).execute(&mut conn).await.unwrap();
+    (mbt, conn)
+}
+
+pub async fn temp_names_mbtiles(file_name: &str, script: &str) -> PathBuf {
```
nyurik: this won't work - you also need to return path and sqliteconnection - otherwise the connection will be dropped on return.  Also, you may want to return a `String` rather than `PathBuf` for simplicity

justfile
```diff
@@ -278,8 +279,27 @@ test-doc *args:
 test-fmt:
     cargo fmt --all -- --check
 
+prepare-fixtures:
```
nyurik: i keep wondering if we need this - it might be good for experimenting with the files i guess - as long as the output does not make the repo dirty

mbtiles/src/metadata.rs
```diff
@@ -231,30 +232,49 @@ impl Mbtiles {
...
+    let mut conn = mbt.open().await.unwrap();
+    sqlx::raw_sql(script).execute(&mut conn).await.unwrap();
+    (mbt, conn)
+}
+
+#[allow(dead_code)]
+pub async fn temp_named_mbtiles(
+    file_name: &str,
+    script: &str,
+) -> (Mbtiles, SqliteConnection, PathBuf) {
```
nyurik, line 247: easier to return `String` here, esp because this is not really a Path anymore - technically it's a URL

```suggestion
) -> (Mbtiles, SqliteConnection, String) {
```

sdf-jkl:
Can't build the martin package on windows, so I can't test locally. Pushing the changes to see the errors here.

nyurik:
CI/Lint are failing - a bunch of small improvements needed. Don't worry about the docs and code coverage

mbtiles/src/metadata.rs
sdf-jkl: This makes sense, but `mbtiles:patcher` and `mbtiles::copier` expect `PathBuf` as input. The helper function is to be used mostly within these two functions.

sdf-jkl:
> CI/Lint are failing - a bunch of small improvements needed. Don't worry about the docs and code coverage

CI/Lint is finally good. Should I look into integration testing?

sdf-jkl:
Keep getting the same issue with integration tests.
It's the last part - comparing the actual output with expected output - that gives me the problem.
```bash
** Comparing actual output with expected output...
Files tests/output/mbtiles/copy_bindiff_diff.txt and tests/expected/mbtiles/copy_bindiff_diff.txt differ
** Expected output does not match actual output
```
This txt file comes from test.sh 620..622
```
sqldiff "$TEST_TEMP_DIR/world_cities_bindiff.mbtiles" \
      ./tests/fixtures/mbtiles/world_cities_bindiff.mbtiles \
      "$TEST_OUT_DIR/copy_bindiff_diff.txt"
```
I replaced the `tests/expected/mbtiles/copy_bindiff_diff.txt` with the output from CI and tried reproducing it myself, but keep the comparison still fails.

@nyurik Can you run `just bless` and see if it will fix the issue? I'm on win, so no local integration tests for me.

I also see that another issue with other CI failing is that they are running `test.sh` without generating fixtures beforehand.

Should we use `just prepare-fixtures` recipe in the CIs or move the fixture preparation step inside `test.sh`?

sdf-jkl:
@nyurik @CommanderStorm Following up on the issue here.

Can someone try running integration tests locally and see what the differences between files are, and `just bless` if possible, to push this through.

CommanderStorm:
I tried and the I did not see any unblessed things.
I need to figure out how to do this in CI
- https://github.com/maplibre/martin/pull/2202

FYI, to unblock you, you can download the output that CI expects:
image

nyurik:
Would it make sense to copy some of these changes to a new PR that has no issues, e.g. the sql files with includes, etc, but keep the original mbtiles until this PR lands? This might make it easier to merge in stages. What do you think?

sdf-jkl:
This PR is definitely a mess, and it could be helpful to split it up, but given it is in its last stage we can try to finish it as is.
If finishing it as is does not work, I am open to splitting it.

nyurik:
sure, up to you. Just thought it might be easier to merge the simpler things first like the new files + doc changes, etc. Totally up to you

sdf-jkl:
It seems like this time it's more than just the usual suspects: 
```
 ** Comparing actual output with expected output...
Files tests/output/auto/save_config.yaml and tests/expected/auto/save_config.yaml differ
Files tests/output/auto/table_and_view_two_schemas1.json and tests/expected/auto/table_and_view_two_schemas1.json differ
Files tests/output/auto/table_and_view_two_schemas1.json.headers and tests/expected/auto/table_and_view_two_schemas1.json.headers differ
Files tests/output/auto/table_and_view_two_schemas2.json and tests/expected/auto/table_and_view_two_schemas2.json differ
Files tests/output/auto/table_and_view_two_schemas2.json.headers and tests/expected/auto/table_and_view_two_schemas2.json.headers differ
Files tests/output/martin-cp/composite_save_config.yaml and tests/expected/martin-cp/composite_save_config.yaml differ
Files tests/output/martin-cp/flat-with-hash_save_config.yaml and tests/expected/martin-cp/flat-with-hash_save_config.yaml differ
Files tests/output/martin-cp/flat_save_config.yaml and tests/expected/martin-cp/flat_save_config.yaml differ
Files tests/output/martin-cp/normalized_save_config.yaml and tests/expected/martin-cp/normalized_save_config.yaml differ
Files tests/output/mbtiles/copy_bindiff_diff.txt and tests/expected/mbtiles/copy_bindiff_diff.txt differ
** Expected output does not match actual output
```
So I am down to split this RP

CommanderStorm:
@sdf-jk Most of these I have fixed a few minutes ago in `:main`. Sorry.
`copy_bindiff_diff` is the only one which is strange.

sdf-jkl:
@CommanderStorm Thanks, so we are back to the same problem.

sdf-jkl:
This time we can actually see the difference:

```diff
Resulting diff (max 100 lines)
diff --recursive --new-file '--exclude=*.pbf' tests/output/mbtiles/copy_bindiff_diff.txt tests/expected/mbtiles/copy_bindiff_diff.txt

1,4c1,4
< DROP INDEX name;
< CREATE UNIQUE INDEX name ON metadata (name);
< DROP INDEX tile_index;
< CREATE UNIQUE INDEX tile_index ON tiles (zoom_level, tile_column, tile_row);
---
> ﻿DROP INDEX name;
> CREATE UNIQUE INDEX name ON metadata (name);
> DROP INDEX tile_index;
> CREATE UNIQUE INDEX tile_index ON tiles (zoom_level, tile_column, tile_row);

error: Recipe `test-int` failed with exit code 1
error: Recipe `coverage` failed with exit code 1
Error: Process completed with exit code 1.
```

Chapgpt suggests it's a "Byte Order Mark (BOM) at the beginning of the file"

sdf-jkl:
Tried rebuilding the bindiff_diff files from scratch and see if there is any difference, but there seems to be nothing again.
```bash
# Import SQL dumps into mbtiles files
sqlite3 ./tests/fixtures/mbtiles/world_cities.mbtiles < ./tests/fixtures/mbtiles/world_cities.sql
sqlite3 ./tests/fixtures/mbtiles/world_cities_modified.mbtiles < ./tests/fixtures/mbtiles/world_cities_modified.sql
sqlite3 ./tests/fixtures/mbtiles/world_cities_bindiff.mbtiles < ./tests/fixtures/mbtiles/world_cities_bindiff.sql

# Generate bindiff1.mbtiles by copying and diffing original and modified mbtiles
./target/debug/mbtiles.exe copy \
  ./tests/fixtures/mbtiles/world_cities.mbtiles \
  --diff-with-file ./tests/fixtures/mbtiles/world_cities_modified.mbtiles \
  ./tests/fixtures/mbtiles/world_cities_bindiff1.mbtiles \
  --patch-type bin-diff-gz

# Compare the generated bindiff1.mbtiles with the expected bindiff.mbtiles, output diff to file
sqldiff.exe ./tests/fixtures/mbtiles/world_cities_bindiff1.mbtiles ./tests/fixtures/mbtiles/world_cities_bindiff.mbtiles > ./tests/fixtures/mbtiles/copy_bindiff_diff.txt

# Check SHA256 sums to confirm files match exactly
sha256sum ./tests/expected/mbtiles/copy_bindiff_diff.txt
sha256sum ./tests/fixtures/mbtiles/copy_bindiff_diff.txt
```

Both times SHA256sum was the same:
e56106ff8270279a90d609c4ac69cd2b5af08efdf5c0ef6c8656def10a860046

CommanderStorm:
lets see where CI precicely sees the diff.
The result printing nonprinting chars is likely what we want anyway

sdf-jkl:
So it has been a windows issue all along:

```diff
diff --recursive --new-file '--exclude=*.pbf' tests/output/mbtiles/copy_bindiff_diff.txt tests/expected/mbtiles/copy_bindiff_diff.txt
1,4c1,4
< DROP INDEX name;
< CREATE UNIQUE INDEX name ON metadata (name);
< DROP INDEX tile_index;
< CREATE UNIQUE INDEX tile_index ON tiles (zoom_level, tile_column, tile_row);
---
> DROP INDEX name;^M
> CREATE UNIQUE INDEX name ON metadata (name);^M
> DROP INDEX tile_index;^M
> CREATE UNIQUE INDEX tile_index ON tiles (zoom_level, tile_column, tile_row);^M
```

CommanderStorm:
Just throwing it out there, have you configured https://docs.github.com/en/get-started/git-basics/configuring-git-to-handle-line-endings?platform=windows or the similar setting in your editor? 🤔

nyurik:
also, highly recommend NOT to use Windows, unless you are inside `WSL/VM/remote/<anything not windows>` :)

nyurik:
for windows, unless you really really need to, i would set git to automatically keep everything with unix line endings rather than the  `auto` converting things to windows line endings.  But that's just my preference - i haven't touched windows in over 15 years

sdf-jkl:
@nyurik I'll do that. I'm only on windows because that's what we have at work.

CommanderStorm:
@nyurik It is weird that this part of CI (which should check for this) seems to be broken, or I don't fully understand what it does:

https://github.com/maplibre/martin/blob/8a0379cab03511b0405994e817c7574f45855236/.pre-commit-config.yaml#L28-L29

nyurik:
I suspect the issue is that the expected file has correct `lf`-only eol, but when running on windows, it is windows output that gets generated with the `crlf`, so it fails comparison?

sdf-jkl:
The current issue is that ci.yml workflow does not have a step to build `.mbtiles` files from `.sql` files created.

We could add a step inside "matrix build" job and "unit test the server" job. What do you think?
image

nyurik:
this step should be a `just` recipe, e.g. `gen-test-mbtiles` -- and `ci-test` should list it as a requirement

sdf-jkl:
I found out that the `ci-test` recipe is not actually used anywhere (therefore adding the `prepare-fixtures` recipe to it didn't do anything). 
I also found that the `lint-rust` job in the CI workflow looks very similar to it:
```yml
lint-rust:
    ...
    steps:
      ...
      - run: just env-info
      - run: just test-fmt
      - run: just clippy
      - run: just check
      - run: just check-doc
```
```just
# Run all tests as expected by CI
ci-test: env-info restart test-fmt clippy check-doc test check prepare-fixtures && assert-git-is-clean

```

nyurik:
sorry, all my other projects use `ci-test` for all testing, making things more consistent in ci and for local dev. For now, just add it to that workflow, but for the future I think we should make CI and local dev to have the same recipe

tests/test.sh
```diff
@@ -290,6 +290,26 @@ echo "::group::Check HTTP server is running"
 $CURL --head "$STATICS_URL/webp2.pmtiles"
 echo "::endgroup::"
 
+# Prepare MBTiles from SQL fixtures
+echo "::group::Prepare .mbtiles fixtures from .sql"
+FOLDERS=("tests/fixtures/files" "tests/fixtures/mbtiles")
+
+for folder in "${FOLDERS[@]}"; do
+    echo "Processing folder: $folder"
```
sdf-jkl, line 298: I moved the mbtiles generation logic inside the test.sh file and it fixed the issue for every test, but the Windows one, because windows doesn't have sqlite preinstalled.

.github/workflows/ci.yml
```diff
@@ -463,7 +463,10 @@ jobs:
       - name: Set up gdal-bin and sqlite3-tools
         if: matrix.os == 'ubuntu-latest'
         run: sudo apt-get update && sudo apt-get install -y gdal-bin sqlite3-tools
-
+      - name: Install sqlite3 and sqldiff on Windows
+        if: matrix.os == 'windows-latest'
+        run: choco install sqlite --yes
+        shell: powershell
```
CommanderStorm: To make pre-commit happy

```suggestion
      - name: Install sqlite3 and sqldiff on Windows
        if: matrix.os == 'windows-latest'
        run: choco install sqlite --yes
        shell: powershell
```

martin/Cargo.toml
```diff
@@ -158,6 +158,7 @@ criterion.workspace = true
 ctor.workspace = true
 indoc.workspace = true
 insta = { workspace = true, features = ["json", "yaml", "redactions"] }
+mbtiles = { workspace = true }
```
CommanderStorm: Is already installed as a regular dependency

```suggestion
```

martin/src/config/args/root.rs
```diff
@@ -323,7 +323,7 @@ mod tests {
         let args = Args::parse_from([
             "martin",
             "../tests/fixtures/pmtiles/png.pmtiles",
-            "../tests/fixtures/mbtiles/json.mbtiles",
+            // "../tests/fixtures/mbtiles/json.mbtiles",
```
CommanderStorm: What is going on here? Should we remove this line instead?


```suggestion
```

martin/src/config/args/root.rs
```diff
@@ -334,11 +334,11 @@ mod tests {
         assert!(err.is_ok());
         insta::assert_yaml_snapshot!(config, @r#"
         pmtiles: "../tests/fixtures/pmtiles/png.pmtiles"
-        mbtiles: "../tests/fixtures/mbtiles/json.mbtiles"
         cog:
           - "../tests/fixtures/cog/rgba_u8_nodata.tiff"
           - "../tests/fixtures/cog/rgba_u8.tif"
         "#);
+        // mbtiles: "../tests/fixtures/mbtiles/json.mbtiles"
```
CommanderStorm: Same question as above

```suggestion
```

mbtiles/src/metadata.rs
```diff
@@ -231,6 +232,26 @@ impl Mbtiles {
     }
 }
 
+#[allow(dead_code)]
```
CommanderStorm:
```suggestion
/// Create an in memory, temporary mbtile connection with the given `script`
```

mbtiles/src/metadata.rs
```diff
@@ -231,6 +232,26 @@ impl Mbtiles {
...
+pub async fn anonymous_mbtiles(script: &str) -> (Mbtiles, SqliteConnection) {
+    let mbt = Mbtiles::new(":memory:").unwrap();
+    let mut conn = mbt.open().await.unwrap();
+    sqlx::raw_sql(script).execute(&mut conn).await.unwrap();
+    (mbt, conn)
+}
+
+#[allow(dead_code)]
```
CommanderStorm:
```suggestion
/// Create a named, in memory, temporary mbtile connection with the given `script`
```

CommanderStorm (approved):
Wow, you actually did it 🚀 

I have a few minor things which I am just going to apply and one question about a code comment

martin/src/config/args/root.rs
sdf-jkl: Thanks for the review! I'm pretty sure I commented it out since the mbtiles were not there anymore.
I think it makes sense to generate an in-memory db for this test to keep it exhaustive.

martin/src/config/args/root.rs
sdf-jkl: Same as above

martin/tests/cors_test.rs
```diff
@@ -48,12 +48,15 @@ macro_rules! create_app {
...
+    let (_mbt, _conn, file) = temp_named_mbtiles("test_cors_explicit_disabled", script).await;
+
+    let app = create_app!(&formatdoc! {"
         cors: false
         mbtiles:
           sources:
-            test: ../tests/fixtures/mbtiles/world_cities.mbtiles
-    "});
+            test: {}
+    ", file.display()});
```
sdf-jkl, line 59: A lot of boilerplate code in this file, could be shrinked into a function/macro

tests/fixtures/files/invalid-tile-format.sql
```diff
@@ -0,0 +1,48 @@
...
...ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff','b0a559d767b45de037d4aefe483d3a274dff23e8'),
```
sdf-jkl, line 20: Does this have to be wall of text?

tests/fixtures/mbtiles/geography-class-jpg-diff.sql
```diff
@@ -0,0 +1,47 @@
...
...37e3f4aaf211c83cfa8aa8be5699128f345c59e7de16f1045e28f0d69baac524520ba81247f25b72abe3e75faab6411d41041ab472587154fc5de199349bc9bc43e1fb6845eac4e6eac638403a87ddc0dc318906cc2b3640c9181b8910e95afdb6b493f924c7716f2986e2de423cc85c1e55c02707b8ec4104641af46325357479ed383e597fc39e936d228b403be3fad35e519f6aa969a8db4d06639a33ff00021486653c89108f6615e61e99619c67839a76fe3dea9b5c20cfef149f453935223b4c06324fa77ff3f8500590f4e593dfbd532db18eff0093fdea55ba8b68ccb1e3b1dd8cd005f0d4e0fcf35556642bfeb14fd1a8f3467e56cfbe7fc8a00b41c62937d44a4e0753f85381249c03c75e28024dfcf5a0b91de9a3040191cd21818763f85004b1dc0ce0f5a91a504553747887dd3cfa8a8beda81b61601bfbbdcd005d2f4071eb5514c8c3215b1ea41a70de79e481d71da802c990628f3463ad55dcedc282ff00414d6f357aa3003d01a00bbe60a4f378e0d51124b9e87143dc94383d4fa734017fcda4f347d2b38de331c2abb1ef85a512499e87e8680347cd18a6f9a327f9550f35b772403e99a559c8ec6802f890669038f5aa8276c127f9537cc90f446c7a906802db4b8efc530cc09c66aa492499fbadf9540f77b4728c39ea54d00681973dea0924f949cf4aa8672e01dae7d30a6a296ed9464a38faa1a0092560467ad729e28f0d43ae859e276b2d5601fe8f7b01daea7aed6e3e64271953907af5008dbbbd41635cbe573d3e5359eda9c67048723b9da69a938bba2651525667ffd9','d8018fba714e93c29500adb778b587a5'),
```
sdf-jkl, line 20: Huge Binary in the script

tests/fixtures/mbtiles/webp.sql
```diff
@@ -0,0 +1,23 @@
...
...c7ae7a39e0bae33214445faecf7274f04a1b00d79a137ea3c143c5a8da2d8f6689efa17902a2ea6bde99eea9f1ddabed0ba52cea6b78d30c60f8cb84180009296d513a596a4f22b3c9038cb16b249df55c87e1a1632f01f3e974acf521095216d083889478e3efe524b48e6883c996c54352946ef673b8d31990389fd634c2f84ab5020a5e6d970bf6794146c6468ef96e3185413835027bd9718fd4b37215ea39590550053a688e8cf5c6435b5d063dfdd5f17e34564e1a867407d2c0eecc90613cb101c73c135a9b9c3a46f1818667408ba00970038575f794bf1f7d1f215e0a424981eadde2f5ac8689d07245af7a52be0c2c166fcb87d4e4a6741a05c14863e818c24a16e04d4022f00cd1dfa1f89344af222c4d95b160c1c67e714eb2c10931f36a98c6a6d2e48bf820bd0ed237c22636c1226c472c8039aa34eda4bdb431057e8a5ec2ab4095867c46ca3656ba6f6ac5cd5176bc99d2f8ff23ed8b7cb1d485472ae5d377a31cf0707556567f83ca4e71ae7c3bc748aa6df58a935bee499502ac0e5681682d23171e4c6bd4748d91680f9046257cc5d0c89162984920643756ca85e731a9cd88d065be0d23899839c6dde066fcf4b7779846a1114192287e0a1bcac8f1e52077952caa2be63311067a5c63d3f5459c9d979e2f977da3459980c3794c56e357aabe7429794a9141019efd7307d0acb0d7d5d656bd067e0a62fdd586c24708baba3ecf54746eb425986c31f6093cd46c33484428710aaff9c0922e9c9756d614d2b46616f378e3f34e7f864f84b4e6ed46363acfcff94f43391160d00d6de25cbacfd9b3e809fe18d1adbb7cc0000');
```
sdf-jkl, line 21: Huge binary again

sdf-jkl (commented):
Self review

CommanderStorm:
@sdf-jkl All of the issues that remain can/should be different PRs 😉 
=> merging as is

This is great work, thank you so much

sdf-jkl:
@CommanderStorm @nyurik Thanks for your help on this, this was definitely too big of a pr to go through and I should have split the work into smaller PRs way earlier.
Happy to get it done.

PR #1876, opened 2025-06-10
[DRAFT] replace cross with a buildx multiarch dockerfile

CommanderStorm:
This is a more complex PR doing 20 things in 1.
I will need to break it apart into the elements/thoughts.

It does:
- throw away cross
- add a new dockerfile using zigbuild + buildx instead of cross + quemu + docker
- more caching for the docker build (lets see if we have to remove this. Github has an fairly small cache)
- adds attestation and sbom generation #1878 
- adds labels #1877


nyurik (commented):
if you can pull it off.... wow :)

CommanderStorm:
well it works on my machine..
Now I only have to convince CI and you guys ^^

```shell
docker buildx build --platform=linux/amd64,linux/arm64 .
```

CommanderStorm:
Todo:
- replace `name: Build ${{ matrix.target }}` with binaries extracted from the docker image (if possible, otherwise .. cry? .. or use native build platforms for this?)

.github/workflows/ci.yml
```diff
@@ -291,13 +261,15 @@ jobs:
...
-          file: .github/files/multi-platform.Dockerfile
+          file: Dockerfile
           push: true
           load: false
           tags: ${{ steps.docker_meta.outputs.tags }}
           annotations: ${{ steps.docker_meta.outputs.annotations }}
           labels: ${{ steps.docker_meta.outputs.labels }}
           platforms: linux/amd64,linux/arm64
+          cache-from: type=gha
+          cache-to: type=gha,mode=min
```
CommanderStorm: Looking at https://github.com/maplibre/martin/actions/caches we should not add more thigns to the cache.

This is also not increasable. I might know that guy.. small world ^^
https://github.com/orgs/community/discussions/42506#discussioncomment-11388788



```suggestion
```

.github/workflows/ci.yml
```diff
@@ -368,17 +338,18 @@ jobs:
...
+      - name: Download build artifact build-x86_64-unknown-linux-gnu
         uses: actions/download-artifact@v4
         with:
-          name: cross-build
+          name: build-x86_64-unknown-linux-gnu
+          path: target_release/x86_64-unknown-linux-gnu
       - run: tests/test-aws-lambda.sh
         env:
-          MARTIN_BIN: x86_64-unknown-linux-musl/martin
+          MARTIN_BIN: target_release/x86_64-unknown-linux-gnu/martin
```
CommanderStorm: I now get why this is done this way. AWS does seem to like their musl build.

@nyurik
could you help me with why this is a different pipeline step?
I have a bit of trouble to understand why this is not just in the `docker-build-test`-step.
All dependencys already exist there => why upload and download here?

(don't look at the rest of the code, it is a WIP and pre-scrubbing)

CommanderStorm:
Took me a while to untangle this, but this is the dependency tree
```mermaid
flowchart TD
    docker["docker"] -- "linux-musl ONLY amd" --> lamda["lamda"]
    docker -- "linux-musl" --> package["package"]
    build["build"] -- "linux-gnu" --> package
    build -- debian --> package & test-with-service["test-with-service"]
    build -- macos --> package
    build -- windows ONLY amd --> package & test-multi-os["test-multi-os"]
    build -- "linux-gnu ONLY amd" --> test-with-service & test-multi-os
    test-multi-os --> package
    test-with-service --> package
```

So basically, the approach that I proposed is quite dumb

CommanderStorm:
a less dumb architecture might be 

```mermaid
flowchart TD
    build["build"] -- "linux-musl ONLY amd" --> lamda["lamda"]
    build -- "linux-musl" --> package["package"] & docker["docker"]
    build -- "linux-gnu" --> package
    build -- debian --> package & test-with-service["test-with-service"]
    build -- macos --> package
    build -- windows ONLY amd --> package & test-multi-os["test-multi-os"]
    build -- "linux-gnu ONLY amd" --> test-with-service & test-multi-os
    test-multi-os --> package
    test-with-service --> package
    docker --> package
```

CommanderStorm:
though this would throw away lots of the things that I like about the reproducible docker files away..

What do you two think? @nyurik @sharkAndshark 
Should I build the musl binaries with docker or outside of docker?

CommanderStorm:
The advantage of the second architecture is that it builds the docker image at roughly double the spped, given that it uses 2 runners instead of one to build the two archs.

nyurik:
i'm ok to fully refresh how we build things.  My original goal was both CI performance and to cover all the release cases, including docker.  Due to limitations of where PG could run, i had to do all sorts of tricks.  Ideally, just run docker with PG in all cases to do tests. I also ran into tons of issues with building multi-arch docker images

CommanderStorm:
I am going to take the ideas from this PR into other PRs, but this PR will not be merged.
there was some valuable learning in here which I used for #1902

PR #1877, opened 2025-06-10
feat(new-dockerfile): add more metadata to the docker build

CommanderStorm:
this PR tries to add every docker metadata tag that we can onto the images.
A tad excessive in terms of metadata depth? maybe ^^

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -170,6 +170,36 @@ jobs:
         with: { port: '5412', output-unix-paths: 'yes' }
       - name: Copy static files
         run: cp -r tests/fixtures/pmtiles2/* ${{ steps.nginx.outputs.html-dir }}
+      - name: Docker meta
+        id: docker_meta
+        uses: docker/metadata-action@v5
+        # https://github.com/docker/metadata-action
+        with:
+          images: ghcr.io/${{ github.repository }}
+          labels: |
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -170,6 +170,36 @@ jobs:
...
       - name: Copy static files
         run: cp -r tests/fixtures/pmtiles2/* ${{ steps.nginx.outputs.html-dir }}
+      - name: Docker meta
+        id: docker_meta
+        uses: docker/metadata-action@v5
+        # https://github.com/docker/metadata-action
+        with:
+          images: ghcr.io/${{ github.repository }}
+          labels: |
+            org.opencontainers.image.created={{commit_date 'YYYY-MM-DDTHH:mm:ss.SSS[Z]'}}
```
nyurik, line 180: i'm ok with it, but kinda strange that we have near-identical label and annotations. Any way to de-dup them, and/or just have one of them?

nyurik (approved):

.github/workflows/ci.yml
CommanderStorm: no clue why labels and annotations exist.
They seem to be the same thing, but at different levels of the docker file.
For now, duplicating this is simple enough.

We will be able to remove the second one after github supports anchors: 
- https://github.com/actions/runner/issues/1182

nyurik: sigh, gitlab has it, and while helpful, it produces such horrible mess at times :(

PR #1878, opened 2025-06-10
feat(ci): add attestation and sbom generation to the docker images

CommanderStorm:
> [!IMPORTANT]
> this PR needs https://github.com/maplibre/martin/pull/1881 to function and thus has it merged in
> => review said PR first to reduce this PRs diff ^^

This PR adds attestation and sbom generation to the docker images.

This allows our users to verify their "authenticity" and better inspect what is in them.
I doubt that a lot of people actually do this though ^^
=> https://github.com/maplibre/martin/attestations

But it is nice to have and should not cause any maintenance overhead. At least it never caused any on the other repos where I enabled it ^^


.github/workflows/ci.yml
```diff
@@ -242,13 +247,24 @@ jobs:
         if: github.event_name != 'pull_request'
```
CommanderStorm, line 288:
```suggestion
        id: docker_push
        if: github.event_name != 'pull_request'
```

.github/workflows/ci.yml
```diff
@@ -242,13 +247,24 @@ jobs:
...
           labels: ${{ steps.docker_meta.outputs.labels }}
           platforms: linux/amd64,linux/arm64
+      - name: Attest
+        if: github.event_name != 'pull_request'
+        uses: actions/attest-build-provenance@v2
+        id: attest
+        with:
+          images: ghcr.io/${{ github.repository }}
+          subject-name: ghcr.io/${{ github.repository }}
+          subject-digest: ${{ steps.push.outputs.digest }}
```
```suggestion
          subject-digest: ${{ steps.docker_push.outputs.digest }}
```

nyurik (approved):
:tada:

PR #1879, opened 2025-06-10
feat(new-dockerfile): try annotating both at index and manifest level

CommanderStorm:
this is a followup to 
- #1877 

Apparently, github does not accept the metadata being there.. ridiculous.
lets add it to the manifest,index

this is the last PR tampering with docker metadata, I swear.


nyurik (approved):
:rofl:

PR #1881, opened 2025-06-10
chore(ci): merged docker load step

CommanderStorm:
This PR merges this into one multiarch docker image build by using the `contained local store`

The reason is in part
- breaking apart #1876 and that
- #1879 does apparently not work on single-arch images.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -119,6 +119,25 @@ jobs:
         env:
           PGPORT: ${{ job.services.postgres.ports[5432] }}
 
+      # we want to test our docker images before we push them to the registry
+      # this is done by
+      # - building multi-arch images
+      # - executing the test script against each arch via docker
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -119,6 +119,24 @@ jobs:
...
+      # > You can enable the containerd image store, or push multi-platform images is to directly push to a registry
+      - name: Set up Docker
+        uses: docker/setup-docker-action@v4
+        with:
+          daemon-config: |
+            {
+              "features": {
+                "containerd-snapshotter": true
+              }
+            }
```
CommanderStorm, line 139: having to do this feels a bit gross.. what a bodge.

nyurik (approved):
if this work, :tada: !

One tiny nit - in the yaml file, one of the lines with URL is a bit too long

PR #1886, opened 2025-06-16
chore: only include the readme in doc.rs builds

CommanderStorm:
Currently, the README.md is considered for any build.
This means, that rebuilds now happen "if we build for docs AND the readme changed", not "if the readme changed"

Fairly simple one, but still an unrelated change. This breaks out a part of:
- https://github.com/maplibre/martin/pull/1876


nyurik (approved):
I usually rely on `features = default` too, but this should be cleaner, thx

PR #1891, opened 2025-06-26
When there are column comments, it will prompt for duplicate table

### LINKED ISSUE #1890: When there are column comments, it will prompt for duplicate table.
When there are column comments, it will prompt for duplicate table.

Image

The result of executing `query_available_tables.sql` is as follows: 

Image

https://github.com/maplibre/martin/blob/c43a3de50055c58cc931f0417fd1270ef83b51c2/martin/src/pg/scripts/query_available_tables.sql#L111

Add `AND pg_description.objsubid = 0` can solve this problem.
```sql
LEFT JOIN pg_description ON cls.oid = pg_description.objoid AND pg_description.objsubid = 0
```

huanglii:
resolve #1890

### DIFF
--- a/tests/fixtures/tables/points2.sql
+++ b/tests/fixtures/tables/points2.sql
@@ -4,10 +4,8 @@ CREATE TABLE points2
     geom GEOMETRY (POINT, 4326)
 );
 
-COMMENT ON TABLE points2 IS '{
-    "description": "description from SQL comment",
-    "attribution": "some attribution from SQL comment"
-}';
+-- this will not be used as tilejson since it is not valid tilejson
+COMMENT ON TABLE points2 IS 'A table with points';
 COMMENT ON COLUMN points2.gid IS 'The primary key';
 COMMENT ON COLUMN points2.geom IS 'The geometry column';
 
--- a/tests/fixtures/tables/points2.sql
+++ b/tests/fixtures/tables/points2.sql
@@ -4,7 +4,8 @@ CREATE TABLE points2
     geom GEOMETRY (POINT, 4326)
 );
 
--- this will not be used as tilejson since it is not valid tilejson
+-- if below were valid tilejson, we would serve it as such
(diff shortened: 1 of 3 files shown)


CommanderStorm:
CI is failing due to the new warning

Could you either allow-list the new warning and add a SQL comment (not table comment) acknowledging that this is won'be used as tilejson since it is not valid tilejson

```
Checking for no other warnings or errors in the log
[2025-06-26T12:26:53Z WARN  martin::pg::query_tables] Unable to deserialize SQL comment on public.points2 as tilejson, the automatically generated tilejson would be used: expected value at line 1 column 1
Log file target/test_logs/auto.txt has unexpected warnings or errors
```

huanglii:
This warning was issued before. https://github.com/maplibre/martin/issues/1890#issue-3177866231
image
I'll try to check it

tests/fixtures/tables/points2.sql
```diff
@@ -4,6 +4,11 @@ CREATE TABLE points2
     geom GEOMETRY (POINT, 4326)
 );
 
+-- this will not be used as tilejson since it is not valid tilejson
```
CommanderStorm:
```suggestion
-- if below were valid tilejson, we would serve it as such
-- for this test case, this is intentionally not the case
```

tests/test.sh
```diff
@@ -211,6 +211,8 @@ validate_log() {
   remove_line "$LOG_FILE" 'Source IDs must be unique'
   remove_line "$LOG_FILE" 'PostgreSQL 11.10.0 is older than the recommended minimum 12.0.0'
   remove_line "$LOG_FILE" 'In the used version, some geometry may be hidden on some zoom levels.'
+  # If a table has comments, it will prompt a warning, but it is not a fatal error.
```
```suggestion
```

CommanderStorm (approved):

nyurik (approved):
thank you!!!

PR #1897, opened 2025-07-01
feat: Web-based UI

CommanderStorm:
This is the drafting pad for building martins new UI.

For what I think would make a good release, we would need the following done and merged:
- [x] (due to the node version) 
  - [x] https://github.com/maplibre/martin/pull/1876
  - [x] https://github.com/maplibre/martin/pull/1902
  - [x] https://github.com/maplibre/martin/pull/1906
- [x] (due to metrics) https://github.com/maplibre/martin/pull/793

Resolves:
- Resolves #1120 
- (works towards, but integration is rudemeentary) #852

It adds:
- [x] (unit, snapshot) Tests for the frontend
- [x] Integrate with the catalog
- [x] Integrate with the metrics
- [x] is clear in communication if it does not implement a feature like font previews

Screenshots are on slack, but they look quite undercooked, so be warned.

To be reviewable, this PR will need to be broken apart and then merged piece by piece

PR #1902, opened 2025-07-03
chore(ci): reduce CI-Time by splitting cross to its own step of the CI build

CommanderStorm:
This PR moves cross to its own step in CI which others (lamda, packaging, docker) can depend on.
It is a hard dependecy of https://github.com/maplibre/martin/pull/1897 as without refactoring this, changing cross for something running a modern node version is quite hard.

As discussed in https://github.com/maplibre/martin/pull/1876, we currently build docker files and two static binaries in the same step.

This is excessively slow (could be paralell) and we need the musl binaries for other steps anyway:

| previously | now |
|--------|--------|
| [32m 20s](https://github.com/maplibre/martin/actions/runs/16037935228) | [19m 26s](https://github.com/maplibre/martin/actions/runs/16073125582/) | 

Moving this step outside is likely a good step.

> [!NOTE]
> We can shave off a few minutes by.. not using cross, but this is a different PR.

This was the dependency tree beforehand:

```mermaid
flowchart TD
    docker -- "linux-musl ONLY amd" --> lamda
    docker -- "linux-musl" --> package
    build -- "linux-gnu" --> package
    build -- debian --> package & test-with-service
    build -- macos --> package
    build -- windows ONLY amd --> package & test-multi-os
    build -- "linux-gnu ONLY amd" --> test-with-service & test-multi-os
    test-multi-os --> package
    test-with-service --> package
```

The proposed dependency tree:

```mermaid
flowchart TD
    build-musl -- "linux-musl ONLY amd" --> lamda
    build-musl -- "linux-musl" --> package & docker
    build -- "linux-gnu" --> package
    build -- debian --> package & test-with-service
    build -- macos --> package
    build -- windows ONLY amd --> package & test-multi-os
    build -- "linux-gnu ONLY amd" --> test-with-service & test-multi-os
    test-multi-os --> package
    test-with-service --> package
    lamda --> package
```


.github/workflows/ci.yml
```diff
@@ -511,7 +533,7 @@ jobs:
   test-with-svc:
     name: Test postgis:${{ matrix.img_ver }} sslmode=${{ matrix.sslmode }}
     runs-on: ubuntu-latest
-    needs: [ build, docker-build-test ]
```
CommanderStorm, line 514: no clue why this was required and why this step in CI calls the rust-cache action.

.github/workflows/ci.yml
```diff
@@ -818,7 +838,7 @@ jobs:
   done:
     name: CI Finished
     runs-on: ubuntu-latest
-    needs: [ package ]
+    needs: [ package, docker-build-test ]
```
CommanderStorm, line 828: I since `docker-build-test` does not produce the musl binaries, it is not required for this step any more.

.github/workflows/ci.yml
```diff
@@ -653,7 +673,7 @@ jobs:
   package:
     name: Package
     runs-on: ubuntu-latest
-    needs: [ lint-debug-test, docker-build-test, test-multi-os, test-with-svc ]
+    needs: [ lint-debug-test, test-multi-os, test-with-svc, test-aws-lambda ]
```
CommanderStorm, line 678: before we package, we should have all tests run through.
Transitively, this still means that musl must have been build.
It does not mean that docker needs to have been build though.

.github/workflows/ci.yml
```diff
@@ -74,10 +74,63 @@ jobs:
...
+          mv target/${{ matrix.target }}/release/mbtiles target_releases/${{ matrix.target }}
+      - name: Save build artifacts to build-${{ matrix.target }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: build-${{ matrix.target }}
+          path: target_releases/*
   docker-build-test:
     name: Build and test docker images
     runs-on: ubuntu-latest
+    needs: [ musl-build, lint-debug-test ]
```
CommanderStorm, line 123: unsure if this should depend on the following instead.
This is kind of the same as `package`, but could also be tests, so this dependency set.

likely, the current version makes CI less dependeent and thus faster, but 🤷‍♂ 


```suggestion
    needs: [ lint-debug-test, test-multi-os, test-with-svc, test-aws-lambda ]
```

.github/workflows/ci.yml
```diff
@@ -74,10 +74,57 @@ jobs:
...
+        if: github.event_name != 'release' && github.event_name != 'workflow_dispatch'
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+        # https://github.com/docker/setup-qemu-action
+        with:
+          platforms: linux/amd64,linux/arm64
+      - name: Set up AWS SAM
+        uses: aws-actions/setup-sam@v2
+        with:
+          use-installer: true
```
CommanderStorm: I don't quite get why the `sam`-CLI was installed in the docker build in the first place.

I think this is not nessesary, right?


```suggestion
```

.github/workflows/ci.yml
CommanderStorm: removing the rust-cache action was a mistake, there is rust-buiding going on for the unit-tests

.github/workflows/ci.yml
```diff
@@ -563,8 +585,6 @@ jobs:
     steps:
       - name: Checkout sources
         uses: actions/checkout@v4
-      - uses: Swatinem/rust-cache@v2
-        if: github.event_name != 'release' && github.event_name != 'workflow_dispatch'
       - name: Run NGINX
```
CommanderStorm: removing this was a mistake as noted in https://github.com/maplibre/martin/pull/1902/files#r2185269914

```suggestion
      - uses: Swatinem/rust-cache@v2
        if: github.event_name != 'release' && github.event_name != 'workflow_dispatch'
      - name: Run NGINX
```

nyurik:
@CommanderStorm one other note you might be interested in as part of this work: https://release-plz.dev/ has been working well to automate the releases.  Basic idea:
* on each merge to main, if the crate version in main is the same as the one in crates.io, automatically create a "release PR".  The release pr proposes to bump crates version to the next one based on semver-checker compatibility (e.g. patch or minor). Also generates changelog based on the commit messages since last release.
* on each merge to main, if the crate version is **different** from the released one, do the release - publishing to crates.io and creating the release page.

I just configured [varnish-rs](https://github.com/varnish-rs/varnish-rs) for this mode (it contains 3 crates). Note that there the version is the same for all 3 crates, so a slightly different configuration might be needed for us (or it might work as is?)

CommanderStorm:
If we move forward with the martin-lib split, yes I agree.
Currently, there is not much benefit in this.

nyurik (approved):
awesome, thx!!!

PR #1903, opened 2025-07-04
fix: clippy::ignore-without-reason

CommanderStorm:
https://rust-lang.github.io/rust-clippy/master/#/ignore_without_reason

one of them is a debugging-test and one idk, so remove the ignore and include it in CI.

This PR blocks https://github.com/maplibre/martin/pull/1897 as CI won't progress far enough without ugly hacks

### DIFF
--- a/mbtiles/tests/copy.rs
+++ b/mbtiles/tests/copy.rs
@@ -578,7 +578,6 @@ async fn diff_and_patch(
 #[rstest]
 #[trace]
 #[tokio::test(flavor = "multi_thread")]
-#[ignore]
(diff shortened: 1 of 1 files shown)


nyurik (approved):
thx!

nyurik:
i think that test used to run for a while, and then became too complicated to maintain and got out of whack.  One day i hope to bring it back, or just delete it - i haven't decided yet :)

CommanderStorm:
Aaaa. I was just about to fix the typo. (somehow, the test does not work in CI, so ignoring is reasonable)

PR #1906, opened 2025-07-06
chore(ci): Replace cross with zigbuild

CommanderStorm:
> [!IMPORTANT]
> This PR builds on top of https://github.com/maplibre/martin/pull/1902 looking at it soft-requires that to be merged to not be super-messy

Debugging actions is a bit hard, so bare with me.

We could also switch to native buids.
Pro:
- less tooling

Con:
- adding more platforms (armv7 <-> armv8 <-> risk5) would be harder, but are people requesting them?


nyurik (approved):
awesome!

PR #1909, opened 2025-07-08
chore: move from eslint to biomejs for formatting/linting

CommanderStorm:
this breaks out another of the components (which are not blocked by node compatability) of #1897 

The PR is more of a preference, but one I would really like to see.
biomejs does have a bit of functionality (such as sorting json and imports) which we currently do manually.

The linting and formatting is also much much faster than the laternative JS-based linters.

The demo frontend does not quite clearing the type-check and also not quite clearing all lints, but it also did not clear that beforehand. No change there.
I fixed a few of the lower hanging fruits


demo/frontend/package.json
```diff
@@ -24,24 +18,27 @@
...
-    "eslint-config-react-app": "7.0.1",
-    "typescript": "^5.2.2",
-    "vite": "^4.5.14",
-    "vite-plugin-mkcert": "^1.16.0",
-    "vite-tsconfig-paths": "^4.2.1"
+    "@vitejs/plugin-react": "^4.6.0",
+    "typescript": "^5.8.3",
+    "vite": "^7.0.0",
+    "vite-plugin-mkcert": "^1.17.8",
+    "vite-tsconfig-paths": "^5.0.0"
```
CommanderStorm, line 29: I had to bump vite a bit because are some incompatabilities with modern nodejs versions in terms of type-checking.

I can break this out to another PR, but this is fairly light in terms of lines changed in the ts/vite config files.

biome.json
```diff
@@ -0,0 +1,32 @@
...
+    },
+    "enabled": true
+  },
+  "formatter": {
+    "enabled": true,
+    "indentStyle": "space",
+    "indentWidth": 2,
+    "lineWidth": 100
+  },
+  "javascript": { "formatter": { "quoteStyle": "double" } },
```
CommanderStorm:
```suggestion
  "javascript": { "formatter": { "quoteStyle": "single" } },
```

nyurik (approved):
lgtm

PR #1910, opened 2025-07-09
fix(pg): handle SRID `4326` bbox tiles spanning over the antimeridian

### LINKED ISSUE #982: Tiles along the antimeridian are always empty for table sources with a positive buffer and a SRID of 4326
I'm working with a polygonal table with srid 4326 in PostGIS, whose extents span the non-contiguous United States. I'd like to be able to visualize the entire dataset at once, which means I need to generate tiles at a very low zoom level. I have Martin set up with the simplest table source configuration imaginable, specifically something like the following:

```
    table_id:
      schema: public
      table: table_name
      geometry_column: geometry
      srid: 4326
      minzoom: 2
      maxzoom: 13
      geometry_type: POLYGON
```

Martin is impressively fast at generating a zoom level 2 tile for the eastern US (tile (2, 1, 1)), but generates an empty tile for the western US (tile (2, 0, 1)), despite there being quite a bit of data within the extents of the tile. 

In looking into it, I've found that seemingly all tiles along the antimeridian (those for which the tile coordinate x = 0 or x = 2^n - 1) end up empty when a positive buffer is specified on a table source with SRID 4326. The crux of the issue is that the search bounding box for such tiles ends up crossing the antimeridian when there is a positive buffer. But SRID 4326 polygons that cross the antimeridian don't operate well with bounding box intersections due to the circular nature of longitudes. So when Martin tries to get geometries within the bounding box, it actually ends up getting features only outside of the actual tile area.

One can find the erroneous code in [table_source.rs](https://github.com/maplibre/martin/blob/73b56e8a90f3617ca04a09f2438dd8a54593ef20/martin/src/pg/table_source.rs#L153). Specifically, the buffered bounding box is calculated as something akin to: `ST_TileEnvelope(z, x, y, margin => buffer / extent)` . The features that go into tile are then queried via something akin to 
`SELECT ... FROM table WHERE geometry && ST_Transform(buffered_bounding_box, srid)`

For the case of tile (2, 0, 1) above, the default buffer of 64, and the default extent of 4096, the post-transform buffered bounding box ends up being `POLYGON((178.59375 -1.406108835435152,178.59375 67.06743335108297,-88.59375 67.06743335108297,-88.59375 -1.406108835435152,178.59375 -1.406108835435152))`, which does not intersect the actual tile in SRID 4326. Indeed, the result of  
` SELECT ST_Transform(ST_TileEnvelope(2, 0, 1, margin => 0.0), 4326) && ST_Transform(ST_TileEnvelope(2, 0, 1, margin => 64.0 / 4096.0), 4326);` is false, which implies that any geometry within the expected tile will not make its way into the resulting tile.

A workaround is to use a buffer of zero, which enforces that the tile envelope doesn't cross the antimeridian. But that is unideal for some purposes, so I am requesting a more permanent solution that properly handles this case.

One possible solution is to use the `ST_ShiftLongitude` function when working with a tile of zoom greater than one along the antimeridian.  Another might be to transform the feature geometry into SRID 3857 (Web Mercator) instead of converting the bounding box to the table's SRID, but I'm unsure of if that would work in all cases.

todtb:
Improves tile query bounding box calculation in table_to_query by applying buffer margins correctly for SRID `3857` (already supported) and `4326` (new) .

For SRID `4326`, this uses `ST_Expand` with degree-based margins to fix antimeridian edge cases.

Resolves #982


todtb:
Reading the SRID projection and unit from `spatial_ref_sys` seems like a workable approach to handling many (though not all) reference systems. 

That said, I think a much simpler conditional for `bbox_search` could handle the vast majority of use cases by just looking at SRID 3857 or SRID 4326

mbtiles/src/summary.rs
```diff
@@ -303,7 +303,7 @@ mod tests {
               - -123.75000000000001
               - -40.97989806962013
               - 179.99999999999997
-              - 61.60639637138628
+              - 61.60639637138627
```
todtb: Not quite sure what this test run change is about, so I want to look into it, even if the precision level here is negligible.

CommanderStorm: In CI at least, this is the following..

```suggestion
              - 61.60639637138628
```

mbtiles/src/summary.rs
```diff
@@ -313,7 +313,7 @@ mod tests {
               - -123.75000000000001
               - -40.97989806962015
               - 180.00000000000003
-              - 61.60639637138628
+              - 61.60639637138627
```
CommanderStorm: CI dooes not like this change.
Unsure why it comes up

```suggestion
              - 61.60639637138628
```

martin/src/pg/config_table.rs
```diff
@@ -25,6 +25,12 @@ pub struct TableInfo {
     /// Geometry SRID
     pub srid: i32,
 
+    /// Projection of SRID if found in `spatial_ref_sys`
+    pub proj: Option<String>,
+
+    /// Projection unit of SRID if found in `spatial_ref_sys`
+    pub proj_unit: Option<String>,
```
CommanderStorm: I am not quite sure about the naming. I usually like to not hamming-code these things. Maybe `srid_projection`, `srid_projection_unit`?
But I think @nyurik has a better taste in this.

An alternative configuration format if these are srid-specific would be to set this as:
`srid: {}`

mbtiles/src/summary.rs
todtb: Just updated the other line below this one as well. Will verify in CI build

CommanderStorm (commented):
Seems like a reasonable approach to me.
I am a but out of my depth in how these projections work though. 😰

I would like a bit more docs on how this can be used for our users.
At lease this (and the parts under /docs) should explain how srid, proj and proj_unit  interact

For us, testing that things work is quite important.
To ensure that this works, I would like:
- one unit test testing proj=Some(..) with proj_unit=None 
- one unit test testing proj=None with proj_unit=Some(..)
- one unit test testing proj=Some(..) with proj_unit=Some(..)
- the last would also make a nice [integration test](https://github.com/maplibre/martin/blob/main/tests/test.sh)

martin/src/pg/config_table.rs
todtb: `proj` is pulled from the `proj4text` string so in this case has a specific meaning:
```
+proj=merc +a=6378137 +b=6378137 +lat_ts=0.0 +lon_0=0.0 +x_0=0.0 +y_0=0 +k=1.0 +units=m +nadgrids=@null +wktext +no_defs
```
Technically I suppose `proj_unit` would just be `units` but that seemed somewhat vague on its own.

Open to the other ideas though.

todtb:
> Seems like a reasonable approach to me. I am a but out of my depth in how these projections work though. 😰
> 
> I would like a bit more docs on how this can be used for our users. At lease this (and the parts under /docs) should explain how srid, proj and proj_unit interact
> 
> For us, testing that things work is quite important. To ensure that this works, I would like:
> 
> * one unit test testing proj=Some(..) with proj_unit=None
> * one unit test testing proj=None with proj_unit=Some(..)
> * one unit test testing proj=Some(..) with proj_unit=Some(..)
> * the last would also make a nice [integration test](https://github.com/maplibre/martin/blob/main/tests/test.sh)

Sure, makes sense.

It looks like `martin/tests/pg_table_source_test.rs` should be testing around the construction of table info from the database and so that should be covering:
- proj=Some(..) with proj_unit=None
- proj=Some(..) with proj_unit=Some(..)

by way of 4326 and 3857.

I'm not sure if there would be a scenario with proj=None unless it was a custom defined SRS. In which case this will always fallback to the existing default behavior anyway.

Yes, further integration tests are always helpful. I'm entirely new to rust and its ecosystem so I would appreciate any guidance on that, but it looks like we may already have a few integration tests with both 4326 and 3857 as a start.

CommanderStorm:
> I'm not sure if there would be a scenario with proj=None [with proj_unit=Some(..)] unless it was a custom defined SRS. In which case this will always fallback to the existing default behavior anyway.

Would it make sense to print a warning in such a case?

todtb:
I think I understand a source of confusion here. There was an oversight on my part which was that table info can be serialized to an auto saved config file. This proj4 data is really only used internally and shouldn't be user configurable (it is based purely on database SRS definitions).  I've marked it to not serialize. 

I did still leave a 3857 table in `martin/tests/pg_table_source_test.rs` as the tests were previously only testing 4326, but this PR should no longer change config.yml output, so this is purely additional unrelated testing.

I think the best integration test additions here would be specifically around tile generation at x=0 (antemeridian). I'll look more into how `test.sh` is working, and see what data I can create to test this (such as in a tile like 2/0/1)

For what it's worth, I have tested these queries with a local data set and the changes are all looking great for 4326 and 3857 data sources, so I have a good idea of what data we can use in the tests to help protect against regressions.

todtb:
@CommanderStorm - I've added integration tests with geometries specifically around the antimeridian. It includes geometries nearby and those that cross, both in the affected SRID 4326. Let me know if this looks good to you or if you have any suggested changes.

@nyurik - Overall, does the approach of reading this proj4 data for guidance on how to expand the bbox work for you?

martin/src/pg/query_tables.rs
```diff
@@ -171,19 +176,34 @@ pub async fn table_to_query(
...
+    } else {
+        ""
+    };
+    let proj_unit = if let Some(proj_unit) = &info.proj_unit {
+        proj_unit
+    } else {
+        ""
+    };
 
     let bbox_search = if buffer == 0 {
```
CommanderStorm, line 189: I would like to see a code comment here explaining the logic (focus: "why?") below.
What is happening there is not quite trivial to parse for me as a younger programmer and I think this goes for a lot of other people too.

martin/src/pg/query_tables.rs
```diff
@@ -171,19 +176,34 @@ pub async fn table_to_query(
...
+    let proj = if let Some(proj) = &info.proj {
+        proj
+    } else {
+        ""
+    };
+    let proj_unit = if let Some(proj_unit) = &info.proj_unit {
+        proj_unit
+    } else {
+        ""
+    };
```
CommanderStorm: nit:
this here seems a bit unconventional.
Would below work?

If we do fold the `None` case anyhow, we might as well not have it and change this to be `String` instead, or am I missing something?

```suggestion
    let proj = info.proj.as_ref().unwrap_or_default();
    let proj_unit = info.proj_unit.as_ref().unwrap_or_default();
```

CommanderStorm (commented):
I will need to look over the testcases in more depth (focus: is this correct).
This is hard to judge from githubs UI on an evening before the conference.

I will meet with @nyurik there and we can have a look if this is reasonable or if a case is left untested.

martin/src/pg/query_tables.rs
todtb: I am by no means a Rustacean (in fact this is the first time I've written it) so if there is a better or more idiomatic way here, I'd be happy to change it. This was more following the pattern established above by the other conditional assignments. The expansion of the more succinct ternary-like syntax was applied by the linter.

martin/src/pg/query_tables.rs
todtb: The bulk of the conditional logic I think @nyurik could elucidate better, specifically the _why_. 

However as I understand it we want to check if we are even using a buffer, then if the database supports the expanded `ST_` function capabilities for margin (see existing comment below). 

The two new clauses only determine the unit, degrees or meters, to use based on projection. Or perhaps you were looking for a more detailed explanation of the `ST_` functions themselves?


---

```
    /// Indicates if `ST_TileEnvelope` supports the margin parameter.
    ///
    /// `true` if running postgis >= `3.1`
    /// This being false indicates that tiles may be cut off at the edges.
    #[must_use]
    pub fn supports_tile_margin(&self) -> bool {
        self.supports_tile_margin
    }
```

todtb:
> I will need to look over the testcases in more depth (focus: is this correct). This is hard to judge from githubs UI on an evening before the conference.
> 
> I will meet with @nyurik there and we can have a look if this is reasonable or if a case is left untested.

Thanks for the review. Looking forward to hearing what you both think and enjoy your conference!

martin/src/pg/query_tables.rs
todtb: I should note that I'm primarily holding off on additional changes here at the moment until we have at least agreed on the overall approach being reasonable.

martin/src/pg/query_tables.rs
```diff
@@ -204,7 +215,7 @@ FROM (
   FROM
     {schema}.{table}
   WHERE
-    {geometry_column} && ST_Transform({bbox_search}, {srid})
```
CommanderStorm, line 207: @nyurik when we looked over this yesterday, we overlooked this part. We are already reprojecting..
Therefore what we discussed is moot

martin/src/pg/query_tables.rs
CommanderStorm: I get the first part.
I have trouble understanding the rest of the conditions.

I think the
- why `ST_Expand` fixes `srid != 3857` would be good to document this way.
- why we  have special handling for `proj == "longlat"`/.. in the first case.

CommanderStorm: the core reason why we are partially talking past each other is likely because I don't quite understand the problem that this PR is solving.

A simple doodle would go a long way to help me understand this

todtb: We discussed this a bit w/ some visuals and came up with comment to help explain the bbox search.

martin/src/pg/query_tables.rs
todtb: Yeah, this was moved due to the conditional wrapping of `ST_Expand`. I also think it helps with the overall clarity of the intersection `&&` in the query by simplifying: `geometry && bbox`

martin/src/pg/query_tables.rs
```diff
@@ -171,19 +175,40 @@ pub async fn table_to_query(
...
+    // (meters) as a unit, that is used. If the SRID uses a non-standard projection or unit, it will
+    // fallback to existing behavior.
+    //
+    // If a geodetic projection such as SRID 4326 were to be used with ST_TileEnvelope and margin
+    // parameter, the resultant bounding box for tiles on the antimeridian would be calculated
+    // incorrectly. For example with a margin of 2 units, the antimeridian edge would transform from
+    // -180 to +178. This results in a bbox that stretches from the easternmost edge of a tile
+    // (plus margin) around the map to the westernmost edge of the tile (minus margin). The
+    // resulting bbox covers none of the original tile. In this example, ST_Expand will result in
+    // a westernmost edge (minus margin) of -182.
```
CommanderStorm: nit: a few rewordings

```suggestion
    // When calculating the bounding box + margin to search within, a few considerations must be made.
    // The `ST_TileEnvelope`-`margin` parameter is for use with SRID 3857.
    // When using a different SRID, `ST_Expand` provides SRID specific units.
    // For longlat projections such as SRID 4326, these are degrees.
    // If the projection specifically provides "m" (meters) as a unit, that is used.
    // If the SRID uses a non-standard projection or unit, it will fallback to existing behaviour.
    //
    // Handling this case is essential because:
    // A geodetic projection (e.g. SRID 4326) used with `ST_TileEnvelope`+`margin` for tiles on the antimeridian results in incorrectly calculated bounding boxes.
    // For example, the antimeridian edge would transform from -180 to +178 with a margin of two units.
    // This results in a bbox that stretches from the easternmost edge of a tile
    // (plus margin) around the map to the westernmost edge of the tile (minus margin).
    // The resulting bbox covers none of the original tile.
    // In this example, `ST_Expand` will result in a westernmost edge (minus `margin`) of -182.
```

tests/fixtures/tables/antimeridian.sql
```diff
@@ -0,0 +1,19 @@
+DROP TABLE IF EXISTS antimeridian;
+
+CREATE TABLE antimeridian
+(
+    gid serial PRIMARY KEY,
+    geom GEOMETRY (GEOMETRY, 4326)
```
CommanderStorm, line 6: I would prefer if we were to test the other case (`proj == "latlon"` vs `proj_unit == "m"`) as well and rename this testcase with the projection in the name.

Having every case tested makes regressions less likely ^^

CommanderStorm (approved):
Really nice work and the explaination really helps.

I approve of the changes, assuming that there is a testcase for the new, untested case 👍🏻

CommanderStorm (approved):
Will be merged once CI passes (not your fault):
- https://github.com/maplibre/martin/pull/1969

PR #1912, opened 2025-07-10
feat: build docker images on PRs

CommanderStorm:
Not sure if we want this.
THis PR adds docker tags such as `ghcr.io/maplibre/martin:pr-1912`

This would allow running #1897 in the demo via just changing the docker tag.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -275,7 +275,6 @@ jobs:
           retention-days: 5
 
       - name: Login to GitHub Docker registry
-        if: github.event_name != 'pull_request'
         uses: docker/login-action@v3
         # https://github.com/docker/login-action
         with:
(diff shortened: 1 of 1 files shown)


CommanderStorm:
image

Makes sense, that approach will not work on second thought 😓
We could publish to ghcr.io/commanderstorm/martin:pr-1912, but not here

nyurik:
this is indeed a good question - i honestly don't know if it will be of a regular use to anyone -- usually these things are done if it is relatively hard to build locally, and there is a CI of sorts to test it, but we pass CI builds between stages using upload/download artifacts rather than the container registry, so that's not needed. Also, doing this pollutes the container store with labels that will be mostly useless very soon.

That said, once in a while it might be useful to create it... so maybe we should do this as a manual workflow trigger, perhaps with a param?  But I honestly don't think we should spend much time on this.

nyurik:
ah, right, we don't want anyone from outside of the org to create something that will be part of our container registry

CommanderStorm:
That is a good point, I did not think about the artefacts published in this way..

PR #1924, opened 2025-07-14
Github Action: auto assign on commenting take

ajita-asthana:
Simplify the process of assigning the issue to the contributor, user's can self assign the issue.
An additional GitHub workflow that would run in response to work take submitted as a comment


CommanderStorm:
We currently don't use the process of assigning users. I don't think we need this management overhead at the scale that we operate at.

=> I don't think we need this, but if one of the other maintainers finds this helpful, we can add this.
Requirement: this is added to the contribution docs

nyurik:
tbh, i am not against this, seems like a useful and simple thing to have. Of course we could try to configure the bot used by the rust project to do the same thing - it has far greater scope in what it is able to do, but i am not certain we need much of it.

So yes, if you fix the CI and add contribution documentation, i'm ok with this

nyurik:
@ajita-asthana when you have a chance, please document this in the https://github.com/maplibre/martin/blob/main/docs/src/development.md as part of this PR, and we can merge it

ajita-asthana:
sure @nyurik will do that.

CommanderStorm:
Given that there is no movement on this PR, I am going to close it preliminary.
We can always still reopen it if you get to adding this to the docs.

PR #1926, opened 2025-07-15
feat: Add "unrecognized key in configuration file" warning

### LINKED ISSUE #1838: Add config file validation for unused values
This is something that I thought we had, but apparently we have not.

> I have lost a couple hours this morning trying to serve my styles, after tracing your pull request where you add this features i find out that i was using the wrong version of martin (copied from this file of the docs).
> 
> **Maybe you could add a conf file validation when the server start, to check if there are some incompatible values.**

From https://github.com/maplibre/martin/issues/1837#issuecomment-2891222386_ CC @Andreal2000

Minimal reproduction:

- `conf.yaml`
  ```yaml
  this-should: be-a-warning
  pmtiles: http://localhost:5412/webp2.pmtiles
  ```
- `cargo run -p martin -- --config config.yaml`

It is of note that we alredy have the underlying infrastructire laid out:
https://github.com/maplibre/martin/blob/d3eba76510c976d82e3187e6ead0c1f8267ac193/martin/src/file_config.rs#L71

Said function returns the unrcognised values for a tile source.

This issues task is to warn if there exists an unrecognised value anywhere instead of silently ignoring them.

I think testing wise, this is simplest asserted by adding `this-should: be-a-warning` on the various levels of the [yaml config we use for integration tests](https://github.com/maplibre/martin/blob/main/tests/config.yaml).

<sup>[our contribution guide is here](https://maplibre.org/martin/development.html)</sup>

ajita-asthana:
This PR adds a warning when encountered with `unrecognized_value` for a tile source. 

Resolves #1838

This change doesn't include the integration tests.


martin/src/file_config.rs
```diff
@@ -171,6 +171,9 @@ impl<T: ConfigExtras> FileConfigEnum<T> {
         if let Self::Config(cfg) = self {
             copy_unrecognized_config(&mut res, prefix, cfg.get_unrecognized());
         }
+        if res.is_empty() {
+            warn!("There exists an unrecognized value");
+        }
```
CommanderStorm: Two things:
- this warning is not really actionable. We need to tell people at minimum which key is unregognised
- Would it not be better to add a warning one level up, i.e. where this is called?

ajita-asthana: I see. I thought if `res` is empty then the value would be triggered. 
side:
I was looking around for adding the test case line in each block of yaml file but it didn't make sense to me.

martin/src/file_config.rs
```diff
@@ -168,8 +168,18 @@ impl<T: ConfigExtras> FileConfigEnum<T> {
...
             copy_unrecognized_config(&mut res, prefix, cfg.get_unrecognized());
+
+            if !res.is_empty() {
+                warn!(
+                    "Unrecognized configuration values found in prefix '{}': {:?}. \
+                    These will be ignored. Please check your configuration file for typos or deprecated keys.",
+                    prefix,
+                    cfg.get_unrecognized()
+                );
+            }
```
CommanderStorm: Much better.

I think below code could be a bit more user friendly.
This way, we are telling people less abstract which value is unexpected.

This also fixes the duplicate reporting bug caused by checking `res` instead of `unrecognized`.

```suggestion
            let unrecognized = cfg.get_unrecognized();
            copy_unrecognized_config(&mut res, prefix, unrecognized);

            for (key, value) in unrecognized.into_iter() {
                warn!("Ignoring unrecognized configuration entry '{prefix}.{key}: {value}'. Please check your configuration file for typos.");
            }
```

ajita-asthana: Thanks @CommanderStorm .

martin/src/file_config.rs
```diff
@@ -168,8 +168,16 @@ impl<T: ConfigExtras> FileConfigEnum<T> {
...
+
         if let Self::Config(cfg) = self {
-            copy_unrecognized_config(&mut res, prefix, cfg.get_unrecognized());
+            let unrecognized = cfg.get_unrecognized();
+            copy_unrecognized_config(&mut res, prefix, unrecognized);
+
+            for (key, value) in unrecognized.into_iter() {
+                warn!(
+                    "Ignoring unrecognized configuration entry '{prefix}.{key}: {value}'. Please check your configuration file for typos."
+                );
```
CommanderStorm: There is a clippy warning about `value` not having `Display` implemented.

Also, on second thought, we are not really objecting to `value`, rather we are objecting on the `key` not being something we recognise.

```suggestion
            for key in unrecognized.keys() {
                warn!(
                    "Ignoring unrecognized configuration key '{prefix}.{key}'. Please check your configuration file for typos."
                );
```

tests/config.yaml
```diff
@@ -169,29 +169,35 @@ postgres:
 
 
 pmtiles:
+  this-should: be-a-warning
```
CommanderStorm: That is a good start 👍🏻

- there are more parts where we should add this above
- we now need to assert that the warning has been raised in `test.sh`

CommanderStorm:
For some reason I cannot push to this branch any more.
Continuing in https://github.com/maplibre/martin/pull/2151

PR #1940, opened 2025-07-22
chore(ui): remove unused debug pages from tests

CommanderStorm:
We added a proper frontend, so the debug pages are no longer nessesary


nyurik:
should we wait to remove these until after we add the new UI?

CommanderStorm:
Both approaches are fine for me.

I don't think the pages were very useful to begin with as they just show the basics. :shrug:

nyurik:
i have actually used the demo page many times to validate various things, and as a prototype for other experiments

CommanderStorm:
🎉 

Now that we don't have a frontend anymore, we might as well remove the debug pages, or should I add a page in the frontend for them?

nyurik:
eventually i hope to have similarly simple experience with our new frontend - where you can explore the data available. Eventually it will be part of maputnik. For now, i don't think we have any alternative simple way to debug data sources - so i don't think we should remove debug.html just yet

CommanderStorm:
@nyurik is there something you are still missing in the new frontend?
I think all the capabilities noted here are included.

What is not included is the deckgl or leaflet ones, but I don't think that we need to have a specific file for them.

CommanderStorm:
@nyurik Could we move forward on this? We really should not have paralell infra lying around ^^

nyurik:
Lets just keep one demo.html - I use it all the time for all sorts of hack-testing

CommanderStorm:
> demo.html

So that would be which? 😅
There currently does not exist such a thing

nyurik:
oops, i meant we should keep just one debug.html - the rest are not needed.  We may want to consolidate anything useful from others into one

nyurik (approved):

justfile
```diff
@@ -153,11 +153,6 @@ coverage *args='--no-clean --open':  (cargo-install 'cargo-llvm-cov') clean star
 cp *args:
     cargo run --bin martin-cp -- {{args}}
```
CommanderStorm, line 154:
```suggestion

# Start Martin server and open a test page (not the integrated UI)
debug-page *args: start
    open tests/debug.html  # run will not exit, so open debug page first
    {{quote(just_executable())}} run {{args}}
```

justfile
```diff
@@ -153,11 +153,10 @@ coverage *args='--no-clean --open':  (cargo-install 'cargo-llvm-cov') clean star
...
 
-# Start Martin server and open a test page
+# Start Martin server and open a test page (not the integrated UI)
 debug-page *args: start
     open tests/debug.html  # run will not exit, so open debug page first
     {{quote(just_executable())}} run {{args}}
```
```suggestion
    {{quote(just_executable())}} run {{args}}

```

tests/demo.html
```diff
@@ -356,4 +356,4 @@
 </script>
 </body>
```
CommanderStorm:
```suggestion
```

tests/demo.html
```diff
@@ -356,4 +356,4 @@
...
 
-</html>
+</html>
```
CommanderStorm:
```suggestion
</html>

```

CommanderStorm:
oops, did not notice your 👍🏻 before re-requesting review

PR #1941, opened 2025-07-23
chore(ci): minor CI formatting

nyurik:
Make YAML files a bit more concise

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,12 +88,10 @@ jobs:
       - run: rustup update stable && rustup default stable
       - name: Checkout sources
         uses: actions/checkout@v4
-        with:
-          set-safe-directory: false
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -88,12 +88,10 @@ jobs:
       - run: rustup update stable && rustup default stable
       - name: Checkout sources
         uses: actions/checkout@v4
-        with:
-          set-safe-directory: false
+        with: { set-safe-directory: false }
```
CommanderStorm, line 91: I am wondering if we actually need this.

CommanderStorm (approved):
Lgtm

.github/workflows/ci.yml
nyurik: I was wondering about that one too - we can always remove it in a separate PR. No idea why it is here

PR #1942, opened 2025-07-23
feat(ui): Add an analytics section via `/_/metrics`

CommanderStorm:
This draft breaks out just the analytics section of the frontend

It is a bit larger, but most of the changes are tests, so likey fine.
The big part is tests.

The commits on this one are a bit of a mess, but I am scared to squash at this point. The rebase would likely result in at least one bug.

# Screenshot

image


nyurik (approved):
thx!!!

PR #1946, opened 2025-07-25
chore(ci): don't accidentally automerge due to skipped checks

### LINKED ISSUE #1811: Optimize CI jobs to only run when needed
It seems some projects use this style of a CI test to check if any job has failed or not, without explicitly waiting for them to pass/fail/be ignored.

```yaml
  ci-passed:
    if: always()
    needs:  [  <list of CI job names to check>  ]
    runs-on: ubuntu-latest
    steps:
      - if: ${{ contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled') }}
        run: exit 1
```

I think we can use the same approach to simplify our CI to only run on relevant changes, while still making our CI ruleset to depend on this one job to tag the PR as passing.

CommanderStorm:
This fixes the stupid problem that our CI allows skipping it in some circumstances

> [!NOTE]
> A job that is skipped will report its status as "Success". It will not prevent a pull request from merging, even if it is a required check.

https://docs.github.com/en/actions/how-tos/writing-workflows/choosing-when-your-workflow-runs/using-conditions-to-control-job-execution

It turns out the reason a lot of people do this is more underlying

Resolves #1811

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -791,7 +791,6 @@ jobs:
     needs: [ package, docker-build-test ]
     if: always()
     steps:
+      - run: echo "${{ toJSON(needs) }}"
       - if: ${{ contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled') }}
         run: exit 1
-      - name: Finished
-        run: echo "CI finished successfully"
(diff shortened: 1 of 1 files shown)


CommanderStorm:
CI proves that the case is fixed via this actions run

https://github.com/maplibre/martin/actions/runs/16527995733/job/46746133426

nyurik (approved):
You are clearly having too much fun with git

PR #1949, opened 2025-07-25
chore(core): unify how we handle errors in modules

CommanderStorm:
One thing that has been consistently complicated is how we handle errors.
It is complicated to learn for others, it is complicated to learn for me.
This needs to go.

Currently, there are 3 (roughly) different styles:
- bubble them up
- wrap them in `FileError`, then bubble up via that
- handle in module level errror, bubble up in that.

This is hard to manage and leads to a nearly circular (not actually) dependency when factoring out things.

This PR proposes:
- one `error.rs` per module
- if unification of errors is nessesary, MartinError is directly chosen, no convoluted skippable "inheritance" via `FileError`

=> less coupling and more cohesion


nyurik (approved):
thx, i was a bit concerned about the extra `map()` for errors, but i guess its not a big deal

PR #1951, opened 2025-07-25
chore(core): make sure that there is only one `TileData`

CommanderStorm:
this is not actively a problem, just something that I noticed during the refactoring excercise towards `martin-core` which also spawned https://github.com/maplibre/martin/pull/1949


nyurik (approved):
a bit weird to have this as a `pub use` here, as this means we are re-publishing it twice, but not a biggie

PR #1954, opened 2025-07-26
feat(ui): Add the font catalog to the dashboard

CommanderStorm:
This breaks the font catalog out of #1897 

Yes, this seems also quite heavy in terms of lines chaneged, but this is the smallest version of this feature sadly.
I hope that from a reviewer load, this is still acceptable though.

It WIP since I still need to do my own review

image

### DIFF
--- a/martin/martin-ui/src/components/catalogs/font.tsx
+++ b/martin/martin-ui/src/components/catalogs/font.tsx
@@ -1,0 +1,168 @@
+import { Eye, Search, Type } from 'lucide-react';
+
+import { ErrorState } from '@/components/error/error-state';
+import { CatalogSkeleton } from '@/components/loading/catalog-skeleton';
+import { Badge } from '@/components/ui/badge';
+import { Button } from '@/components/ui/button';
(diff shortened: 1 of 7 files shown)


martin/martin-ui/src/components/catalogs/font.tsx
```diff
@@ -0,0 +1,168 @@
...
+                  </div>
+                </div>
+                <div className="flex flex-col md:flex-row items-center gap-2 mt-4">
+                  <CopyLinkButton
+                    className="flex-1 bg-transparent w-full"
+                    link={buildMartinUrl(`/font/${name}/{range}`)}
+                    size="sm"
+                    toastMessage="Font link copied!"
+                    variant="outline"
+                  />
```
CommanderStorm, line 127: This is clearly not what users want.

not quite sure what users want instead 🤷🏻‍♂️
I doubt that they would want to get this URL.

image

nyurik (approved):
looks good overall!  Small styling improvements can be done either here or in another pr

PR #1955, opened 2025-07-26
feat(ui): Add sprite catalog without dialogs to the dashboard

CommanderStorm:
> [!IMPORTANT]
> This PR bases on https://github.com/maplibre/martin/pull/1954 Said PR should be reviewed first.

This PR implements the sprite catalog.

To remain managable, I have split this one in two.

image

PR #1956, opened 2025-07-26
feat(ui): Sprite catalog with dialogs for the dashboard

CommanderStorm:
> [!IMPORTANT]
> This PR bases on https://github.com/maplibre/martin/pull/1955 Said PR should be reviewed first.

This PR implements the full dialogs for the sprite catalog.

image

### DIFF
--- a/justfile
+++ b/justfile
@@ -64,6 +64,10 @@ biomejs-martin-ui:
 # Run integration tests and save its output as the new expected output (ordering is important)
 bless: restart clean-test bless-insta-martin bless-insta-mbtiles bless-frontend bless-int
 
+# Bless the frontend tests
+bless-frontend:
+    npm run test:update-snapshots
+
 # Run integration tests and save its output as the new expected output
 bless-insta-cp *args:  (cargo-install 'cargo-insta')
     cargo insta test --accept --bin martin-cp {{args}}
@@ -82,10 +86,6 @@ bless-int:
     rm -rf tests/temp
     tests/test.sh
     rm -rf tests/expected && mv tests/output tests/expected
-
-# Bless the frontend tests
-bless-frontend:
-    npm run test:update-snapshots
 
 # Build and open mdbook documentation
 book:  (cargo-install 'mdbook') (cargo-install 'mdbook-alerts')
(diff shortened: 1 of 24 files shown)


nyurik (approved):
same, looks good, styling cleanup can be done separate if wanted

nyurik:
pretty soon maputnik will just copy some of this code, or will want some independent lib that we will reuse in both :)

CommanderStorm:
I am not quite sure if this would fit into mapuntiks design language (this being a dialog).
Maybe adding the icons next to the text in maputniks selector might be possible though.

The code doing the fetching of sprites is simple enough.
In the case of SDF-Sprites, that becomes a bit more complicated and for maputnik you would need this.
For maputnik, you would also need to support those, since you can't just hand-wave them away via using png spritesheets.

With sdf support, likely a small lib would make sense.

@HarelM what do you think?

HarelM:
Not sure I'm following, I'm also not very familiar with martin and the are which you are discussing here.
The url which you put in the style.json and in Maputnik doesn't know about x2 or SDF.
The x2 will be attempted always with devices with high pixel ratio, and SDF definition is inside the relevant .json file next to the sprite.

So I'm probably missing something basic here. In any case, I'm sure you know what you are doing...

CommanderStorm:
Yes, that was a bad proposal and this should have been an issue instead.

The basic idea is "maybe maputnik should allow preview sprites".
Having this feature in the tile server is nice, but having it in the style-editor would be nicer.
You have worked with styles a lot more than I have, is this something that you would agree?

image

For regular sprites (1x/2x png), this is fairly straightforward. No lib needed.
For sdf sprites however, htis becomes a bit more complex.
If we were to integrate a sprite preview into maputnik and martin, one would need to handle sdf's. => a lib would make sense

CommanderStorm:
I think if such a feature makes sense boils down to if the expressions in poi layers are common or not. Something for which I don't have a good intuiton:

image

HarelM:
I don't have usage statistics, but an auto complete with icon preview would probably make a good user experience in maputnik. 
Isn't tinysdf the library you are taking about or does it need to do something different?

PR #1957, opened 2025-07-27
feat(ui): Add an Tile catalog to the dashboard

CommanderStorm:
This PR implements the tile catalog.

I have not split it like the sprite catalog since I think this is a bit simpler code and most importantly a lot less.

image


nyurik (approved):
thx!

PR #1959, opened 2025-07-27
Update Martin Logo

wayofthefuture:
per request of Yuri...


nyurik:
Thanks, I like the new look, but obviously this is not just my call, so would be happy to get some more feedback.

About the technical side:
* why is there 2 different PNG logo files that look similar?
* can we make them significantly smaller?  400kb seems excessive for such a simple image
* we probably would need a tiny favicon, probably just with the `M` symbol
* we probably would need a regular-sized image of just `M`, possibly with transparent background
* why do we need a PSD file? We probably will want to work with it until we settle on what we need, but i don't think we would want to store it in the repo, esp 2MB

once again, looks good, thanks!

CommanderStorm:
Really nice work ❤️ 

I have a few design comments

image

wayofthefuture:
Github noob here. It is however clear that our limitation is not ideology but time. I figured you guys would want to make modifications, hence the PSD, evidenced by @CommanderStorm's comments. If your vision is upward mobility it really sucks to lose the high-res copy... sorry didn’t realize size mattered so much outside of the build. I usually push at least 2X the desired resolution so it looks awesome on iPhones, that fine touch some notice. Readme.md currently provides for a large logo display. I pushed the favicon not sure why you aren’t seeing it. I will say whoever programmed the original SVG logo created a programmatic work of art. Thanks guys you are the rockstars of the internet. Anyway, back to my original point…

CommanderStorm:
Regarding size, this can be significantly reduced if you run it through https://tinypng.com/
After one run through that, we should be fine.

Regarding the PSD file, I agree with your reasoning. I want access to that, there will be a point in the future when we need it.
If possible, I would prefer an SVG. I don't have a photoshop subscription ^^

the `.ico` looks awsome, githubs UI is a bit confusing, no action needed there.

wayofthefuture:
Wow huge reduction in size... so much for Adobe "save for web" feature.

nyurik:
agree - I would love to  keep the SVG around, but definitely not PSD - as that requires a subscription that few of us have (or want to have :) )

wayofthefuture:
I did a little research and it appears that GIMP is the most popular free alternative to Photoshop. I was able to convert it to *.XCF which is GIMPs format. I can upload that as a replacement (and it's 50% smaller). If I convert it to SVG you will lose the following effect on the tiles:

image


Please let me know if the GIMP XCF will work.

CommanderStorm:
Gimp should work as well. Not prefered, since it is not vector based, but hey.

wayofthefuture:
Ok so I replaced the PSD with XCF and placed it into the same directory as logo.png. The png utility @CommanderStorm recommended reduced logo.png by about 300KB and the XCT reduced the editable version by about 1MB.

CommanderStorm:
Hmm.. Aparently there was a fuckup in communication on my side.
When I was thinking about "gimp would work" I was thinking about a non-destructivly layered gimp document with effects, not a png on a single layer.

If this is the intermediate, that does not work for me.
The issues that I noted above need to be fixed in this PR.
Without those issues being fixed, that would imo be a downgrade, even if we then don't have an icon.

Lets go back:
What effect would be lost if you export to svg?

wayofthefuture:
I can export the two layers separately and add them to the gimp file as separate layers. The text would be modifiable in vector format. However, the M would still remain as a rasterized image within the file (as a separate layer). If I tried to switch the M to SVG, all tiles would be the same fill color and it would basically look flat 1 color. Your bottom 2 pink arrows means that the top of the line would then not intersect exactly at the top right corner, and the only way to make it work would be to modify the slope of the triangle, which is bound to the slope of the inner legs of the M. Otherwise, here is a modified draft addressing some of your issues:

image

CommanderStorm:
Much better.

I think most of the color can be uniform.
Some tiles a bit lighter, some a bit darker, but we don't actually need an complex gradient likely.
The important part is being able to edit it and fit it into the places where we need that.

I most of those places actually I think that a black-white version of that logo might just work better. (or not? would need to check) 🤔 
Maybe even a subtle pulse animation..

For example, currently the frontpage would look like this.. If the `MARTIN`-Text on the right would be whitened and replace with the `Martin`-logo on the left, that would be better...

image

wayofthefuture:
I think the banner may be a different concept than the logo? See:

image

wayofthefuture:
Vectorizable:

image

nyurik:
sigh, i actually kinda liked the gradient color inside each square :(

wayofthefuture:
Lol. Me too, but whatever you guys want. I think you need two .xcf though, one for the banner and one for the logo.

nyurik:
but svg supports gradients, don't they??

nyurik:
but obviously we wouldn't want to do all of that by hand... i wonder if there is some tool for that, perhaps overlay a semitransparent image on top?

wayofthefuture:
I think the question is: will you ever need a size larger than the one originally provided? Otherwise, svg isn't needed IMO.

nyurik:
that's a good question... i think we need
* favicon (tiny, for serving)
* larger icon (mid-size, for docs/readme)
* large icon + name (large, for the book or the main site)
* both icon + named icon should work for dark and light backgrounds, i.e. the background must be transparent

All of these could be either JPG or PNG, whichever is smaller.  SVG on the other hand is like a wildcard for future maintenance - so that all others can be easily regenerated.    I suspect JPG does not support transparency, thus we might have to  use PNG

wayofthefuture:
JPEG does not support transparency. PNG is great for transparency. Old GIF did as well without the color support. Regardless, you said "book". You need vector based. I'll make it happen. Hold my beer.

wayofthefuture:
SVG:

m

nyurik:
Love it!!!  Unless anyone else objects, i'm all for using it in the above cases

CommanderStorm:
gosh darn it. so close. I have two things for the logo left that I would like to see changed:

- These corners are not rounded.
- There also seems to be a weird misalignment in trangle-angle here, which makes things feel "slightly crooked"

image

nyurik:
it is so good not to have perfect 20/20...

CommanderStorm:
For the banner, we should change the font to white, or at least pink.
Purple on dark purple is not very accessible..

image

CommanderStorm:
Regarding the font of "artin", I am not quite convinced, Helvetica-ish does not seem like a 100% fit next to the now quite nice M.
Helvetica always feels so "oldfashioned".

How about something like [`Anta`](https://fonts.google.com/specimen/Anta?preview.text=Martin%20MARTIN&preview.size=183&categoryFilters=Feeling:%2FExpressive%2FFuturistic;Sans+Serif:%2FSans%2FGeometric&script=Latn):

image

or maybe something more futuristic like [`Monomaniac One`](https://fonts.google.com/specimen/Monomaniac+One?preview.text=Martin%20MARTIN&preview.size=183&categoryFilters=Sans+Serif:%2FSans%2FRounded;Technology:!%2FTechnology%2F*;Feeling:%2FExpressive%2FFuturistic)

image

wayofthefuture:
https://fonts.adobe.com/fonts/nasalization

or

https://fonts.adobe.com/fonts/monomaniac-one

CommanderStorm:
Fonts are hard and without putting them into context you always choose the wrong one.

I trust your judgement on this one.
Your taste in this is likely better

wayofthefuture:
I also like Ubuntu Bold not sure how you feel about that:

https://fonts.adobe.com/fonts/ubuntu

CommanderStorm:
I like that much more than monomanic one.
Could definitely work.

Not sure of this or Anta works better, I trust your judgement on that one ^^
It is not Helvetica xD

nyurik:
is it me or is the distance between `M` and `artin` a bit too wide in the logo?

CommanderStorm:
And you are saying I am nitpicking? 😆

I don't feel strongly about that aspect.
What spacing works depends on the final font that gets chosen and what then feels good between the custom M and the regular font. ^^

wayofthefuture:
martin-new

wayofthefuture:
Had to delete last post the svg didn't load properly, here is new one:

martin-new

wayofthefuture:
Or:

martin-new - Copy

wayofthefuture:
The font decision is a hard one

CommanderStorm:
First one turns out to not work out, the font choice does not match the M enough for this.
=> for this, a lot of changes would need to happen on the M to make it not look like a different font. Lets not do that work.

For the second one, that defintively works better.
Lower uppercase feels a bit weird in a logo, especially if I zoom out.
I wonder:
- have you tried the lowercase variants and
- making the uppercase variant?

I keep wondering if Anta would not work out better. 
That bit futuristic style could either work really nicely or not. 🤔

If you were to attach the svg, I could give better feedback than the (likely inffuriating for you 😅) suggestion-ish mode.

wayofthefuture:
The lowercase was Anta. The original was Segoi UI fyi. I think the uppercase Anta at like 50-75% height will be the best. I attached the SVG will try to attach it a different way tomorrow when I get another chance to work on it.

wayofthefuture:
And yes I have tried many, many variants...

wayofthefuture:
75%, with slightly rounded corners! I think this is the best:
<br>
martin-anta
<br>
[martin-anta.zip](https://github.com/user-attachments/files/21593706/martin-anta.zip)

wayofthefuture:
well based on the silence i will assume you hated it lol... found another font will give this one last shot otherwise no worries i can pull the pr if u want:

nyurik:
@wayofthefuture my appologies - i was away on vacation, trying to desperately catch up.  The font looks good to me too!  And I def want to get at least the M in.  The last one actually is my fav?  Also, do make sure to advertise options in the martin slack channel to raise awareness

CommanderStorm:
The silence is mostly based on me being extremely buisy with an CPP-Practicum.

I would be happy both of the new options. They look great!

sharkAndshark:
> For the banner, we should change the font to white, or at least pink.
Purple on dark purple is not very accessible

I also worried about this. It looks a little bit somber and hard to read。

wayofthefuture:
@sharkAndshark For the banner wouldn't it be better to make the background transparent and use github's natural off-black instead? The blocky background looks kind of old to me... no block would look better IMO...

@nyurik I agree the last option is my fav too.

CommanderStorm:
I can't quite visualise how an transparent bg would look like here.

We would need both an image for githubs dark and the light mode though
=> one with dark font, and one with light.

If an image on GH has `#gh-dark-mode-only`/`#gh-light-mode-only` appened, it can be selectively triggered for only a certain theme.

If you think this is better, we should try it out ^^

wayofthefuture:
A good logo should place nicely on both dark and light backgrounds?

Github Light:
image

Github Dark:
image


Maybe brighten "artin" to same level as M to make it fit better on both? I do like the depth that the darker "artin" gives though...

A big dark purple box on a desired light theme doesn't look that great imo... a logo and banner that are versatile to both themes would be consistent and ideal i think

nyurik:
ok, i think we are good with these, lets use them!  And yes, lets brighten them a bit to match the logo.  @wayofthefuture are you on OSM-US slack?  I posted it in the martin channel

CommanderStorm:
The contrast of the dark purple to black is still pretty bad..
I think that we should have that background and change the main color to white, as this is the simplest fix.

Tbh, I don't see an issue with the purple box on white. Maybe I am getting old, asking on slack is likely a better call..
I am a junior dev with barely experience.

wayofthefuture:
Why should a background affect the logo just modify the background...?

nyurik:
I don't think we want to make non-transparent background.  Should be enough to make `artin` a bit lighter, reusing the lighter colors of `M`

wayofthefuture:
Ok guys, I did my best. I had to force push to get my work username out of there. You have commit access to make any changes you want. You'll notice logo.png in addition to logo.svg as there was the og social media tag so I think we needed both. Nevertheless I scrubbed sizes too. The favicon is in there. The martin front page was very hard to work with but I'm pretty sure I got it close. I placed image sources into a "graphics" folder as I'm a big presentation buff as you can tell imo deserves its own directory... Hope you like it! I can't spend any more time on this so I'm owwwwwwttt... :)

CommanderStorm:
Thanks a ton. ❤️ 

I will have a look if there is something left to do and do that work sometime in the later half of this/next week ^^

nyurik:
I made the "artin" a bit lighter (maybe too much?). I would be totally fine to merge this, but it seems there are other non-logo related HTML changes that prevent CI from passing. @CommanderStorm or @wayofthefuture, do you have any idea of why its failing?

CommanderStorm:
The error message does not seem related to the changes in this PR, lets rerun CI and see if this fixes it.

wayofthefuture:
Not too sure Cap! I went back and forth with the brightness endlessly and never did like the full bright. No idea why the build is failing. Your SVG reduced the stroke which I used for the impossible task of @CommanderStorm's desired rounded corners 🍭 . Remember it's just a logo guys souls only glance at it!

image

nyurik (approved):

nyurik:
We now also need to update the main site to use the new images

PR #1963, opened 2025-07-27
feat(ui) Style Catalog without the integration guide

CommanderStorm:
This PR breaks out the style catalog from #1897

image
image


martin/martin-ui/src/components/dashboard-content.tsx
```diff
@@ -1,7 +1,9 @@
 import { type ErrorInfo, useEffect, useState } from 'react';
```
CommanderStorm:
```suggestion
import { type ErrorInfo, useEffect, useState, useCallback } from 'react';
```

nyurik (approved):
nice

PR #1967, opened 2025-07-28
chore: refactor how `get_unrecognized_keys` is handled to enable unused config detection

CommanderStorm:
During https://github.com/maplibre/martin/pull/1926 I noticed that the way we currently handle this is not super working out.

- `get_unrecognized(&self) -> &UnrecognizedValues` is not owned => if I need to merge, this is not quite possible
- `UnrecognizedValues` is not quite the information that I need. If I just need the keys, we should just ship around the keys
- some of the functions related to this are not documented
- `is_default` related to this is always `true`.


martin/src/cog/config.rs
```diff
@@ -16,8 +16,8 @@ pub struct CogConfig {
 }
 
 impl ConfigExtras for CogConfig {
-    fn get_unrecognized(&self) -> &UnrecognizedValues {
-        &self.unrecognized
+    fn get_unrecognized_keys(&self) -> UnrecognizedKeys {
+        self.unrecognized.keys().cloned().collect()
```
nyurik: why do we need to clone it in every fn?  Why not just return the ref to the original, and if needed, the caller can do the `.keys().cloned().collect()` on the result of the `get_unrecognized()`?

CommanderStorm: I need this to be owned as otherwise this trait does not work on every level (think enums, or a struct with two sub-structs).
Especially at the top level this otherwise does not work

I considered an iterator (I can chain them, not owned) but this neither works nicely with the prefixes, nor enums.

I doubt that cloning for the warning case is a huge issue.
This is one malloc/free per warning more.

nyurik (approved):

PR #1970, opened 2025-07-30
fix(ci): re-enable caches for `x86_64-apple-darwin`

CommanderStorm:
I missed this in the previous PR.

I just assumed that all runners are roughly equivalently fast.
They are not. Not by a long shot. MacOS x86 is lagging behind the others by a lot.

In the previous PR I somehow did not noticed this.

If we enable Caching just for this one, our CI speeds up again.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,11 +2,11 @@ name: CI
 
 on:
   push:
-    branches: [main]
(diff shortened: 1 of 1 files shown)


nyurik (approved):

nyurik (approved):
lol, note to self - don't review from the phone

PR #1972, opened 2025-07-30
chore: Feature gate tests with dependencies

### LINKED ISSUE #1739: Make all PG-requiring tests feature-gated
We need to simplify our testing process, and that means that `cargo test` should pass without any external dependencies, i.e. without requiring PG to be running, or some env var to be set.

I think PostgreSQL tests are the only complex ones, so lets mark them all as requiring some `test-pg` feature (name?).  Note that `cargo test --all-features` should still include the postgreSQL tests, and thus require some connection string to be set.  This will be done by CI, but will make development simpler.

todtb:
Resolves https://github.com/maplibre/martin/issues/1739

### DIFF
--- a/martin/Cargo.toml
+++ b/martin/Cargo.toml
@@ -66,6 +66,8 @@ pmtiles = ["dep:aws-config", "dep:pmtiles"]
 postgres = ["dep:deadpool-postgres", "dep:enum-display", "dep:json-patch", "dep:postgis", "dep:postgres", "dep:postgres-protocol", "dep:regex", "dep:rustls-native-certs", "dep:rustls-pemfile", "dep:semver", "dep:tokio-postgres-rustls"]
 sprites = ["dep:spreet", "tokio/fs"]
 styles = ["dep:walkdir", "tokio/fs"]
+test-pg = ["postgres"]
+test-pmtiles = ["pmtiles"]
 webui = ["dep:actix-web-static-files", "dep:static-files", "dep:walkdir"]
 
 [dependencies]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,6 +75,8 @@ jobs:
           cargo test --package mbtiles --no-default-features
           cargo test --package mbtiles
           cargo test --package martin
+          cargo test --features "test-pg" --no-default-features
+          cargo test --features "test-pmtiles" --no-default-features
           cargo test --doc
         env:
           DATABASE_URL: postgres://${{ env.PGUSER }}:${{ env.PGUSER }}@${{ env.PGHOST }}:${{ job.services.postgres.ports[5432] }}/${{ env.PGDATABASE }}?sslmode=require
--- a/martin/tests/pg_function_source_test.rs
+++ b/martin/tests/pg_function_source_test.rs
@@ -1,4 +1,4 @@
-#![cfg(feature = "postgres")]
+#![cfg(feature = "test-pg")]
 
 use ctor::ctor;
 use indoc::indoc;
--- a/martin/tests/pg_server_test.rs
+++ b/martin/tests/pg_server_test.rs
@@ -1,4 +1,4 @@
-#![cfg(feature = "postgres")]
+#![cfg(feature = "test-pg")]
 
 use actix_http::Request;
 use actix_web::http::StatusCode;
--- a/martin/tests/pg_table_source_test.rs
+++ b/martin/tests/pg_table_source_test.rs
@@ -1,4 +1,4 @@
-#![cfg(feature = "postgres")]
+#![cfg(feature = "test-pg")]
 
 use ctor::ctor;
 use indoc::indoc;
--- a/martin/tests/pmt_server_test.rs
+++ b/martin/tests/pmt_server_test.rs
@@ -1,3 +1,5 @@
+#![cfg(feature = "test-pmtiles")]
+
 use actix_web::http::header::{ACCEPT_ENCODING, CONTENT_ENCODING, CONTENT_TYPE};
 use actix_web::test::{TestRequest, call_service, read_body, read_body_json};
 use ctor::ctor;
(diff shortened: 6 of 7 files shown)


martin/Cargo.toml
```diff
@@ -66,6 +66,8 @@ pmtiles = ["dep:aws-config", "dep:pmtiles"]
 postgres = ["dep:deadpool-postgres", "dep:enum-display", "dep:json-patch", "dep:postgis", "dep:postgres", "dep:postgres-protocol", "dep:regex", "dep:rustls-native-certs", "dep:rustls-pemfile", "dep:semver", "dep:tokio-postgres-rustls"]
 sprites = ["dep:spreet", "tokio/fs"]
 styles = ["dep:walkdir", "tokio/fs"]
+test-pg = ["dep:deadpool-postgres", "dep:enum-display", "dep:json-patch", "dep:postgis", "dep:postgres", "dep:postgres-protocol", "dep:regex", "dep:rustls-native-certs", "dep:rustls-pemfile", "dep:semver", "dep:tokio-postgres-rustls"]
```
CommanderStorm:
```suggestion
test-pg = ["postgres"]
```

martin/Cargo.toml
```diff
@@ -66,6 +66,8 @@ pmtiles = ["dep:aws-config", "dep:pmtiles"]
...
+test-pmtiles = ["dep:aws-config", "dep:pmtiles"]
```
CommanderStorm:
```suggestion
test-pmtiles = ["pmtiles"]
```

CommanderStorm (commented):
I did not quite read that this is still a draft when I started looking at it.

We now need to re-enabable them in ci. 
Otherwise they won't get executed.

todtb:
> I did not quite read that this is still a draft when I started looking at it.

I was looking over the `ci.yml` while letting this run to see if anything was affected there.

Yes, the `ci.yml` as far as I can tell targets specific features.

> We now need to re-enabable them in ci. Otherwise they won't get executed.

Was there recent work to disable them in ci/cd? Something I can easily restore?

CommanderStorm:
What I meant by that is that the `test-*` feature-gated tests are now not in `default` anymore.
So we need to re-add them to the section of the justfile executing them. => activating them.

Note that packages are not features. (the package `mbtiles` != the feature `mbtiles`, yes confusing)

todtb:
I was referring more to the CI at this point. Since then I've added this alongside the other `cargo test` commands:

```yml
          cargo test --features "test-pg" --no-default-features
          cargo test --features "test-pmtiles" --no-default-features
```

As for the `justfile` I was thinking the `--all-targets` would handle it, but it sounds like you're saying we need to add `--all-features` as well. 

I guess the question then is: would this be the right place/just command to include that for a typical user? How are you using this `justfile` currently in day to day development? Given that `just test` will run `test-int` I think it's fair to assume we should be running the `test-pg` and `test-pmtiles` as well, but maybe you have other thoughts?

```just
# Run all tests using a test database
# TODO add --all-features as well?
test: start (test-cargo '--all-targets') test-doc test-frontend test-int

# Run Rust unit tests (cargo test)
test-cargo *args:
    cargo test {{args}}
```

CommanderStorm:
`test` should specifically not run the postgres/nginx tests. Otherwise the testing process is not really simpler (or at least I don't think it is, right?)

What @nyurik suggested in another issue is to move things into `test-ci` and `test`.
`start` and `(test-cargo '--all-features --all-targets')` can thus be included there.

What is important is that we also have CI further down in the pipeline.
For example in [Test postgis:11-3.0-alpine sslmode=disable](https://github.com/maplibre/martin/actions/runs/16634947303/job/47074423059?pr=1972#logs)
where we test if the sslmode of postgres works as expected.

todtb:
> `test` should specifically not run the postgres/nginx tests. Otherwise the testing process is not really simpler (or at least I don't think it is, right?)

I guess I don't know what the typical testing process is. I'm running `just test` (including all DB and integration tests) locally and don't really find it that burdensome. Sure it takes a few seconds longer, but I could also just use `cargo test "name"` or the equivalent if I was only focused on one part of code. Back to what I was saying earlier, what would somebody normally run to test locally that their changes didn't introduce regressions?

I don't want to take up too much of your time talking through that though, and I'm probably slowing down whatever vision you have for where this goes from here. Happy to step back from this one if that's the case. I'm also entirely new to rust and its paradigms as well as the project obviously, so maybe there's patterns established here I'm not aware of.

I can, however, help in a separate issue/pr with adding in additional `OSTYPE` support for `test/test.sh` which I slightly modified locally to run on macos. Currently it guards against running on anything but linux.

CommanderStorm:
I don't fully have a strong oppinion on this. @nyurik could you help us allgn? 

I expect of our contributors to not execute the testcases locally as I don't.
We have a lot of them, that takes ages and CI is faster than my laptop[^1] anyhow.

[^1]: Yes, I should likely upgrade that one. I have been putting off migrating for a few years now.

So my take would be
- `just test` => just the local unit tests + clippy + doc tests => everything that does not depend on an external system.
- `just test-ci` => `just test` + the other steps that currently lives next to it in CI
- `just test-int` => the end to end tests, but also the `test-*` integration tests.

todtb:
Just to follow up here, the reason I started looking into this was because in the course of adding SRID 4326 support, I wanted to run _all_ tests (including those with dependencies and including integration tests in `test.sh`) locally.

I was able to do so by making a few small uncommitted code changes and the full test suite ran fairly quickly. I was ultimately hoping to be able to contribute those changes and allow others who wish to run such tests in a development lifecycle.

It seems like there are broader unresolved discussions to be had here though (particularly around which tests are run in CI and which tests are run in locally in the course of typical work). For what it's worth, my personal opinion is that CI should run _all_ tests available to best protect against regressions. But in any case, I will close this for now.

If we're interested in being able to run `just test` locally across multiple platforms/`OSTYPE`s perhaps a separate issue is in order.

Issue #1977, opened 2025-07-31
Allow configuring Martin to serve under a subpath for Traefik and proxy compatibility


CommanderStorm:
Reference:
- https://github.com/maplibre/martin/discussions/1184

Currently, to run Martin behind a Traefik proxy with a subpath (e.g., `/tiles`), the following configuration is required:

```yaml
...
  martin:
    image: ghcr.io/maplibre/martin:latest
    restart: unless-stopped
    command:
      - --config
      - /map/config.yaml
    labels:
      - "traefik.enable=true"
      - "traefik.http.routers.martin.entrypoints=websecure"
      - "traefik.http.routers.martin.tls.certresolver=letsencrypt-prod"
      - "traefik.http.routers.martin.rule=Host(`example.com`) && (PathPrefix(`/tiles`) || PathPrefix(`/_/`))"
      - "traefik.http.routers.martin.middlewares=martin-stripprefix@docker"
      - "traefik.http.middlewares.martin-stripprefix.stripprefix.prefixes=/tiles"
      - "traefik.http.services.martin.loadbalancer.server.port=3001"
```

Currently, the `base_path` configuration option in Martin only affects the TileJSON URLs, not the actual API endpoints:

```yaml
# Set TileJSON URL path prefix. This overrides the default of respecting the X-Rewrite-URL header.
# Only modifies the JSON (TileJSON) returned, martins' API-URLs remain unchanged. If you need to rewrite URLs, please use a reverse proxy.
# Must begin with a `/`.
# Examples: `/`, `/tiles`
base_path: /tiles
```

There is currently no setting to configure the actual path under which Martin listens and serves the frontend.
This makes it difficult to serve Martin under a subpath using a proxy like Traefik, and causes compatibility issues with some frontend setups.

lexhuismans:
In my situation I am having Cloudfront > Lambda (martin) > S3. In Cloudfront I have set-up a behavior to route `/tiles` to the Lambda (martin). However, I get `404`'s in the Lambda logs, which I think is due to the `/tiles/<tile_name>.pmtiles` prefix.

I think the above feature request would also solve this situation of hosting Lambda (martin) behind Cloudfront on a subpath.

If there are any workarounds I would be interested to hear them. One option I believe is to have Lambda@edge deployed to rewrite the path, but I would rather avoid doing so. A configuration option in Martin would be very neat.

CommanderStorm:
I think this this makes sense to add and should be very simple to contribute as noted above 😉

[...]

PR #1978, opened 2025-07-31
fix(ui): improve subpath deployment support

CommanderStorm:
This improves https://github.com/maplibre/martin/issues/1977 a bit and makes this at least work on the baseline for people running martin on a base-path:

```diff
-      - "traefik.http.routers.navigatum-martin.rule=Host(`nav.tum.de`) && (PathPrefix(`/tiles`) || PathPrefix(`/_/`) || PathPrefix(`/catalog`))"
+      - "traefik.http.routers.navigatum-martin.rule=Host(`nav.tum.de`) && (PathPrefix(`/tiles`) || PathPrefix(`/_/`))"
```

The issue is mainly, that otherwise one would need to confiugre every source in traefik.. not pheasable for lots of our users.
By using `href` instead of `origin` this works as users expect.

### DIFF
--- a/martin/martin-ui/__tests__/lib/api.test.ts
+++ b/martin/martin-ui/__tests__/lib/api.test.ts
@@ -20,7 +20,7 @@ describe('getMartinBaseUrl', () => {
 
     // window.location.href is "http://localhost"
     const result = getMartinBaseUrl();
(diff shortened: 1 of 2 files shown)


nyurik (approved):
lgtm

PR #1995, opened 2025-08-01
feat(martin-cp): better error message for invalid coodinate bboxes

### LINKED ISSUE #1991: `martin-cp` raises an `assertion failed: min_y <= max_y` if called with bad bboxes instead of an explanatory error message
As I noticed in https://github.com/maplibre/martin/issues/1987, when a user passes `0.829000,72.004000,55.827000,137.834000` as a bbox, we abort with an assertion.

Said bbox is not possible, since bboxes follow the schema:

```rust
Bounds {
        left: 0.829,
        bottom: 72.004,
        right: 55.827,
        top: 137.834,
    }
```

so by that logic, that bbox would go past the 85° northwards, which is not possible.

We should abort with an error message explaining that this bounding box does not make sense instead.

> [!NOTE]
> This is amplified by some tooling like http://bboxfinder.com/#0.829000,72.004000,55.827000,137.834000 which has this in their URLs..

CommanderStorm:
This PR adds a new error message for the case when one tries to call `martin-cp` with an defective bbox.

I did not add an e2e test for this as I think this is a simple enough to remain at the parameter validation step

image

Resolves https://github.com/maplibre/martin/issues/1991


martin/src/bin/martin-cp.rs
```diff
@@ -246,6 +270,10 @@ enum MartinCpError {
         "More than one source found, please specify source using --source.\nAvailable sources: {0}"
     )]
     MultipleSources(String),
+    #[error(
+        "{0} of bounding box '{1}' must fit into {2:?}. Please check that your bounding box is in the `min_lon,min_lat,max_lon,max_lat` format."
+    )]
+    InvalidBoundingBox(String, String, RangeInclusive<f64>),
```
nyurik: why strings?  that requires allocation. For .0, use `&'static str`, and for the second string - perhaps the copy of a bounding box (if possible?)

CommanderStorm: I don't think this super matters, but good call

nyurik (approved):
thx!

PR #1996, opened 2025-08-02
chore(martin-cp)!: change the concurrencly of martin-cp to be the number of cpus, not `1`

### LINKED ISSUE #1990: warn if `martin-cp`s `--concurrency` is `1` (the default) and if we think more cores were benefitial
I think that the current setting of `martin-cp`s' `--concurrency` is a bit contra-intuitive.

In all other cases, we try to be highly performant. For this however, we don't use multiple cores.
I think we should be more smart about this if the parameter is not set and default to `num_cpu` instead.

This shifts the load to the database and makes our tool not be the bottleneck.
This comes at increased potentially increased memory usage, wo we would need to call this out in the release notes.

Alternatively, we can set this to `--pool-size` which is default 20.

CommanderStorm:
When I originally looked at martin-cp, I thought it was super slow. The reason is that we by default only use a single thread, somethign that does not really go hand in hand with our goals.

We should really use multiple threads here and I think the number of CPUs is a good bet.

Resolves https://github.com/maplibre/martin/issues/1990

> [!WARNING]
> Yes, this is a breaking change and yes this needs to be put at the top of the release notes with such a warning-alert.

### DIFF
--- a/martin/src/bin/martin-cp.rs
+++ b/martin/src/bin/martin-cp.rs
@@ -90,8 +90,7 @@ pub struct CopyArgs {
     /// Allow copying to existing files, and indicate what to do if a tile with the same Z/X/Y already exists
     #[arg(long, value_enum)]
     pub on_duplicate: Option<CopyDuplicateMode>,
-    /// Number of concurrent connections to use.
-    #[arg(long, default_value = "1")]
+    #[arg(help = format!("Number of concurrent connections to use. [DEFAULT: {}]", num_cpus::get()), long)]
     pub concurrency: Option<usize>,
     /// Bounds to copy, in the format `min_lon,min_lat,max_lon,max_lat`. Can be specified multiple times. Overlapping regions will be handled correctly.
     #[arg(long)]
@@ -306,7 +305,7 @@ fn check_sources(args: &CopyArgs, state: &ServerState) -> Result<String, MartinC
 #[allow(clippy::too_many_lines)]
 async fn run_tile_copy(args: CopyArgs, state: ServerState) -> MartinCpResult<()> {
     let output_file = &args.output_file;
-    let concurrency = args.concurrency.unwrap_or(1);
+    let concurrency = args.concurrency.unwrap_or(num_cpus::get());
 
     let source = check_sources(&args, &state)?;
 


nyurik:
i am not so certain this is a good default. The issue is that this is usually IO bound -- either waiting for PG to return the tile (so the value would be the number of PG server CPUs, possibly with a multiplier), or disk bound (e.g. mbtiles or pmtiles as a source). In either case, core count is mostly useluss

CommanderStorm:
When I tested it, I noticed quite significant uplift in performance.
I saw neither martin, nor PG using the full systems resources.

Slamming PG with more paralelel requests is sadly somethign that one has to do as otherwise a lot of performance is left on the table.
See the htop in this blog post to show the difference between dbms => https://cedardb.com/blog/ode_to_postgres/

Either way, you really want more than one request ongoing with any of the sources that we have.
With modern SSDs + Storage being what it is, using one thread to request data is not enough to max out the bandwith by a long shot.
- For PG it should be clear that one would want have multiple requests in all stages of flight at all times.
- In terms of Cloud storage bandwith, even with `io_uring`, you are still spending 50% of your CPU on networking.
- Local might be different. That highly depends on the size of data read and storage medium it is stored on.
  - Modern SSDs are wickedly fast, but they benefit from paralell outstanding requests, so under our current model this would also benefit from multiple requests.
  - If your database is still on spinning rust (first of all, why?) this is bad for performance due to the seeking

Maybe just documenting this better (https://github.com/maplibre/martin/pull/1989) is enough, but I feel that the default here is too low.
Maybe increasing this just slightly would be a good middle ground.
How about `roundup(num_cpu / 4)` as an compromise? That should not increse memory too much and increase the performance for the users who don't configure away from the detfaults.

nyurik:
@CommanderStorm I am **not** saying it should be 1. The problem is that the number has nothing to do with the local CPU count.

I think what we should do instead is simply warn the users that the concurrency is set to 1, and that users should configure it based on fairly complex heuristic (i.e. the warning should give a link to some doc page)

CommanderStorm:
fair enough, warning is good enough => lets regroup

PR #2011, opened 2025-08-05
update readme to include auto assign

ajita-asthana:


nyurik:
@ajita-asthana this should be part of the #1924 - each PR should be "self-contained" as much as possible - e.g. so it can be reverted all together

CommanderStorm:
Yes, lets merge it into https://github.com/maplibre/martin/pull/1924 instead.
I am going to close this one as having two PRs open for the same thing seems unnessesary.

PR #2023, opened 2025-08-08
 chore(ci): pre-install `cargo-binstall` in CI until `cargo-sort` installation works upstream better

nyurik:


CommanderStorm:
This issue is specifically, that the version which gets installed currently does not get updated correctly.
It is not a complile fail, but I would like to not experience the issues that likely come with this diff.

We can obviously reopen once that issue gets resolved upstream

nyurik:
@CommanderStorm not sure i understood - is that an issue with the tool or with the github action or with what?

nyurik:
Also, if there is an issue with installation, perhaps we should install `cargo-binstall` as part of the CI - this way justfile will install it faster

.github/workflows/ci.yml
```diff
@@ -47,7 +47,7 @@ jobs:
     steps:
       - run: rustup update stable && rustup default stable
       - uses: taiki-e/install-action@v2
-        with: { tool: just }
+        with: { tool: 'just,cargo-sort' }
```
```suggestion
        with: { tool: 'just,cargo-binstall' } # todo: add cargo-sort, once https://github.com/taiki-e/install-action/pull/997 is resolved
```

CommanderStorm:
The issue is that when we add it to the action the "last non-broken" version gets installed.
The latest version is not actually broken though.

CommanderStorm (approved):

PR #2027, opened 2025-08-10
feat(martin-cp): add a warning if `--concurrency 1` and an error if `--concurrency 0`

### LINKED ISSUE #1990: warn if `martin-cp`s `--concurrency` is `1` (the default) and if we think more cores were benefitial

CommanderStorm:
This PR implements a slightly more smart algorithm for detecting if we should warn of `--concurrency 1`.

If we are copying locally, having more threads is likely not very effective, unless we have to re-encode.
I am not warning of the re-encoding case to not overcomplicate things for this PR review wise.

image

Becausse of the nature of this functionality, I did not attach tests

Resolves #1990


martin/src/pmtiles/mod.rs
```diff
@@ -172,6 +172,12 @@ macro_rules! impl_pmtiles_source {
             fn clone_source(&self) -> TileInfoSource {
                 Box::new(self.clone())
             }
+            
+            fn benefits_from_concurrent_scraping(&self) -> bool {
+              // having multiple http requests in flight is beneficial
+              // when using local disks, the benefits are less clear
+              stringify!($backend) != "MmapBackend"
+            }
```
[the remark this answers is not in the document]
CommanderStorm: LOL

martin/src/srv/server.rs
```diff
@@ -265,6 +265,10 @@ pub mod tests {
             Box::new(self.clone())
         }
 
+        fn benefits_from_concurrent_scraping(&self) -> bool {
```
nyurik: if the source already has a default impl, there is no need to implement it here or in bench.rs

CommanderStorm: If there are very few usages of this, I prefer to be explicit in how things work.
It is not like this is an complex implementation

nyurik (approved):

martin/src/srv/server.rs
nyurik: i think more code is worse here - any boilerplate `{ false }` in case of a test just gets in a way.  At the same time, having a comment explaining why it is needed that way is important - so ok to have them there. Merging...

PR #2030, opened 2025-08-12
docs: change demo readme to use `https`

lukasalexanderweber:


CommanderStorm:
No, this won't work.
It is not a public site with letsencrypt/..

lukasalexanderweber:
@CommanderStorm thanks for your feedback. `http` doesn't work at all (at least for me).. Yes, `https` is insecure, but at least I can access the demo.

CommanderStorm:
I guess, yes.
You are correct.. I am not happy about this though.

CommanderStorm (approved):
I guess, this is better than before. Not ideal though.

lukasalexanderweber:
If you can give me a bit more insight in how you set up your demo maybe we find a better sollution with working http or secure https

CommanderStorm:
Sorry

@nyurik will need to give that context. I was not part of the project when the docker-compose for this part was setup.

To unblock you:
I just run `npm run dev` from the frontend directory if I need to make changes. 🤷🏻‍♂️ if this is correct, but it works good enough.

nyurik:
i am so confused by this - usually the demo.html file would be opened via `file://....` protocol.  https shouldn't work locally

CommanderStorm:
not `tests/demo.html`, the `/demo`

PR #2048, opened 2025-08-17
chore(core): migrate environment tracking and testing (2/n)

CommanderStorm:
This PR migrates the environment support into `martin_core`.


martin-core/src/config/env.rs
```diff
@@ -0,0 +1,122 @@
...
+
+    fn get(&'a self, key: &str) -> Option<Self::Value> {
+        self.0.borrow_mut().insert(key.to_string());
+        std::env::var(key).ok()
+    }
+}
+
+/// Test implementation with configurable environment variables.
+#[derive(Debug, Default)]
+pub struct FauxEnv(pub std::collections::HashMap<&'static str, OsString>);
```
CommanderStorm, line 77: apparently, `#[cfg(test)]` acts as if the struct is not public any more.
So strange.

martin-core/src/config/env.rs
```diff
@@ -0,0 +1,122 @@
...
+///
+/// Tracks which variables are accessed via [`VariableMap`] using interior mutability.
+#[derive(Debug, Default)]
+pub struct OsEnv(RefCell<HashSet<String>>);
+
+impl Env<'_> for OsEnv {
+    fn var_os(&self, key: &str) -> Option<OsString> {
+        #[allow(unused_qualifications)]
+        std::env::var_os(key)
+    }
```
CommanderStorm, line 59: This sees to bypass the internal hashset for tracking.
Sees like a bug.

nyurik (approved):

martin/src/utils/tests.rs
```diff
@@ -1,62 +0,0 @@
...
-#[must_use]
-pub fn some(s: &str) -> Option<String> {
-    Some(s.to_string())
-}
-
-#[allow(clippy::unnecessary_wraps)]
-#[must_use]
-pub fn os(s: &str) -> OsString {
-    OsString::from(s)
-}
```
CommanderStorm, line 19: I inlined these in https://github.com/maplibre/martin/pull/2048/commits/3fccfa80b338acb6bb63cded8be6d64e88b1fdc5 as otherwise I would need to move them util too `martin_core` too or do other things (f.ex. exact feature-gate) to pass CI.

Yes, the code is a few chars longer this way, but it is imo more clear what is happening for new devs.

PR #2052, opened 2025-08-18
chore: use let-if chains

nyurik:


martin/src/cog/image.rs
```diff
@@ -155,10 +155,10 @@ fn ensure_pixels_valid(
                     //   So the visual result is the same, but the component values are different.
 
                     let value = data[(idx_chunk + component_idx) as usize];
-                    if let Some(v) = nodata {
-                        if value == v {
-                            continue 'outer;
-                        }
+                    if let Some(v) = nodata
+                        && value == v
```
CommanderStorm: 🤔 I think this should be the following instead

```suggestion
                    if nodata == Some(value)
```

martin/src/pg/builder.rs
```diff
@@ -126,10 +126,10 @@ impl PgBuilder {
         let mut pending = Vec::new();
         for (id, cfg_inf) in &self.tables {
             // TODO: move this validation to serde somehow?
-            if let Some(extent) = cfg_inf.extent {
-                if extent == 0 {
-                    return Err(InvalidTableExtent(id.to_string(), cfg_inf.format_id()));
-                }
+            if let Some(extent) = cfg_inf.extent
+                && extent == 0
+            {
```
CommanderStorm: I think the following would be identical

```suggestion
            if cfg_inf.extent == Some(0) {
```

CommanderStorm (approved):
LGTM, two oppertunities to simplify further but nothing big

PR #2063, opened 2025-08-23
docs: Reformat config example comments to remove overlong lines

### LINKED ISSUE #2062: Reformat config page comments
The [config doc](https://maplibre.org/martin/config-file.html) page uses comments that are too long - resulting in the need to horizontally scrolling. We should reformat all comments there to keep things more readable.

Image

furmak331:
Fixes #2062 .
I wrapped long comment lines at 80 characters and big paragraphs into smaller chunks .
No content or meaning was changed

### DIFF
--- a/docs/src/config-file.md
+++ b/docs/src/config-file.md
@@ -107,9 +107,8 @@ postgres:
 
   # Specify how bounds should be computed for the spatial PG tables [default: quick]
   # 'calc' - compute table geometry bounds on startup.
-  # 'quick' - same as 'calc', but the calculation will be aborted if it takes
-  # more than 5 seconds.
-  # 'skip' - do not compute table geometry bounds on startup.
+  # - `quick` same as 'calc', but the calculation will be aborted after 5 seconds.
+   # - `skip` does not compute table geometry bounds on startup.
   auto_bounds: quick
 
   # Enable automatic discovery of tables and functions.
--- a/docs/src/config-file.md
+++ b/docs/src/config-file.md
@@ -138,8 +138,7 @@ postgres:
   # If a list of strings is given, the first found column will be treated as a
   # feature ID.
       id_columns: feature_id
-  # Boolean to control if geometries should be clipped or encoded as is,
-  # optional, default to true
+  # Controls if geometries should be clipped or encoded as is [default: true]
       clip_geom: true
   # Buffer distance in tile coordinate space to optionally clip geometries,
   # optional, default to 64
--- a/docs/src/config-file.md
+++ b/docs/src/config-file.md
@@ -108,7 +108,9 @@ postgres:
   max_feature_count: null # either a positive integer, or null=unlimited (default)
 
   # Specify how bounds should be computed for the spatial PG tables [default: quick]
-  # 'calc' - compute table geometry bounds on startup.
+   #
+   # Options:
+   # - `calc` compute table geometry bounds on startup.
   # - `quick` same as 'calc', but the calculation will be aborted after 5 seconds.
    # - `skip` does not compute table geometry bounds on startup.
   auto_bounds: quick
--- a/docs/src/config-file.md
+++ b/docs/src/config-file.md
@@ -26,7 +26,7 @@ keep_alive: 75
 # The socket address to bind [default: 0.0.0.0:3000]
 listen_addresses: '0.0.0.0:3000'
 
-## Set TileJSON URL path prefix.
+# Set TileJSON URL path prefix.
 # This overrides the default of respecting the X-Rewrite-URL header.
 # Only modifies the JSON (TileJSON) returned; Martin's API-URLs remain unchanged.
 # If you need to rewrite URLs, please use a reverse proxy.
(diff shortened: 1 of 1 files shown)


docs/src/config-file.md
```diff
@@ -26,8 +26,10 @@ keep_alive: 75
 # The socket address to bind [default: 0.0.0.0:3000]
 listen_addresses: '0.0.0.0:3000'
 
-# Set TileJSON URL path prefix. This overrides the default of respecting the X-Rewrite-URL header.
-# Only modifies the JSON (TileJSON) returned, martins' API-URLs remain unchanged. If you need to rewrite URLs, please use a reverse proxy.
+## Set TileJSON URL path prefix.
```
```suggestion
# Set TileJSON URL path prefix.
```

docs/src/config-file.md
```diff
@@ -38,10 +40,15 @@ worker_processes: 8
 # Amount of memory (in MB) to use for caching tiles [default: 512, 0 to disable]
 cache_size_mb: 1024
 
-# If the client accepts multiple compression formats, and the tile source is not pre-compressed, which compression should be used. `gzip` is faster, but `brotli` is smaller, and may be faster with caching.  Default could be different depending on Martin version.
+# If the client accepts multiple compression formats, and the tile source is not
+# pre-compressed, which compression should be used. `gzip` is faster, but
+# `brotli` is smaller, and may be faster with caching. Default could be
+# different depending on Martin version.
```
CommanderStorm: Let's split the first sentence, it is quite long

Let's also align indents

```suggestion
  # Which compression should be used if the
  # - client accepts multiple compression formats, and
  # - tile source is not pre-compressed.
  # 
  # `gzip` is faster, but `brotli` is smaller, and may be faster with caching.
  # Default could be different depending on Martin version.
```

docs/src/config-file.md
```diff
@@ -38,10 +40,15 @@ worker_processes: 8
...
 preferred_encoding: gzip
 
-# Enable or disable Martin web UI. At the moment, only allows `enable-for-all` which enables the web UI for all connections. This may be undesirable in a production environment. [default: disable]
+# Enable or disable Martin web UI. At the moment, only allows `enable-for-all`,
+# which enables the web UI for all connections. This may be undesirable in a
+# production environment. [default: disable]
```
CommanderStorm: Let's move information around a bit on this one.

```suggestion
# Enable or disable Martin web UI. [default: disable]
#
# At the moment, only allows `enable-for-all`, which enables the web UI for all connections.
# This may be undesirable in a production environment
```

docs/src/config-file.md
```diff
@@ -107,31 +119,37 @@ postgres:
...
-      # If a list of strings is given, the first found column will be treated as a feature ID.
+  # A table column to use as the feature ID
+  # If a table has no column with this name, `id_column` will not be set for
+  # that table.
+  # If a list of strings is given, the first found column will be treated as a
+  # feature ID.
       id_columns: feature_id
-      # Boolean to control if geometries should be clipped or encoded as is, optional, default to true
+  # Boolean to control if geometries should be clipped or encoded as is,
+  # optional, default to true
```
```suggestion
  # Controls if geometries should be clipped or encoded as is [default: true]
```

docs/src/config-file.md
```diff
@@ -81,22 +88,27 @@ postgres:
...
+  # included in the tile and the result will look "cut off"/incomplete.
+  # This feature allows you to put a maximum latency bound on tiles with an
+  # extreme amount of detail at the cost of not returning all data.
+  # It is sensible to set this limit if you have user generated/untrusted
+  # geodata, e.g. a lot of data points at [Null Island]
+  # (https://en.wikipedia.org/wiki/Null_Island).
   max_feature_count: null # either a positive integer, or null=unlimited (default)
 
   # Specify how bounds should be computed for the spatial PG tables [default: quick]
   # 'calc' - compute table geometry bounds on startup.
```
CommanderStorm: Maybe the following is more online with how the rest is formatted.

```suggestion
   #
   # Options:
   # - `calc` compute table geometry bounds on startup.
```

docs/src/config-file.md
```diff
@@ -81,22 +88,27 @@ postgres:
...
-  # 'quick' - same as 'calc', but the calculation will be aborted if it takes more than 5 seconds.
+  # 'quick' - same as 'calc', but the calculation will be aborted if it takes
+  # more than 5 seconds.
   # 'skip' - do not compute table geometry bounds on startup.
```
```suggestion
  # - `quick` same as 'calc', but the calculation will be aborted after 5 seconds.
   # - `skip` does not compute table geometry bounds on startup.
```

CommanderStorm (changes requested):
Currently this PR changes the indentation of comments in a few places. Could you re-add that indentation?
I feel that if comment and value is indented the same it makes it more obvious that they belong together ^^

furmak331 (commented):
Thanks for the documentation suggestions. Definitely helped

CommanderStorm (approved):
Nice, thanks a ton ♥️

furmak331:
Glad it helped :) Looking forward to contributing more

PR #2065, opened 2025-08-24
docs: improve doc comment for `IdResolver::resolve` to remove ambiguitiy if we should log for reserved names

CommanderStorm:
This fixes the confusion on slack if we log for reserved keywords by documenting this fact better.


martin/src/utils/id_resolver.rs
```diff
@@ -22,9 +22,34 @@ impl IdResolver {
         }
     }
 
-    /// If source name already exists in the self.names structure,
-    /// try appending it with ".1", ".2", etc. until the name is unique.
-    /// Only alphanumeric characters plus dashes/dots/underscores are allowed.
+    /// Makes sure that every source has a unique, non-reserved name
+    ///
+    /// Replace non-alphanumeric characters or dashes/dots/underscores with dashes
```
CommanderStorm:
```suggestion
    /// Replace non-alphanumeric characters or dashes/dots/underscores with dashes.
```

nyurik (approved):

PR #2089, opened 2025-08-27
test(martin-cp): add a testcase for composite sources in `martin-cp`

CommanderStorm:
Primary motivation: Test if our impl is broken in this part => I have not noticed anything in this direction.

This ensures that the feature noted in https://github.com/maplibre/martin/discussions/2087 does not break.


nyurik (approved):
thanks!!

PR #2090, opened 2025-08-27
docs(martin-cp): add a comment that composite sources are supported for martin-cp

CommanderStorm:


nyurik (approved):

docs/src/martin-cp.md
```diff
@@ -23,6 +23,10 @@ martin-cp  --output-file tileset.mbtiles \
            postgresql://postgres@localhost:5432/db
 ```
 
+> [!TIP]
+> Next to regular sorces, `--source <SOURCE>` does support [composite sources](sources-composite.md).
+> This means `martin-cp` can be used to merge two different sources into one `mbtiles` archive.
+
 > [!TIP]
```
CommanderStorm:
```suggestion
> This means `martin-cp` can be used to merge two different sources into one `mbtiles` archive.

If performance is a concern, you should also consider

 > [!TIP]
```

PR #2097, opened 2025-08-28
fix(cp): make sure that the cache is always disabled for martin-cp

CommanderStorm:
In the previous attempt, I removed the CLI option for this, as caching tiles does not make sense for tile copying.

- https://github.com/maplibre/martin/pull/2026

I forgot that we can also be configured by the config file.


nyurik:
I am not so sure about this one - if copying from a remote pmtiles source, cache is used for directory caching - you don't want to re-request pmtile indexes on each tile request

nyurik:
we may need to rethink the whole caching strategy. I think i wrote some of these thoughts in some other ticket, but not sure

* for serving, we need to cache the final tile, sprite, and font results - thus speeding up common tile serving. These are "complex processing -> single URL"
  * we can cache this pre-compression or after-compression -- because the client may want brotli vs gzip - so we can cache the result "as-is" (from the backend), and/or "as-served" (in the compression requested by the client)
  * composite results should be able to re-use the cache
* for pmtiles we need to cache indexes - internal metadata per tile source, whose results are shared by multiple tile requests
* for contour / hillshading, we may have to cache internal elevation map which could be used for generating two outputs

CommanderStorm:
Even if this makes sense for PMTiles directories, it does not make sense to cache PMTiles' tiles.

We definitively need to decouple these caches as pmtiles directory caching imo never makes sense to not have - assuming that we implement the ETAG based invalidation.

CommanderStorm:
Yea, what you said. I don't like to have so many refactorings ongoing at the same time. I would like to push this refactoring after `martin_core`.

Not sure if this brings us closer or farther from the caching-refactoring.

nyurik:
sure, but that still leaves the question of martin-cp caching - it seems at least for now we should allow it, or else we are killing pmtiles access performance. Overall for caching, i would prefer a single cache (max MB), but with the ability to configure what gets cached

Issue #2054, opened 2025-08-19
Generate MVT tiles from GeoJSON


nyurik:
Per discussion in #1458, in order to support arbitrary geometry data providers (e.g. databases), we need Martin to be able to convert arbitrary geometries into tiles.  GeoJSON is the simplest data provider, and thus can be used as the first step.

## MVT encoding
Build/find a crate to encode arbitrary geometries as MVT. This is similar to what [mapbox/wagyu](https://github.com/mapbox/wagyu) library does (used by [ST_AsMVT](https://postgis.net/docs/ST_AsMVT.html) in Postgis).

* [ ] convert geometry to webmercator
* [ ] scale / offset to the correct tile coordinates
* [ ] clip geometries to the tile boundaries + configurable buffer 
* [ ] fix any invalid geometries resulting from converting coordinates to integers or clipping

## Serve GeoJSON

* [ ] Decide GeoJSON tile source configuration - probably similar to .mbtiles files + extras like buffer size
* [ ] Decide CLI-provided geojson sources

styrowolf:
Would it be possible to use/adapt a port of geojson-vt from mapbox, such as https://github.com/maxammann/geojson-vt-rs, for this?

CommanderStorm:
Seems like a reasonable starting point.
There seem to be a few TODO comments in there which might be dicy, but we can contribute PRs for the issues we find.

@maxammann what do you think about this?

maxammann:
Yeah definitely. The post is done and tested. There is no compatibility layer on top of it to make it more rust idiomatic but PRs welcome.

Or a fresh crate that just builds a nice API on top of it.

PR #2098, opened 2025-08-28
feat(geojson): add mvt tiles generation from GeoJSON

styrowolf:
This is a quickly hacked PR to implement MVT tile generation from GeoJSON files (#2054). I used [geojson-vt-rs](https://github.com/maxammann/geojson-vt-rs) to slice GeoJSON into tiles and integrated the `mvt` module from [geozero](github.com/georust/geozero) for encoding those tiles in MVT format. I needed access to some of the internals from `geozero`'s `mvt`, which were not public -- that's why I decided to vendor in the module independently.

There are still rough edges and lots of TODOs left on this PR. In my initial testing, the tile endpoint was abnormally slow. I did not yet get around to why that is the case. 

I also had to remove `#![forbid(unsafe_code)]` from the library for an unsafe trait impl in the `mvt` module. It might be a better idea to just break out that module into a seperate crate.

I would like to get guidance on how to move forward with this PR. Let me know what works and what does not. Thank you for your time.


martin/src/geojson/mod.rs
```diff
@@ -0,0 +1,117 @@
...
+
+mod config;
+mod mvt;
+
+pub use config::GeoJsonConfig;
+
+pub struct GeoJsonSource {
+    id: String,
+    path: PathBuf,
+    inner: Arc<Mutex<GeoJSONVT>>,
```
CommanderStorm: I am not a fan of Mutex for this.

If we can, an RWLock would be much better.

martin/src/geojson/mod.rs
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    fn clone_source(&self) -> TileInfoSource {
+        // TODO: implement clone for GeoJsonSource (GeoJSONVT is not cloneable, which is a blocker)
+        Box::new(Self::new(self.id.clone(), self.path.clone()).unwrap())
+    }
+
+    fn benefits_from_concurrent_scraping(&self) -> bool {
+        // TODO: figure out if this is a good idea
+        false
```
CommanderStorm: If we switch to `RWLock`, the answer should be `true`. If we cannot, the answer is `false` with a reference to the `Mutex`

```suggestion
        // Extracting information from JSON is fairly CPU intensitve
        // => we can use more IO by using more threads
        true
```

nyurik:
this is great, thank you!  One of us will definitely review it shortly. Being a fairly complex feature, i'm sure there will be some back and forth about it, so please don't be dissuaded if it doesn't get merged right away :)

martin/build.rs
```diff
@@ -101,9 +101,46 @@ fn webui() {
...
+        let mut file = std::fs::OpenOptions::new()
+            .write(true)
+            .truncate(true)
+            .open("src/geojson/mvt/vector_tile.rs")?;
+        file.write_all("// This file was automatically generated through the build.rs script, and should not be edited.\n// Remove this file to force a rebuild.\n\n".as_bytes())?;
+        file.write_all(buffer.as_bytes())?;
+    }
+
+    // As the proto file is checked in, the build should not fail if the file is not found
+    Ok(())
```
CommanderStorm: In build scripts, the usual guidance of "don't unwrap" does not apply.
For them, you want to use `.expect` instead, as returning an uncatched error is effectively the same and retains the context of where the error occured.

martin/src/geojson/mod.rs
```diff
@@ -0,0 +1,117 @@
...
+        let tile_info = TileInfo::new(Format::Mvt, martin_tile_utils::Encoding::Uncompressed);
+        let geojson_file =
+            File::open(&path).map_err(|e: std::io::Error| FileError::IoError(e, path.clone()))?;
+
+        // TODO: better error handling
+        let geojson = geojson::GeoJson::from_reader(geojson_file)
+            .map_err(|e| FileError::InvalidFilePath(path.clone()))?;
+
+        // TODO: see if options default is goood enough
+        let inner = GeoJSONVT::from_geojson(&geojson, &Options::default());
```
CommanderStorm: There are a few options which might be nice to have not hard-coded => added to `GeoJsonConfig` and passed to `GeoJsonSource::new`.
We can just expose these options in the config file.

I am not sure if `index_max_points` works like our existing option, or what that index is.

```rust
Options {
    max_zoom: 18,          // max zoom to preserve detail on; can't be higher than 24
    index_max_zoom: 5,    // max zoom in the tile index
    index_max_points: 100000, // max number of points per tile in the tile index
    generate_id: false,     // whether to generate feature ids, overriding existing ids
    tile: TileOptions {
        tolerance: 3.,     // simplification tolerance (higher means simpler)
        extent: 4096,        // tile extent
        buffer: 64,        // tile buffer on each side
        line_metrics: false, // enable line metrics tracking for LineString/MultiLineString features
    }
}
```

More a personal interest:
I would be interested if this changes the performance of the code as the compiler now can no longer constant-fold things.

martin/src/geojson/mod.rs
```diff
@@ -0,0 +1,117 @@
...
+
+        let tilejson = tilejson! {
+            tiles: vec![],
+            minzoom: 0,
+            maxzoom: 18,
+        };
```
CommanderStorm: This looks a bit sparse. I am not sure if this tilejson works, as it does not specify what is contained features.. 🤔

martin/src/geojson/mod.rs
```diff
@@ -0,0 +1,117 @@
...
+    tile_info: TileInfo,
+}
+
+impl Debug for GeoJsonSource {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "GeoJsonSource {{ id: {}, path: {:?} }}",
+            self.id, self.path,
+        )
```
CommanderStorm: nit: I think this is a bit more rusty. Your version is equivalent functionality wise ^^.

```suggestion
        f.debug_struct("GeoJsonSource")
         .field("id", &self.id)
         .field("path", &self.path)
         .finish()
```

martin/src/geojson/mod.rs
```diff
@@ -0,0 +1,117 @@
...
+        let mut builder = LayerBuilder::new(self.id.clone(), 4096);
+        for feature in &tile.features.features {
+            builder.add_feature(feature);
+        }
+        let mvt_layer = builder.build();
+        let mvt_tile = mvt::vector_tile::Tile {
+            layers: vec![mvt_layer],
+        };
+
+        return Ok(mvt_tile.encode_to_vec());
```
CommanderStorm: nit: I don't like explicit `drop`, this can too seasily get buggy => lets use a scope instead.

```suggestion
        {
          let mut guard = self.inner.lock().await;
          let tile = guard.get_tile(xyz.z, xyz.x, xyz.y).clone();
        }

        let mut builder = LayerBuilder::new(self.id.clone(), 4096);
        for feature in &tile.features.features {
            builder.add_feature(feature);
        }
        let mvt_tile = mvt::vector_tile::Tile {
            layers: vec![builder.build()],
        };

        Ok(mvt_tile.encode_to_vec())
```

martin/Cargo.toml
```diff
@@ -143,10 +176,18 @@ tokio-postgres-rustls = { workspace = true, optional = true }
 url.workspace = true
 walkdir = { workspace = true, optional = true }
 xxhash-rust.workspace = true
+# feat: geojson
```
CommanderStorm: We alphabetically sort these => lets remove this and run `just fmt-toml` afterwards.

This also occurs a few more times in this file

```suggestion
```

CommanderStorm (changes requested):
Firstly, this is nice work 👍🏻
Thanks for tackling this!
This is very straightforward code you wrote.

> I needed access to some of the internals from geozero's mvt, which were not public

Instead of effectively forking, let's try to just make the fields public that we need.
For an experimental impl, current version is fine.
For getting merged, I would like to see a PR to make the values we need `pub` in `geozero`.

> In my initial testing, the tile endpoint was abnormally slow. I did not yet get around to why that is the case.

Just throwing it out there: did you run it in release mode?. This burnt me once 😉

To benchmark things, we have the following two functionalities. If you think performance could be an issue, benchmarking it is probably a good idea.

https://github.com/maplibre/martin/blob/872a0e734afabe22dc394bb542818013737e449e/justfile#L41-L59

> I would like to get guidance on how to move forward with this PR.
> Let me know what works and what does not.

There are three larger aspects:

Testing. We try to test what we can. This prevents unwanted surprises later on and makes working generally more pleasant. Lets 
- add integration tests in `test.sh`
  - add a minimal GeoJSON fixture with a polygons, points, and lines  (each with one over tile border, one in tile border)
  - test if it does the splitting at tile borders works as expected
  - if the options (see inline comment) are correctly passed.

- add unit tests for that
  - `file.geojson` CLI argument parsing works as expected
  - the config file is serialised as expected
  - since there is not much "actual code" (= code which can have bugs without the dependency having a bug), I don't think we need much in terms of unit testing

Upstreaming the `geozero`-`pub` change necessary.
Vendoring (=forking) is a last resort, and we should not do this, unless as a last resort.
You can change where a dependency points to via `git = "..", branch = ".."`.

Adjusting the existing documentation (README and /docs, for the individual source and the config file respectively)
=> we should tell people about the capability you are adding 😉

martin/build.rs
styrowolf: I set up a branch of geozero now with the mvt module exposed and fully removed my changes from build.rs

styrowolf:
Thanks for the quick review. I've taken out `mvt` module code and replaced it with `geozero` with its internals exposed (I put the dependency from my fork for now to restructure the code, I'll put the PR in and I'll replace the dep when the changes are upstream).

martin/src/geojson/mod.rs
styrowolf: I looked at the `cog` module's one and took that as an example. That's just like this `let tilejson = tilejson! {
            tiles: vec![],
            minzoom: min_zoom,
            maxzoom: max_zoom
        };`

I'll look into this.

martin/src/geojson/mod.rs
styrowolf: The problem is `GeoJSONVT`'s `get_tile` method takes a `&mut self`, so I don't think it would make a huge difference. The `GeoJSONVT` struct internally caches tiles in a HashMap, as tiles at higher zoom are created by splitting tiles at lower zooms. The `&mut self` makes it so that we can get one tile at a time, as far as I understand.

I can use [`geojson_to_tile()`](https://docs.rs/geojson-vt-rs/0.1.1/geojson_vt_rs/fn.geojson_to_tile.html) directly but that means losing the caching. Another option would be to move the RwLock into `GeoJSONVT` itself -- not sure 100%.

@maxammann is my understanding correct? what do you think?

CommanderStorm: Yea, using `GeoJSONVT` then does not quite work for us then.

We already have a multithreaded cache outside of this function
=> lets use [`geojson_to_tile`](https://docs.rs/geojson-vt-rs/0.1.1/geojson_vt_rs/fn.geojson_to_tile.html) with either an `Arc` or `Arc<RWLock<GeoJSON>>`
=> nicely paralell.

Max does not need paralellisability as wasm does not have good multithreading support.

martin/src/geojson/mod.rs
styrowolf: Added the required vector_layers field as per https://github.com/mapbox/tilejson-spec/blob/master/3.0.0/README.md#332-fields. I viewed it in Maputnik editor and it seems to be working. We should probably add bbox etc. some time.

styrowolf:
@CommanderStorm I've replaced the GeoJSONVT struct with the function. [geojson_to_tile](https://docs.rs/geojson-vt-rs/0.1.1/geojson_vt_rs/fn.geojson_to_tile.html) reprojects the entire dataset on every call, which might be a significant overhead. Maybe we could do a PR to that crate so that we can project once and keep that as state in `GeoJsonSource`.

CommanderStorm:
Yea, reprojecting on every tile would not be great for performance

styrowolf:
Current TODOs for this PR
- [x]  Expose `geojson_vt_rs::TileOptions` to `GeoJsonConfig`
- [x]  upstream and use MVT feature writing in geozero https://github.com/georust/geozero/pull/264
- [x]  Return info for more TileJSON fields (especially bounds)
- [x]  Reproject once, tile multiple times https://github.com/maxammann/geojson-vt-rs/pull/3
- [ ]  Add GeoJSON error type and fix error handling
- [ ]  Add GeoJSON fixtures
- [ ]  Unit tests
- [ ]  Integration tests
- [ ]  Benchmarking

CommanderStorm:
> there are multiple ways of accomplishing this

I would prefer option 1 of using `geozero` to do the feature writing, if this is okay with you.
Having this code live here does seem fairly hacky.
This code living in `geozero` has ecosystem advantages for other people too 😉

I would prefer `geozero` over `mvt`, but I have not looked into this. That is mostly based around size of the community and "trust"/having talked with the people running it.

I know that this is a big ask.
If you want me to take over some part (besides encoding, you have dug into this already ^^), feel free to task me away.
I currently have semester breaks.

styrowolf:
> I would prefer geozero over mvt, but I have not looked into this. That is mostly based around size of the community and "trust".

I fully understand and agree with your point, though `mvt` also seems like a pretty popular crate with around 74K downloads -- the file headers also indicate `Copyright (c) 2019–2024 Minnesota Department of Transportation`.

> I would prefer option 1 of using `geozero` to do the feature writing, if this is okay with you. Having this code live here does seem fairly hacky. This code living in `geozero` has ecosystem advantages for other people too 😉

I agree with you on the ecosystem advantages part. I've briefly looked over some `geozero` code to see how to approach this. I'm currently thinking of implementing something like a `MvtLayerWriter` which implements `FeatureProcessor`.

> If you want me to take over some part (besides encoding, you have dug into this already ^^), feel free to task me away.
I currently have semester breaks.

Thank you so much! I really appreciate the help. I’ll stay in touch as I put together a basic plan to move forward.

CommanderStorm:
Since writing docs is something that takes iterations, I thought you might appreciate me doing the legwork and writing up a first rough draft to iterate on later => https://github.com/maplibre/martin/commit/c3b1728abc861f974dc4b08bcdcf93be4dd3163e

styrowolf:
I'm currently waiting on two PRs: geozero MVT feature-writing https://github.com/georust/geozero/pull/264 and geojson-vt-rs preprocessed geojson https://github.com/maxammann/geojson-vt-rs/pull/3. I will update this PR as those PRs move forward.

styrowolf:
Both of those PRs got merged, so I'll try to continue work on this again.

CommanderStorm:
We have done a bit of light refactoring to support the v1 release via `martin-core`.
I have merged the changes into your PR, a few files have changed where they are located. Nothing big.

styrowolf:
Thanks Frank. I just pushed a few (rough) commits to show how the code is with the new PRs. I still have to test, add fixtures, and improve error handling. The function naming is a bit rough too. Looking forward to what you think.

docs/src/sources-geojson.md
```diff
@@ -0,0 +1,30 @@
...
+
+You may also want to generate a [config file](config-file.md) using `--save-config my-config.yaml`.
+The config file can then be used via the `--config my-config.yaml` option.
+
+> [!WARNING]
+> Serving these files is less efficient compared to [PMTiles or MBTiles](sources-files.md).
+> This is because to serve GeoJSON, martin needs to:
+>
+> - parse JSON
+> - reproject geometrys
```
sdf-jkl:
```suggestion
> - reproject geometries
```

CommanderStorm:
I rebased it again as our changes (should now be mostly done) in main were kind of touching this code.

I and @nyurik discussed this offline and came to the conclusion that
- vendoring geozero, or
- your original version are good.

The current path that we all hoped/assumed we can get geozero to make a release via internal chanels is apparently more a longer term plan 😞

We can also rename the feature `geojson-unstable` (default-disabled) and merge the done parts (f.ex. configuration) to reduce the PRs' scope.

styrowolf:
My academic term has finished, so I have more time to work on this PR. I merged the main branch and did some refactoring. Let me know how can I continue, or whether we should shelve the PR.

Cargo.toml
```diff
@@ -55,6 +55,9 @@ env_logger = "0.11"
 flate2 = "1"
 flume = "0.11"
 futures = "0.3"
+geojson = "0.24.2"
+geojson-vt-rs = "0.2.0"
+geozero = { git = "https://github.com/georust/geozero.git", rev = "9975851" } # TODO: put version in when geozero cuts new release
```
CommanderStorm, line 60: @styrowolf FYI, we finally got this dependency released.
Sorry for the long time

```suggestion
geozero = "0.15.1"
```

CommanderStorm:
lets tackle this via https://github.com/maplibre/martin/pull/2538 instead

Thanks for doing the work in this PR and sorry that it did not lead to a merge.

PR #2112, opened 2025-09-03
fix(pg): join of `geometry_columns` with `pg_class` should take into account not only the table names but al…

danielgerigk:
…so the table schemas.


CommanderStorm:
So basically this fixes the table `foo` existing in namespaces `baz` and `bar`?

danielgerigk:
@CommanderStorm yes, exactly.

CommanderStorm:
Oof, yikes.

Ideally, we should add an integration test to ensure that this  is handled correctly and that we don't accidentally regress.
If you want me to take this over, this would be ok too.

Our integration tests are under /test and for this one creating an schema and a same-name table should be all that is required to test this case

nyurik:
yeah, would be great to get a simple integration test and test the catalog output

danielgerigk:
@CommanderStorm that would be great if you could take care of the integration test(s).

CommanderStorm:
Shure, will do.
I would need to push this into the first half next week due to  vacation reasons, but if this is blocking you I have my laptop with me ^^

danielgerigk:
@CommanderStorm np, this is not blocking me/us at all. actually, this bug didn't even come up as an actual bug in our martin server deployment, but i saw the bug in the sql code when i inspected that query. the reason why i inspected the query was that it takes quite a long time in our db to run - about 15 minutes - because we have about 70000 tables and/or views...

CommanderStorm:
That is a lot.
Does this change improve this?

danielgerigk:
no, this "bug" is completely unrelated to the query running so long in our db.
the guys who designed our db had the great idea to create a view for every single column of every table in our `public` schema. not sure about the reason why they did that. recently i saw that it is possible to define a pg function that takes a json as an argument (for example, specifying which table and columns the caller is interested in) and returns an mvt layer containing exactly those columns. i don't have much experience with martin, but i assume we could use that one function instead of the 60000+ views. or am i mistaken?
right now, every time we restart the martin, we have to wait for 30 mins to be able to get layers served again. that's why they deployed a second martin server that takes over while the first one restarts... 🤣

CommanderStorm:
Could you run `explain`,  `explain timing` and `explain verbose` this query and share the results?

Maybe there are low hanging fruits in this query where pg is not smart or something else is obvious.

CommanderStorm:
For security reasons allowing arbitrary table access via the function you talked about seems.. suboptimal.

60k is also not that much, we must have a performance bug in the query somewhere.

danielgerigk:
sure, you'd have to check and approve the table given as argument first before returning any data.

i'll run `explain` on the query later today (probably in the evening) and post my results here.
that would indeed be great if that query could be improved to not take 15 mins. 👍

danielgerigk:
i ran the query again, manually, without restarting martin, and now the query takes only about 7 secs. yesterday when i ran that same query it took 15 mins. so now my guess is that there is another, more complicated reason why the query took so long. either some kind of deadlock that we will have to investigate, or the roughly 10,000 mat views that i stumbled upon that all don't have indices on their geometry columns... 🤔
so i think the query itself is fine.
but if we find out the reason why our martin restarts take so long, and if the reason is related to this query, then i'll write a note here.

CommanderStorm (approved):
Thanks for the PR and helping to reproduc the bug.

@nyurik FYI, the following happens without this:

```
Checking for no other warnings or errors in the log
[2025-09-13T15:12:29Z WARN  martin::pg::query_tables] Unexpected duplicate table schema_a.table_and_view_two_schemas.a_geom
[2025-09-13T15:12:29Z WARN  martin::pg::query_tables] Table schema_b.table_and_view_two_schemas has no spatial index on column a_geom
[2025-09-13T15:12:29Z WARN  martin::pg::query_tables] Unexpected duplicate table schema_b.table_and_view_two_schemas.a_geom
Log file target/test_logs/auto.txt has unexpected warnings or errors
```

martin/tests/pg_table_source_test.rs
```diff
@@ -132,7 +132,7 @@ async fn table_source() {
     geometry_column: geog
     bounds:
       - -2
-      - 0
+      - -1
```
CommanderStorm:
```suggestion
      - 0
```

danielgerigk:
see this pull request for further discussion: https://github.com/maplibre/martin/pull/2257

danielgerigk:
> i ran the query again, manually, without restarting martin, and now the query takes only about 7 secs. yesterday when i ran that same query it took 15 mins. so now my guess is that there is another, more complicated reason why the query took so long. either some kind of deadlock that we will have to investigate, or the roughly 10,000 mat views that i stumbled upon that all don't have indices on their geometry columns... 🤔 so i think the query itself is fine. but if we find out the reason why our martin restarts take so long, and if the reason is related to this query, then i'll write a note here.

actually, now i know that what i ran here was not the v0.13.0 version of the query, but the v0.18.0 version.

PR #2130, opened 2025-09-08
chore: use Path instead of PathBuf for strs

nyurik:


CommanderStorm (approved):
I did not know that `Path::new` exists, but makes sense that it does.

PR #2149, opened 2025-09-12
chore: add a testcase for multiple schemas with the same name

CommanderStorm:
This is a testcase that asserts the case that #2112 claims it fixes.

No matter how much I debug, I can't find a case in which our existing logic fails.
@danielgerigk could you have a look if I have missed a case?
What does https://github.com/maplibre/martin/pull/2112 fix precicely?


danielgerigk:
one situation where the fixed query gives a different result compared to the original query is the following:
```
CREATE SCHEMA baz;
CREATE TABLE baz.foo (geom GEOMETRY);

CREATE SCHEMA bar;
CREATE VIEW bar.foo AS (SELECT * FROM baz.foo);
```

nyurik (approved):
Thanks!!

PR #2151, opened 2025-09-12
feat(config): Implement unrecognised value in config file warning

CommanderStorm:
this PR pushes us towards having unrecognised value in the config file warnings.


tests/expected/configured/save_config.yaml
```diff
@@ -2,6 +2,12 @@ cache_size_mb: 8
 keep_alive: 75
 listen_addresses: localhost:3111
 worker_processes: 1
+cors:
```
nyurik, line 5: do we want to actually output `cors` here if there are no configurations?  Esp when we have `max_age: null` style params

nyurik: oh, this is being printed because you modified the input, never mind

CommanderStorm: And I need to modify the input to have the `warning: 'due to not expected'`, yes.

Would doing this in a separate test be better? Not sure.

nyurik (commented):
seems like a lot of boilerplate code... anyway to dedup?

CommanderStorm:
We could use a derive macro, but that did not quite seem like a much better idea.
At minimum, this would mean another crate for this abstraction, as dreive macros are proc-maros

nyurik (approved):

PR #2152, opened 2025-09-12
chore(config): unify the `unrecognized`-handling

CommanderStorm:
This PR pulls the handling of `unrecognized` from https://github.com/maplibre/martin/pull/2151 as this is kind of unrelated and litters in that PR a bit.


nyurik (approved):
i'm ok with this, but i do have one concern - do we actually want to skip serializing unrecognized values?  The use case of `--config inp.yml --save-config outp.yml` would remove all values it cannot recognize... I guess it should be ok

CommanderStorm:
I think not serialising them is less confusing.. Otherwise this would be more confusing and get people down the wrong debugging path.

PR #2155, opened 2025-09-14
fix(ui): inspections of image sources not working

CommanderStorm:
image

Beforehand, this was just white which is not very satisfactory.

Thanks @tombh for reporting the bug


nyurik (approved):

nyurik:
hmm, coverage seems to keep failing - anything we can do?

CommanderStorm:
I have not had the time to fully investigate why codecov is acting up.

PR #2158, opened 2025-09-14
chore(core): change the errors of the Source trait to be dyn based

CommanderStorm:
The `Source` trait currently returns `MartinError`s.
This ties this trait to the `martin`-crate as I don't want to move the errors.

Moving the erros would be pointless as others cannot add an error here.
I benchmarked the performance with the existing bench and the performance is about the same (-1.5%, but that is likely just noise)

image


nyurik:
Perhaps we should move the errors to core, and add a boxed variant for extensibility, but keep all the core-features as is for perf and clarity reasons?

CommanderStorm:
I am not a huge fan of this.

First of all, because I am not entirely buying your "clearer API" claims 😉 
And secondly, because that essentially means that the replacement PR will be ginormous as I cannot reference error types across the crate boundary as would be needed.

Also regarding performance, I think I have shown that the good case is not affected.
Here is the updated benchmaked also showing the bad side:

image

How about this compromise: we go with my solution for a few PRs, then switch to your solution for the performance boost in the error case.

nyurik:
as mentioned in another PR - if we wrap errors in a box, they become opaque - it won't be possible for the caller to act on specific error details - i.e. they would have to parse the error text after formatting it into a string - defeating the whole idea of the `thiserror`... we may as well just switch to anyhow

CommanderStorm:
Doing `martin-core` with `thiserror` is the better API.
You have convinced me on the clairity aspect.

I would still like to progress with the `Box<dyn Error>` for a few PRs as this makes them much simpler.

nyurik (approved):

PR #2167, opened 2025-09-14
chore(core): move `Source` to the core

CommanderStorm:
This PR builds on https://github.com/maplibre/martin/pull/2158 and moves the `Source`-trait to the core.

> [!IMPORTANT]
> - https://github.com/maplibre/martin/pull/2158  needs to be reviewed first


nyurik (commented):
are you introducing boxed error here? I am a bit uncertain about this yet - lets keep that to a separate pr maybe?

CommanderStorm:
it already is a separate PR..

martin-core/src/tiles/source.rs
```diff
@@ -0,0 +1,77 @@
...
+    /// Retrieves tile data for the given coordinates.
+    ///
+    /// # Arguments
+    /// * `xyz` - Tile coordinates (x, y, zoom)
+    /// * `url_query` - Optional query parameters for dynamic tiles
+    async fn get_tile(
+        &self,
+        xyz: TileCoord,
+        url_query: Option<&UrlQuery>,
+    ) -> Result<TileData, Box<dyn std::error::Error>>;
```
nyurik: but you have it here and elsewhere?

CommanderStorm:
Ah, now I see what you mean.

What I meant is that 
- https://github.com/maplibre/martin/pull/2158 is said separate PR and needs to be reviewed first

The alternative would be to have a thiserror-enum in martin-core, but that enum would also have to have at least one `Box<dyn Errror>` variant to be extendable.
Not ideal and seems messy.

Lets keep that discussion in https://github.com/maplibre/martin/pull/2158 though

nyurik (approved):

PR #2168, opened 2025-09-14
chore(bench): add an benchmark that tests the impact of the error variant

CommanderStorm:
This benchmark is to check the claims in #2158 more thorughly


nyurik:
@CommanderStorm do note that the main issue with the boxed error implementation is that we loose the error type when boxed - so the caller no longer can analyze what caused the error - they can only print it

nyurik (approved):

PR #2171, opened 2025-09-14
chore(core): migrate mbtiles to the core

CommanderStorm:
This needs https://github.com/maplibre/martin/pull/2167 merged first


martin-core/src/tiles/error.rs
```diff
@@ -1,5 +1,18 @@
 /// A convenience [`Result`] for tiles coming from `martin-core`.
-pub type MartinCoreResult<T> = Result<T, MartinCoreError>;
+pub type MartinCoreResult<T> = Result<T, BoxedMartinCoreError>;
 
 /// Temporary error type for integration purposes.
-pub type MartinCoreError = Box<dyn std::error::Error>;
+pub type BoxedMartinCoreError = Box<dyn std::error::Error>;
```
nyurik, line 2: logically, error should be before the result because it uses it, but no biggie

nyurik (approved):

PR #2176, opened 2025-09-15
chore(core): inline `InfoMap` and remove `json_to_hashmap`

CommanderStorm:
This PR tries to start untangling the dependency tree of the `pg` module with how it is configured.

### DIFF
--- a/martin/src/config/file/pg/config_function.rs
+++ b/martin/src/config/file/pg/config_function.rs
@@ -1,11 +1,13 @@
+use std::collections::BTreeMap;
+
 use serde::{Deserialize, Serialize};
 use tilejson::{Bounds, TileJSON};
 
 use super::config::PgInfo;
 use crate::config::file::UnrecognizedValues;
-use crate::pg::utils::{InfoMap, patch_json};
+use crate::pg::utils::patch_json;
 
-pub type FuncInfoSources = InfoMap<FunctionInfo>;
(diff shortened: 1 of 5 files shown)


martin/src/config/file/pg/config_function.rs
```diff
@@ -5,7 +5,7 @@ use tilejson::{Bounds, TileJSON};
 
 use super::config::PgInfo;
 use crate::config::file::UnrecognizedValues;
-use crate::pg::utils::patch_json;
+use crate::config::file::pg::utils::patch_json;
```
nyurik, line 8: kinda weird that we have `config/file/pg`  -- i would have thought that `/file` is for pmtiles/mbtiles/... other file-based sources, whereas `config/pg` is explicit for pg. Not a biggie, just a bit confusing

nyurik (approved):

martin/src/config/file/pg/config_function.rs
CommanderStorm: My reasoning is that instead of splitting by config type (mbtiles, Pg, ..) we split by config format (file, args, env) first.

The reason is that not everything can be configured every way and it is slightly more organised this way.
In the end, if you think the other way works better, we can still change that, but let's finish one refactoring first.

I also don't think of those two as so closely related that they should share..

PR #2179, opened 2025-09-15
chore(core): move the cache to the core

CommanderStorm:
I tried experimenting with replacing the cache, but that was not very successfull.

This PR just moves the cache to the core, which enables pmtiles to be moved there as well in two PRs.
pmtiles currently has parts of its state in the config file, which is suboptimal.


martin-core/src/lib.rs
```diff
@@ -1,6 +1,6 @@
 #![doc = include_str!("../README.md")]
 #![forbid(unsafe_code)]
-#![deny(missing_docs)]
+#![forbid(missing_docs)]
```
nyurik: these configurations should probably be done as part of Cargo.toml.  Also, never use `forbid` for things like this -- it should be a WARNING, and that warning should be treated as error in the CI, but we do not want to require each new function to have documentation while you are developing.

nyurik (approved):

PR #2186, opened 2025-09-16
chore(core): move the `PgPool` to the core

CommanderStorm:
after merging https://github.com/maplibre/martin/pull/2185, the pool can be moved to the core.

This PR is either to be reviewed by just looking at the last commit or by waithing for the dependent PRs to be merged.

PR #2191, opened 2025-09-16
chore(core): move postgres' `PgPool` and `PgSource` to the core

CommanderStorm:
For simplicity, this bases off https://github.com/maplibre/martin/pull/2186

As with other PRs, only the last commit is actually belonging to this PR, the rest will merge out.


nyurik (approved):

nyurik:
btw, i'm not a big fan of `use super::...` anywhere except unit test blocks - i prefer absolute declarations most of the time - a bit more readable imo

PR #2192, opened 2025-09-16
chore(core) final push in the pg-move

CommanderStorm:
This is the final push in the postgres move to core.

For the function/table resolver, this needs a bit of an refactoring before these would be publicly usefull. I am clasifying them as "configuration"-stuff and moving them to the config-file area.

I am not sure if this is the final resting place, but for now this seems an OK place to push this to.

This PR also fixes two unrelated issues:
- some of our constants are `static` instead of `const`
- the `postgres` configuration module (in the config file "postgres") is named `pg`


nyurik (approved):

tests/test.sh
```diff
@@ -403,8 +403,8 @@ test_pbf table_and_view_two_schemas2_0_0_0  table_and_view_two_schemas.1/0/0/0
 
 kill_process "$MARTIN_PROC_ID" Martin
 
-test_log_has_str "$LOG_FILE" 'WARN  martin::pg::query_tables] Table public.table_source has no spatial index on column geom'
-test_log_has_str "$LOG_FILE" 'WARN  martin::pg::query_tables] Table public.table_source_geog has no spatial index on column geog'
+test_log_has_str "$LOG_FILE" 'WARN  martin::postgres::query_tables] Table public.table_source has no spatial index on column geom'
+test_log_has_str "$LOG_FILE" 'WARN  martin::postgres::query_tables] Table public.table_source_geog has no spatial index on column geog'
```
```suggestion
test_log_has_str "$LOG_FILE" 'WARN  martin::postgres::resolver::query_tables] Table public.table_source has no spatial index on column geom'
test_log_has_str "$LOG_FILE" 'WARN  martin::postgres::resolver::query_tables] Table public.table_source_geog has no spatial index on column geog'
```

tests/test.sh
```diff
@@ -515,8 +515,8 @@ test_jsn fnc_comment_cfg  function_Mixed_Name
 test_metrics "metrics_1"
 
 kill_process "$MARTIN_PROC_ID" Martin
-test_log_has_str "$LOG_FILE" 'WARN  martin::pg::query_tables] Table public.table_source has no spatial index on column geom'
-test_log_has_str "$LOG_FILE" 'WARN  martin::pg::query_tables] Table public.table_source_geog has no spatial index on column geog'
+test_log_has_str "$LOG_FILE" 'WARN  martin::postgres::query_tables] Table public.table_source has no spatial index on column geom'
+test_log_has_str "$LOG_FILE" 'WARN  martin::postgres::query_tables] Table public.table_source_geog has no spatial index on column geog'
```
```suggestion
test_log_has_str "$LOG_FILE" 'WARN  martin::postgres::resolver::query_tables] Table public.table_source has no spatial index on column geom'
test_log_has_str "$LOG_FILE" 'WARN  martin::postgres::resolver::query_tables] Table public.table_source_geog has no spatial index on column geog'
```

PR #2198, opened 2025-09-17
feat(pmtiles): Embed version in tilejson tiles URL

keichan34:
This is more of a discussion than a PR, but I wanted to get some feedback on the possibility of embedding version information in the tilejson `tiles` URL. This would help a lot with cache-busting and increasing cache hits by extending the cache age maximum.

In the case of pmtiles, I've used the technique of setting the version in the URL to the value of the etag on the pmtiles archive, so minor tile updates clear the cache as well -- how does that sound?

If the reception is good on this, I'd like to continue (config options? I'm using `v` now, but if that overlaps with something else, it could produce surprising results), but this is basically just a bare-bones prototype to test the waters. Let me know what you think.


martin/src/srv/tiles_info.rs
```diff
@@ -45,11 +46,21 @@ async fn get_source_info(
...
+        .iter()
+        .filter_map(|s| s.get_version())
+        .collect::<Vec<_>>()
+        .join("-");
     let query_string = req.query_string();
-    let path_and_query = if query_string.is_empty() {
+    let query = (!versions.is_empty())
+        .then(|| Cow::Owned(format!("v={versions}")))
+        .into_iter()
+        .chain((!query_string.is_empty()).then(|| Cow::Borrowed(query_string)))
```
CommanderStorm:
```suggestion
        .chain((!query_string.is_empty()).then_some(Cow::Borrowed(query_string)))
```

CommanderStorm (commented):
what happens if the query provided already contains `v=` => now there are two `v=` is a bit problematic.

Could we move this to an `ETAG`/similar header instead?
Would this still solve your problem?

Also, to be mergable, this needs tests 😉

keichan34:
@CommanderStorm thanks for the review. Yeah, I know about the overlapping `v` - that's what I was referring to in the disclaimer. Etags don't solve the problem, because even if the etag on the tilejson request returns a 304, the individual tiles will still be requested. The end goal of this PR is to be able to, ideally, cache tiles with this `v` attribute indefinitely -- because when the `v` changes, it's a new URL, and new tiles are loaded. So it needs to be on the URL.

For my use case, a simple `v` attribute would suffice. Maybe in the case of collision a suffix or prefix could be set? I'm imagining these could be used in CDN configs to control CDN cache keys as well -- if what query string params are incorporated in the cache key, a nondeterministic cache key wouldn't be appropriate.

Tests will also be on the way! I was having some trouble with getting snapshot tests passing, will be troubleshooting those next.

CommanderStorm:
🤔 That seems not great. We are pre-v1.0, so changing `v=` to be a different value is also possible, but lets discuss the underlying assumptions:

Which rendering engine are you using?
Is the `version` field of tilejson handled by their cache?
=> If it is allowed to cache the tiles, maplibre should do so

keichan34:
Using maplibre-gl-js in the browser, relying on the browser and CDN cache. Version field in the tilejson has nothing to do with the cache (I'm just using that now because it's convenient to do so -- in the case of pmtiles, the etag of the upstream archive, if available, may be more appropriate).

The main reason why I want version strings on the tile requests is because of the CDN cache. Here's what that looks like:

1. maplibre requests tilejson, tilejson contains the tiles URL. This is always up-to-date.
2. maplibre requests each tile it requires, with the tiles URL in the tilejson.

This way, we can set `Cache-Control: public, max-age=..., immutable` on the tile responses to optimize browser and CDN cache hit rates, while also making sure that old tiles aren't served when a new tileset is deployed (when the new tileset is deployed, the version param is updated, and a new request is issued)

A config option like `tilejson_url_version_param: v`, defaulting to None (meaning no version param at all), could mitigate automatically overriding the `v` parameter.

What do you think?

CommanderStorm:
Yes, that seems like a good and logical option. 👍🏻
I did not think about CDNs.

Alternatively, I would also be fine with a longer key than `v=`, but making this configurable seems more sensible (and simple to do).

So to be clear on requirements:
- [x] add the option to the `SrvConfig`
- [ ] document the new option in the [config file documentation](https://maplibre.org/martin/config-file.html) with an explanation, in which case this is beneficial (`max_feature_count` is likely a good reference)
- [x] add this parameter to one of our integration tests under `tests/` and `just bless` the results.

I am not sure if this needs more testing or documentation, but if you think an area benefits from it regarding this feature, that is always appreciated 😉.

nyurik:
note that some CDNs, e.g. AWS one, have various policies by default, and some of them actually could remove the query param

keichan34:
@CommanderStorm thanks for the quick reply, I'll update the PR with this.

@nyurik Yeah, the default CachingOptimized policy will ignore query string parameters by default. Using martin with CORS, especially [with the Lambda setup](https://maplibre.org/martin/run-with-lambda.html) required a little wrangling so I think a quick mention there should be OK, when we get around to adding the CloudFront CDN docs.

keichan34:
@CommanderStorm OK - I added the option and updated the integration tests. 

There were quite a few differences in the lengths of floats - I'm guessing this is a platform issue... How should we proceed with this? I committed the outputs of `just bless` on my system, but if necessary I can edit it down to only the changes that occurred in this patch.

So, should I ignore the differences or is there a way we can make explicit the accuracy of output floats? Has this issue occurred before to your knowledge? A cursory search here on GitHub didn't really bring anything up...

CommanderStorm:
Thanks. Lets remove that change and fix it in a separate PR.
I am not sure why all of the sudden the floats round different or how to precicely fix this 🤷🏻‍♂️

CommanderStorm:
The float rounding issue should be resolved now.

docs/src/config-file.md
```diff
@@ -307,4 +307,11 @@ styles:
      some_style_name: /path/to/this/style.json
      #  Publish specific file as `other_style_name`
      other_style_name: /path/to/other_style.json
+
+# If set, the version of the tileset (as specified in the MBTiles or PMTiles metadata)
+# will be embedded in the TileJSON `tiles` URL, with the set identifier. For example, if
+# the setting here is `version`, and the PMTiles tileset version is `1.0.0`, the
+# TileJSON will be:
+# { ..., "tiles": [".../{z}/{x}/{y}?version=1.0.0"], ... }
```
CommanderStorm, line 325:
```suggestion
# If set, the version of the tileset (as specified in the MBTiles or PMTiles metadata)
# will be embedded in the TileJSON `tiles` URL, with the set identifier.
# This is useful to
#
# For example, if
# - the setting here is `version`, and
# - the PMTiles tileset version is `1.0.0`, the
# TileJSON will be:
# { ..., "tiles": [".../{z}/{x}/{y}?version=1.0.0"], ... }
```

docs/src/config-file.md
```diff
@@ -307,4 +307,15 @@ styles:
      some_style_name: /path/to/this/style.json
      #  Publish specific file as `other_style_name`
      other_style_name: /path/to/other_style.json
+
+# If set, the version of the tileset (as specified in the MBTiles or PMTiles metadata)
+# will be embedded in the TileJSON `tiles` URL, with the set identifier.
+# This is useful to
```
CommanderStorm: Only question remaining is how we document this part

```suggestion
# This is useful to give clients a better way to cache-bust a CDN:
# 1. maplibre requests tilejson, tilejson contains the tiles URL. This is always up-to-date.
# 2. maplibre requests each tile it requires, with the tiles URL in the tilejson.
# 3. `Control: public, max-age=..., immutable` on the tile responses can bow be set to optimize browser and CDN cache hit rates, while also making sure that old tiles aren't served when a new tileset is deployed
#
# To work correctly, this needs the CDN to be configured to allow query parameters and take them into acount for cacheing purposes.
# Some common CDN configurations ignore query parameters by default.
```

CommanderStorm (approved):
Thanks for the feature. I am going to merge it

I would appreciate it if you give us feedback once you have deployed it if this improves your caching/..

PR #2200, opened 2025-09-17
Allow test.sh to run on darwin

todtb:
This allowed me to run integration tests locally. Mainly it is portability around `sed` and `tr`, and allowing darwin to use `ogrmerge` (will still require that setup just like linux) and `vtzero`.

---

There is another OS check in the `justfile` which ultimately diffs the files after the script runs. I haven't updated that yet, as
most everything there looked fine, but I was seeing some json header diffs. Namely, the expected is brotli encoding whereas the default is gzip. 

I didn't see anything in the test suite setup (such as args, or `tests/config.yaml`) that would be setting `preferred_encoding` to brotli, so I'm wondering why that is expected to be the case as opposed to the default gzip. Was it never `bless`ed? Am I missing somewhere in the test config where that is being changed?


tests/test.sh
```diff
@@ -100,13 +108,13 @@ test_metrics() {
   URL="$MARTIN_URL/_/metrics"
 
   echo "Testing $1 from $URL"
-  $CURL --dump-header  "$FILENAME.headers" "$URL" | sed -E 's/^(martin_.*?) [\.0-9]+$/\1 NUMBER/g' > "$FILENAME.txt"
+  $CURL --dump-header  "$FILENAME.headers" "$URL" | $SED -E 's/^(martin_.*?) [\.0-9]+$/\1 NUMBER/g' > "$FILENAME.txt"
```
```suggestion
  $CURL --dump-header  "$FILENAME.headers" "$URL" | $SED --regexp-extended 's/^(martin_.*?) [\.0-9]+$/\1 NUMBER/g' > "$FILENAME.txt"
```

tests/test.sh
```diff
@@ -100,13 +108,13 @@ test_metrics() {
...
   clean_headers_dump "$FILENAME.headers"
-  $CURL --dump-header  "$FILENAME.fetched_with_compression.headers" --compressed "$URL" | sed -E 's/^(martin_.*?) [\.0-9]+$/\1 NUMBER/g' > "$FILENAME.fetched_with_compression.txt"
+  $CURL --dump-header  "$FILENAME.fetched_with_compression.headers" --compressed "$URL" | $SED -E 's/^(martin_.*?) [\.0-9]+$/\1 NUMBER/g' > "$FILENAME.fetched_with_compression.txt"
```
```suggestion
  $CURL --dump-header  "$FILENAME.fetched_with_compression.headers" --compressed "$URL" | $SED --regexp-extended 's/^(martin_.*?) [\.0-9]+$/\1 NUMBER/g' > "$FILENAME.fetched_with_compression.txt"
```

tests/test.sh
```diff
@@ -119,11 +127,16 @@ test_pbf() {
 
   if [[ $OSTYPE == linux* ]]; then
     ./tests/fixtures/vtzero-check "$FILENAME"
-    # see https://gdal.org/en/stable/programs/ogrmerge.html#ogrmerge
-    ogrmerge.py -o "$FILENAME.geojson" "$FILENAME" -single -src_layer_field_name "source_mvt_layer" -src_layer_field_content "{LAYER_NAME}" -f "GeoJSON" -overwrite_ds
-    jq --sort-keys '.features |= sort_by(.properties.source_mvt_layer, .properties.gid) | walk(if type == "number" then .+0.0 else . end)' "$FILENAME.geojson" > "$FILENAME.sorted.geojson"
-    mv "$FILENAME.sorted.geojson" "$FILENAME.geojson"
+  elif [[ $OSTYPE == darwin* ]]; then
+    ./tests/fixtures/vtzero-check-darwin "$FILENAME"
```
CommanderStorm, line 131: For the record: I don't think we should have binaries in our codebase.
Another motivation for refactoring this subsystem someday

CommanderStorm (commented):
I enabled macos in CI, lets see where this breaks.
It would be great to get this re-enabled. This way, we can prevent the level of breakage of our test-suite for macos.

Lets also make those sed-commans more like the others -> usng the long names so people know what is happening

nyurik:
yep, we clearly need to do a `tile-tool` -- something that can easily dump MVT and MLT tiles, encode them, possibly a few other basic things.  A bit like geozero, but simpler and cleaner

tests/test.sh
todtb: This is fine. For what it's worth, I suspect `-E` was originally used here to provide portability between GNU and BSD `sed` variants, however the expression will only work with GNU variant anyway, hence the tooling check added above.

CommanderStorm: no, it was used because I only rememberd `-E` as the "make sed behave like regexes should" flag

todtb: My apologies, I didn't notice you had made the change to `lefuturiste`'s PR. In any case, the long option is fine here.

todtb:
```
==> Installing postgis
==> Pouring postgis--3.6.0.ventura.bottle.tar.gz
🍺  /usr/local/Cellar/postgis/3.6.0: 1,801 files, 82MB
==> No outdated dependents to upgrade!
Run psql -v ON_ERROR_STOP=1 -c 'CREATE EXTENSION IF NOT EXISTS postgis;' '***localhost:34837/test'
  
ERROR:  extension "postgis" is not available
DETAIL:  Could not open extension control file "/usr/local/opt/postgresql@16/share/postgresql@16/extension/postgis.control": No such file or directory.
HINT:  The extension must first be installed on the system where PostgreSQL is running.
Error: Process completed with exit code 1.
```

[allows set postgis version for macos](https://github.com/nyurik/action-setup-postgis/pull/15#top) may help here, but I'm still looking into this

nyurik:
@todtb yeah, that PR looked promising, but I am not sure if @sharkAndshark has time to finish it. Feel free to clone it and re-submit as a new PR

P.S. One aspect that I would love for it to fix is not to maintain the version list if possible

.github/workflows/ci.yml
```diff
@@ -413,9 +413,8 @@ jobs:
       fail-fast: true
       matrix:
         include:
-          # temporary disabled due to an issue with homebrew which is difficult to debug
-          # - target: x86_64-apple-darwin
-          #   os: macos-13
+          - target: x86_64-apple-darwin
+            os: macos-13
```
CommanderStorm: lets disable this again and move it to a new PR

```suggestion
          # temporary disabled due to an issue with homebrew which is difficult to debug
          # - target: x86_64-apple-darwin
          #  os: macos-13
```

.github/workflows/ci.yml
```diff
@@ -415,7 +415,7 @@ jobs:
         include:
           # temporary disabled due to an issue with homebrew which is difficult to debug
           # - target: x86_64-apple-darwin
-          #   os: macos-13
+          #  os: macos-13
```
```suggestion
          #   os: macos-13
```

CommanderStorm (approved):
Lets merge this as is and move preventing regressions via adding this to CI into anothr PR

PR #2203, opened 2025-09-18
chore: move publishing testcase to stable rust

CommanderStorm:
As noted, this was a workaround until `v1.90.0`.
Since it is released now, we can remove the workaround


nyurik (approved):
nice!

PR #2205, opened 2025-09-19
chore: reduce panics in result-returning functions

### LINKED ISSUE #2204: Enable `clippy::unwrap_used` to better handle panics
I think these restrictions should improve our code quality - i.e. we shouldn't panic (unless its a test).  To enable, modify root `Cargo.toml` and add these lines, and `just clippy` it...

```diff
 [workspace.lints.clippy]
 # Restrictions
 panic_in_result_fn = "warn"
 todo = "warn"
+unwrap_used = "warn"
 # Pedantics without being too noisy
 pedantic = { level = "warn", priority = -1 }
```

Raj-G07:
Closes: #2204


CommanderStorm:
There are a few cases clippy found where the new lint is triggering that need to be adjusted ^^

Raj-G07:
Instead of | Use in Library/Test Code | Use in build.rs
-- | -- | --
unwrap() | ok_or_else() / expect() | ok_or_else() / expect()
assert!(), assert_eq!() | return Err(...) | eprintln!() + exit (if fatal)
panic!() | return Err(...) | eprintln!() + std::process::exit(1)

Is it right?

CommanderStorm:
It really depends on the context of the code.
If we are in a result, returning an error might be better.
Or not, if this is a case that cannot happen => we can expect it.

In build.rs, `expect`-ing is the way to go.
There should not be a lot of change here.

There might also be cases that I am missing...

Raj-G07:
In build.rs 
```diff
-  assert!(
-       src.is_dir(),
-       "source for the copy operation is not an existing directory"
-    );
+ if !src.is_dir(){
+   eprintln!("warning=source for the copy operation is not an existing directory" );
+       std::process::exit(1);
+    }

- panic!(
-            "failed to create destination directory {}: {e}",
-            dst.display()
-        )
+  eprintln!(
+           "warning=failed to create destination directory {}: {}",
+           dst.display(), e
+        );
+     std::process::exit(1);
```

Can we apply these ?

CommanderStorm:
Negative, those are fine.
Am I talking to a human btw (your answers seem slightly robotic)?

Raj-G07:
Yes, i am 100% human btw , actually i am following **James Pearce** , which gives me idea how to talk to maintainers ?  
Am i doing right?

CommanderStorm:
Got it, thanks for clarifying 🙂

Following best practices is definitely a good idea, but don’t worry too much about sounding “perfect” — it’s better to just write naturally so we can understand your thinking

CommanderStorm:
This PR has been inactive for a while, so I’m closing it for now to keep things tidy.
If you’d like to continue the work, feel free to reopen it anytime — we’d be happy to pick it up again! 😊

Issue #2136, opened 2025-09-09
Improve warnings in case of a bad sprite dir data


nyurik:
As described in #1336, if a provided sprite dir has invalid files - e.g. the pre-generated sprite files instead of the source `*.svg` files, we need to explain to the users that no SVG files were found.

PR #2211, opened 2025-09-19
Bad sprite dir warnings and svg validation

nuts-rice:
Addresses #2136 by:
- Adding errors for empty dirs, invalid svg format + Failed Dir validation
- Doing basic SVG format validation and warning on fails for individual files
- Then validates whole directory , warns if empty, and provides summary of validated or failed files, tracks files that appear to be pre-generated
-  comprehensive warning messages if files appear to be pre-generated

Please let me know if this ought to be edited or the validation could be moved elsewhere. 
Thanks!


martin-core/src/resources/sprites/mod.rs
```diff
@@ -71,27 +71,285 @@ impl SpriteSources {
...
+
+        info!(
+            "Validated sprite directory {disp_path}: found {svg_count} SVG files out of {total_files} total files"
+        );
+        Ok(())
+    }
+
+    /// Validates an individual SVG file for format.
+    async fn validate_svg_file(&self, path: &PathBuf) -> Result<(), SpriteError> {
+        let on_err = |e| SpriteError::IoError(e, path.clone());
```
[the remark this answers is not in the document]
nuts-rice: `let content = content.trim();`
is already before the check, correct?

martin-core/src/resources/sprites/mod.rs
```diff
@@ -71,27 +71,273 @@ impl SpriteSources {
...
-            match self.0.entry(id) {
-                Entry::Occupied(v) => {
+            return;
+        }
+        if !path.exists() {
+            warn!("Sprite source {id} path doesn't exist: {disp_path}");
+            return;
+        }
+
+        match std::fs::read_dir(&path) {
```
CommanderStorm: Let's move this validation to a new function. It is kind of messy to read otherwise

CommanderStorm (changes requested):
Why is there so much duplicate code? This does not seem very maintainable.
Please do a polishing pass 😉

Let's also likely change add_sprite to be a Result<(), SpritError> function as users likely want this instead.

martin-core/src/resources/sprites/mod.rs
```diff
@@ -71,27 +71,236 @@ impl SpriteSources {
...
             warn!("Ignoring non-directory sprite source {id} from {disp_path}");
-        } else {
-            match self.0.entry(id) {
-                Entry::Occupied(v) => {
-                    warn!(
-                        "Ignoring duplicate sprite source {} from {disp_path} because it was already configured for {}",
-                        v.key(),
-                        v.get().path.display()
-                    );
+            return Err(SpriteError::DirectoryValidationFailed(
```
nyurik: a warning is mutually exclusive with an error - either it fails, in which case the error message will be printed or handled by the caller, or it continues the execution but prints a warning message. Both do not make sense

martin-core/src/resources/sprites/mod.rs
```diff
@@ -71,27 +71,236 @@ impl SpriteSources {
...
+                path.clone(),
+                "No SVG element found".to_string(),
+            ));
+        }
+
+        Ok(())
+    }
+
+    /// Validates all configured sprite sources and logs a summary.
+    pub async fn validate_all_sources(&self) -> Result<(), SpriteError> {
```
CommanderStorm: `validate_all_sources` is never used.
As I asked before, please do a polishing pass.

martin-core/src/resources/sprites/mod.rs
```diff
@@ -71,27 +71,165 @@ impl SpriteSources {
...
-            match self.0.entry(id) {
-                Entry::Occupied(v) => {
-                    warn!(
-                        "Ignoring duplicate sprite source {} from {disp_path} because it was already configured for {}",
-                        v.key(),
-                        v.get().path.display()
-                    );
+            return Err(SpriteError::DirectoryValidationFailed(
+                path,
+                "Path is not a directory".to_string(),
```
nyurik: a minor nit, but makes for a better pattern -- if you have several similar errors with some magical string, easier to just have dedicated errors one for each. This way they are much faster to create (you don't need to allocate a new string on each error), and all content of an error is in the error description only

martin-core/src/resources/sprites/mod.rs
```diff
@@ -71,27 +71,153 @@ impl SpriteSources {
...
+
+        Ok(())
+    }
+
+    /// Validates a sprite source directory to ensure it contains valid SVG files.
+    /// Checks include:
+    /// - Directory existence and accessibility
+    /// - Presence of SVG files
+    /// - Basic SVG format validation
+    pub async fn validate_source_directory(&self, path: &PathBuf) -> Result<(), SpriteError> {
```
CommanderStorm, line 108: Let's not add dead code.
As I said above, this is never used in the codebase.

martin-core/src/resources/sprites/mod.rs
```diff
@@ -191,11 +317,11 @@ mod tests {
     #[tokio::test]
     async fn test_sprites() {
         let mut sprites = SpriteSources::default();
-        sprites.add_source(
+        let _ = sprites.add_source(
```
CommanderStorm: Let's unwrap instead to make sure this is not an error

CommanderStorm (commented):
Sorry, but this still needs a deeper scrubbing.

Please add a unit test for each new behaviour (to prevent regressions and go over the code you are trying to add again.

CommanderStorm:
I just auto-merged a fairly related warning.
I think this PR would still be an improvement, but needs a bit of scrubbing to get into a mergable state.

Lets regroup on this one

PR #2214, opened 2025-09-22
chore: simplify the `just check` command

CommanderStorm:
Imo, we can simplify this command and ensure that we actually check every feature instead of the below, which for example only checks martin-core transitively.


nyurik (approved):
Need to fix codecov :(

PR #2215, opened 2025-09-22
chore(core): be consistent in posgres naming

CommanderStorm:
This renames to Pg -> Postgres, which is consistent with all of our other tile sources, resources, ...

The counterpoint is that we are also naming everythin COG `cog`.. Not sure which way to go.
We could also rename the other way Postgres -> Pg


nyurik:
tbh, i like `pg` for all the internal code use, but for config `postgres` might be better (although we can do aliasing here).  Lastly, `postgresql` is too verbose, despite being "the correct" one.  Note that `psql` understands both the `postgres://...` and `postgresql://...` urls

CommanderStorm:
So what would this mean for martin-core => `Postgres..` naming for structs? 🤔
I am really not sure which way is better, I just think it is likely cleaner to go one of both ways in terms of naming.

Not a big deal, but just one of those polishing things.

nyurik:
I guess everything could be `PostgresFooBar`

martin/src/config/args/postgres.rs
```diff
@@ -224,7 +228,7 @@ impl PgArgs {
 }
 
 #[must_use]
-fn is_postgresql_string(s: &str) -> bool {
+fn is_postgres_connection_string(s: &str) -> bool {
     s.starts_with("postgresql://") || s.starts_with("postgres://")
 }
```
CommanderStorm, line 233: this is a little more than just applying the renaming where it `pg` and `postresql`

nyurik (approved):

PR #2229, opened 2025-09-24
chore(tests): improve the style fixtures

CommanderStorm:
Currently two of the styles don't work locally becasue they would need acces keys.
By moving them over to openfreemap (and adding a semi-minimal example of a map which won't load) this makes the test data more realistic.

And to answer the question:
What does maplibre-native-rs do in this case?
Well, it throws an exception.


nyurik:
that just means we need to wrap all C++ calls in MLN-RS with `catch -> Result ` code

CommanderStorm:
I am aware of this, but lets move step by step.

I think having this available to do test cases against will be nice for other PRs. For solving the panicing, we will need to review the precursor first:
- https://github.com/maplibre/maplibre-native-rs/pull/66

nyurik (approved):

.github/workflows/ci.yml
```diff
@@ -63,6 +63,7 @@ jobs:
           -c "exec docker-entrypoint.sh postgres -c ssl=on -c ssl_cert_file=/etc/ssl/certs/ssl-cert-snakeoil.pem -c ssl_key_file=/etc/ssl/private/ssl-cert-snakeoil.key"
     steps:
       - run: rustup update stable && rustup default stable
+      - run: rustup component add clippy rustfmt
```
```suggestion
```

PR #2234, opened 2025-09-26
chore(ci): Redact last decimal from saved configs

CommanderStorm:
THis should fix our issue with codecov 🤞🏻


nyurik (approved):

tests/test.sh
```diff
@@ -585,6 +585,13 @@ else
   echo "Skipping martin-cp tests"
 fi
 
+# If we don't do this, rounding differences on CI and local machines are a problem
+echo "::group::redact unnecessary precision in *_config.yaml"
+for file in $(find ./tests/ -name "*_config.yaml" -type f); do
+    echo "truncating floats in $file"
+    "$SED" --regexp-extended --in-place 's/(-?[0-9]+\.0*[1-9][0-9]{0,10})[0-9]+/\1/g' "$file"
```
nyurik: just FYI - if you want, [ruplacer](https://github.com/your-tools/ruplacer) is a very good in-place search/replacer tool

tests/test.sh
```diff
@@ -589,7 +589,7 @@ fi
 echo "::group::redact unnecessary precision in *_config.yaml"
 for file in $(find ./tests/ -name "*_config.yaml" -type f); do
     echo "truncating floats in $file"
-    "$SED" --regexp-extended --in-place 's/(-?[0-9]+\.0*[1-9][0-9]{0,10})[0-9]+/\1/g' "$file"
+    "$SED" --regexp-extended --in-place 's/(-?[0-9]+\.0*[1-9][0-9]{0,10})[0-8]+/\1/g' "$file"
```
nyurik: you are trimming everything except the 9s at the end??? You must at least have a long comment explaining it here :)

CommanderStorm: I meant to adjust the 10 next to it 😉
Apparently 10 digits is not stable still. Kind of strange but in the cases where it is happening reasonable.

170.03 => 4 digits of precision, but my regex is counting it as 1

tests/test.sh
```diff
@@ -589,7 +589,7 @@ fi
 echo "::group::redact unnecessary precision in *_config.yaml"
 for file in $(find ./tests/ -name "*_config.yaml" -type f); do
     echo "truncating floats in $file"
-    "$SED" --regexp-extended --in-place 's/(-?[0-9]+\.0*[1-9][0-9]{0,9})[0-9]+/\1/g' "$file"
+    "$SED" --regexp-extended --in-place 's/(-?[0-9]+\.[0-9]{10})[0-9]+$/\1 # truncated to 10 digits/g' "$file"
```
nyurik, line 592: do you want to do another `sed` run to truncate trailing zeroes, or try to do it in a single regex?

PR #2242, opened 2025-09-26
chore(release): improve release-plz config

CommanderStorm:
This PR fixes a few bugs in our release plz config:
- martins changelog not containing its own changes
- dependency versions being bumped wrong


nyurik (approved):
be careful when changing package versions - release-plz will try to publish the moment it sees a mismatch between the crate version and what's published

CommanderStorm:
And that was the reason why I set

```yaml
release_always = false
```

PR #2249, opened 2025-09-27
chore(ci): set `persist-credentials: false` in CI and make the order of commands more consistent 

CommanderStorm:
this pr also fixes a few cases of inconsistent workflow item ordering

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,18 +66,23 @@ jobs:
       - name: Checkout sources
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with: { persist-credentials: false }
-      - uses: taiki-e/install-action@cc60de1d6831d7e9c4342f618ce7a5d6a9f223a4 # v2.61.6
-        with: { tool: 'just' }
       - name: Init database
         run: tests/fixtures/initdb.sh
         env:
           PGPORT: ${{ job.services.postgres.ports[5432] }}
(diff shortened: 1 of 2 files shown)


nyurik (commented):
why all of a sudden we have direct cargo commands in the CI? The whole idea of the justfile is to have identical CI and local dev to use the same scripts, and in a docker too.  I would like to remove all `cargo` commands from the workflows

CommanderStorm:
Not sure why this was originally done this way.
I just know that CI inexplicably fails without this => https://github.com/maplibre/martin/pull/2249/commits/2bacb20cc0c46af8ef09dbad880f3c0a98c3e5fa

This is why I pulled that part of the PR.

I think `just` is tampering/not forwarding the environment variables

.github/workflows/ci.yml
```diff
@@ -76,11 +77,12 @@ jobs:
           cargo test --package mbtiles --no-default-features
           cargo test --package mbtiles
           cargo test --package martin
-          cargo test --doc
+          cargo test --package martin-core
```
nyurik, line 80: what about doc tests?

nyurik (approved):

.github/workflows/ci.yml
CommanderStorm: What about them?
They are in the new section below because they are kind of different and imo fail differently.

If we merge all the unit tests into one just command they should be included. Currently this makes more sense to me  being split

nyurik: ah, thx, didn't see it

PR #2251, opened 2025-09-27
feat(pmtiles)!: change pmtiles to base the implementation on `object_storage` instead

CommanderStorm:
Works towards #2180 

This primarily swaps out our internall implementation of pmtiles with one based on the `object_storage` crate.

Pro:
- more cloud providers
- more AWS specific options
- most environment variables and all config key continute to work
- muuuch more options to configure

Con:
- `AWS_PROFILE` would need a larger refactoring as this makes the `fn finish(&mut self) -> ConfigurationResult<()>` to a `async fn finish(&mut self) -> ConfigurationResult<()>` with all the follow on effects

image

PR #2252, opened 2025-09-27
fix(tests): missing cfg making testing just one feature more cumbersome

CommanderStorm:
A few of our tests where incorrectly scoped in terms of the features required to run them

I found this during https://github.com/maplibre/martin/pull/2251

### DIFF
--- a/martin/src/config/args/root.rs
+++ b/martin/src/config/args/root.rs
@@ -205,7 +205,7 @@ pub fn parse_file_args<T: crate::config::file::ConfigExtras>(
 
 #[cfg(test)]
 mod tests {
-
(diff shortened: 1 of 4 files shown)


martin/src/config/args/root.rs
```diff
@@ -205,7 +205,7 @@ pub fn parse_file_args<T: crate::config::file::ConfigExtras>(
 
 #[cfg(test)]
 mod tests {
-
+    #[cfg(all(feature = "pmtiles", feature = "mbtiles", feature = "cog"))]
     use insta::assert_yaml_snapshot;
```
nyurik: might be better to inline it - i.e. call it as `insta::assert_yaml_snapshot!(...);` rather than using `use` with a complex cfg

nyurik (approved):

PR #2254, opened 2025-09-27
chore(ci): add two obscure windows specific filename checks to pre-commit

CommanderStorm:
In response to https://github.com/maplibre/martin/pull/1868 I explored if we check for some of these things and found that we don't for filename specific things, but otherwise we are OK.

I also sorted a few things alphabetically

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -18,9 +18,9 @@ demo/
 **/dist/
 **/node_modules/
 *.profraw
-.DS_Store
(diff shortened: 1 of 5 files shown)


nyurik (approved):
sort is a weird thing - you used case insensitive, i used default (case sensitive).  No biggie - do note that IDE's sort lines tool might be different between IDEs

PR #2263, opened 2025-09-28
feat(pg): Add benchmark for source discovery timing

CommanderStorm:
This adds a benchmark for how we do source discovery.

It tries to remain relatively grounded in reality when it comes to the features we support and that might trigger a longer analysis time.

@danielgerigk could you look over the code if something odd jumps out at you?

Interestingly, function discovery seems to be much, much faster than tables 🤔 
500 tables take approx. 400ms on my 2018 laptop with bad cooling
500 functions take approx. 100ms on the same machine


CommanderStorm:
I ordered a benchmark to be run in GHA ([link](https://github.com/CommanderStorm/martin/actions/runs/18074929096)).

Lets see what the results are with the following sizing:
```rust
const SIZES: &[usize] = &[10, 100, 200, 400, 800, 1600, 3200, 6400, 12800, 25600];
```

This might take a wile

martin/benches/postgres_discovery.rs
```diff
@@ -0,0 +1,378 @@
...
+use pprof::criterion::{Output, PProfProfiler};
+use testcontainers_modules::postgres::Postgres;
+use testcontainers_modules::testcontainers::ImageExt;
+use testcontainers_modules::testcontainers::runners::AsyncRunner;
+
+// Benchmark sizes
+const SIZES: &[usize] = &[10, 100, 200, 400, 800, 1600, 3200, 6400, 12800, 25600];
+
+/// Setup PostGIS container
+async fn setup_postgres_container() -> (
```
danielgerigk: `setup_postgres_container` or `setup_postgis_container`?

CommanderStorm: They are essentially the same thing.
We internally always use postgres, a bit confusing yes

danielgerigk:
> @danielgerigk could you look over the code if something odd jumps out at you?

looks fine to me

CommanderStorm:
I looked into a number of factors, but getting PG to be reliably been faster is not as simple as I thought and goes over the bar where I kind of can't justify the effort.

Being below half a second for less than 5k of tables, this is likely fine for the time beeing.

The optimisation that saw benefit was `join lateral`, but I am not sure how correct that impl was.
In any case, it would only have been a 15% improvement.

CommanderStorm:
the main problem is that the statistics of PG are way of for the base tables, so the rest does not really have a good chance.

Maybe we can stream the values in. Hard to say what is a suitable solution.
Not sure if this would even be beneficial.
Without being able to add indexes, we are always pretty limited.

(if we could add indexes, this would be a different question => this is likely the better solution to your problem)

danielgerigk:
> I looked into a number of factors, but getting PG to be reliably been faster is not as simple as I thought and goes over the bar where I kind of can't justify the effort.
> 
> Being below half a second for less than 5k of tables, this is likely fine for the time beeing.
> 
> The optimisation that saw benefit was `join lateral`, but I am not sure how correct that impl was. In any case, it would only have been a 15% improvement.

i agree. perhaps the query can be optimized by rewriting it with less joins, but that has to be done very carefully. and is indeed not really worth the effort.

nyurik:
Yeah, in theory, a custom filtering rust algorithm running in martin may be faster than to keep all the processing in sql

CommanderStorm:
Given how stupid the query order (and how inefficient the execution engine) is, that is a good probability.

Sounds fun to develop..

nyurik (approved):

PR #2270, opened 2025-09-30
chore: Add build for `aarch64-unknown-linux-gnu`

### LINKED ISSUE #2269: Distribute binary for arm64/linux GNU
Thanks for the continued work on the tool!

I'm currently developing an Ubuntu docker image that uses Martin, among other things. I'm developing locally on Apple silicon, so it would be convenient to have a native arm64 GNU binary to work with. Currently I'm stuck compiling it myself (slow startup) or emulating my whole docker container in amd64 (slow performance). But if you don't think anyone else would find this useful I can manage without it.

quincylvania:
Closes #2269.

I'm no expert in this stuff but I think something like this should work.

I also switched the Homebrew package on Linux to use the `gnu` binaries instead of `musl` since I'm _pretty_ sure they're [expected](https://docs.brew.sh/Support-Tiers) to use `glibc` by default. Not 100% sure though.

### DIFF
--- a/.github/files/homebrew.martin.rb.j2
+++ b/.github/files/homebrew.martin.rb.j2
@@ -23,11 +23,11 @@ class Martin < Formula
   on_linux do
     on_arm do
       sha256 "{{ linux_arm_sha256 }}"
-      url "https://github.com/maplibre/martin/releases/download/v#{current_version}/martin-aarch64-unknown-linux-musl.tar.gz"
+      url "https://github.com/maplibre/martin/releases/download/v#{current_version}/martin-aarch64-unknown-linux-gnu.tar.gz"
     end
     on_intel do
       sha256 "{{ linux_intel_sha256 }}"
-      url "https://github.com/maplibre/martin/releases/download/v#{current_version}/martin-x86_64-unknown-linux-musl.tar.gz"
+      url "https://github.com/maplibre/martin/releases/download/v#{current_version}/martin-x86_64-unknown-linux-gnu.tar.gz"
     end
   end
 
--- a/.github/files/homebrew.martin.rb.j2
+++ b/.github/files/homebrew.martin.rb.j2
@@ -23,11 +23,11 @@ class Martin < Formula
   on_linux do
     on_arm do
       sha256 "{{ linux_arm_sha256 }}"
-      url "https://github.com/maplibre/martin/releases/download/v#{current_version}/martin-aarch64-unknown-linux-gnu.tar.gz"
+      url "https://github.com/maplibre/martin/releases/download/v#{current_version}/martin-aarch64-unknown-linux-musl.tar.gz"
     end
     on_intel do
       sha256 "{{ linux_intel_sha256 }}"
-      url "https://github.com/maplibre/martin/releases/download/v#{current_version}/martin-x86_64-unknown-linux-gnu.tar.gz"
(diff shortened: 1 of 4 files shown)


.github/files/homebrew.martin.rb.j2
```diff
@@ -23,11 +23,11 @@ class Martin < Formula
   on_linux do
     on_arm do
       sha256 "{{ linux_arm_sha256 }}"
-      url "https://github.com/maplibre/martin/releases/download/v#{current_version}/martin-aarch64-unknown-linux-musl.tar.gz"
+      url "https://github.com/maplibre/martin/releases/download/v#{current_version}/martin-aarch64-unknown-linux-gnu.tar.gz"
```
CommanderStorm: This opens us up to this breaking due to different libc versions.

I'd like to avoid this, unless you have benchmarked the performance and for you, this makes a difference.

Open to alternative opinions ^^

docs/src/config-file.md
```diff
@@ -313,10 +313,11 @@ styles:
 # This is useful to give clients a better way to cache-bust a CDN:
 # 1. maplibre requests tilejson, tilejson contains the tiles URL. This is always up-to-date.
 # 2. maplibre requests each tile it requires, with the tiles URL in the tilejson.
-# 3. `Control: public, max-age=..., immutable` on the tile responses can bow be set to optimize browser and CDN cache hit rates, while also making sure that old tiles aren't served when a new tileset is deployed
+# 3. Set `Control: public, max-age=..., immutable` on tile responses can now optimize browser and CDN cache hit rates.
```
```suggestion
# 3. `Control: public, max-age=..., immutable` on tiles to optimize browser/CDN cache hit rates
```

.github/files/homebrew.martin.rb.j2
quincylvania: Okay I reverted this change, it doesn't affect me. From my read it seemed like using musl is unexpected by Homebrew on Linux but if it's working IRL then no need to change it.

CommanderStorm (approved):
Thanks for the enhancement 🎉 

This should be available as the artefact from the build on main (or this PR).
If you need this in a release for let me know and I can cut an release, with the new process that is not much work anymore.

docs/src/config-file.md
```diff
@@ -313,10 +313,11 @@ styles:
 # This is useful to give clients a better way to cache-bust a CDN:
 # 1. maplibre requests tilejson, tilejson contains the tiles URL. This is always up-to-date.
 # 2. maplibre requests each tile it requires, with the tiles URL in the tilejson.
-# 3. `Control: public, max-age=..., immutable` on the tile responses can bow be set to optimize browser and CDN cache hit rates, while also making sure that old tiles aren't served when a new tileset is deployed
+# 3. `Control: public, max-age=..., immutable` on tiles to optimize browser/CDN cache hit rates
+#    It also ensures that old tiles aren’t served when a new tileset is deployed.
```
CommanderStorm: So anoying to churn on this, anyhow.

```suggestion
# 3. Add `Control: public, max-age=..., immutable` on the tile responses 
#    optimize browser/CDN cache hit rates, while also making sure that
#    old tiles aren't served when a new tileset is deployed.
```

.github/files/homebrew.martin.rb.j2
CommanderStorm: I am not sure if we have a user for this, so reverting is likely a better call, thanks 👍🏻

quincylvania:
Thanks for the quick turnaround! A release would be helpful to me but no rush.

PR #2272, opened 2025-10-01
fix(ci): update the macos runner label to coninue having macos CI

CommanderStorm:


CommanderStorm:
It is getting picked up, good.
Lets merge it to unblock other PRs

PR #2283, opened 2025-10-05
chore(lints): applly `clippy::panic_in_result_fn` and `clippy::todo` as warnings

CommanderStorm:
This PR tighens our lints a bit

This applies two of the lints from
- https://github.com/maplibre/martin/issues/2204


mbtiles/src/copier.rs
```diff
@@ -1041,12 +1041,11 @@ mod tests {
         verify_copy_with_zoom_filter(opt, 2).await;
     }
 
-    #[actix_rt::test]
+    #[tokio::test]
```
CommanderStorm: different pr

```suggestion
    #[actix_rt::test]
```

martin/src/config/file/file_config.rs
```diff
@@ -66,6 +66,9 @@ pub enum ConfigFileError {
     #[cfg(feature = "fonts")]
     #[error("Failed to load fonts from {1}: {0}")]
     FontResolutionFailed(#[source] martin_core::fonts::FontError, PathBuf),
+
+    #[error("init_parsing should only be called once")]
```
nyurik: i know there is a lint for panicking in a result function, but in this case i'm not certain which is better -- i.e. Result is something legit, whereas calling init function twice implies a bug in the code...

mbtiles/src/copier.rs
```diff
@@ -1228,17 +1227,17 @@ mod tests {
                        ON t1.zoom_level = t2.zoom_level AND t1.tile_column = t2.tile_column AND t1.tile_row = t2.tile_row")
             .await.unwrap();
 
-        // Ensure all entries in expected_tiles are in tiles and vice versa
+        let expected_tiles_or_tiles = query(
+            "SELECT * FROM expected_tiles EXCEPT SELECT * FROM tiles
```
nyurik: multiline inline SQL statements should probably be formatted like this:  https://github.com/maplibre/martin/blob/45e0c7bf841147a985c6b3bda6e913b65a891ba3/mbtiles/src/summary.rs#L128-L131

nyurik (approved):

mbtiles/src/copier.rs
CommanderStorm: Lets split the sql too.

This way its bug (it does not do symmetric difference) becomes more apparent and the code simpler to understand

PR #2285, opened 2025-10-06
fix(cog)!: rename `cog` feature to `unstable-cog`

CommanderStorm:
Alternative would be to finish the cog refactoring in
- #1820

### DIFF
--- a/martin-core/Cargo.toml
+++ b/martin-core/Cargo.toml
@@ -12,7 +12,7 @@ rust-version.workspace = true
 homepage.workspace = true
 
 [features]
-default = ["fonts", "styles", "sprites", "mbtiles", "pmtiles", "cog", "postgres"]
+default = ["fonts", "styles", "sprites", "mbtiles", "pmtiles", "experimental-cog", "postgres"]
 postgres = [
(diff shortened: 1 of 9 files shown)


martin-core/Cargo.toml
```diff
@@ -12,7 +12,15 @@ rust-version.workspace = true
 homepage.workspace = true
 
 [features]
-default = ["fonts", "styles", "sprites", "mbtiles", "pmtiles", "cog", "postgres"]
+default = [
+    "fonts",
```
nyurik, line 16: sort?

nyurik (approved):
should we rename it to `cog-experimental` ? This way it keeps the ordering.  Also, please sort all the multi-line things to keep my OCD happy? :)

CommanderStorm:
I renmaed it to `unstable-cog` instead.
This seems to be the consensus in the ecosystem and some of our tooling has support for this:

https://github.com/obi1kenobi/cargo-semver-checks/blob/feb8310d6ef4914b325070d79613e03df6d00885/src/rustdoc_gen.rs#L109-L158

PR #2287, opened 2025-10-06
chore(lints): enable `clippy::unimplemented` and `clippy::panic`

CommanderStorm:


martin/build.rs
```diff
@@ -1,3 +1,8 @@
+#![allow(
```
nyurik: should this be `expect` ?

martin/src/config/file/postgres/resolver/query_functions.rs
```diff
@@ -59,7 +59,10 @@ pub async fn query_available_function(pool: &PostgresPool) -> PostgresResult<Sql
                     assert_eq!(t, &["bytea", "text"]);
                 }
                 (None, None) => {}
-                _ => panic!("Invalid output record names or types: {output_record_names:?} {output_record_types:?}"),
+                #[allow(clippy::panic, reason = "Can only happen if postgres changes their code. We have tests against this")]
```
nyurik: expect?

nyurik (approved):
i'm ok with this, but should we use `expect` more often?

martin/build.rs
CommanderStorm: I now use `allow`, when a warning is conditional on `cfg`-flags and `expect`.
The conditional allow is bad to read IMO and likely also maintain.

Would love your opinion on this generally though. ^^

martin/src/config/file/postgres/resolver/query_functions.rs
CommanderStorm:
```suggestion
                #[expect(clippy::panic, reason = "Can only happen if postgres changes their code. We have tests against this")]
```

mbtiles/src/metadata.rs
```diff
@@ -244,6 +244,10 @@ pub async fn anonymous_mbtiles(script: &str) -> (Mbtiles, SqliteConnection) {
 }
 
 /// Create a named, in memory, temporary mbtile connection with the given `script`
+#[allow(
```
CommanderStorm:
```suggestion
#[expect(
```

mbtiles/tests/copy.rs
```diff
@@ -780,7 +780,14 @@ async fn dump(conn: &mut SqliteConnection) -> MbtResult<Vec<SqliteEntry>> {
                                     )
                                 )
                             }),
-                            _ => panic!("Unknown column type: {typ}"),
+
+                            #[allow(
```
CommanderStorm:
```suggestion
                            #[expect(
```

PR #2288, opened 2025-10-06
chore(lints): audit all allows, add reasons and remove unnessesary ones

CommanderStorm:
I went thorugh all of our allows and marked them as expect if possible.

I also added a few reason =.. blocks

Both are not by far enough to enable the respective restriction, but baby steps


nyurik (approved):
:trophy:

PR #2292, opened 2025-10-07
build: Make mbtiles dependency properly optional again

jleedev:


CommanderStorm:
I don't think this will work.
`martin-cp` depends on this dependency.
Not sure if we can mark this as enabled when building for one --bin target but not for another

https://github.com/maplibre/martin/blob/af86800f13810dc1ddf687077067bff31814222c/martin/src/bin/martin-cp.rs#L28

nyurik:
https://doc.rust-lang.org/cargo/reference/cargo-targets.html#the-required-features-field

jleedev:
Already is: https://github.com/maplibre/martin/blob/af86800f13810dc1ddf687077067bff31814222c/martin/Cargo.toml#L76

jleedev:
Adding something like `cargo check --workspace --no-default-features --all-targets` to the unit test step would exercise that — removing the required-features line would break it. The server doesn't start with no features, but all targets should at least build.

justfile
```diff
@@ -101,6 +101,7 @@ book:  (cargo-install 'mdbook') (cargo-install 'mdbook-alerts') (cargo-install '
 # Quick compile without building a binary
 check: (cargo-install 'cargo-hack')
     cargo hack check --all-targets --each-feature --workspace
+    cargo hack check --all-targets --no-default-features --workspace^
```
```suggestion
    cargo hack check --all-targets --no-default-features --workspace
```

CommanderStorm (approved):
thanks ❤️

justfile
```diff
@@ -101,6 +101,7 @@ book:  (cargo-install 'mdbook') (cargo-install 'mdbook-alerts') (cargo-install '
 # Quick compile without building a binary
 check: (cargo-install 'cargo-hack')
     cargo hack check --all-targets --each-feature --workspace
+    cargo hack check --all-targets --no-default-features --workspace
```
nyurik, line 104: why do we need `cargo hack` for a basic check run?  why not `cargo check`?  Also, are we certain `cargo hack check` doesn't include `no-default-features` mode?

nyurik: @CommanderStorm and @jleedev when you have a sec - please see if we actually need it?

jleedev: It appears that hack will already test no features based on https://github.com/taiki-e/cargo-hack/blob/5da681fec58315e6327dd3130f414a9b8fbe2996/tests/test.rs#L472

PR #2293, opened 2025-10-07
feat(core): enable overriding of the automatic hashing for source traits

CommanderStorm:
On the path to eliminating the hashing for some of our code paths, this needs to live where sources can override it -> in the source trait.


nyurik (approved):
one concern i have with our current etags is that it is base-10, which is wasteful from bandwith perspective (partially address by gzip, but still). Instead, we should target base64, perhaps the [RFC 4648 §5](https://datatracker.ietf.org/doc/html/rfc4648#section-5): base64url (URL- and filename-safe standard)[[a]](https://en.wikipedia.org/wiki/Base64#cite_note-common-8) -- the one that uses `-` and `_` variant for 63 and 64th chars (although any would work because they are double-quoted anyway

martin-core/src/tiles/tile.rs
```diff
@@ -21,12 +21,33 @@ pub struct Tile {
...
     /// Metadata about the tile's format and encoding
     pub info: TileInfo,
+    /// Pre-computed etag/hash for the tile data (empty for empty tiles)
+    pub etag: String,
 }
 
 impl Tile {
     /// Creates a new tile with the given tile data and metadata.
+    ///
+    /// For empty tiles, etag will be "0", otherwise [`xxh3_128(data)`](xxhash_rust::xxh3::xxh3_128).
```
CommanderStorm:
```suggestion
    /// For empty tiles, etag will be "base64(0)", otherwise [`base64(xxh3_128(data))`](xxhash_rust::xxh3::xxh3_128).
```

nyurik (approved):

nyurik:
thanks!

PR #2295, opened 2025-10-08
feat(cache): implement sprite caching

### LINKED ISSUE #1342: sprite generation is uncached causing problems with big svgs
Hello, for several months, I have tried many things to host my own map layers. So, I decided to test Martin. I can now host my mbtiles and also SVGs via Martin using Docker. However, when I use large SVGs (size 500x500 each one for example), accessing the sprite via Maputnik (or maplibre-gl) takes a lot of time (about 20 seconds to load an image of only 1MB). Why? Is this .png image not generated when launching the Docker container? Or is it generated on each call to the sprite URL? I need to have excellent image quality...

CommanderStorm:
This PR adds sprite caching support to Martin.

Resolves #1342


martin/src/config/file/main.rs
```diff
@@ -182,44 +196,95 @@ impl Config {
...
+                tile_cache_size_mb: 256,
+                #[cfg(feature = "pmtiles")]
+                pmtiles_cache_size_mb: 128,
+                #[cfg(feature = "sprites")]
+                sprite_cache_size_mb: 64,
+                #[cfg(feature = "fonts")]
+                font_cache_size_mb: 64,
+            }
+        }
+    }
```
CommanderStorm, line 300: @nyurik Could you review the basic concept of this work?

I split the cache into multiple parts.
The PR is not yet in a reviewable state, but if the basic concept of this work is something we can sync on would still be quite nice to know 😉

nyurik: seems reasonable.  We could of course have simpler division of tiles vs everything else, but giving a bit more control is probably ok.  Reducing size is probably not worth it if we do not pre-allocate memory. The only optimization here might be due async cache purging - if we have a dedicated process to purge cache (doubt it), it might be worth it to not even start that cache sub-system. Apart from that, probably not worth it.

martin-core/src/resources/sprites/cache.rs
```diff
@@ -0,0 +1,131 @@
+use actix_web::web::Bytes;
+use moka::future::Cache;
+
+/// Sprite cache for storing generated sprite sheets.
+
```
CommanderStorm:
```suggestion
```

martin/src/config/file/resources/fonts.rs
```diff
@@ -1,5 +1,3 @@
-use std::collections::BTreeMap;
-
 use martin_core::fonts::FontSources;
```
CommanderStorm, line 1:
```suggestion
use std::collections::BTreeMap;
use martin_core::fonts::FontSources;
```

martin-core/src/resources/sprites/cache.rs
```diff
@@ -0,0 +1,130 @@
...
+            );
+        } else {
+            log::trace!("Sprite cache MISS for ids={ids}, sdf={as_sdf}, json={as_json}");
+        }
+
+        result
+    }
+
+    /// Inserts a sprite sheet into the cache.
+    pub async fn insert(&self, ids: &str, as_sdf: bool, as_json: bool, bytes: Bytes) {
```
nyurik: if you need an owned param, you should ask for it. I.e. - rule of thumb - if you have to do `ids.to_string()`, just ask for `String` as the param

martin/src/config/file/cache.rs
```diff
@@ -18,8 +21,11 @@ impl CacheConfig {
     #[must_use]
     pub fn create_tile_cache(&self) -> Option<martin_core::tiles::TileCache> {
         if self.tile_cache_size_mb > 0 {
+            log::info!(
+                "Initializing tile cache with maximum size {} MiB",
```
nyurik: can we please just use `MB`? Makes life soo much more readable :)

nyurik (approved):
minor nits

martin/src/config/file/cache.rs
CommanderStorm: Read:
- change to base `1000` too or
- stick with base `1024` and incorrectly/windows-style mark this as `MB`

CommanderStorm: Our frontend also uses `MiB`, so better change there too

CommanderStorm: mbtiles also uses `KiB` => via `size_format::SizeFormatterBinary`

nyurik: sigh, just remember that our config values are `_mb` -- so we want consistency.  Also, very few ppl use `MiB` - this is one of those "technically accurate, but who cares" type of things :)

nyurik (approved):

martin/src/config/file/cache.rs
CommanderStorm: The difference is a 5% larger or smaller cache.

I adjusted it to be option 1 => change to base 1000 too.
That seems to be the most correct thing to do

PR #2301, opened 2025-10-11
docs: clearer pg function documentation

CommanderStorm:
I got bitten a bit by there not existing a lot of documentation on how postgres sources have to be written and why.

This PR tries to improve this


nyurik (approved):
thanks!

PR #2315, opened 2025-10-27
chore(admin): move functionality into better modules

CommanderStorm:
To make the reorganisation of the admin UI port handling not a bigger mess, this PR moves:
- all tile srv functionality into its own module
- all admin  srv functionality into its own module

### DIFF
--- a/martin/src/srv/server.rs
+++ b/martin/src/srv/server.rs
@@ -10,18 +10,13 @@ use actix_web::{App, HttpResponse, HttpServer, Responder, middleware, route, web
 use futures::TryFutureExt;
 #[cfg(feature = "lambda")]
 use lambda_web::{is_running_on_lambda, run_actix_on_lambda};
-use serde::{Deserialize, Serialize};
 
 #[cfg(all(feature = "webui", not(docsrs)))]
 use crate::config::args::WebUiMode;
 use crate::config::file::ServerState;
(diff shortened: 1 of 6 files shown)


martin/src/srv/server.rs
```diff
@@ -10,19 +10,14 @@ use actix_web::{App, HttpResponse, HttpServer, Responder, middleware, route, web
 use futures::TryFutureExt;
 #[cfg(feature = "lambda")]
 use lambda_web::{is_running_on_lambda, run_actix_on_lambda};
-use serde::{Deserialize, Serialize};
 
 #[cfg(all(feature = "webui", not(docsrs)))]
 use crate::config::args::WebUiMode;
 use crate::config::file::ServerState;
 use crate::config::file::srv::{KEEP_ALIVE_DEFAULT, LISTEN_ADDRESSES_DEFAULT, SrvConfig};
+use crate::srv::admin::Catalog;
```
nyurik, line 18: so is this `admin` or `webui`?

nyurik (approved):
a bit of a naming confusion

martin/src/srv/server.rs
CommanderStorm: ???

Not sure what you mean.

If we move all endpoints that allow access to internal state, this includes /catalog.

CommanderStorm: not just the frontend and metrics

nyurik: i guess i was confused for a bit if admin and webui refer to the same concept. NVM

PR #2326, opened 2025-11-01
fix: restrict zooming and panning on the data inspector (#2325)

### LINKED ISSUE #2325: Restrict zooming and panning on the data inspector
Currently, we just show 0/0/0 in the inspector in the fronted.

This works for global datasets, but does not for more local datasets.
For these smaller datasets, we should restrict our zoom to the extent of the dataset.

As a visualisation:

Image

Currently shows as:

Image

On the frontend, we currently don't store the bbox + min/maxzoom.
We need to either:
- take data that maplibre-gl-js already fetched
- or fetch the tilejson ourselves

The data about this is not avaliable in the catalog, so won't be avaliable to the frontend without extra work.

harshasiddartha:
## Description

Fixes issue #2325 by restricting zooming and panning in the data inspector to the dataset's extent.

Currently, the inspector shows 0/0/0 (zoom level 0/0/0) which works for global datasets but not for local/smaller datasets. This PR implements zoom and pan restrictions based on TileJSON metadata.

## Changes

- **Fetch TileJSON** when the inspector dialog opens from the API endpoint `/{source_ids}`
- **Extract metadata** from TileJSON: `bounds`, `minzoom`, `maxzoom`, and `center`
- **Configure map** with:
  - `initialViewState` using center coordinates if available
  - `minZoom` and `maxZoom` restrictions
  - `maxBounds` to prevent panning outside the dataset extent
  - `fitBounds()` to automatically fit the map to the dataset bounds when available
- **Handle both raster and vector sources** with appropriate configuration
- **Update tests** to mock TileJSON fetch calls

## Technical Details

The implementation:
1. Fetches TileJSON asynchronously when the dialog opens
2. Applies restrictions through both React props (`initialViewState`, `minZoom`, `maxZoom`, `maxBounds`) and direct map API calls (`setMinZoom`, `setMaxZoom`, `setMaxBounds`, `fitBounds`)
3. Handles cases where TileJSON loads after the map has already initialized
4. Gracefully continues without restrictions if TileJSON fetch fails

## Testing

- Updated existing tests to mock TileJSON fetch
- Added test to verify TileJSON is fetched when dialog opens
- All existing tests continue to pass

## Screenshots

Before: Inspector showed blank map with no restrictions (0/0/0)
After: Inspector restricts zoom/pan to dataset bounds and fits to extent

## Related Issues

Closes #2325

### DIFF
--- a/martin/martin-ui/src/components/dialogs/tile-inspect.tsx
+++ b/martin/martin-ui/src/components/dialogs/tile-inspect.tsx
@@ -79,8 +79,8 @@ export function TileInspectDialog({ name, source, onCloseAction }: TileInspectDi
           [east, north],
         ],
         {
-          padding: 50,
           maxZoom: tileJSON.maxzoom,
+          padding: 50,
         },
       );
     } else if (tileJSON.center) {
@@ -176,25 +176,15 @@ export function TileInspectDialog({ name, source, onCloseAction }: TileInspectDi
           <section className="border rounded-lg overflow-hidden">
             {isImageSource ? (
               <MapLibreMap
-                ref={mapRef}
-                reuseMaps={false}
-                onLoad={() => {
-                  // Configure bounds for raster sources after map loads
-                  if (tileJSON) {
-                    configureMapBounds();
-                  }
-                }}
                 initialViewState={
                   tileJSON?.center
                     ? {
-                        longitude: tileJSON.center[0],
                         latitude: tileJSON.center[1],
+                        longitude: tileJSON.center[0],
                         zoom: tileJSON.center[2] ?? 0,
                       }
                     : undefined
                 }
-                minZoom={tileJSON?.minzoom}
-                maxZoom={tileJSON?.maxzoom}
                 maxBounds={
                   tileJSON?.bounds
                     ? [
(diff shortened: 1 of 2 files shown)


martin/martin-ui/src/components/dialogs/tile-inspect.tsx
```diff
@@ -76,6 +176,31 @@ export function TileInspectDialog({ name, source, onCloseAction }: TileInspectDi
...
+                    : undefined
+                }
+                maxZoom={tileJSON?.maxzoom}
+                minZoom={tileJSON?.minzoom}
+                onLoad={() => {
+                  // Configure bounds for raster sources after map loads
+                  if (tileJSON) {
+                    configureMapBounds();
+                  }
+                }}
```
CommanderStorm, line 203: You always seem to configure this twice:
- once in the react component
- once in the onLoad callback

Could you explain this? 🤔
Seems a bit strange

martin/martin-ui/src/components/dialogs/tile-inspect.tsx
```diff
@@ -27,6 +44,86 @@ export function TileInspectDialog({ name, source, onCloseAction }: TileInspectDi
...
+      }
+    }
+  }, [tileJSON]);
+
+  // Fetch TileJSON when dialog opens
+  useEffect(() => {
+    let cancelled = false;
+
+    fetch(buildMartinUrl(`/${name}`))
+      .then((response) => {
```
CommanderStorm, line 100: not sure if this is very react-y.

I think the current recomendation is to use `<Suspense` + `asyncData` and just wrap the components where we need the data.

At least this is explained like this here:

https://stackoverflow.com/questions/53332321/react-hook-warnings-for-async-function-in-useeffect-useeffect-function-must-ret

What do you think? 🤔

CommanderStorm (changes requested):
Nice!

I tested the code and the first case (a raster layer) works.
The second case however (a vector map) does not seem to work nicely:

image


You can reproduce this by running:

```
VITE_MARTIN_BASE=https://nav.tum.de/martin npm run dev
```

and then visting <http://localhost:3001/?inspect=shortbread&tab=tiles>

CommanderStorm:
Lets close this PR since it has become a bit stale.
Thank you for this work nevertheless!

PR #2340, opened 2025-11-05
fix: fix assertion

YichiZhang0613:
In martin/martin-tile-utils/src/lib.rs, for get_zoom_precision's assertion, the message requires zoom <= MAX_ZOOM while predicate requires zoom < MAX_ZOOM. Judging by how MAX_ZOOM is used elsewhere, the assertion here should be changed to `assert!(zoom <= MAX_ZOOM, "zoom {zoom} must be <= {MAX_ZOOM}");`
```rust
/// Compute precision of a zoom level, i.e. how many decimal digits of the longitude and latitude are relevant
#[must_use]
#[expect(clippy::cast_possible_truncation, clippy::cast_sign_loss)]
pub fn get_zoom_precision(zoom: u8) -> usize {
    assert!(zoom < MAX_ZOOM, "zoom {zoom} must be <= {MAX_ZOOM}");
...
}
```

### DIFF
--- a/martin-tile-utils/src/lib.rs
+++ b/martin-tile-utils/src/lib.rs
@@ -312,7 +312,7 @@ pub fn bbox_to_xyz(left: f64, bottom: f64, right: f64, top: f64, zoom: u8) -> (u
 #[must_use]
 #[expect(clippy::cast_possible_truncation, clippy::cast_sign_loss)]
 pub fn get_zoom_precision(zoom: u8) -> usize {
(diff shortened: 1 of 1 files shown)


CommanderStorm (approved):
Fair, defined as valid here for example:

https://github.com/maplibre/martin/blob/f87ac4fd36c1245532ba7f247041a8076ae393bc/mbtiles/src/validation.rs#L384

Thanks for the deep audit 😉

PR #2350, opened 2025-11-07
fix: libdeflater build for musl

CommanderStorm:
this PR tests if libdeflater needs this feature flag in the newer version


CommanderStorm:
does not work

PR #2352, opened 2025-11-07
docs(config): Move config file to a separate file

CommanderStorm:
If our config file from the docs is currently copy pasted, it does not work as it is not valid yaml.

This fixes this, hopefully before someone was affected by this.


nyurik (approved):
thx, i have been wanting to do that for a while now :)

CommanderStorm:
Yes, still not tested, but not sure if this is possible.

PR #2370, opened 2025-11-24
feat(martin-core): add an pmtiles example

CommanderStorm:
This is one of the examples that we talked about

### DIFF
--- a/martin-core/Cargo.toml
+++ b/martin-core/Cargo.toml
@@ -96,6 +96,7 @@ walkdir = { workspace = true, optional = true }
 xxhash-rust.workspace = true
 
 [dev-dependencies]
+actix-web.workspace = true
 approx.workspace = true
 env_logger.workspace = true
 futures.workspace = true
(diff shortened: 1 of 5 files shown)


martin-core/examples/pmtiles-tileserver.rs
```diff
@@ -0,0 +1,133 @@
...
+                "source": "pmtiles"
+            }
+        ]
+    });
+
+    Ok(HttpResponse::Ok().json(style))
+}
+
+async fn index() -> ActixResult<HttpResponse> {
+    let html = r#"<!DOCTYPE html>
```
nyurik: Would it make sense to place this code into `pmtiles-tileserver.html` and use an include macro? I am ok either way, just a thought

nyurik (approved):

nyurik (approved):
thx!

martin-core/Cargo.toml
```diff
@@ -97,11 +97,18 @@ xxhash-rust.workspace = true
 
 [dev-dependencies]
 approx.workspace = true
```
CommanderStorm, line 100:
```suggestion
actix-web.workspace = true
approx.workspace = true
```

Issue #2212, opened 2025-09-20
martin-cp: better nullness/bbox handling for vastly increased performance


CommanderStorm:
Currently, when martin-cp sees the following, the green area will be attempted to copy.

<img width="1364" height="810" alt="Image" src="https://github.com/user-attachments/assets/ed7a9466-a36b-4ea6-ac0c-00a228309a27" />

This is suboptimal, as the green area cannot ever have tiles.
We should either 
- if the query is taking very long (more than 10 minutes, or something) calculate the bboxes for each source and see if we can skip tiles
- use the bboxes of sources to pre-filter sources (This might also apply to martin itsself, I have not checked)

The motivation for this is the following observation:

<img width="1306" height="141" alt="Image" src="https://github.com/user-attachments/assets/e616c459-b2c4-47de-966c-e21b51b98dc1" />

todtb:
@CommanderStorm, this is something I've thought a bit about but wasn't sure if there is a universal approach on how to handle this. I think it also depends quite a bit on the extent of the data (global vs local) and how the mbtiles are being used (offline, "cache" of static data, etc).

## Global vs local
For small, locally constrained data sets, I've been able to use existing `ST_` functions to generate the list of bboxes, so it should be possible to use similar logic to eliminate generating empty tiles. 

However, for global data sets (think adm0) bbox unions quickly become very large and are further complicated by geometries spanning the antimeridian (US, RU, KI, FJ, possibly more depending on sources/definitions and their bboxes).

Perhaps the adm examples are more niche than I imagine, but the fundamental challenge remains.


## Approaches
The way I'm currently using the tool is simply passing in the bboxes I'm concerned with:
```
martin-cp  --output-file /tiles/out.mbtiles --mbtiles-type normalized \
"--bbox=..." \
"--bbox=..." \
"--bbox=..." \
...
--config /config/mbtile_config.yaml --source source --on-duplicate override
```

These bboxes are generated, but manually validated or adjusted as needed.

I imagine you were thinking about something a little less manual, but I'm wondering if you can describe a rough heuristic of what you were thinking?

CommanderStorm:
Yea, pretty much what I was thinking.
Having to add those boxes for acceptable performance is not what we want, way too error-proawn. 😅

My thinking was a bit more simple/naiive.
taking one bbox per source would in fact already be an improvement and we have this data in the tilejson.

todtb:
[...]

PR #2385, opened 2025-11-26
feat(martin-cp): infer default bounds from configured sources for better performance

todtb:
#2212

- Adds more dynamic default bounds (bbox) to `martin-cp`
- If no bounds passed as args, and a single source exists with defined bounds, use them. Fallback to global as previously done.

### DIFF
--- a/martin/src/bin/martin-cp.rs
+++ b/martin/src/bin/martin-cp.rs
@@ -363,7 +363,7 @@ fn default_bounds(src: &DynTileSource) -> Vec<Bounds> {
                 source
                     .get_tilejson()
                     .bounds
-                    .unwrap_or_else(|| Bounds::MAX_TILED)
+                    .unwrap_or(Bounds::MAX_TILED)
             })
             .collect::<Vec<Bounds>>();
 
--- a/martin/src/bin/martin-cp.rs
+++ b/martin/src/bin/martin-cp.rs
@@ -691,10 +691,10 @@ mod tests {
     #[rstest]
     #[case::one_source(one_source(), "test_source", vec![Bounds::from_str("-120.0,30.0,-110.0,40.0").unwrap()])]
     #[case::many_sources(many_sources(), "test_source,test_source2", vec![Bounds::from_str("-110.0,20.0,-120.0,80.0").unwrap(), Bounds::from_str("-130.0,40.0,-170.0,10.0").unwrap()])]
-    #[case::many_sources_rev(many_sources(), "test_source2,test_source", vec![Bounds::from_str("-110.0,20.0,-120.0,80.0").unwrap(), Bounds::from_str("-130.0,40.0,-170.0,10.0").unwrap()])]
+    #[case::many_sources_rev(many_sources(), "test_source2,test_source", vec![Bounds::from_str("-130.0,40.0,-170.0,10.0").unwrap(), Bounds::from_str("-110.0,20.0,-120.0,80.0").unwrap()])]
     #[case::many_sources_only_unbounded(many_sources(), "unbounded_source", vec![Bounds::MAX_TILED])]
     #[case::many_sources_bounded_and_unbounded(many_sources(), "test_source,unbounded_source", vec![Bounds::from_str("-110.0,20.0,-120.0,80.0").unwrap(), Bounds::MAX_TILED])]
-    #[case::many_sources_bounded_and_unbounded_rev(many_sources(), "unbounded_source,test_source", vec![Bounds::from_str("-110.0,20.0,-120.0,80.0").unwrap(), Bounds::MAX_TILED])]
+    #[case::many_sources_bounded_and_unbounded_rev(many_sources(), "unbounded_source,test_source", vec![Bounds::MAX_TILED, Bounds::from_str("-110.0,20.0,-120.0,80.0").unwrap()])]
     #[case::source_wo_bounds(source_wo_bounds(), "test_source", vec![Bounds::MAX_TILED])]
     fn test_default_bounds(
         #[case] src: TileSources,
--- a/martin/src/bin/martin-cp.rs
+++ b/martin/src/bin/martin-cp.rs
@@ -353,22 +353,25 @@ fn check_sources(args: &CopyArgs, state: &ServerState) -> Result<String, MartinC
 }
 
 fn default_bounds(src: &DynTileSource) -> Vec<Bounds> {
-    if src.sources.len() == 1 {
-        match src.sources.first().unwrap().get_tilejson().bounds {
-            Some(bounds) => {
-                info!("No bbox specified, using source bounds: {}", bounds);
-                vec![bounds]
-            }
-            None => {
-                info!(
-                    "No configured bounds for source, using: {}",
-                    Bounds::MAX_TILED
-                );
-                vec![Bounds::MAX_TILED]
-            }
-        }
-    } else {
+    if src.sources.is_empty() {
         vec![Bounds::MAX_TILED]
+    } else {
+        let mut source_bounds = src.sources.iter().filter_map(|source|
+            source.get_tilejson().bounds
+        ).collect::<Vec<Bounds>>();
+        
+        source_bounds.dedup_by_key(|bounds| bounds.to_string());
+        
+        if source_bounds.is_empty() {
+            info!("No configured bounds for source, using: {}", Bounds::MAX_TILED);
+            vec![Bounds::MAX_TILED]
+        } else {
+            info!(
+                "No bbox specified, using source bounds: {}",
+                source_bounds.iter().map(|s| format!("[{s}]")).collect::<Vec<String>>().join(", ")
+            );
+            source_bounds
+        }
     }
 }
 
@@ -645,6 +648,11 @@ mod tests {
                 tj: tilejson! { tiles: vec![], bounds: Bounds::from_str("-130.0,40.0,-170.0,10.0").unwrap() },
                 data: Vec::default(),
             }),
+            Box::new(MockSource {
+                id: "unrequested_source",
+                tj: tilejson! { tiles: vec![], bounds: Bounds::from_str("-150.0,40.0,-120.0,10.0").unwrap() },
+                data: Vec::default()
+            })
         ]])
     }
 
@@ -668,7 +676,7 @@ mod tests {
 
     #[rstest]
     #[case::one_source(one_source(), "test_source", vec![Bounds::from_str("-120.0,30.0,-110.0,40.0").unwrap()])]
-    #[case::many_sources(many_sources(), "test_source,test_source2", vec![Bounds::MAX_TILED])]
+    #[case::many_sources(many_sources(), "test_source,test_source2", vec![Bounds::from_str("-110.0,20.0,-120.0,80.0").unwrap(), Bounds::from_str("-130.0,40.0,-170.0,10.0").unwrap()])]
     #[case::source_wo_bounds(source_wo_bounds(), "test_source", vec![Bounds::MAX_TILED])]
     fn test_default_bounds(
         #[case] src: TileSources,
--- a/martin/src/bin/martin-cp.rs
+++ b/martin/src/bin/martin-cp.rs
@@ -193,13 +193,7 @@ async fn start(copy_args: CopierArgs) -> MartinCpResult<()> {
     run_tile_copy(copy_args.copy, sources).await
 }
 
-fn check_bboxes(args: &CopyArgs) -> MartinCpResult<Vec<Bounds>> {
-    let boxes = if args.bbox.is_empty() {
-        vec![Bounds::MAX_TILED]
-    } else {
-        args.bbox.clone()
-    };
-
+fn check_bboxes(boxes: Vec<Bounds>) -> MartinCpResult<Vec<Bounds>> {
     for bb in &boxes {
         let allowed_lon = Bounds::MAX_TILED.left..=Bounds::MAX_TILED.right;
         if !allowed_lon.contains(&bb.left) || !allowed_lon.contains(&bb.right) {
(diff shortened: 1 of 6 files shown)


martin/src/bin/martin-cp.rs
```diff
@@ -566,12 +588,97 @@ async fn main() {
...
+    use tilejson::{TileJSON, tilejson};
 
-    use insta::assert_yaml_snapshot;
-    use rstest::rstest;
+    #[derive(Debug, Clone)]
+    pub struct MockSource {
+        pub id: &'static str,
+        pub tj: TileJSON,
+        pub data: TileData,
+    }
```
todtb, line 620: It looks like there's a couple ways mocks and fixtures are organized in the project. Any thoughts on these for bins?

martin/src/bin/martin-cp.rs
```diff
@@ -343,20 +337,41 @@ fn iterate_tiles(tiles: Vec<TileRect>) -> impl Iterator<Item = TileCoord> {
...
-        if let Some(source) = sources.first() {
-            if sources.len() > 1 {
-                return Err(MartinCpError::MultipleSources(sources.join(", ")));
+        let source_ids = state.tiles.source_names();
+        if let Some(source_id) = source_ids.first() {
+            if source_ids.len() > 1 {
+                return Err(MartinCpError::MultipleSources(source_ids.join(", ")));
             }
-            Ok(source.clone())
+            Ok(source_id.clone())
```
todtb, line 350: This is simply a variable rename to reflect what it represents.

However, I was curious how this relates to composite sources.

martin/src/bin/martin-cp.rs
```diff
@@ -343,20 +337,41 @@ fn iterate_tiles(tiles: Vec<TileRect>) -> impl Iterator<Item = TileCoord> {
...
+                    "No configured bounds for source, using: {}",
+                    Bounds::MAX_TILED
+                );
+                vec![Bounds::MAX_TILED]
+            }
+        }
+    } else {
+        vec![Bounds::MAX_TILED]
+    }
+}
```
CommanderStorm, line 387: How about we iterate over sources and take the bound of the source if available for all, else the world?.

I think the multi-source merging case is better supported this way and composite sources should work much better.

martin/src/bin/martin-cp.rs
CommanderStorm: Bins? Not sure I follow what you mean.

Current tests you added look very good if this is what you are asking 👍

CommanderStorm: Bins? Not sure I follow what you mean.

Current tests you added look very good if this is what you are asking 👍

CommanderStorm (commented):
I am not sure if you want a review now.

Code wise this looks very good. I think if we change the default bbox generation a bit, composite sources are supported (the original reason for the issue) too.

martin/src/bin/martin-cp.rs
todtb: Sorry, I suppose that was an odd way of putting it 😄 

I meant that I had included the mock structs and fixtures within the  `src/bin/*.rs` files. I saw in some other areas the mocks were exposed via `mod.rs` and included/used.

But if this works, great!

todtb:
> I am not sure if you want a review now.
> 
> Code wise this looks very good. I think if we change the default bbox generation a bit, composite sources are supported (the original reason for the issue) too.

I honestly wasn't expecting it with the timezone diff, but thanks for the feedback. I hope to have those changes to the bbox fn made within a couple days.

martin/src/bin/martin-cp.rs
todtb: So were you thinking essentially a union of bounds (min left/bottom, max right/top) if every `Source` has defined bounds? 

If I understand it correctly the `DynTileSource` is analogous to what is used by maplibergl (ie `map.addSource`), whereas `DynTileSource.sources` (`Vec<BoxedSource>`) are the _data_ sources themselves (ie db tables, functions, etc).

So `if src.sources.len() > 1` then `src.source.iter()...` and construct the union bounds?

I think there's still a use case for handling `src.sources.len() == 1` as well, mainly that if there is a provided `--source` arg, the user only needs to specify `--bbox` arg if overriding what the `--source` actually is configured for. And thus for many simple use cases, it avoids running for the global bbox.

CommanderStorm: A union would not solve the problem that you are trying to solve in https://github.com/maplibre/martin/issues/2212 I think?
Passing on the correct bboxes where data exists is still valuable, right?

If this metadata does NOT exist for one, this is still infectious => `vec![Bounds::MAX_TILED]`

todtb: Yup, you are 100% correct. I completely forgot we're dealing with a vec of bounds here when I was typing that.

Re:
> If I understand it correctly the DynTileSource is analogous to what is used by maplibergl (ie map.addSource), whereas DynTileSource.sources (Vec<BoxedSource>) are the data sources themselves (ie db tables, functions, etc).

Is that an accurate mental model here?

CommanderStorm: Yes your model is correct.

One composite source `/foo,bar` is a `DynTileSource`, the source actually requested.
This `DynTileSource` has access to all all data sources themselves (f.ex. `foo`, `bar`, `bazinga`) and the sources that are requested `foo,bar`.

So thinking about it, we don't quite want all the tile json bboxes, but only the ones which are requested.
The union of all is still likely an improvement over current way this works.

🤔 Another improvement would be to make sure that the requested bbox is fully within the avaliable bbox and both warn the user of this and automatically crop the bbox to what can be requested.

todtb: > This DynTileSource has access to all all data sources themselves (f.ex. foo, bar, bazinga) and the sources that are requested foo,bar.

It looks like the `DynTileSource` constructor is limiting sources to the source_ids passed in here: `martin/src/srv/tiles/content.rs:87`.

So ultimately only the sources provided by the `--source` arg or the fallback as determined here: `martin/src/bin/martin-cp.rs:391`.

I've manually confirmed this but also covered via test: 
```rs
    #[fixture]
    fn many_sources() -> TileSources {
        TileSources::new(vec![vec![
            Box::new(MockSource {
                id: "test_source",
                tj: tilejson! { tiles: vec![], bounds: Bounds::from_str("-110.0,20.0,-120.0,80.0").unwrap() },
                data: Vec::default()
            }),
            Box::new(MockSource {
                id: "test_source2",
                tj: tilejson! { tiles: vec![], bounds: Bounds::from_str("-130.0,40.0,-170.0,10.0").unwrap() },
                data: Vec::default()
            }),
            Box::new(MockSource {
                id: "unrequested_source",
                tj: tilejson! { tiles: vec![], bounds: Bounds::from_str("-150.0,40.0,-120.0,10.0").unwrap() },
                data: Vec::default()
            })
        ]])
    }
...
#[case::many_sources(many_sources(), "test_source,test_source2", vec![Bounds::from_str("-110.0,20.0,-120.0,80.0").unwrap(), Bounds::from_str("-130.0,40.0,-170.0,10.0").unwrap()])]
```

martin/src/bin/martin-cp.rs
```diff
@@ -566,12 +600,102 @@ async fn main() {
...
+            Box::new(MockSource {
+                id: "test_source2",
+                tj: tilejson! { tiles: vec![], bounds: Bounds::from_str("-130.0,40.0,-170.0,10.0").unwrap() },
+                data: Vec::default(),
+            }),
+            Box::new(MockSource {
+                id: "unrequested_source",
+                tj: tilejson! { tiles: vec![], bounds: Bounds::from_str("-150.0,40.0,-120.0,10.0").unwrap() },
+                data: Vec::default(),
+            }),
```
CommanderStorm, line 666:
```suggestion
            Box::new(MockSource {
                id: "unrequested_source",
                tj: tilejson! { tiles: vec![], bounds: Bounds::from_str("-150.0,40.0,-120.0,10.0").unwrap() },
                data: Vec::default(),
            }),
            Box::new(MockSource {
                id: "unbounded_source",
                tj: tilejson! { tiles: vec![] },
                data: Vec::default(),
            }),
```

martin/src/bin/martin-cp.rs
```diff
@@ -566,12 +600,102 @@ async fn main() {
...
+        TileSources::new(vec![vec![Box::new(MockSource {
+            id: "test_source",
+            tj: tilejson! { tiles: vec![] },
+            data: Vec::default(),
+        })]])
+    }
+
+    #[rstest]
+    #[case::one_source(one_source(), "test_source", vec![Bounds::from_str("-120.0,30.0,-110.0,40.0").unwrap()])]
+    #[case::many_sources(many_sources(), "test_source,test_source2", vec![Bounds::from_str("-110.0,20.0,-120.0,80.0").unwrap(), Bounds::from_str("-130.0,40.0,-170.0,10.0").unwrap()])]
```
CommanderStorm, line 695: lets also test the case for multi source with one being without a bound

```suggestion
    #[case::many_sources(many_sources(), "test_source,test_source2", vec![Bounds::from_str("-110.0,20.0,-120.0,80.0").unwrap(), Bounds::from_str("-130.0,40.0,-170.0,10.0").unwrap()])]
    #[case::many_sources_rev(many_sources(), "test_source2,test_source", vec![Bounds::from_str("-110.0,20.0,-120.0,80.0").unwrap(), Bounds::from_str("-130.0,40.0,-170.0,10.0").unwrap()])]
    #[case::many_sources_only_unbounded(many_sources(), "unbounded_source", vec![Bounds::MAX_TILED])]
    #[case::many_sources_bounded_and_unbounded(many_sources(), "test_source,unbounded_source", vec![Bounds::from_str("-110.0,20.0,-120.0,80.0").unwrap(), Bounds::MAX_TILED])]
    #[case::many_sources_bounded_and_unbounded_rev(many_sources(), "unbounded_source,test_source", vec![Bounds::from_str("-110.0,20.0,-120.0,80.0").unwrap(), Bounds::MAX_TILED])]
```

martin/src/bin/martin-cp.rs
```diff
@@ -343,20 +337,53 @@ fn iterate_tiles(tiles: Vec<TileRect>) -> impl Iterator<Item = TileCoord> {
...
 }
+
+fn default_bounds(src: &DynTileSource) -> Vec<Bounds> {
+    if src.sources.is_empty() {
+        vec![Bounds::MAX_TILED]
+    } else {
+        let mut source_bounds = src
+            .sources
+            .iter()
+            .filter_map(|source| source.get_tilejson().bounds)
```
CommanderStorm: 🤔 Is this correct?

If a source does not have a bound, I think we need to assume the following

```suggestion
            .map(|source| source.get_tilejson().bounds.unwrap_or_else(|| Bounds::MAX_TILED))
```

what do you think?

CommanderStorm (approved):
Code LGTM, except one question and the associated testcase.

Once that question is resolved, we can merge.

martin/src/bin/martin-cp.rs
todtb: This would cover a scenario where:
- The user has not specified args for `bbox`
- The user has configured at least one source with bounds (for example, by `config.yml`)
  - If none configured, that is covered by check below after filter map:
  
   ```rs
   if source_bounds.is_empty() {
    info!(
       "No configured bounds for source, using: {}",
       Bounds::MAX_TILED
     );
    vec![Bounds::MAX_TILED]
   ```
- The user has omitted bounds for at least one source (for example, by omitting them in `config.yml`)

---

I had assumed under such a scenario to use just the bounds configured and available, logging to INFO what was being used.

I suppose we could have the secondary fallback here where if _any_ source does not have configured bounds, add to the vec (and thus effective use) the global bounds. The only drawback is it adds another scenario where the tiles for the global bounds need to be processed.

I'll leave it to you to decide which is more intuitive for an inferred default as up to this point all of my use of the `martin-cp` tool has been either very explicit via multiple bbox args/each source's bounds defined in configuration, or just global, depending on the data set I was using it for.

All that said, I don't have a strong preference either way, because either solution still gives the user ample opportunity to specify bounds via args or `config.yml`

martin/src/bin/martin-cp.rs
todtb: Will hold off updating tests until the approach is decided above, but I was curious here if ordering matters? Looks like that was the intention with the `_rev` tests.

martin/src/bin/martin-cp.rs
CommanderStorm: I think my approach (effectively global bounds) might be more correct for some cases.

Lets prioritise correctness over speed..

martin/src/bin/martin-cp.rs
CommanderStorm: ordering should not matter in downstream code, but this makes us check this if the code changes.

martin/src/bin/martin-cp.rs
```diff
@@ -343,20 +337,58 @@ fn iterate_tiles(tiles: Vec<TileRect>) -> impl Iterator<Item = TileCoord> {
...
+        vec![Bounds::MAX_TILED]
+    } else {
+        let mut source_bounds = src
+            .sources
+            .iter()
+            .map(|source| {
+                source
+                    .get_tilejson()
+                    .bounds
+                    .unwrap_or_else(|| Bounds::MAX_TILED)
```
CommanderStorm: clippy suggestion

```suggestion
                    .unwrap_or(Bounds::MAX_TILED)
```

martin/src/bin/martin-cp.rs
```diff
@@ -566,12 +605,111 @@ async fn main() {
...
+        })]])
+    }
+
+    #[rstest]
+    #[case::one_source(one_source(), "test_source", vec![Bounds::from_str("-120.0,30.0,-110.0,40.0").unwrap()])]
+    #[case::many_sources(many_sources(), "test_source,test_source2", vec![Bounds::from_str("-110.0,20.0,-120.0,80.0").unwrap(), Bounds::from_str("-130.0,40.0,-170.0,10.0").unwrap()])]
+    #[case::many_sources_rev(many_sources(), "test_source2,test_source", vec![Bounds::from_str("-110.0,20.0,-120.0,80.0").unwrap(), Bounds::from_str("-130.0,40.0,-170.0,10.0").unwrap()])]
+    #[case::many_sources_only_unbounded(many_sources(), "unbounded_source", vec![Bounds::MAX_TILED])]
+    #[case::many_sources_bounded_and_unbounded(many_sources(), "test_source,unbounded_source", vec![Bounds::from_str("-110.0,20.0,-120.0,80.0").unwrap(), Bounds::MAX_TILED])]
+    #[case::many_sources_bounded_and_unbounded_rev(many_sources(), "unbounded_source,test_source", vec![Bounds::from_str("-110.0,20.0,-120.0,80.0").unwrap(), Bounds::MAX_TILED])]
```
CommanderStorm: seems like the data I suggested had the wrong order.

```suggestion
    #[case::many_sources_rev(many_sources(), "test_source2,test_source", vec![Bounds::from_str("-130.0,40.0,-170.0,10.0").unwrap(), Bounds::from_str("-110.0,20.0,-120.0,80.0").unwrap()])]
    #[case::many_sources_only_unbounded(many_sources(), "unbounded_source", vec![Bounds::MAX_TILED])]
    #[case::many_sources_bounded_and_unbounded(many_sources(), "test_source,unbounded_source", vec![Bounds::from_str("-110.0,20.0,-120.0,80.0").unwrap(), Bounds::MAX_TILED])]
    #[case::many_sources_bounded_and_unbounded_rev(many_sources(), "unbounded_source,test_source", vec![Bounds::MAX_TILED, Bounds::from_str("-110.0,20.0,-120.0,80.0").unwrap()])]
```

PR #2387, opened 2025-11-28
chore(docs): fix martin-cp bbox docs

todtb:
- Fixes existing docs around allowable bbox range
- Also clarifies behavior added in https://github.com/maplibre/martin/pull/2385


CommanderStorm (approved):
LGTM

PR #2388, opened 2025-11-30
docs(dev-guide): Add some note about alternative Docker hosts/runtimes

yutannihilation:
I get this error when I tried to run `cargo test`, and found this is because I use an alternative Docker host (Colima), which locates its docker socket on a different location (`${HOME}/.colima/default/docker.sock`).

```
failures:

---- tiles::postgres::pool::tests::parse_version stdout ----

thread 'tiles::postgres::pool::tests::parse_version' (107274) panicked at martin-core/src/tiles/postgres/pool.rs:239:14:
container launched: Client(Init(SocketNotFoundError("/var/run/docker.sock")))
```

This can be solved by several ways according to the testcontainers crate's documentation.

https://docs.rs/testcontainers/latest/testcontainers/index.html#docker-host-resolution

This pull request adds a few notes about this. I'm not sure if this is a common issue or not, so please feel free to remove or modify any part that you feel is a bit too specific!

### DIFF
--- a/docs/src/development.md
+++ b/docs/src/development.md
@@ -4,7 +4,7 @@
 
 Install [docker](https://docs.docker.com/get-docker/) and [docker-compose](https://docs.docker.com/compose/)
 
-Note that alternative container hosts/runtimes should also work, but you might need some tweak such as setting `DOCKER_HOST` envvar in order to let [the testcontainers module](https://docs.rs/testcontainers/latest/testcontainers/index.html#docker-host-resolution) detect your container. For example, [Colima uses a different socket location](https://github.com/abiosoft/colima/blob/main/docs/FAQ.md#cannot-connect-to-the-docker-daemon-at-unixvarrundockersock-is-the-docker-daemon-running).
+Note that alternative container hosts/runtimes should also work, but you might need some tweaks such as setting `DOCKER_HOST` envvar in order to let [the testcontainers module](https://docs.rs/testcontainers/latest/testcontainers/index.html#docker-host-resolution) detect your container. For example, [Colima uses a different socket location](https://github.com/abiosoft/colima/blob/main/docs/FAQ.md#cannot-connect-to-the-docker-daemon-at-unixvarrundockersock-is-the-docker-daemon-running).
 
 ## just
 
--- a/docs/src/development.md
+++ b/docs/src/development.md
@@ -4,6 +4,8 @@
 
 Install [docker](https://docs.docker.com/get-docker/) and [docker-compose](https://docs.docker.com/compose/)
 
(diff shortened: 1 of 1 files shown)


docs/src/development.md
```diff
@@ -4,6 +4,8 @@
 
 Install [docker](https://docs.docker.com/get-docker/) and [docker-compose](https://docs.docker.com/compose/)
 
+Note that alternative container hosts/runtimes should also work, but you might need some tweaks such as setting `DOCKER_HOST` envvar in order to let [the testcontainers module](https://docs.rs/testcontainers/latest/testcontainers/index.html#docker-host-resolution) detect your container. For example, [Colima uses a different socket location](https://github.com/abiosoft/colima/blob/main/docs/FAQ.md#cannot-connect-to-the-docker-daemon-at-unixvarrundockersock-is-the-docker-daemon-running).
```
CommanderStorm:
```suggestion
> [!TIP]
> Alternative container hosts/runtimes work, but you need to set the `DOCKER_HOST` environment variable to let [the testcontainers module](https://docs.rs/testcontainers/latest/testcontainers/index.html#docker-host-resolution) detect your container.
> For example, [Colima uses a different socket location](https://github.com/abiosoft/colima/blob/main/docs/FAQ.md#cannot-connect-to-the-docker-daemon-at-unixvarrundockersock-is-the-docker-daemon-running).
```

CommanderStorm (approved):
Thanks and sorry for the many macos specific trouble you are having

nyurik (commented):
Can we automate it with the justfile?

yutannihilation:
Thanks for polishing and merging!

@nyurik 
I was hoping so, but it seems there's no API or convention to know the Docker socket location. According to Colima's document, it assumes there are multiple Docker servers and it's the user's responsibility to explicitly choose which Docker server is the default one. So, I don't think we can automate this, unfortunately... But, I think this is okay. I guess those who don't know well about these details can just use Docker Desktop (and I guess most of them already do so).

> Colima uses Docker contexts to allow co-existence with other Docker servers and sets itself as the default Docker context on startup.
https://github.com/abiosoft/colima/blob/main/docs/FAQ.md#cannot-connect-to-the-docker-daemon-at-unixvarrundockersock-is-the-docker-daemon-running

yutannihilation:
Btw, I hit another error at the same code location on WSL (Ubuntu 24.04) on Windows.

```
failures:

---- tiles::postgres::pool::tests::parse_version stdout ----

thread 'tiles::postgres::pool::tests::parse_version' (2379) panicked at martin-core/src/tiles/postgres/pool.rs:239:14:
container launched: Client(StartContainer(HyperLegacyError { err: hyper_util::client::legacy::Error(SendRequest, hyper::Error(IncompleteMessage)) }))
```

It seems this one is a bug on Docker's side and we can just wait for the patch release planned in next week. Just for sharing just in case this error hits this repository's CI or local Linux machines.

https://github.com/testcontainers/testcontainers-rs/issues/884

PR #2408, opened 2025-12-09
fix: allow `az://` URL schemes in discovery

### LINKED ISSUE #2407: `az://` protocol wrongly listed as valid in documentation
According to https://maplibre.org/martin/sources-files.html#mbtiles-and-pmtiles-file-sources , 

> For Azure, you need to provide the account name, container and path. For example:
>
> `martin  az://account/container.pmtiles`
>
> The available url schemes are:
>
   > `az://account/container/path.pmtiles`
   > `adl://account/container/path.pmtiles`
   > `azure://account/container/path.pmtiles`
   > `abfs://account/container/path.pmtiles`
   > `abfss://account/container/path.pmtiles`

However, attempting to use `az://` gives the error "Unrecognizable connection strings". Looking in source, I don't see `az` in the available options for protocols: 

https://github.com/maplibre/martin/blob/738c55e9cee1d7cafff030ab36ebd370f7affee7/martin/src/config/args/root.rs#L156-L157

https://github.com/maplibre/martin/blob/738c55e9cee1d7cafff030ab36ebd370f7affee7/martin/src/config/file/file_config.rs#L478-L480

I would think either `az` needs to be added to the code or removed from the docs.

CommanderStorm:
Realted code in the next layer

https://docs.rs/object_store/latest/src/object_store/parse.rs.html#113

- Resolves #2407


martin/src/config/args/root.rs
```diff
@@ -273,7 +273,7 @@ mod tests {
             }),
             ..Default::default()
         };
-        let meta = MetaArgs {
+        let meta = MeftaArgs {
```
CommanderStorm: typo

```suggestion
        let meta = MetaArgs {
```

nyurik (approved):

PR #2409, opened 2025-12-10
feat(perf): filter available tables query

todtb:


martin/src/config/file/tiles/postgres/resolver/query_tables.rs
```diff
@@ -22,13 +22,23 @@ pub type SqlTableInfoMapMapMap = BTreeMap<String, BTreeMap<String, BTreeMap<Stri
 const DEFAULT_EXTENT: u32 = 4096;
 const DEFAULT_BUFFER: u32 = 64;
 const DEFAULT_CLIP_GEOM: bool = true;
+const QUERY_AVAILABLE: &str = include_str!("scripts/query_available_tables.sql");
```
CommanderStorm: lets inline, personal preference ^^

```suggestion
```

martin/src/config/file/tiles/postgres/resolver/query_tables.rs
```diff
@@ -22,13 +22,23 @@ pub type SqlTableInfoMapMapMap = BTreeMap<String, BTreeMap<String, BTreeMap<Stri
...
 
 /// Queries the database for available tables with geometry columns.
-pub async fn query_available_tables(pool: &PostgresPool) -> PostgresResult<SqlTableInfoMapMapMap> {
+pub async fn query_available_tables(
+    pool: &PostgresPool,
+    filtered_tables: Option<Vec<&str>>,
+) -> PostgresResult<SqlTableInfoMapMapMap> {
+    let mut query = QUERY_AVAILABLE.to_string();
```
CommanderStorm:
```suggestion
    let mut query = include_str!("scripts/query_available_tables.sql").to_string();
```

martin/src/config/file/tiles/postgres/resolver/query_tables.rs
```diff
@@ -22,13 +22,23 @@ pub type SqlTableInfoMapMapMap = BTreeMap<String, BTreeMap<String, BTreeMap<Stri
...
+
+    if let Some(table_names) = filtered_tables {
+        query.push_str(&format!("WHERE name in ('{}')\n", table_names.join("','")));
```
CommanderStorm: to fix the clippy lint

```suggestion
        query.push_str("WHERE name in ('");
        query.push_str(&table_names.join("','"));
        query.push_str("')");
```

martin/src/config/file/tiles/postgres/builder.rs
```diff
@@ -133,7 +133,14 @@ impl PostgresAutoDiscoveryBuilder {
...
         // FIXME: this function has gotten too long due to the new formatting rules, need to be refactored
-        let mut db_tables_info = query_available_tables(&self.pool).await?;
+
+        let filter_config_tables: Option<Vec<&str>> = if self.auto_tables.is_none() {
+            Some(self.configured_tables())
+        } else {
+            None
+        };
+
+        let mut db_tables_info = query_available_tables(&self.pool, filter_config_tables).await?;
```
CommanderStorm:
```suggestion
        let configured_tables_filter = if self.auto_tables.is_none() {
            // a user might have configured thousands of tables.
            // filtering by thousands of tables is likely to be a perf degradation
            // => We only apply this optimisation arbitrairly up to 100 tables
            let configured_tables = self.configured_tables();
            if configured_tables.len() > 100 {
                Some(configured_tables)
            } else {
                None
            }
        } else {
            None
        };

        let mut db_tables_info = query_available_tables(&self.pool, configured_tables_filter).await?;
```

martin/src/config/file/tiles/postgres/resolver/query_tables.rs
```diff
@@ -22,13 +22,23 @@ pub type SqlTableInfoMapMapMap = BTreeMap<String, BTreeMap<String, BTreeMap<Stri
 const DEFAULT_EXTENT: u32 = 4096;
 const DEFAULT_BUFFER: u32 = 64;
 const DEFAULT_CLIP_GEOM: bool = true;
+const QUERY_AVAILABLE: &str = include_str!("scripts/query_available_tables.sql");
 
 /// Queries the database for available tables with geometry columns.
-pub async fn query_available_tables(pool: &PostgresPool) -> PostgresResult<SqlTableInfoMapMapMap> {
+pub async fn query_available_tables(
+    pool: &PostgresPool,
+    filtered_tables: Option<Vec<&str>>,
```
CommanderStorm:
```suggestion
    configured_tables_filter: Option<Vec<&str>>,
```

martin/src/config/file/tiles/postgres/resolver/query_tables.rs
```diff
@@ -22,13 +22,23 @@ pub type SqlTableInfoMapMapMap = BTreeMap<String, BTreeMap<String, BTreeMap<Stri
...
+) -> PostgresResult<SqlTableInfoMapMapMap> {
+    let mut query = QUERY_AVAILABLE.to_string();
+
+    if let Some(table_names) = filtered_tables {
```
CommanderStorm:
```suggestion
    if let Some(table_names) = configured_tables_filter {
```

CommanderStorm (approved):
Looks like a nice improvement.

@nyurik suggested on slack that we should limit the where clause to a "resonable" amount of tables to not slow down PG. I think this is fair.

martin/src/config/file/tiles/postgres/builder.rs
```diff
@@ -133,7 +133,22 @@ impl PostgresAutoDiscoveryBuilder {
...
     pub async fn instantiate_tables(&self) -> PostgresResult<(Vec<BoxedSource>, TableInfoSources)> {
         // FIXME: this function has gotten too long due to the new formatting rules, need to be refactored
-        let mut db_tables_info = query_available_tables(&self.pool).await?;
+
+        let configured_tables_filter = if self.auto_tables.is_none() {
+            // a user might have configured thousands of tables.
+            // filtering by thousands of tables is likely to be a perf degradation
+            // => We only apply this optimisation arbitrairly up to 100 tables
+            let configured_tables = self.configured_tables();
+            if configured_tables.len() > 100 {
```
CommanderStorm, line 142: this limit is a bit from thin air and we can obviously also decrease this. 
Not sure, would like some oppinion on this

If you @todtb have an oppinion on where such a limit should lie, we can adjust this ^^

martin/src/config/file/tiles/postgres/builder.rs
```diff
@@ -133,7 +133,23 @@ impl PostgresAutoDiscoveryBuilder {
...
```
CommanderStorm, line 142: To protect against this properly (and fix the logging bug), we likely need to have a second check before logging 🤔

nyurik:
lets do this properly - let's implement a proper rust-based filtering - assuming SQL returns all available tables/views. Once we have the rust-based filtering, we can reconsider additional perf optimization - i.e. passing table names to the query as parameters.

Note that the "proper" way may be to create a temporary per-session table and joining with it (which will remove the 100 limit) - assuming the user has permissions (which might not be the case).  Alternatively, it could be CTE-based, e.g.

```sql
WITH tmp AS (
    SELECT *
    FROM (VALUES
        ('schema1', 'table1'),
        ('schema1', 'table2')
    ) AS v(schema_name, table_name)
)
...
```

Also note that you are not escaping table names - https://xkcd.com/327/

todtb:
> lets do this properly - let's implement a proper rust-based filtering - assuming SQL returns all available tables/views. Once we have the rust-based filtering, we can reconsider additional perf optimization - i.e. passing table names to the query as parameters.
> 
> Note that the "proper" way may be to create a temporary per-session table and joining with it (which will remove the 100 limit) - assuming the user has permissions (which might not be the case). Alternatively, it could be CTE-based, e.g.
> 
> ```sql
> WITH tmp AS (
>     SELECT *
>     FROM (VALUES
>         ('schema1', 'table1'),
>         ('schema1', 'table2')
>     ) AS v(schema_name, table_name)
> )
> ...
> ```
> 
> Also note that you are not escaping table names - https://xkcd.com/327/

Again, sounds good to me. This draft PR was only intended as a way to show some pseudo implementation (code fragments speak a thousand words) of a SQL approach, and has since had some other changes pushed to it. I'm going to close this out and carry forward with passing a list of configured tables to the `query_available_tables` function. After the original SQL statement executes and returns all rows, we can filter that down with rust.

@CommanderStorm - sound good for now? We can revisit any SQL changes you'd like to see in a different PR?

CommanderStorm:
Yes. That would be awesome 👍

PR #2411, opened 2025-12-11
fix: filter available tables result when auto publish false

todtb:
Fix to `query_available_tables` allowing the function to take a list of configured tables if `auto_publish: false`.
Prevents configuration instantiation of tables that will not be used and addresses issue where WARN level logs appear for tables not used .

### DIFF
--- a/martin/src/config/file/tiles/postgres/builder.rs
+++ b/martin/src/config/file/tiles/postgres/builder.rs
@@ -133,7 +133,14 @@ impl PostgresAutoDiscoveryBuilder {
     #[expect(clippy::too_many_lines)]
     pub async fn instantiate_tables(&self) -> PostgresResult<(Vec<BoxedSource>, TableInfoSources)> {
         // FIXME: this function has gotten too long due to the new formatting rules, need to be refactored
-        let mut db_tables_info = query_available_tables(&self.pool).await?;
+
+        let filter_config_tables: Option<Vec<&str>> = if self.auto_tables.is_none() {
+            Some(self.configured_tables())
+        } else {
+            None
+        };
+
+        let mut db_tables_info = query_available_tables(&self.pool, filter_config_tables).await?;
 
         // Match configured sources with the discovered ones and add them to the pending list.
         let mut used = HashSet::<(&str, &str, &str)>::new();
@@ -334,6 +341,10 @@ impl PostgresAutoDiscoveryBuilder {
         let tilejson = pg_info.to_tilejson(id.clone());
         let source = PostgresSource::new(id, sql_info, tilejson, self.pool.clone());
         sources.push(Box::new(source));
+    }
+
+    fn configured_tables(&self) -> Vec<&str> {
+        self.tables.values().map(|t| t.table.as_str()).collect::<Vec<&str>>()
     }
 }
 
--- a/martin/src/config/file/tiles/postgres/builder.rs
+++ b/martin/src/config/file/tiles/postgres/builder.rs
@@ -134,13 +134,13 @@ impl PostgresAutoDiscoveryBuilder {
     pub async fn instantiate_tables(&self) -> PostgresResult<(Vec<BoxedSource>, TableInfoSources)> {
         // FIXME: this function has gotten too long due to the new formatting rules, need to be refactored
 
-        let filter_config_tables: Option<Vec<(&str, &str)>> = if self.auto_tables.is_none() {
+        let restrict_to_tables = if self.auto_tables.is_none() {
             Some(self.configured_tables())
         } else {
             None
         };
 
-        let mut db_tables_info = query_available_tables(&self.pool, filter_config_tables).await?;
+        let mut db_tables_info = query_available_tables(&self.pool, restrict_to_tables).await?;
 
         // Match configured sources with the discovered ones and add them to the pending list.
         let mut used = HashSet::<(&str, &str, &str)>::new();
@@ -343,11 +343,11 @@ impl PostgresAutoDiscoveryBuilder {
         sources.push(Box::new(source));
     }
 
-    fn configured_tables(&self) -> Vec<(&str, &str)> {
+    fn configured_tables(&self) -> HashSet<(&str, &str)> {
         self.tables
             .values()
             .map(|t| (t.schema.as_str(), t.table.as_str()))
-            .collect::<Vec<(&str, &str)>>()
+            .collect()
     }
 }
 
--- a/martin/src/config/file/tiles/postgres/builder.rs
+++ b/martin/src/config/file/tiles/postgres/builder.rs
@@ -343,10 +343,10 @@ impl PostgresAutoDiscoveryBuilder {
         sources.push(Box::new(source));
     }
 
-    fn configured_tables(&self) -> HashSet<(&str, &str)> {
+    fn configured_tables(&self) -> HashSet<(String, String)> {
         self.tables
             .values()
-            .map(|t| (t.schema.as_str(), t.table.as_str()))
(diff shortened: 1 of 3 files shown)


martin/src/config/file/tiles/postgres/builder.rs
```diff
@@ -133,7 +133,14 @@ impl PostgresAutoDiscoveryBuilder {
...
         // FIXME: this function has gotten too long due to the new formatting rules, need to be refactored
-        let mut db_tables_info = query_available_tables(&self.pool).await?;
+
+        let filter_config_tables: Option<Vec<(&str, &str)>> = if self.auto_tables.is_none() {
+            Some(self.configured_tables())
+        } else {
+            None
+        };
+
+        let mut db_tables_info = query_available_tables(&self.pool, filter_config_tables).await?;
```
CommanderStorm: nit:

```suggestion
        let restrict_to_tables = if self.auto_tables.is_none() {
            Some(self.configured_tables())
        } else {
            None
        };

        let mut db_tables_info = query_available_tables(&self.pool, restrict_to_tables).await?;
```

martin/src/config/file/tiles/postgres/builder.rs
```diff
@@ -335,6 +342,13 @@ impl PostgresAutoDiscoveryBuilder {
         let source = PostgresSource::new(id, sql_info, tilejson, self.pool.clone());
         sources.push(Box::new(source));
     }
+
+    fn configured_tables(&self) -> Vec<(&str, &str)> {
+        self.tables
+            .values()
+            .map(|t| (t.schema.as_str(), t.table.as_str()))
+            .collect::<Vec<(&str, &str)>>()
```
CommanderStorm: nit:

```suggestion
            .collect()
```

martin/src/config/file/tiles/postgres/resolver/query_tables.rs
```diff
@@ -36,6 +39,15 @@ pub async fn query_available_tables(pool: &PostgresPool) -> PostgresResult<SqlTa
...
         let schema: String = row.get("schema");
         let table: String = row.get("name");
+
+        // If a list of configured tables was provided to the function, because auto_publish set
+        // to false, skip processing the tables that are not in the list.
+        if let Some(ref table_names) = filtered_tables {
+            if !table_names.contains(&(schema.as_str(), table.as_str())) {
+                continue;
+            }
+        }
```
CommanderStorm: because of clippy

```suggestion
        if let Some(ref table_names) = restrict_to_tables
            && if !table_names.contains(&(schema.as_str(), table.as_str())) {
            continue;
        }
```

martin/src/config/file/tiles/postgres/resolver/query_tables.rs
```diff
@@ -24,7 +24,10 @@ const DEFAULT_BUFFER: u32 = 64;
 const DEFAULT_CLIP_GEOM: bool = true;
 
 /// Queries the database for available tables with geometry columns.
```
CommanderStorm, line 26:
```suggestion
/// Queries the database for available tables with geometry columns.
///
/// The reported tables are filtered by the `restrict_to_tables` parameter.
```

martin/src/config/file/tiles/postgres/resolver/query_tables.rs
```diff
@@ -24,7 +24,10 @@ const DEFAULT_BUFFER: u32 = 64;
...
-pub async fn query_available_tables(pool: &PostgresPool) -> PostgresResult<SqlTableInfoMapMapMap> {
+pub async fn query_available_tables(
+    pool: &PostgresPool,
+    filtered_tables: Option<Vec<(&str, &str)>>,
```
CommanderStorm: As noted in https://github.com/maplibre/martin/pull/2411#discussion_r2612494809 just to be sure.

This needs an import as well

```suggestion
    restrict_to_tables: Option<HashSet<(&str, &str)>>,
```

martin/src/config/file/tiles/postgres/builder.rs
```diff
@@ -335,6 +342,13 @@ impl PostgresAutoDiscoveryBuilder {
         let source = PostgresSource::new(id, sql_info, tilejson, self.pool.clone());
         sources.push(Box::new(source));
     }
+
+    fn configured_tables(&self) -> Vec<(&str, &str)> {
```
CommanderStorm: Does not make a performance difference, unless number of configured table is pretty insane.

We want to only check for membership => lets choose a hashset

```suggestion
    fn configured_tables(&self) -> HashSet<(&str, &str)> {
```

martin/src/config/file/tiles/postgres/resolver/query_tables.rs
```diff
@@ -36,6 +39,15 @@ pub async fn query_available_tables(pool: &PostgresPool) -> PostgresResult<SqlTa
     for row in &rows {
         let schema: String = row.get("schema");
         let table: String = row.get("name");
+
+        // If a list of configured tables was provided to the function, because auto_publish set
+        // to false, skip processing the tables that are not in the list.
```
CommanderStorm: nit: I think this can be more explicit

```suggestion
        // If auto_publish is not active, a user can provide a set of tables.
        // The query above might find more tables than the statically configured ones.
        // Therefore we filter to limit us to the subset of available and configured tables
```

CommanderStorm (commented):
LGTM, nice!

CommanderStorm:
It seems like `pg_tables_feature_id` is failing, but I have not had the time to investigate deper why `no_id` in it gets filtered away

todtb:
> It seems like `pg_tables_feature_id` is failing, but I have not had the time to investigate deper why `no_id` in it gets filtered away

My guess is that as `auto_publish` isn't specified and defaults to false, it doesn't get added to the TableInfo. Haven't checked yet either but will take a look when I continue working on this.

Thanks for the feedback thus far.

todtb:
After looking into it, it appears to just be that the hashset contains is not factoring in that the config definition and the actual schema/table name in DB are case insensitive. Will look into the fix.

martin/src/config/file/tiles/postgres/resolver/query_tables.rs
todtb: This doesn't read right to me. We wouldn't restrict to the tables found by the query, but rather the tables in the config.

CommanderStorm: Fair. I have updated the phrasing, calling the subset-behaviour out is nice as the next generation of devs might not notice this otherwise

CommanderStorm (approved):
Lets merge once the last comment (about the comment) is in

martin/src/config/file/tiles/postgres/resolver/query_tables.rs
todtb: > a user can provide a set of tables.

In a way I suppose, through defining them in the config.yml and setting auto_publish to false.

How about:
```
// Within the config, if auto_publish is false or omitted, the list of schema and table names set explicitly 
// under the tables key is provided to the function. As the query above may return more tables than 
// explicitly defined, these are filtered out below.
```

But since this comment should be first and foremost to help people who didn't author this bit of code, I'll leave it to your discretion on what makes the most sense.

CommanderStorm: Sure, your version is better

CommanderStorm: Sure, your version is better

Issue #2395, opened 2025-12-03
Improvement: make startup resilient to bad file paths


gabeschine:
**Current:**

- If a config YML references a bad file path, ie for an `.mbtiles` or `.pmtiles` file, startup fails completely.
- In contrast, if the config references a bad Postgres _table_, startup warns but continues to initialize & serve the remaining sources.

**Desired state:**

- Consistently ignore bad sources, allow startup to finish, and serve good known sources.

**Reasoning:**

- Better production resilience: a single bad path should not "take down" the serving infrastructure.

gabeschine:
See: https://github.com/maplibre/martin/pull/1689 for relevant context and code patterns.

CommanderStorm:
Once configurable, the default should likely be `warn`.

Aborting might still be valuable, especially in CI 😉 

CC @nyurik because I think this was your decision.

gabeschine:
> Once configurable, the default should likely be `warn`.
> 
> Aborting might still be valuable, especially in CI 😉
> 
> CC [@nyurik](https://github.com/nyurik) because I think this was your decision.

Copied from Slack discussion:

I opted to go simple, and not provide any configuration, for this round. I noticed the PR you sent me was never merged - while I appreciate the ability to configure what to do on a config with validation problems, my view is that the default for running martin should be a best-effort initialization (warn, not abort).

nyurik:
`martin foo.mbtiles` -- should it fail to start if foo.mbtiles does not exist, i.e. nothing to serve?  What if the connection is a DB server that may later come back up?  What about the dev experience vs production?  I think dev experience should be simpler - if you gave wrong params, fail right away.  In a prod, devops might be ok to configure a bit more complex configuration, e.g. start and continuously refresh. Lastly, if a file is missing from a local disk, the chances of it becoming available are much smaller than for a DB server or an S3 bucket to come back up / get the needed file.

In a devops env, e.g. kubernetes, the best "signal" of something is bad is for the service to crash - k8s would automatically restart the failed pod, often fixing the problem. If a service starts but fails to deliver needed results may require complex configuration using [liveness / rediness / startup probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/).

In short, I don't have a magical answer. We should make on-boarding experience simple, i.e. crash rather than show some warning. And yet we should account for the long tail of the edge cases that are possible.  This may also intertwine with the automtaic refresh - #288

gabeschine:
[...]

PR #2412, opened 2025-12-12
feat(config): optionally fail config loading/resolution for missing sources

gabeschine:
Fixes issue #2395

Adds a config option `on_invalid`, which selects from one of two policies to handle misconfigured sources during startup:

- `warn`: logs warnings, but continues startup
- `abort`: logs warnings as error-level logs, then aborts startup

Example:

```
...
on_invalid: abort
```

The change also:

- exposes what were warn- or info-level logs in the `file_config` mod and the `postgres` mod up the stack for inclusion in the warnings list
- adds an integration test to the `postgres` config mod to exercise exposing these warnings
- small change to `just validate-tools` to check for `gsed` on Darwin, and suggest its installation

Task list:

- [x] Return sources + warnings from file resolver (pmtiles, mbtiles, cog)
- [x] Return sources + warnings from pg resolver
- [x] Write unit/integration tests
- [x] Add config documentation + examples


martin/src/config/file/main.rs
```diff
@@ -66,11 +66,15 @@ pub struct Config {
...
     /// Can be overridden by [`tile_cache_size_mb`](Self::tile_cache_size_mb) or similar configuration options.
     pub cache_size_mb: Option<u64>,
+
     /// Maximum size of the tile cache in megabytes (0 to disable)
     ///
     /// Overrides [`cache_size_mb`](Self::cache_size_mb)
     pub tile_cache_size_mb: Option<u64>,
 
+    #[serde(default)]
+    pub fail_on_source_resolution_warnings: bool,
```
CommanderStorm: I kind of would like this to be an enum as following, as this way this is not a super specific option, but rather a more genererl setting.

This way we are a bit more explicit what happens.

```suggestion
    #[serde(default)]
    pub on_invalid: OnInvalid,
```

```rust
#[derive(PartialEq, Eq, Debug, Clone, Copy, Default, Serialize, Deserialize, ValueEnum)]
#[serde(rename_all = "lowercase")]
pub enum OnInvalid {
    /// Print warning message, and abort if the error is critical
    #[default]
    Warn,
    /// Do not start Martin on any warnings
    Abort,
    /// Ignore even critical errors
    Ignore,
}
```

martin/src/config/file/main.rs
```diff
@@ -218,15 +222,32 @@ impl Config {
...
+                }
+                return Err(MartinError::TileResolutionWarningsIssued);
+            }
+            log::warn!(
+                "Warnings issued during source resolution. These sources will not be available:"
+            );
+            for warning in &warnings {
+                log::warn!("- {}", warning);
+            }
+        }
```
CommanderStorm: nit: I think writing it this way might be more elegant.
Not sure, current version is likely fine too.

```suggestion
        #[cfg(feature = "_tiles")]
        let tiles = {
            let (tiles, warnings) = self
                .resolve_tile_sources(
                    &resolver,
                    #[cfg(feature = "pmtiles")]
                    pmtiles_cache,
                )
                .await?;
            if !warnings.is_empty() {
                OnInvalid::handle_tiles(warnings)?;
            }
            tiles
        }
```

martin/src/error.rs
```diff
@@ -74,4 +74,13 @@ pub enum MartinError {
...
+
+    #[error("one or more warnings issued during tile source resolution")]
+    TileResolutionWarningsIssued,
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum TileSourceWarning {
+    #[error("{id}: {error}")]
+    SourceError { id: String, error: MartinError },
 }
```
CommanderStorm: nit: I'd move it to where it is first used

martin/src/error.rs
```diff
@@ -74,4 +74,13 @@ pub enum MartinError {
     #[cfg(feature = "metrics")]
     #[error("could not initialize metrics: {0}")]
     MetricsIntialisationError(#[source] Box<dyn std::error::Error>),
+
+    #[error("one or more warnings issued during tile source resolution")]
+    TileResolutionWarningsIssued,
```
CommanderStorm:
```suggestion
    #[error("warnings issued during tile source resolution")]
    TileResolutionWarningsIssued,
```

CommanderStorm (commented):
General aspect looks good.

I added a comment on how to configure this.

I know this is bikeshedding, but since the way things are in the config/CLI are unlikely to be changable, this is hopefully OK.
I am a bit torn on the exact wording of how to configure this on the CLI
- `--on-invalid warn`/... or
- `--verify`/`--ignore-invalid`

martin/src/config/file/main.rs
gabeschine: I like it. Keeps more of the model & policy in the same place.

martin/src/error.rs
gabeschine: I moved it to `config/file/main.rs`, if that's what you meant.

gabeschine:
@CommanderStorm @nyurik status update and some open questions.

Done:
- Enum config in code + handling logic
- Returning warnings in `file_config.rs`
- Unit tests for `file_config.rs`

To do:
- Returning warnings for postgres
- Tests for postgres
- Documentation

Open questions, for postgres sources: I see the code in `builder.rs` does a lot of property discovery using `find_info()` in `utils.rs`.

- Is it correct that I should be converting the `warn!` lines in `find_info_kv()` into a return value that I can then use in `builder.rs`'s functions, to then propagate the reason a source was skipped up the stack?
- Are there any cases in `TableInfo::append_cfg_info()` that I should also expose in a similar fashion?
- For tests, I'm assuming test cases in `test.sh` are the best / only place I can test against a Postgres instance?
- Anything else I'm missing for Postgres?

Thanks!

CommanderStorm:
> Open questions, for postgres sources: I see the code in `builder.rs` does a lot of property discovery using `find_info()` in `utils.rs`.
> 
> * Is it correct that I should be converting the `warn!` lines in `find_info_kv()` into a return value that I can then use in `builder.rs`'s functions, to then propagate the reason a source was skipped up the stack?

Yes, that would be awsome.
Honestly, I am not 100% a fan of how `find_info_kv` + surounding code works, it is kind of clunky and non-intuitive to read.

We can do it here or in a followup.
It would defintively be nice to have. ^^

> * Are there any cases in `TableInfo::append_cfg_info()` that I should also expose in a similar fashion?

You are likely refering to `Table {self.format_id()} has geometry type={src}, but source {new_id} has {cfg}`.
I think, plumbing this through to the warning system in this PR would be resonable.
As above, we can do it in this PR or one PR at a time.

> * For tests, I'm assuming test cases in `test.sh` are the best / only place I can test against a Postgres instance?

Our e2e tests in `test.sh` is kind of hard to debug and reason about.[^1]
Yes, it is testing via the public API, thus likely how we should test, but writing those tests is not very ergonomic.

An alternative is to use a postgres testcontainer and write new tests.
Such integration tests are likely much simpler to write.
We currently don't have integration tests like this, it is likely relatively simple to add them though.

Here is an example of a test using a testcontainer

https://github.com/maplibre/martin/blob/2a2caf41bee4279cf7aaf98a9f0ca9f3eb294149/martin-core/src/tiles/postgres/pool.rs#L233-L276

> * Anything else I'm missing for Postgres?

Not to my knowlege


[^1]: We have talked a bit on this, but have not come to a good redesign what should replace this. It is on the todo ist so to speak.

martin/src/config/file/main.rs
```diff
@@ -365,6 +393,38 @@ impl Config {
...
+            OnInvalid::Warn => {
+                warn!("Tile source resolution warnings:");
+                for warning in warnings {
+                    warn!("  - {warning}");
+                }
+                Ok(())
+            }
+            OnInvalid::Ignore => Ok(()),
+        }
+    }
```
CommanderStorm, line 440: thee requests on the current code:

- (optional) if there is only one warning, lets add a special case with better log formatting
- `OnInvalid::Abort` should log at error level (currently this swallows the log)
- `OnInvalid::Ignore` should likely log at debug or trace level (-> better debuggability)

gabeschine:
> Yes, that would be awsome. Honestly, I am not 100% a fan of how `find_info_kv` + surounding code works, it is kind of clunky and non-intuitive to read.
> 
> We can do it here or in a followup. It would defintively be nice to have. ^^

I'll do it here. I think having `info_info*()` return a `Result` instead of `Optional` will work for this case, and the caller can decide to ignore the error (to preserve today's logic) or do something else (make a tile source warning).

> > * Are there any cases in `TableInfo::append_cfg_info()` that I should also expose in a similar fashion?
> 
> You are likely refering to `Table {self.format_id()} has geometry type={src}, but source {new_id} has {cfg}`. I think, plumbing this through to the warning system in this PR would be resonable. As above, we can do it in this PR or one PR at a time.

Yep, that's what I'm referring to. If it's straight-forward I'll do it here.

> > * For tests, I'm assuming test cases in `test.sh` are the best / only place I can test against a Postgres instance?
> 
> Our e2e tests in `test.sh` is kind of hard to debug and reason about.[1](#user-content-fn-1-c434db11ce7bf4e0e807d152c7ca9a70) Yes, it is testing via the public API, thus likely how we should test, but writing those tests is not very ergonomic.
> 
> An alternative is to use a postgres testcontainer and write new tests. Such integration tests are likely much simpler to write. We currently don't have integration tests like this, it is likely relatively simple to add them though.

I definitely like the inline Postgres container option. I'll see how hard it is to write a test for this new behavior specifically. I find that once someone breaks through the ice by writing one test, everyone else piles on, so hopefully I can do that here.

gabeschine:
> I'll do it here. I think having `info_info*()` return a `Result` instead of `Optional` will work for this case, and the caller can decide to ignore the error (to preserve today's logic) or do something else (make a tile source warning).

Please see https://github.com/maplibre/martin/pull/2426 - this is a no-op refactor that will reduce the diff here in this PR.

docs/src/config.yaml
```diff
@@ -52,6 +52,17 @@ preferred_encoding: gzip
 # This may be undesirable in a production environment
 web_ui: disable
 
+# The policy for handling invalid sources during startup. [default: warn]
```
nyurik: are we certain we want to `warn` by default? I like the current default clarity of "fail on mis-configuration" approach - as it gives instant clear feedback to novice users. More advanced users can set this param to   warn/ignore as needed.  Also, I keep thinking if `warn/abort/ignore` is the right approach.  How about `auto/abort/ignore` instead?  auto will use arbitrary internal "magic" approach, where martin itself decides what to do based on some heuristics, and leaves us flexibility to change it in the future.  The user-supplied `abort` would fail on any incorrect configuration - with the option to be able to configure this per source/per source type in the future. Lastly, the `ignore` will print a warning, but continue.

gabeschine: Right, in fact I originally intended to make `abort` the default. Will update.

I'm not sure what `auto` would do, or when I'd want to use it. Can you give some more specific examples? As a dev, I'm hesitant to use an option I don't fully understand.

nyurik: p.s. i'm not certain if it should be named `ignore` or `warn` mode, but regardless of that, a warning should always be logged - configuration should NOT control the logging level of a problematic source.

nyurik: there are different types of warnings, some of which could be OK in most cases, but not in production. For example, we have extensive `.mbtiles` [validation](https://github.com/maplibre/martin/blob/main/mbtiles/src/validation.rs), and one of those validations fail - e.g. https://github.com/maplibre/martin/blob/151a6c69df2e3faadc61e2a96fca191778c03d5e/mbtiles/src/validation.rs#L210-L212

In theory, this is not a blocker to start martin, but it could be an indicator that the invalid mbtiles file is being used - something you wouldn't want in production.  So you would ignore (log) this warning in `auto`, but you may want to abort in production.  Hence the potential for more nuanced configuration at a later date - where we may want to ignore certain types warnings for specific sources / source types. The later is TBD, but for now we may want to have at least several root levels.

gabeschine: > p.s. i'm not certain if it should be named ignore or warn mode, but regardless of that, a warning should always be logged - configuration should NOT control the logging level of a problematic source.

Makes sense. And that would make `warn` and `ignore` identical policy-wise, so I'll eliminate `ignore`. To me, `ignore` is ambiguous if there will be any evidence / output logged, whereas `warn` is clear. Happy to be told to do otherwise.

CommanderStorm: True.
Also, f someone (who?) maybe (why?) wants to disable warnings being logged, they can always set the log level to `off`.

Only point that I can see is maaaybe performance, but that at this point is academic and that person could just as well fuse these log levels off when compiling the binary.
Logging is not very much of our runtime anyhow.

docs/src/config.yaml
```diff
@@ -2,7 +2,7 @@
 keep_alive: 75
 
 # The socket address to bind [default: 0.0.0.0:3000]
-listen_addresses: '0.0.0.0:3000'
+listen_addresses: "0.0.0.0:3000"
```
gabeschine, line 5: Sorry, my formatter changed these to double quotes. If this is a bother, LMK and I'll revert.

docs/src/config.yaml
```diff
@@ -52,6 +52,16 @@ preferred_encoding: gzip
 # This may be undesirable in a production environment
 web_ui: disable
 
+# The policy for handling invalid sources during startup. [default: abort]
+#
+# Invalid sources are those that are missing (file not found, table doesn't exist, ...),
+# reference columns that don't exist, and so on.
```
CommanderStorm, line 58:
```suggestion
# Invalid sources are those that are missing (file not found, table doesn't exist, ...),
# reference columns that don't exist, and so on.
# Currently limited to tile sources; broader rollout is planned.
```

martin/src/config/file/main.rs
```diff
@@ -365,6 +393,45 @@ impl Config {
     }
 }
 
+/// Describes the action to take during startup when configuration is found to be invalid
+/// but Martin could still startup in a degraded state (ie, some sources not served).
+#[derive(PartialEq, Eq, Debug, Clone, Copy, Default, Serialize, Deserialize, ValueEnum)]
+#[serde(rename_all = "lowercase")]
+pub enum OnInvalid {
+    /// Log warning messages, abort if the error is critical
+    Warn,
```
CommanderStorm, line 410: Two choices (at your liberty).
We shoud have an alias for user-firendlyness though.

```suggestion
    #[serde(alias = "warnings", alias = "warning"))]
    Warn,
```

```suggestion
    #[serde(alias = "warnings", alias = "warning", alias = "continue", alias = "ignore"))]
    Warn,
```

martin/src/config/file/main.rs
```diff
@@ -365,6 +393,45 @@ impl Config {
...
+            for warning in warnings {
+                log!(level, "  - {warning}");
+            }
+        }
+
+        match policy {
+            OnInvalid::Abort => Err(MartinError::TileResolutionWarningsIssued),
+            OnInvalid::Warn => Ok(()),
+        }
+    }
```
CommanderStorm: nit: lets change this to the imo more idiomatic `policy.handle_tile_warnings(warnings)` and switch the way warnings are matched.

```suggestion
    pub fn handle_tile_warnings(self, warnings: &[TileSourceWarning]) -> MartinResult<()> {
        if warnings.is_empty() {
            return Ok(());
        }
        let level = match self {
            OnInvalid::Warn => Level::Warn,
            OnInvalid::Abort => Level::Error,
        };
        match warnings {
            [warning] => log!(level, "Tile source resolution warning: {warning}"),
            warnings => {
                log!(level, "Tile source resolution warnings:");
                for warning in warnings {
                    log!(level, "  - {warning}");
                }
            }
        }

        match self {
            OnInvalid::Abort => Err(MartinError::TileResolutionWarningsIssued),
            OnInvalid::Warn => Ok(()),
        }
    }
```

martin/src/config/file/main.rs
```diff
@@ -218,15 +233,21 @@ impl Config {
...
+        let (tiles, warnings) = self
+            .resolve_tile_sources(
+                &resolver,
+                #[cfg(feature = "pmtiles")]
+                pmtiles_cache,
+            )
+            .await?;
+
+        #[cfg(feature = "_tiles")]
+        OnInvalid::handle_warnings(self.on_invalid, &warnings)?;
```
CommanderStorm: to match below + above

```suggestion
        self.on_invalid.unwrap_or_default().handle_tile_warnings(&warnings)?;
```

martin/src/config/file/main.rs
```diff
@@ -66,11 +77,15 @@ pub struct Config {
...
     /// Can be overridden by [`tile_cache_size_mb`](Self::tile_cache_size_mb) or similar configuration options.
     pub cache_size_mb: Option<u64>,
+
     /// Maximum size of the tile cache in megabytes (0 to disable)
     ///
     /// Overrides [`cache_size_mb`](Self::cache_size_mb)
     pub tile_cache_size_mb: Option<u64>,
 
+    #[serde(default)]
+    pub on_invalid: OnInvalid,
```
CommanderStorm: Lets do it this way.

The reason is that otherwise the files like `tests/exepected/auto_mini/save_config.yaml` contain this when serialized again.

```suggestion
    pub on_invalid: Option<OnInvalid>,
```

CommanderStorm (approved):
LGTM from my side.

I have a bit of code comments, none of which should be hard to address.

Auspicus:
I'm glad this finally made it in, sorry I wasn't able to take it over the line :)

PR #2424, opened 2025-12-15
docs: document the default value for `auto_publish`

todtb:


docs/src/config.yaml
```diff
@@ -114,7 +114,9 @@ postgres:
   # - `skip` does not compute table geometry bounds on startup.
   auto_bounds: quick
 
-  # Enable automatic discovery of tables and functions.
+  # Enable automatic discovery of tables and functions. [default: null]
+  # null will attempt automatic discovery if no tables or functions are explicitly set
+  # in postgres.tables or postgres.functions.
   # You may set this to `false` to disable.
```
CommanderStorm: how about we mention true / false / null instead?
I think this would be less confusing to read.

```suggestion
  # Enable automatic discovery of tables and functions. [default: null]
  # 
  # Options are:
  # - `true` to run automatic discovery
  # - `false` to disable automatic discovery
  # - `null` to run automatic discovery if postgres.tables is `null` and postgres.functions is `null`
```

CommanderStorm (approved):
good catch.
Lets mention `true` as well for completeness

CommanderStorm:
@todtb was closing intentional?

PR #2426, opened 2025-12-16
refactor(config): no-op refactor to pull out "build_one_(table|function)_info"

gabeschine:
This will make reviewing the diff in https://github.com/maplibre/martin/pull/2412 easier.

Also:

- Converts the `find_info*()` methods to use `Result` instead of `Option`.


martin/src/config/file/tiles/postgres/builder.rs
```diff
@@ -154,28 +153,14 @@ impl PostgresAutoDiscoveryBuilder {
...
-            let Some(db_geo_columns) = find_info(db_tables, &cfg_inf.table, "table", id) else {
-                continue;
-            };
-            let Some(db_inf) = find_info(
-                db_geo_columns,
-                &cfg_inf.geometry_column,
-                "geometry column",
-                id,
-            ) else {
+            let Ok(merged_inf) = self.instantiate_one_table(&db_tables_info, id, cfg_inf) else {
```
CommanderStorm: I like this, much simpler to read!

martin/src/config/file/tiles/postgres/builder.rs
```diff
@@ -262,31 +247,20 @@ impl PostgresAutoDiscoveryBuilder {
...
-            if db_funcs.is_empty() {
-                warn!(
-                    "No functions found in schema {}. Only functions like (z,x,y) -> bytea and similar are considered. See README.md",
-                    cfg_inf.schema
-                );
-                continue;
-            }
-            let func_name = &cfg_inf.function;
-            let Some((pg_sql, db_inf)) = find_info(db_funcs, func_name, "function", id) else {
+            let Ok((merged_inf, pg_sql_info)) =
```
CommanderStorm: nit/unrelated to your change:

One thing that makes this part of the codebase harder to parse is its insistance to hamming code the variable names.

This was previously a quite short name and not very self-explanatory.
Could we lengthen it to something more expressive like `merged_db_and_cfg_info`

martin/src/config/file/tiles/postgres/utils.rs
```diff
@@ -11,49 +11,42 @@ pub fn normalize_key<T>(
...
+            "Unable to configure source {id} because {info} '{key}' was not found.  Possible values are: {}",
+            map.keys().map(String::as_str).join(", ")
+        )),
+        Ok(Some(result)) => map.get(result).map(|v| (result.as_str(), v)).ok_or(format!(
+            "For source {id}, {info} '{key}' was not found, but found '{result}' instead."
+        )),
+        Err(multiple) => Err(format!(
+            "Unable to configure source {id} because {info} '{key}' has no exact match and more than one potential matches: {}",
+            multiple.join(", ")
+        )),
```
CommanderStorm, line 57: I am not 100% sure this here is entirely what we want.

As this currently is implemented, this means that we don't logg errors/warnings, as they get sallowed in the match arm after `instantiate_one_table`.

martin/src/config/file/tiles/postgres/builder.rs
CommanderStorm: much simpler to read, if we can get this to log the same way 👍🏻

As it is implmeneted (ignoring the error/warning/info log) this is a harder sell.

martin/src/config/file/tiles/postgres/builder.rs
```diff
@@ -326,6 +301,52 @@ impl PostgresAutoDiscoveryBuilder {
         Ok((res, info_map))
     }
 
+    fn instantiate_one_table(
+        &self,
+        db_tables_info: &std::collections::BTreeMap<
+            String,
+            std::collections::BTreeMap<String, std::collections::BTreeMap<String, TableInfo>>,
+        >,
```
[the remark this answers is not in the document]
CommanderStorm: Lets import `BTreeMap` at least ^^

martin/src/config/file/tiles/postgres/builder.rs
```diff
@@ -326,6 +300,50 @@ impl PostgresAutoDiscoveryBuilder {
         Ok((res, info_map))
     }
 
+    fn instantiate_one_table(
```
CommanderStorm: nit: 
if possible, I would like to establish that new functions have at least a very short doc comment as this makes future work much simpler.

I for example have no clue from reading the function what is happening, but I also have not looked at this code deeper, so not supprising.

CommanderStorm (changes requested):
Looks like a good start.
If we can get it to not swallow the logs, I am all for it.

martin/src/config/file/tiles/postgres/utils.rs
CommanderStorm: That CI is green on this one is not unsuprising, but also not great btw ^^

martin/src/config/file/tiles/postgres/builder.rs
gabeschine: Ah good catch! Will update it to log the warning for now, and in the next PR the same warnings will propagate up further.

martin/src/config/file/tiles/postgres/builder.rs
gabeschine: 💯  agreed. I was trying to stick to the existing style, but I'd be _happy to_ improve the naming.

gabeschine: Actually, would you like me to do a naming pass in a follow-up PR? I don't want to make this diff too big. Up to you since you're the reviewer.

martin/src/config/file/tiles/postgres/utils.rs
gabeschine: I've updated the code so it logs these at warn-level (in the calling code). That loses the definition in the log levels here. I'm not sure if losing that definition is impactful or not.

I've noticed a slight diff on line 47 (in the base branch) and will address that as well, so the `info!()` is preserved.

gabeschine: I'm not sure if I've resolved your concerns here, or not. Please resolve the thread if so.

martin/src/config/file/tiles/postgres/builder.rs
gabeschine: Done. After writing doc-comments, renamed to `build_one_<foo>_info` as that seems to better describe what it's doing.

martin/src/config/file/tiles/postgres/utils.rs
CommanderStorm: I am going to bed now, will do so tomorrow

martin/src/config/file/tiles/postgres/builder.rs
CommanderStorm: I am fine with either, what is simpler for you.

GitHub has a (semi-working) feature where I can go commit by commit when reviewing ^^

gabeschine: I've renamed the variables in the two new functions, left the rest for later.

CommanderStorm (approved):
LGTM now, much better.

Could you resolve the clippy lints? we can merge after that

martin/src/config/file/tiles/postgres/utils.rs
```diff
@@ -11,49 +11,50 @@ pub fn normalize_key<T>(
...
             info!("For source {id}, {info} '{key}' was not found, but found '{result}' instead.");
-            Some((result.as_str(), map.get(result)?))
-        }
-        Err(multiple) => {
-            error!(
-                "Unable to configure source {id} because {info} '{key}' has no exact match and more than one potential matches: {}",
-                multiple.join(", ")
-            );
-            None
+            let value = map.get(result).unwrap(); // guaranteed to be in the map
```
CommanderStorm:
```suggestion
            let value = map.get(result).expect("guaranteed to be in the map");
```

gabeschine:
> LGTM now, much better.
> 
> Could you resolve the clippy lints? we can merge after that

Thanks. I'll have to work `just clippy` into my pre-review tasks. Done.

PR #2202, opened 2025-09-17
chore: add an action that on labeled bless blesses the PR

CommanderStorm:

PR #2427, opened 2025-12-17
feat(martin-ui): Add click to copy

### LINKED ISSUE #1982: styling extends beyond its box for long names
To reproduce:   `just start run`  and open the web UI:

Image

todtb:
Resolves #1982

Adds click to copy functionality on component titles. 

- This adds it pretty much everywhere it could be truncated, feel free to remove if not desired
- Let's discuss if they are all really necessary 

Other notes:
- Does not add to fonts - this already has a copy link button which is likely what somebody would want anyway
- There's an existing issue with the tabs in the style integration guide modal that can be handled elsewhere, but if that tabs div is not rendered truncation works fine

image

image

image

image

image


martin/martin-ui/src/components/dialogs/sprite-download.tsx
```diff
@@ -94,9 +95,12 @@ export function SpriteDownloadDialog({ name, sprite, onCloseAction }: SpriteDown
...
     <Dialog onOpenChange={(v: boolean) => !v && onCloseAction()} open={true}>
       <DialogContent className="max-w-2xl w-full max-h-[90vh] overflow-auto">
-        <DialogHeader>
+        <DialogHeader className="truncate">
           <DialogTitle className="text-2xl">
-            Download <code className="font-mono">{name}</code>
+            Download{' '}
+            <code className="font-mono">
+              <TooltipCopyText side="bottom" text={name} />
+            </code>
```
CommanderStorm: this is already one level down ^^

```suggestion
            <TooltipCopyText side="bottom" text={name} />
```

martin/martin-ui/src/components/ui/tooltip-copy-text.tsx
```diff
@@ -0,0 +1,64 @@
...
+        <div
+          aria-disabled="true"
+          className="text-lg font-mono cursor-pointer truncate"
+          onClick={handleCopy}
+          onKeyDown={handleKeyDown}
+          role="button"
+          tabindex="0"
+        >
+          {text}
+        </div>
```
CommanderStorm: Works, but lets simplify this a bit (plus improve a11y).

Interactive buttons should really be buttons, so the version below works a bit better with assistive tooling.
This also allows us to remove some code ^^

```suggestion
  const handleCopy = async () => {
    try {
      await navigator.clipboard.writeText(text);
      toast({
        description: `"${text}"`,
        title: 'Copied!',
      });
    } catch (err) {
      console.error(`Failed to copy "${text}"`, err);
      toast({
        description: 'Failed to copy to clipboard',
        title: 'Error',
        variant: 'destructive',
      });
    }
  };

  return (
    <Tooltip>
      <TooltipTrigger
        aria-label={`Copy ${text} to clipboard`}
        className="text-lg font-mono cursor-pointer truncate"
        onClick={handleCopy}
        type="button"
        >
        <code>{text}</code>
```

martin/martin-ui/src/components/ui/tooltip-copy-text.tsx
```diff
@@ -0,0 +1,64 @@
+import * as React from 'react';
```
CommanderStorm:
```suggestion
```

CommanderStorm (approved):
Nice work!

CommanderStorm:
Don't worry about the linting, that will be fixed once we merge the following:
- #2429

martin/martin-ui/src/components/ui/tooltip-copy-text.tsx
todtb: I agree w/ the a11y bit and also semantic HTML in general. I use keyboard based navigation myself and was able to focus and interact with this just fine. That said my knowledge of other tooling such as screen readers is lacking.

But there are a couple things:
1) The suggestion above won't work as intended (but we may have work arounds)
image

`<button>`s (AFAIK) would need fixed width to use text overflow CSS. I can poke a little at this and see if tailwind has some utilities for it, but otherwise we'll need to figure out some custom CSS to accomplish it.


2) Biomejs seems to be throwing false positives but that may be addressed as you mentioned below

martin/martin-ui/src/components/dialogs/sprite-download.tsx
todtb: This was a format auto-correct, but are you saying just remove the `<code>` tag here?

martin/martin-ui/src/components/ui/tooltip-copy-text.tsx
todtb: It looks like `w-full` would work

however

We would need to allow `asChild` in some scenarios (see https://github.com/radix-ui/primitives/issues/2248) but this results in `type=button` vs `<button>`. Ongoing discussion in that thread re: a11y. 

But let me know if you have a nice solution or workaround.

CommanderStorm: Good intuition, thanks for double checking me.

I'd prefer the button based version with the overflow fixed on the basis that it is less code.
I'd also merge your version if you think yours is better ^^

todtb: Third option: we can evaluate how many areas we need this click-to-copy functionality. I'm not 100% sure of the typical work flow for users of a martin server who aren't already aware of the underlying configuration.

If we can avoid using this component everywhere the title is, problem solved, we'll just use your suggestion.

@nyurik - you mentioned click-to-copy was critical on the catalog cards:

image

Is it also critical in the modals? 
image

If not critical in the modals, but nice to have, could it be added as a button below in future work?

nyurik:
@CommanderStorm i thought linting should work with this?

CommanderStorm:
@nyurik What do you mean?
The check that is failing is a snapshot test on the frontend.
-> Running `just bless` resolves this.

Shameless plug:
- #2202

PR #2440, opened 2025-12-23
chore(ci): Tighten CI security a bit more by enforcing zizmore in CI

CommanderStorm:
The changes to `ci.yml` are to prevent "suspected template injections" (not exploitable).

I also added a zizmore action because now we can do this.

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -4,7 +4,7 @@ on:
   schedule:
     - cron: "0 0 * * *"
   pull_request:
-    branches: [ main ]
+    branches: ["main"]
 
 permissions:
   contents: read
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -20,7 +20,7 @@ jobs:
       steps:
         - name: Checkout repository
           uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
-          with: {persist-credentials: false}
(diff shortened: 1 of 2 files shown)


.github/workflows/security.yml
```diff
@@ -3,13 +3,27 @@ name: Security audit
 on:
   schedule:
     - cron: "0 0 * * *"
+  push:
+    branches: ["main"]
   pull_request:
-    branches: [ main ]
+    branches: ["**"]
```
nyurik: if this is for all pull requests, i think you can just do `pull_requests: ~` or something similar?

.github/workflows/security.yml
```diff
@@ -3,13 +3,27 @@ name: Security audit
...
+  zizmor:
+      runs-on: ubuntu-latest
+      permissions:
+        security-events: write
+        contents: read
+        actions: read
+      steps:
+        - name: Checkout repository
+          uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+          with: {persist-credentials: false}
```
nyurik:
```suggestion
          with: { persist-credentials: false }
```

nyurik (approved):

.github/workflows/security.yml
CommanderStorm: Fair, I think it can be removed

```suggestion
```

PR #2447, opened 2025-12-25
feat(mbtiles): improve mbtiles summary output

nyurik:
* support for `json` and `json-pretty` output format
* print when there are no tiles in .mbtiles
* improve web-ui blessing to not wait for the user


justfile
```diff
@@ -67,6 +67,7 @@ bless: restart clean-test bless-insta-martin bless-insta-martin-core bless-insta
 # Bless the frontend tests
 [working-directory: 'martin/martin-ui']
 bless-frontend:
+    npm install
```
CommanderStorm: this will also update the lockfile.

How about we do this instead?

```suggestion
    npm clean-install
```

mbtiles/src/summary.rs
```diff
@@ -44,23 +45,28 @@ pub struct Summary {
 
 impl Display for Summary {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-        writeln!(f, "Schema: {}", self.mbt_type)?;
+        writeln!(f, "{:15} {}", "File path:", self.file_path)?;
+        writeln!(f, "{:15} {}", "Schema:", self.mbt_type)?;
```
CommanderStorm, line 49: I would prefer something like below, because:
- Your impl is a bit brittle towards ignoring newly added fields (like `file_path`), my way we get a compilation warning
- I find it harder to read compared to spelling out where the spaces go manually

```suggestion
        let { file_path, mbt_type } = self;
        writeln!(f, "File path:     {file_path}")?;
        writeln!(f, "Schema:        {mbt_type}")?;
```

tests/test.sh
```diff
@@ -654,6 +659,8 @@ if [[ "$MBTILES_BIN" != "-" ]]; then
   set -x
 
   $MBTILES_BIN summary ./tests/fixtures/mbtiles/world_cities.mbtiles 2>&1 | tee "$TEST_OUT_DIR/summary.txt"
+  $MBTILES_BIN summary --format json-pretty ./tests/fixtures/mbtiles/world_cities.mbtiles 2> "$TEST_OUT_DIR/summary.pretty.json.err" | cleanup_json | tee "$TEST_OUT_DIR/summary.pretty.json"
+  $MBTILES_BIN summary --format json ./tests/fixtures/mbtiles/world_cities.mbtiles 2> "$TEST_OUT_DIR/summary.json.err" | cleanup_json | tee "$TEST_OUT_DIR/summary.json"
```
CommanderStorm: In no other place do we not log the `.err` log.
Is adding it intentional?

-> I think we should only add this snapshot if we expect something to be therere, or?

mbtiles/src/bin/mbtiles.rs
```diff
@@ -33,11 +35,26 @@ pub struct Args {
...
 
+#[derive(
+    Default, Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize, EnumDisplay, ValueEnum,
+)]
+#[enum_display(case = "Kebab")]
+pub enum OutputFormat {
+    #[default]
+    Text,
+    Json,
+    JsonPretty,
```
CommanderStorm, line 47: Because people will misremember

```suggestion
    #[value(alias("pretty-json"))]
    JsonPretty,
```

CommanderStorm (approved):
LGTM, I have noted 3-4 comments below, but nothign major/blocking merging

mbtiles/src/summary.rs
nyurik: I am already not printing all the available fields - only those I care about to print. JSON is the one that keeps it consistent.  Also, formatting with your approach is a bit less readable because it does not line up inside the `if` blocks.  Lastly, this approach is the same as used right below that - with the table formatting.  If we really want to clean that up (and I don't think it is worth it), we may want to use one of many table-formatting crates, rather than what I used.

tests/test.sh
nyurik: thx, fixed

PR #2448, opened 2025-12-26
fix(mbtiles): meta-all with empty mbtiles

nyurik:
* Separate metadata from the detected tile format
* treat it as an error when creating a source for compatibility
* print it as `null` if the file is empty


CommanderStorm (approved):
In an ideal world, we would have a testcase to test Empty mbtiles, but I am fine to not have that.
It is very much an edge case.

PR #2451, opened 2025-12-28
Improve README: add H1 title and refine wording

raoqingyun0629:
- Add '# Martin' as an H1 heading after the badges section
- Change 'able to generate' to 'that can generate' for cleaner phrasing


CommanderStorm:
Looks akward when I look at https://github.com/maplibre/martin/blob/d0732c12119ffcd06ca6f4f190761777686fea88/README.md

I think we are going to stay as-is

PR #2452, opened 2025-12-31
chore(ci): migrate `mdbook-mermaid` to `mdbook-mermaid-ssr`

### LINKED ISSUE #2428: remove vendored mermaid diagram code
https://github.com/maplibre/martin/pull/2422 added the achitecture docs, which contains mermaid diagrams.

For this, the (as of the time of writing this) only way to do this in mdbook is to include the mermaid.js runtime in our codebase.
It would be much better, if this were rendered on the build side to a svg.

I have laid out a few approaches how to do this in
- https://github.com/badboy/mdbook-mermaid/issues/63#issuecomment-3664761403

> Currently, `mdbook-mermaid` relies on the Mermaid JS runtime on the client side. While this works, it requires shipping ~2 MB of JavaScript to every reader, which isn’t ideal—especially for static docs, low-bandwidth scenarios, or environments with strict JS policies.
> 
> An alternative would be to render Mermaid diagrams at build time / server-side and embed the resulting SVGs. I see two main implementation options:
> 
> * Call out to [`mermaid-cli`](https://github.com/mermaid-js/mermaid-cli) to generate SVGs
> * Headless Chrome–based rendering, similar to the approach used by [`mermaid-rs`](https://crates.io/mermaid-rs) (the code there is relatively small and straightforward)
> 
> Each approach has different trade-offs (dependencies, performance, portability), so supporting them behind a configurable backend (`cfg` / feature flag) could make sense.

CommanderStorm:
Resolves #2428 

Still needs a PR to taiki-e/install-action, but that should be simple enough:
- https://github.com/taiki-e/install-action/pull/1400


.github/workflows/build-deploy-docs.yml
```diff
@@ -35,8 +35,10 @@ jobs:
           file-path: './README.md'
           config-file: '.github/files/markdown.links.config.json'
 
+      - name: Install Chromium
+        uses: browser-actions/setup-chrome@b94431e051d1c52dcbe9a7092a4f10f827795416 # v2.1.0
       - uses: taiki-e/install-action@69e777b377e4ec209ddad9426ae3e0c1008b0ef3 # v2.64.0
-        with: { tool: 'mdbook,mdbook-tabs,mdbook-mermaid' }
+        with: { tool: 'mdbook,mdbook-tabs,mdbook-mermaid-ssr' }
```
CommanderStorm:
```suggestion
        with: { tool: 'mdbook,mdbook-tabs,cargo-binstall' }
      - run: cargo binstall -y mdbook-mermaid-ssr

```

nyurik (approved):

.github/workflows/build-deploy-docs.yml
```diff
@@ -35,8 +35,11 @@ jobs:
           file-path: './README.md'
           config-file: '.github/files/markdown.links.config.json'
 
+      - name: Install Chromium
+        uses: browser-actions/setup-chrome@b94431e051d1c52dcbe9a7092a4f10f827795416 # v2.1.0
       - uses: taiki-e/install-action@69e777b377e4ec209ddad9426ae3e0c1008b0ef3 # v2.64.0
-        with: { tool: 'mdbook,mdbook-tabs,mdbook-mermaid' }
+        with: { tool: 'mdbook,mdbook-tabs,cargo-binstall' }
+      - run: cargo binstall -y mdbook-mermaid-ssr
```
CommanderStorm:
```suggestion
      - run: cargo binstall -y mdbook-mermaid-ssr # TODO: move above once merged https://github.com/taiki-e/install-action/pull/1400
```

PR #2476, opened 2026-01-03
test(mbtiles): remove inner testing loop from `diff_and_patch`

CommanderStorm:
this test is already fairly hard to read and this inner testing loop is not helping with this.

There are other things that need to happen before this can be better debugged, but this is one of them.

Yes, the tests will be a bit slower, I don't care. They are fast enough.


CommanderStorm:
now I get why this is here, it is not about speed

PR #2478, opened 2026-01-03
test(mbtiles): remove the prefix-ism around how files are named for binary diff copy and simpify their naming

CommanderStorm:
What is relatively suboptimal for understandability is the `prefix` is used to hide things such as the criss-cross of `a`, `b`, `a`, `b` in the names.

In tests, DRY is not a concearn. Being understandable is much more of one.


mbtiles/tests/copy.rs
```diff
@@ -507,32 +507,32 @@ async fn convert(
...
 async fn diff_and_patch(
     #[values(Flat, FlatWithHash, Normalized)] a_type: MbtTypeCli,
     #[values(Flat, FlatWithHash, Normalized)] b_type: MbtTypeCli,
     #[values(None, Some(Flat), Some(FlatWithHash), Some(Normalized))] dif_type: Option<MbtTypeCli>,
-    #[values(&[Flat, FlatWithHash, Normalized])] destination_types: &[MbtTypeCli],
-    #[values(
-        ("v1", "v2", "dif"),
-        ("v1", "v1_clone", "dif_empty"))]
-    tilesets: (&'static str, &'static str, &'static str),
+    #[values(DB_WITH_DIFF, DB_EMPTY_DIFF)] tilesets: (&'static str, &'static str, &'static str),
```
CommanderStorm: results in the much more resonable nanes

```diff
+ diff_and_patch::a_type_3_Normalized::b_type_3_Normalized::dif_type_1_None::tilesets_1_DB_WITH_DIFF
- diff_and_patch::a_type_3_Normalized::b_type_3_Normalized::dif_type_1_None::destination_types_1___Flat_FlatWithHash_Normalized_::tilesets_1____v1_____v2_____dif___
```

nyurik:
eek... yeah, i think i over-optimized this one :)  I think this is a good path, just may need a bit of cleaning up to get it to run properly.  Thank you so much for working on this!!!

CommanderStorm:
> eek... yeah, i think i over-optimized this one :) I think this is a good path, just may need a bit of cleaning up to get it to run properly. Thank you so much for working on this!!!

It is not over-optimisation that is the problem.
The tests seem to be good.
They just do quite a lot in very little space, which makes them rather DENSE...

CommanderStorm:
@nyurik did you mean for this to be an 👍🏻 (you did not approve of the PR)

I think refactoring this testcase needs a few iterations to make it workable in the long run, so that would be appreciated.
Alternatively, which part needs cleaning up?

mbtiles/tests/copy.rs
```diff
@@ -510,32 +510,33 @@ async fn convert(
...
+    let short_b_type = shorten(b_type);
+    let short_a_type = shorten(a_type);
 
     eprintln!(
         "TEST: Compare {a_db} with {b_db}, and copy anything that's different (i.e. mathematically: {b_db} - {a_db} = {dif_db})"
     );
-    let (dif_mbt, mut dif_cn) = open!(diff_and_patch, "{prefix}__{dif_db}");
+    let (dif_mbt, mut dif_cn) = open!(
+        diff_and_patch,
+        "{a_db}_{short_b_type}--{b_db}_{short_a_type}={dif}__{dif_db}"
```
nyurik: I agree with copilot - shouldn't the type of "a" match the name of "a"?

CommanderStorm: technically I am weith you "two", but then the whole test fails.
I don't know why and would like to debug this from a clean slate

CommanderStorm: @nyurik this is now fixed. Remember that if you locally get test fails in this area, that you need to run cargo clean once to rename the files.

nyurik (approved):
thanks!!

PR #2482, opened 2026-01-04
fix: update copy text to include icon for better visibility

nyurik:
This adds a `⧉ Click to copy` text.

My only concern (possibly todo in another PR) is that `Click to copy` text is bigger than the label text.  I'm not a designer, but it seems the label should be bigger and more pronounced than the command hint, which should be grayer and smaller.


martin/martin-ui/src/components/sprite/SpriteCanvas.tsx
```diff
@@ -67,7 +67,7 @@ const SpriteCanvas = ({ meta, image, label, previewMode = false }: SpriteCanvasP
                   {label}
                 </code>
                 <br />
-                <span className="text-xs text-gray-500">Click to copy</span>
+                <span className="text-xs text-gray-500">⧉ Click to copy</span>
```
CommanderStorm: I'd like to have

https://lucide.dev/icons/copy

Plus `import { Copy } from 'lucide-react';` at the top

```suggestion
                <span className="text-xs text-gray-500"><Copy /> Click to copy</span>
```

nyurik: While it does look cleaner, do you really want to have a separate dependency and an image just for this?  Unicode symbol is so much simpler

nyurik: P.S. This is totally up to you - feel free to drive it any way you feel is better if you feel strongly

CommanderStorm: We are already using proper icons for all proper icons, I would like to not add a second class of icons.

CommanderStorm (approved):
I also took the liberty to clean up the tests surounding this a bit

nyurik:
@CommanderStorm is this what you wanted?  I got this by running `just start && just run`

image

I think the icon and the text should be smaller than the name of the label, and the icon might be best to be placed on the same line as the instruction text...  Perhaps left align it too? Or what would be best?

CommanderStorm:
jikes..

CommanderStorm:
defintively 100% intentional

PR #2483, opened 2026-01-05
fix: the `Copy` icon displaying nicely, next to the text and with enough padding ot all items

CommanderStorm:


nyurik:
Looks great!  One minor nit - it seems the icon is just a bit raised above the text - perhaps by 1px (icon is 24px, and the text box is 22px?).  This is really minor, so if you feel like it's just my ACD acting up, don't worry.

image

nyurik (approved):

PR #2491, opened 2026-01-09
feat: improve martin-cp progress output time estimate

amandasaurus:
`martin-cp` will print a simple progress bar. This change makes the elapsed and remaining time be displayed in a human readable hours/minutes/seconds display instead of just a raw number of seconds.

Namely, instead of this

```
[2026-01-09T16:40:22Z INFO  martin_cp] [71.5s] 44% @ 547.8/s | ✓ 7393 □ 31800 | 88s left
[2026-01-09T16:40:24Z INFO  martin_cp] [73.8s] 46% @ 555.1/s | ✓ 7754 □ 33200 | 84s left
[2026-01-09T16:40:26Z INFO  martin_cp] [75.9s] 47% @ 551.9/s | ✓ 8000 □ 33882 | 82s left
```

You get something like this:

```
[2026-01-09T16:30:01Z INFO  martin_cp] [1m05s] 29% @ 400.5/s | ✓ 5382 □ 20300 | 2m35s left
[2026-01-09T16:30:03Z INFO  martin_cp] [1m07s] 36% @ 486.0/s | ✓ 6000 □ 26173 | 1m54s left
[2026-01-09T16:30:05Z INFO  martin_cp] [1m09s] 40% @ 514.3/s | ✓ 6592 □ 28500 | 1m42s left
```

### DIFF
--- a/martin/src/bin/martin-cp.rs
+++ b/martin/src/bin/martin-cp.rs
@@ -319,7 +319,10 @@ fn write_duration(f: &mut impl Write, secs: f32) -> std::fmt::Result {
         let (weeks, days) = (days / 7, days % 7);
 
         if years > 0 {
-            write!(f, "{years}y{weeks:02}w{days:02}d{hrs:02}h{mins:02}m{secs:02}s")
+            write!(
+                f,
+                "{years}y{weeks:02}w{days:02}d{hrs:02}h{mins:02}m{secs:02}s"
+            )
         } else if weeks > 0 {
             write!(f, "{weeks}w{days:02}d{hrs:02}h{mins:02}m{secs:02}s")
         } else if days > 0 {
--- a/martin/src/bin/martin-cp.rs
+++ b/martin/src/bin/martin-cp.rs
@@ -298,6 +298,42 @@ enum MartinCpError {
     InvalidBoundingBox(&'static str, Bounds, RangeInclusive<f64>),
 }
 
+fn write_duration(f: &mut impl Write, secs: f32) -> std::fmt::Result {
+    if !secs.is_normal() || secs < 0. {
+        // Nonsense input
+        f.write_str("???")
+    } else if secs < 1. {
+        write!(f, "<1s")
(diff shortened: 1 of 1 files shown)


martin/src/bin/martin-cp.rs
```diff
@@ -312,9 +312,21 @@ impl Display for Progress {
         } else {
             0.0
         };
+        write!(f, "[")?;
+        let secs = elapsed_s.ceil() as u64;
+        let (mins, secs) = (secs / 60, secs % 60);
```
nyurik: i am not too happy about the dup code - perhaps a helper function?

nyurik (changes requested):
* clippy is unhappy
* a single helper function would be better to avoid code dup
* I wonder if it would be worth using one of the available crates for formatting?  Especially because the generation time could also be days rather than hours, or even weeks and years might be helpeful to print - so that people can cancel it right away.
* perhaps use one of the off-the-shelf formatting crates? Like https://lib.rs/crates/humantime or https://lib.rs/crates/human_format -- the  later one might even replace the currently used `size_format` dependency

amandasaurus:
Thanks for the feedback. 👍🏻
I've fixed the clippy errors. I think there's no way to write the `f32` → `u64` code that _doesn't_ generate a clippy warning. 😅
I didn't use an external dependency because I thought that a few lines would solve the issue, and some projects are (rightly) cautious about adding new dependencies.

CommanderStorm (approved):
Nice, thank you

CommanderStorm:
In case you are curious, longer term (this seems to require some work on the underlying CI infra), I want to move this to
- https://github.com/maplibre/martin/pull/2495

I hope that you are not offended if your code does not stay for more than a release 😉

amandasaurus:
I'm not offended. I think something like indicatif is a good thing. I just wanted to start with a simple PR

nyurik (approved):
Thanks!

PR #2495, opened 2026-01-10
feat(martin-cp): indicativ based progress bar

CommanderStorm:
This PR migrates us from our own progress bar to indicativ.

Needs tracing done first

This produces the following output:

```
00:00:02 -> eta: 10w [░                                        0%] 611973/1,466,015,503,701 (249,001.8169/s) | ✓ 5 □ 509994 
```

and when done

```
0:00:01 -> eta: 0s [████████████████████████████████████████ 100%] 5/5 (137.9386/s) | ✓ 5 □ 0
```

I was resonably uncreative with the template, might be a better option, not sure -> https://docs.rs/indicatif/latest/indicatif/#templates

Also note that this is not the be-all/end-all of solutions, this for example has the following problems:
It does not use all of indicatif/indicatif-tracins' functionality -> for example we could [use a trace](https://github.com/emersonford/tracing-indicatif/blob/main/examples/progress_bar.rs) or a [multiprocess-style](https://github.com/emersonford/tracing-indicatif/tree/main/examples), ...
=> iterative development


martin/src/logging/mod.rs
```diff
@@ -73,6 +76,78 @@ impl LogFormat {
...
+                        fmt_layer()
+                            .json()
+                            .with_span_events(FmtSpan::NONE)
+                            .with_writer(indicatif_layer.get_stderr_writer()),
+                    )
+                    .with(indicatif_layer)
+                    .init();
+            }
+        }
+    }
```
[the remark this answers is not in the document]
CommanderStorm: if I knew how..

martin/src/logging.rs
```diff
@@ -102,22 +177,47 @@ impl FromStr for LogFormat {
...
+            tracing::Level::INFO => log::LevelFilter::Info,
+            tracing::Level::WARN => log::LevelFilter::Warn,
+            tracing::Level::ERROR => log::LevelFilter::Error,
+            tracing::Level::TRACE => log::LevelFilter::Trace,
+        };
+        log_builder = log_builder.with_max_level(max_level);
+    }
+    // Use try_init to avoid panicking if logger is already initialized
+    // (e.g., by test harness or other initialization code)
+    let _ = log_builder.init();
```
CommanderStorm:
```suggestion
    log_builder
        .init()
        .expect("Failed to initialize log -> tracing bridge");
```

nyurik:
what is the output in case there is no terminal? e.g. running in CI

martin/src/logging/mod.rs
```diff
@@ -104,22 +179,42 @@ impl FromStr for LogFormat {
...
+        let max_level = match max_level {
+            Level::DEBUG => log::LevelFilter::Debug,
+            Level::INFO => log::LevelFilter::Info,
+            Level::WARN => log::LevelFilter::Warn,
+            Level::ERROR => log::LevelFilter::Error,
+            Level::TRACE => log::LevelFilter::Trace,
+        };
+        log_builder = log_builder.with_max_level(max_level);
+    }
+    let _ = log_builder.init();
```
CommanderStorm: this needs to be the following.

`tracing-indicatif` enables the `log` feature of `tracing`, which we want disabled.

```suggestion
    log_builder
        .init()
        .expect("Failed to initialize log -> tracing bridge");
```

CommanderStorm:
> what is the output in case there is no terminal? e.g. running in CI

This currently just prints nothing at all.
Not exactly sure where this requires work upstream or if this is possible without major achitecutral changes.

- https://github.com/console-rs/indicatif/issues/530#issuecomment-1826413120

nyurik:
heh, i guess i did look into it a while back :)   So we could try to implement it upstream, or at least have a local workaround.  Proposal: if not terminal, print what we print now, using TIME is the only indicator when to print. E.g. print every five minutes.  Alternatively we could print on a log scale, e.g. on this schedule:
* 5 second
* 10 second
* 30 second
* 1 minute
* 2 minute
* 5 minute
* every 5 minute until 1 hr
* every 30 min ever after

CommanderStorm:
I think all issues are resolved 👍🏻

nyurik (approved):
thanks!!

PR #2497, opened 2026-01-12
chore(deps): shear our dependencys

CommanderStorm:
🐑 + ✂️ = 🤟🏻 

not a large gain, more to the point that it runs the GC that we forgot
helps with compile times a tiny bit, but not much


nyurik (approved):
I wonder how shear differs from udeps

CommanderStorm:
udeps does not work for us, we use newer rust and workspaces.

https://github.com/Boshen/cargo-shear?tab=readme-ov-file#technique

PR #2503, opened 2026-01-15
fix: extract Azure account name from URL path

### LINKED ISSUE #2498: `az://` PMTiles source fails with "Account must be specified", but it is specified
Martin v1.2, running on Azure App Service. When I try to add a PMTiles source in the command line that I have saved on Azure Blob Storage, I get the error:

`2026-01-13T15:51:56.6663045Z [2026-01-13T15:51:56Z ERROR martin::config::file::main] Tile source resolution warning: Path az://<accountname>/<containername>/us_protomaps.pmtiles: Failed to parse object store URL of us_protomaps: Generic MicrosoftAzure error: Account must be specified`

From my reading of the docs, Martin should be parsing the account name from the `az://` URL I provided.

I'm currently able to work around this by disabling mandatory HTTPS on the Blob Storage account and using the HTTP URL instead.

VedantMadane:
## Summary

- Fixes Azure URL parsing for `az://`, `azure://`, and `adl://` schemes where account is in the URL path
- Extracts the account name from the first path segment when using documented URL format `az://account/container/path.pmtiles`

## Details

The Martin documentation shows `az://account/container/path.pmtiles` as a valid Azure URL format. However, the underlying `object_store` crate expects `az://container/path` with the account name configured separately via options or environment variables.

This PR bridges that gap by:
1. Detecting Azure URL schemes (`az`, `azure`, `adl`)
2. Extracting the account name from the first path segment
3. Reconstructing the URL as `az://container/path`
4. Passing the extracted account name as an option to `object_store`

This makes the documented URL format work as expected without requiring users to separately configure `account_name`.

## Test plan

- [ ] CI tests pass
- [ ] Manual testing with Azure Blob Storage (if available)
- [ ] Verify existing Azure configurations still work (when account_name is already provided, it's not overwritten)

Fixes #2498

### DIFF
--- a/martin/src/config/file/tiles/pmtiles.rs
+++ b/martin/src/config/file/tiles/pmtiles.rs
@@ -228,14 +228,43 @@ impl TileSourceConfiguration for PmtConfig {
         self.new_sources_url(id, url).await
     }
 
-    async fn new_sources_url(&self, id: String, url: Url) -> MartinResult<BoxedSource> {
+    async fn new_sources_url(&self, id: String, mut url: Url) -> MartinResult<BoxedSource> {
         use std::sync::LazyLock;
         use std::sync::atomic::{AtomicUsize, Ordering};
 
         static NEXT_CACHE_ID: LazyLock<AtomicUsize> = LazyLock::new(|| AtomicUsize::new(0));
         let cache_id = NEXT_CACHE_ID.fetch_add(1, Ordering::SeqCst);
 
-        let (store, path) = object_store::parse_url_opts(&url, &self.options)
+        let mut options = self.options.clone();
+
+        // Handle Azure URL format documented as az://account/container/path
+        // The object_store crate expects az://container/path with account configured separately.
+        // Extract account from the first path segment and set it as an option.
+        // See: https://github.com/maplibre/martin/issues/2498
+        if matches!(url.scheme(), "az" | "azure" | "adl") {
+            if let Some(segments) = url.path_segments() {
+                let segments: Vec<_> = segments.collect();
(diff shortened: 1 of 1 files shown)


CommanderStorm (changes requested):
I would really like to see this fixed properly instead.

Here is the current handling of this in the library we use.
It might be fixed already by upgrading the `object_store` and `pmtiles` dependency

https://docs.rs/crate/object_store/0.13.0/source/src/azure/builder.rs#662

This might be a documentation issue on our side, but I know too little about how azure builds their URs.

VedantMadane:
Thanks for the feedback! I investigated this further:

**Current state:**
- Martin uses `pmtiles = 0.18.2` which depends on `object_store = 0.12`
- `object_store` 0.13.0 has PR https://github.com/apache/arrow-rs-object-store/pull/604 that fixes Azure URL parsing

**Solution:**
- Upgrade `pmtiles` to **0.19.x** which uses `object_store = 0.13`
- Update documentation to use the new URL format: `az://container@account.blob.core.windows.net/path`

The new `object_store` 0.13 supports formats like:
- `az://container@account.blob.core.windows.net/path`
- `abfs://filesystem@account.dfs.core.windows.net/path`

I can update this PR to instead:
1. Bump pmtiles to 0.19.x
2. Update Azure URL documentation to reflect the proper format

Would that be the preferred approach?

CommanderStorm:
Lets do both the bump in the dependency and the documentation update to arrows' current state.
Since this is fairly manual (the docs need manual syncing on bumps), I did this myself

- https://github.com/maplibre/martin/pull/2504

Could you review that PR if anything pokes out at you?

PR #2510, opened 2026-01-17
feat(unstable-cog): Change tile path semantics for COG sources to match other sources, expose COG bounds, center and tileSize in TileJSON

Auspicus:
This change tightens COG support and introduces breaking validation behavior: Martin now only accepts Cloud Optimized GeoTIFFs that are fully compatible with WebMercatorQuad. COGs must be tiled (no strips), use Chunky planar configuration, RGB/RGBA 8-bit imagery or YCbCr, and one of None/LZW/Deflate/ModernJPEG/WEBP compression methods, and must declare ProjectedCRSGeoKey = EPSG:3857. Zoom levels are derived from image resolution and tile size rather than IFD order, tile coordinates are computed in absolute WebMercator space (including non-zero tile origins), and all images must use a consistent tile size. TileJSON generation is updated to include accurate bounds, center, zoom range, and tileSize. As a result, previously accepted but ambiguous or non-WebMercator COGs may now fail to load.


sharkAndshark:
Great work! Some thoughts here:

> reading a tile could be a lot more efficient as we can skip re-projecting pixels from file model space to the grid.

yes and not.. Actually only works for the last zoom, eg, if the zoom range is 8 -20, we could only skip re-projecting at zoom 20 with that gdal option. 

> a requirement for the COG to be projected in EPSG:3857

Yeah. For web-frinendly eanbled, it should be enforced. 

> removes the previous ability to read a COG source using arbitrary model space indexing.

It's a breaking change. Unfortunally our server rely on  this feature with a local CRS(without a epsg code). It's pretty fast as no clipping and merging.

 Option1

Remove it. As Most user only use 3857. My company could keep our martin branch and someday contributing back upstream once martin figure out how to tell user the inner tilegrid of COG. [I think `/catalog` is the best endpoint to do this](https://github.com/maplibre/martin/issues/343#issuecomment-2699719483)

Option 2

Keep it. And make web-friendly default. But We should keep the code as it's the most efficient way to serve tiles, no clipping and merging. And can serve any CRS.


I perfer option1. It's the most easy way. Once  #343 is resolved, we could add it back.

> it appears that the full COG file is loaded into memory

Is it true?

Auspicus:
From original PR description

---

*notes*
- if the COG is aligned to the mercator grid aka. "Web Friendly" (each chunk in the file starts and ends at the boundaries of a martin tile), for instance by using `gdal_translate cog.tif cog_aligned.tif -of COG -co TILING_SCHEME=GoogleMapsCompatible`, reading a tile could be a lot more efficient as we can skip re-projecting pixels from file model space to the grid
- this PR adds a requirement for the COG to be projected in EPSG:3857 which seemed to have been required before but was not enforced, the OGC COG spec makes no such assertion but we probably don't want to have to support re-projecting COGs written in arbitrary CRS. downloading a sample COG from AWS USDA NAIP mirror which was projected in NAD83 (EPSG:26910) doesn't work with the current assumptions without prior reprojection to EPSG:3857
- this PR removes the previous ability to read a COG source using arbitrary model space indexing (where `/cog_source/0/0/0` does not align with expected mercator tiles) which is not backwards compatible but since this feature is unstable, perhaps this is OK?
- ~~it appears that the full COG file is loaded into memory (which can be > 4GB with the BIGTIFF extension) and might be avoidable as COGs are assumed to be tiled in chunks of 256x256 or 512x512 so theoretically each tile request only needs to load up to 4 tiles at worst and 1 tile at best (depending on boundaries)~~ I was looking crudely at the total memory usage and because it was near the size of the COG I was loading I assumed this was the cause. After further digging it appears disabling the fonts feature drops the memory usage significantly. I believe due to the precomputed masks

---

> yes and not.. Actually only works for the last zoom, eg, if the zoom range is 8 -20, we could only skip re-projecting at zoom 20 with that gdal option.

I'm not sure why that would be the case if the internal tile blocks were aligned 1:1 with the WebMercator grid. Since a request for `/cog_source/10/10/10` would correspond 1:1 with an entire tile block in the COG in that instance. Can you elaborate?

> Yeah. For web-frinendly eanbled, it should be enforced.

I think to get this over the line and "stable-ish" we could aim for MVP that's fast but maybe doesn't support all the variations of COGs just yet. I will work on a full list of requirements for the COG file to work within the current system but it sounds like we're in alignment on:
- rgb/rgba 8 bit color (3 or 4 bands)
- projected in epsg:3857
- tiled not striped, a COG requirement (maybe requiring 256, 512, 1024 blocks)?
- compression?

> It's a breaking change. Unfortunally our server rely on this feature with a local CRS(without a epsg code). It's pretty fast as no clipping and merging.
> 
> Option1
> 
> Remove it. As Most user only use 3857. My company could keep our martin branch and someday contributing back upstream once martin figure out how to tell user the inner tilegrid of COG. https://github.com/maplibre/martin/issues/343#issuecomment-2699719483
> 
> Option 2
> 
> Keep it. And make web-friendly default. But We should keep the code as it's the most efficient way to serve tiles, no clipping and merging. And can serve any CRS.
> 
> I perfer option1. It's the most easy way. Once https://github.com/maplibre/martin/issues/343 is resolved, we could add it back.

I've gone with option 1 here. Keeping support will require extra complexity to the feature and it seems likely it will confuse users who want COGs to "just work".

> > it appears that the full COG file is loaded into memory
>
> Is it true?

It seems as though there was a change between v1.0.0 and v1.2.0 that made the fonts feature take a lot more memory on boot. It took about as much memory as my COG file so I thought it was that. Even when no fonts are loaded. Removing the pre-computed masks drops this significantly. I haven't dug in far enough to get to the bottom of it. I run martin without the fonts feature. TLDR; not sure how the tiff crate handles buffering, etc. yet but it doesn't seem to be the immediate culprit for this spike.

martin-core/src/tiles/cog/source.rs
```diff
@@ -254,20 +298,43 @@ fn verify_requirements(
...
 
 fn get_image(
     decoder: &mut Decoder<File>,
     path: &Path,
     ifd_index: usize,
+    origin: [f64; 3],
+    extent: [f64; 4],
+    (width_in_model, length_in_model): (f64, f64),
 ) -> Result<Image, CogError> {
     let (tile_width, tile_height) = (decoder.chunk_dimensions().0, decoder.chunk_dimensions().1);
```
Auspicus: Should we enforce that all chunks are square? ie. blocks of 256x256, 512x512, ...?

CommanderStorm:
> It seems as though there was a change between v1.0.0 and v1.2.0 that made the fonts feature take a lot more memory on boot. It took about as much memory as my COG file so I thought it was that. Even when no fonts are loaded

We noticed an increase in this in the folliwing PR, but we apparently should have looke deeper into this, rahter than just accepting it.
- https://github.com/maplibre/martin/pull/2438#issuecomment-3730865201

sharkAndshark:
> we could only skip re-projecting at zoom 20 with that gdal option.

If we have a COG file whose extent is this red square:

image


we could  make it 1:1 to webmercator quad at its nearest zoom level, let's suppose  it's 8:

this red bbox would be aligned and padded to four tiles big in web mercator quad:

|z|x|y|
|-|-|-|
|8|213|111|
|8|214|111|
|8|213|110|
|8|214|110|

then let's make it 1:1 to webmercator quad at zoom level 7:

|zoom 8| zoom 7|
|-------|--------|
|8/213/111|7/106/55|
|8/214/111|7/107/55|
|8/213/110|7/106/55|
|8/214/110|7/107/55|

The combination of `/7/106/55` and `/7/107/55` is much bigger than the combination of `(/8/213/1111 && /8/214/111 && /8/213/11 && /8/214/110)`. 

It's not allowed there are different extents in one single COG. I believe in gdal didn't align it at all zoom-levels and I remember I made a concept prove about this that I found I could only skip reprojection at the lagest zoom-level.


@Auspicus

Auspicus:
@CommanderStorm I have a proposal to help with the memory usage here: https://github.com/maplibre/martin/pull/2519 but I'm not sure if it gets it all the way across the line. I think there's more that could be done.

Auspicus:
> It's not allowed there are different extents in one single COG. I believe in gdal didn't align it at all zoom-levels and I remember I made a concept prove about this that I found I could only skip reprojection at the lagest zoom-level.
> 
> @Auspicus

@sharkAndshark I think I see what you mean. Let me try a few things out and report back. Ultimately it'd be a premature optimization to even try this without seeing how much the pixel by pixel copy costs. FWIW, where you've gotten it to feels pretty fast on my machine. Might not be necessary to go further :)

martin-core/src/tiles/cog/source.rs
sharkAndshark: >  A COG file SHALL have square tiles (TileWidth=TileLength)

Yes, COG spec ask this though tiff spec didn't.  See [Requirement 7 in COG spec](https://docs.ogc.org/is/21-026/21-026.html#_requirement_small_tiles)

Auspicus:
@sharkAndshark I've found that using the following options with GDAL produces a COG that aligns at all generated zoom levels so that all tile requests have exactly 1 intersecting tile in the source file:

```
gdal_translate original.tif compatible.tif \
-b 1 -b 2 -b 3 -of COG -co BIGTIFF=YES \
-co TILING_SCHEME=GoogleMapsCompatible \
-co ALIGNED_LEVELS=5 -co OVERVIEW_COUNT=4 \
-co OVERVIEWS=IGNORE_EXISTING -co COMPRESS=LZW \
-co ADD_ALPHA=YES -co NUM_THREADS=ALL_CPUS \
-co ZOOM_LEVEL_STRATEGY=LOWER -co BLOCKSIZE=512
```

The key options are `ALIGNED_LEVELS` and `OVERVIEW_COUNT`, you must request `ALIGNED_LEVELS=OVERVIEW_COUNT+1` so that every overview and the full resolution image are aligned. GDAL imposes a conservative maximum value for `ALIGNED_LEVELS` to 10 but it should be possible to push further.

I'm working on a commit to adjust the clipping and zoom level approximation to account for this. I will push that soon. It also appears that currently the only supported compression codecs are LZW, DEFLATE and NONE.

Also, the `/z/x/y` values are dependent on `tileSize` so if we correctly publish the internal tile size of the COG in the TileJSON then the requests are properly aligned by the client.

image

martin-core/src/tiles/cog/README.md
```diff
@@ -1,29 +1,38 @@
...
-| 8   | 4           | 320        | 0    |
-
-### Tile grid code representation
-
-```rust, ignore
-let images = vec![ image_0, image_1, image_2, image_3, image_4 ];
-let minzoom = 0;
-let zoom_of_image = image_count - 1 - idx_in_vec;  # TODO: what is this?
-let maxzoom = images.len() - 1;
+| 1   | 1           | 40         | 3    |
```
Auspicus: These should be multiples of 2.

martin-core/src/tiles/cog/README.md
```diff
@@ -1,29 +1,38 @@
...
+| 3   | 3           | 160        | 1    |
+| 4   | 4           | 320        | 0    |
+
+### COG file requirements enforced by Martin
+
+Due to the flexibility of the COG, GEOTIFF and TIFF file formats and the assumptions of Martin, not all COG files will be compatible. The following are a few requirements that Martin has of the COG file, some of which are defined in the COG or TIFF spec, some are constraints imposed by assumptions of Martin. If your file conforms to these requirements, it's more likely to work with Martin:
+
+* File MUST define the ProjectedCRS GeoKey with a value of 3857 (EPSG:3857)
+* File MUST use PlanarConfiguration=1 aka. "Contiguous" or "Chunky"
+* File MUST use Compression=1 "None", 5 "LZW" or 8 "Deflate" [See GDAL comrpession tag definitions](https://github.com/OSGeo/gdal/blob/c7d41bf263a1a3951546c5cfa66872fc05dfc8cc/frmts/gtiff/libtiff/tiff.h#L182-L219)
```
Auspicus:
```suggestion
* File MUST use Compression=1 "None", 5 "LZW" or 8 "Deflate" [See GDAL compression tag definitions](https://github.com/OSGeo/gdal/blob/c7d41bf263a1a3951546c5cfa66872fc05dfc8cc/frmts/gtiff/libtiff/tiff.h#L182-L219)
```

martin-core/src/tiles/cog/source.rs
```diff
@@ -75,59 +76,106 @@ impl CogSource {
...
     }
 }
 
+/// Find a zoom level of [WebMercatorQuad](https://docs.ogc.org/is/17-083r2/17-083r2.html#72) that
+/// is within the error tolerance difference from expected WebMercatorQuad zoom levels.
+fn web_mercator_zoom(model_resolution: f64, tile_size: u32) -> Option<u8> {
+    for z in 0..=MAX_ZOOM {
+        let resolution_in_web_mercator =
+            EARTH_CIRCUMFERENCE / f64::from(1_u32 << z) / f64::from(tile_size);
+        println!("{z:?} {resolution_in_web_mercator:?}");
```
Auspicus:
```suggestion
```

martin-core/src/tiles/cog/source.rs
```diff
@@ -163,17 +211,14 @@ impl Source for CogSource {
...
-        let mut decoder =
-            Decoder::new(tif_file).map_err(|e| CogError::InvalidTiffFile(e, self.path.clone()))?;
-        decoder = decoder.with_limits(tiff::decoder::Limits::unlimited());
-
         let image = self.images.get(&(xyz.z)).ok_or_else(|| {
             CogError::ZoomOutOfRange(xyz.z, self.path.clone(), self.min_zoom, self.max_zoom)
         })?;
 
-        let bytes = image.get_tile(&mut decoder, xyz, self.nodata, &self.path)?;
+        let file = File::open(&self.path).map_err(|e| CogError::IoError(e, self.path.clone()))?;
```
Auspicus, line 226: Opening a file descriptor per request seems like a bad idea. Perhaps the source should maintain a file descriptor pool?

martin-core/src/tiles/cog/source.rs
```diff
@@ -254,20 +319,50 @@ fn verify_requirements(
...
+        tile_size,
+    ))
+}
+
+/// Calculates the origin of the first tile
+fn get_tiles_origin(tile_size: u32, resolution: f64, origin: [f64; 2]) -> Option<(u32, u32)> {
+    let tile_size_mercator_metres = f64::from(tile_size) * resolution;
+    let tile_origin_x_f = (origin[0] + (EARTH_CIRCUMFERENCE / 2.0)) / tile_size_mercator_metres;
+    let tile_origin_y_f = ((EARTH_CIRCUMFERENCE / 2.0) - origin[1]) / tile_size_mercator_metres;
+    let tile_origin_x = tile_origin_x_f.floor() as u32;
```
Auspicus: Should use a fallible cast and return None when failing to cast.

martin-core/src/tiles/cog/source.rs
sharkAndshark: A single PR about file descriptor pool might be better

sharkAndshark:
> The key options are ALIGNED_LEVELS and OVERVIEW_COUNT

Good job! I would try this and inspect the inner layout of result file. If the aligning is from top down then there would be no different extents. But if the aligning is from full resolution to the last overview, there would be. So maybe gdal work from top to down when aligned levels count bigger than 1.

martin-core/src/tiles/cog/source.rs
CommanderStorm: Currently for example the styles are also handled similarly. With caching, this is likely still fine.

Future optimisation in any case

martin-core/src/tiles/cog/source.rs
```diff
@@ -418,36 +509,54 @@ fn get_extent(
...
+        longitude: -121.34674072265622,
+        latitude: 41.96765920367816,
+        zoom: 17,
+    }, Bounds {
+        left: -121.34948730468746,
+        top: 41.97174336327965,
+        right: -121.34399414062497,
+        bottom: 41.96357478222515,
+    }, 16, 18, 256)]
+    #[case("usda_naip_512_deflate_z2".to_string(), Center {
```
Auspicus, line 563: Added some test fixtures generated using GDAL from [USDA NAIP](https://registry.opendata.aws/usda-naip/). It's free for public use with attribution. I'm not sure how to properly attribute in this context other than putting the source in the file name.

CommanderStorm:
One thing which you can decide (I am fine with both ways) is if we want to stabilise this feature and take it into the default compilation.
Basically, renaming `unstable-cog` -> `cog`.

The effect of this is that it makes breaking changes slower (since they require a SemVer-major bump), so if you think this needs some in the short-term future, we should not do that.

Auspicus:
@sharkAndshark 

> Good job! I would try this and inspect the inner layout of result file. If the aligning is from top down then there would be no different extents. But if the aligning is from full resolution to the last overview, there would be. So maybe gdal work from top to down when aligned levels count bigger than 1.

From the GDAL docs it says:

> ALIGNED_LEVELS=<integer>: Number of resolution levels for which GeoTIFF tile and tiles defined in the tiling scheme match. When specifying this option, padding tiles will be added to the left and top sides of the target raster, when needed, so that a GeoTIFF tile matches with a tile of the tiling scheme. Only taken into account if [TILING_SCHEME](https://gdal.org/en/stable/drivers/raster/cog.html#drivers/raster/cog-co-TILING_SCHEME) is different from CUSTOM. Effect of this option is only visible when setting it at 2 or more, since the full resolution level is by default aligned with the tiling scheme. For a tiling scheme whose consecutive zoom level resolutions differ by a factor of 2, care must be taken in setting this value to a high number of levels, as up to 2^(ALIGNED_LEVELS-1) tiles can be added in each dimension. The driver enforces a hard limit of 10.

So this will work but requires padding of the higher zooms to align them with the lower zooms, and you're right to be hesitant about that. This will generate a larger file than necessary for sure. I wonder if `SPARSE_OK=TRUE` would help in this case:

> SPARSE_OK=[TRUE​/​FALSE]: (GDAL >= 3.2) Defaults to FALSE. Should empty blocks be omitted on disk? When this option is set, any attempt of writing a block whose all pixels are 0 or the nodata value will cause it not to be written at all (unless there is a corresponding block already allocated in the file). Sparse files have 0 tile/strip offsets for blocks never written and save space; however, most non-GDAL packages cannot read such files. On the reading side, the presence of a omitted tile after a non-empty one may cause optimized readers to have to issue an extra GET request to the TileByteCounts array.

Testing with this option on a small area, with SPARSE_OK=TRUE the file is ~4MB with 10 zoom levels. Without SPARSE_OK=TRUE the file is closer to 80MB. Both were using WEBP compression which likely makes this difference smaller than it would be with PNG, or JPEG.

Auspicus:
Amazingly, SPARSE_OK=TRUE actually kinda works in this MR. But you get an error logged for all the missing tiles. I'll see if we can rectify that. I generated an LZW compression COG and the difference was much greater between SPARSE_OK=TRUE/FALSE. Still around ~4MB with SPARSE_OK=TRUE but ~300MB with SPARSE_OK=FALSE.

image

martin-core/Cargo.toml
```diff
@@ -33,7 +33,15 @@ postgres = [
...
     "_tiles",
 ]
-unstable-cog = ["dep:png", "dep:tiff", "_tiles"]
+unstable-cog = [
+    "dep:png",
+    "dep:tiff",
+    "dep:tokio",
+    "_tiles",
+    "dep:image",
+    "dep:imageproc",
```
Auspicus: imageproc isn't used anymore, IIRC

martin-core/Cargo.toml
```diff
@@ -33,7 +33,15 @@ postgres = [
     "dep:serde_json",
     "_tiles",
 ]
-unstable-cog = ["dep:png", "dep:tiff", "_tiles"]
+unstable-cog = [
+    "dep:png",
+    "dep:tiff",
+    "dep:tokio",
```
Auspicus: tokio isn't used anymore IIRC

Auspicus:
@CommanderStorm I think we should keep this as unstable for now and not include it in the default build. There's still a few low hanging fruit like WebP and JPEG support which might require further breakage.

Auspicus:
@sharkAndshark do you mind casting another eye over this when you have some time?

martin-core/src/tiles/cog/image.rs
```diff
@@ -3,192 +3,128 @@ use std::io::BufWriter;
...
+            tiles_origin,
             tiles_across,
             tiles_down,
+            tile_size,
         }
     }
 
-    /// Retrieves a tile from the image, decodes it, and converts it to PNG format.
-    #[expect(clippy::cast_sign_loss, clippy::cast_possible_truncation)]
+    #[allow(clippy::cast_possible_truncation)]
```
CommanderStorm: this way this is an error if the truncation is not there 😉 

```suggestion
    #[expect(clippy::cast_possible_truncation)]
```

martin-core/src/tiles/cog/image.rs
```diff
@@ -3,192 +3,128 @@ use std::io::BufWriter;
...
+        let Some(idx) = self.get_chunk_index(xyz) else {
+            Err(CogError::NoImagesFound(path.to_path_buf()))?
+        };
         let color_type = decoder
             .colortype()
             .map_err(|e| CogError::InvalidTiffFile(e, path.to_path_buf()))?;
 
-        let (tile_width, tile_height) = decoder.chunk_dimensions();
-        let (data_width, data_height) = decoder.chunk_data_dimensions(tile_idx);
+        // @todo: can we replace this with reading the raw bytes and
```
CommanderStorm: nit: I think this is more common and at least highligted differently by my IDE.
Not sure about `@todo`

```suggestion
        // TODO: can we replace this with reading the raw bytes and
```

martin-core/src/tiles/cog/image.rs
```diff
@@ -3,192 +3,128 @@ use std::io::BufWriter;
...
-                    // Copy this component to the result vector
-                    result_vec[(idx_result + component_idx) as usize] = value;
-                }
-                if add_alpha {
-                    result_vec[(idx_result + 3) as usize] = 255; // opaque
-                }
-            }
+        let x = i64::from(xyz.x) - i64::from(self.tiles_origin.0);
+        let y = i64::from(xyz.y) - i64::from(self.tiles_origin.1);
+        if x < 0 || x >= i64::from(self.tiles_across) || y < 0 || y >= i64::from(self.tiles_down) {
```
CommanderStorm: nit: I think it is slightly more redable if we do it like this

```suggestion
        if 0 > x || x >= i64::from(self.tiles_across) || 0 > y || y >= i64::from(self.tiles_down) {
```

martin-core/src/tiles/cog/model.rs
```diff
@@ -104,10 +117,47 @@ impl ModelInfo {
                 )
             })
             .ok();
+
+        let mut projected_crs: Option<u16> = None;
+        // See: https://docs.ogc.org/is/19-008r4/19-008r4.html#_requirements_class_geokeydirectorytag
+        if let Ok(geokeys) = decoder.get_tag_u16_vec(Tag::GeoKeyDirectoryTag) {
+            let mut i = 0;
+            for chunk in geokeys.chunks_exact(4) {
```
CommanderStorm: the code below is a bit weird.
I think it could be a bit more ideomatic.

- So first, we discard the remainder which are not len%4
- then the chunk is always lenght 4
- and we are checking if things are in there

CommanderStorm (approved):
Nice work, looks good to me.
I have a few nits, nothing major

Obviously, I would like a review from @sharkAndshark if possible

sharkAndshark:
Yeah I will review it these days @Auspicus would you mind update the description of this PR either?

martin-core/src/tiles/cog/model.rs
Auspicus: I found a better way to represent the requirement here. Firstly we need to ensure the GeoKey version matches expectation (first 4 u16). Then we need to find the corresponding GeoKey for ProjectedCRS. Hopefully now the logic is a bit clearer.

Auspicus:
I managed to add in support for WEBP and JPEG without much fuss. It might be worth waiting to merge until upstream [`image-tiff` library change lands](https://github.com/image-rs/image-tiff/pull/336) in a release which adds `CompressionMethod::WEBP` so we don't need to hardcode it. Even with this I don't think this is quite ready for stable but it should cover a lot more use cases now.

CommanderStorm:
That is cool. Both ways are fine with me.

Auspicus:
I've created this discussion: https://github.com/maplibre/martin/discussions/2540 to cover support for other projections than EPSG:3857. I think this will require a bit of consideration.

martin-core/src/tiles/cog/image.rs
```diff
@@ -1,194 +1,272 @@
 use std::fs::File;
-use std::io::BufWriter;
+use std::io::{BufWriter, Read, Seek, SeekFrom};
```
CommanderStorm:
```suggestion
use std::io::{BufWriter, Read as _, Seek as _, SeekFrom};
```

PR #2511, opened 2026-01-19
feat: internal cache expiry

### LINKED ISSUE #2505: Making internal cache expiry more configurable
Martin has many concepts similar to Varnish - so we may need at least some of the same cache controls to configure it. We can even view Martin as a proxy, especially if it gets content from a http-hosted pmtiles.

For the good in-depth analysis of Varnish's TTL, Grace, and keep control see https://www.varnish-software.com/developers/tutorials/object-lifetime/
 
`TTL`, `Grace`, and `Keep controls` determine how long an object stays in Martin's internal cache, and what happens if the TTL has expired, .... This could be as simple as TTL only, but note that for max performance, we want to allow "respond while refreshing" - i.e. if a user asks for a tile, they may get an expired tile while Martin starts refreshing the tile cache in the background.  This way you don't get big delays in responses every time a popular tile is expired.

Currently, our caches don't expire in the classical sense, unless the LRU+LFU "think" this cache-item is not worth keeping around.

We likely can also make our caches expirable in a similar way -> https://docs.rs/moka/latest/moka/future/struct.CacheBuilder.html#method.expire_after

### See also
For headers sent to the client, see https://github.com/maplibre/martin/issues/749

CommanderStorm:
this PR adds TTL and TTI controls to our cache.

Imo, those are all the controls anybody would need, I feel someone will disagree on that.
That is fine 😉
This PR is progress to wards the goal, not the goal.

The PR looks unmanagbly large.
It is not.
That is mostly tests + docs.
The rest is plumbing work to get things from the config to the cache.

Resolves #2505


martin/src/config/file/main.rs
```diff
@@ -275,49 +326,96 @@ impl Config {
...
+                sprite_cache_idle_timeout: sprite_cache_idle_timeout.or(self.cache_idle_timeout),
                 #[cfg(feature = "fonts")]
                 font_cache_size_mb,
+                #[cfg(feature = "fonts")]
+                font_cache_expiry: font_cache_expiry.or(self.cache_expiry),
+                #[cfg(feature = "fonts")]
+                font_cache_idle_timeout: font_cache_idle_timeout.or(self.cache_idle_timeout),
             }
         } else {
             // TODO: the defaults could be smarter. If I don't have pmtiles sources, don't reserve cache for it
```
CommanderStorm: this functions code is the only part that can actualy be error prown.

Not sure if I did this right, likely not.
Need to check when not tired.

martin/src/config/file/main.rs
```diff
@@ -273,52 +328,174 @@ impl Config {
...
+    }
+
+    #[cfg(feature = "_tiles")]
+    fn get_tiles_cache_config(&self) -> Option<SubCacheSetting> {
+        let size_mb = self.tile_cache_size_mb;
+        // TODO: the defaults could be smarter. If I don't have a specific source, don't reserve cache for it
+        // default to 50% of the total cache size
+        let size_mb = size_mb.or(self.cache_size_mb.map(|c| c / 2));
+        // or default to 128MB
+        let size_mb = size_mb.unwrap_or(128);
```
CommanderStorm:
```suggestion
        // or default to 256MB
        let size_mb = size_mb.unwrap_or(256);
```

docs/src/config.yaml
```diff
@@ -309,6 +355,16 @@ fonts:
   # Defaults to cache_size_mb / 4
```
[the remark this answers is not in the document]
CommanderStorm:
```suggestion
  # Defaults to cache_size_mb / 8
```

CommanderStorm:
TODO: 
Consolidate cache configuration to `cache:`

CommanderStorm:
The change done in this PR is solid.
Closing as this will be 1-2 PRs until we can get the cache config refactored to accept this

PR #2516, opened 2026-01-21
fix: improve error message if no SVG sprite files are present

thomfuhrmann:
Closes issue #2136 by checking if the returned SVG file paths vector is empty. Also, the error message includes the word SVG now to indicate to the user that they should put SVG files into the folder.


nyurik (approved):
thx!

PR #2519, opened 2026-01-22
fix(perf): Remove FontSources.masks as they were consuming large amounts of memory even when no font sources were set

### LINKED ISSUE #2518: Lower memory usage of fonts feature
Running:

```bash
cargo b --release --no-default-features --features fonts --features unstable-cog
valgrind --tool=dhat ./target/release/martin -C 0 -W 16 ./tests/fixtures/cog/rgba_u8.tif
```

Results in:

```
Total:     6,075,775,587 bytes in 102,801 blocks
At t-gmax: 5,156,637,478 bytes in 77,539 blocks
At t-end:  5,156,358,707 bytes in 77,525 blocks
Reads:     5,471,865,799 bytes
Writes:    5,781,010,937 bytes
```

Compiling without the fonts feature:

```bash
cargo b --release --no-default-features --features unstable-cog
valgrind --tool=dhat ./target/release/martin -C 0 -W 16 ./tests/fixtures/cog/rgba_u8.tif
```

Results in:

```
Total:     15,500,664 bytes in 23,314 blocks
At t-gmax: 2,042,262 bytes in 5,680 blocks
At t-end:  63,153 bytes in 173 blocks
Reads:     8,370,650 bytes
Writes:    14,974,969 bytes
```

Adding an `Arc<T>` around `FontSources.masks` ie.

```rs
/// Thread-safe font manager for discovery, cataloging, and serving fonts as Protocol Buffers.
#[derive(Debug, Clone)]
pub struct FontSources {
    // ...
    /// Pre-computed bitmasks for each 256-character Unicode range.
    masks: Arc<Vec<BitSet>>,
}
```

And rebuilding, new report:

```
Total:     931,702,207 bytes in 43,917 blocks
At t-gmax: 608,550,346 bytes in 11,138 blocks
At t-end:  80,193 bytes in 228 blocks
Reads:     322,806,781 bytes
Writes:    636,577,571 bytes
```

Since this feature is on by default, it would be good if it didn't eat up as much memory. I'll post an accompanying PR for review.

Auspicus:
fixes #2518 

This drops the RSS value after the bitsets are computed. Unfortunately the bitset computation still spikes the memory usage quite high.

UPDATE: Swapping out BitSet for BTreeSet results in much lower peak memory usage after startup. I don't think BitSet is doing what we want here.

UPDATE: Replaced BTreeSet back with BitSet for the font codepoints for lower peak memory usage for a font that provides every available codepoint.


martin-core/src/resources/fonts/mod.rs
```diff
@@ -181,19 +180,17 @@ impl FontSources {
             return Err(FontError::InvalidFontRange(start, end));
         }
 
-        let mut needed = self.masks[(start as usize) / CP_RANGE_SIZE].clone();
+        let needed: &BTreeSet<usize> = &self.masks[(start as usize) / CP_RANGE_SIZE];
```
CommanderStorm: @nyurik Why did you use `masks` in the first place here instead of something like

```suggestion
        // Create a BitSet for the requested range (start..=end)
        let mut needed = BitSet::with_capacity(end as usize);
        for cp in start..=end {
            needed.insert(cp as usize);
        }
```

In the testing I have done, this works fairly well, but there must be a reason why this is done this way that I am not getting..

nyurik: tbh, it has been a long time since i did it, so no clue :)

martin-core/src/resources/fonts/mod.rs
```diff
@@ -247,7 +244,7 @@ pub struct FontSource {
     /// Face index within the font file (for .ttc collections).
     face_index: isize,
     /// Unicode codepoints this font supports.
-    codepoints: BitSet,
+    codepoints: BTreeSet<usize>,
```
CommanderStorm: Unsure if there is a better way.
Lets brainstorm a bit.

so we need to know which if the 256 bits -> [`BitSet256`](https://docs.rs/cbitset/latest/cbitset/type.BitSet256.html) is set.

While we curently store this as one very big `BitSet`, that is likely not nessesary, since this range is likely fairly compressible.
No font designer would leave weird gaps in their font intentionally I think.

- `Vec<BitSet256>` works and weighs in at ~1.1 megabytes (`MAX_UNICODE_CP` Bits)
- `BTreeSet<u32>` works and is 32B * number of font-items -> Likely lower than above.
- `HashMap<u16, BitSet256>` ((start / 256) -> next 256 code points) should works and is a compromise between the two?

CommanderStorm: Noticed that only u32 is needed for the index, that is another halfing of that memory usage.

This also stacks with the improvement above.

martin-core/src/resources/fonts/mod.rs
Auspicus: I can imagine BitSet would work well for a small range like 0-255 but if you support the whole unicode range you need a lot of memory per bitset.

martin-core/src/resources/fonts/mod.rs
Auspicus: Yeah there was a mix of usize and u32 around so definitely could standardise on u32 and avoid a bunch of casting

Auspicus: The problem with using a BitSet256 is you need to range it from 0-255 but the codepoints are 0-0x10FFFF so you either need to translate the requested codepoints back to 0-255 before masking (losing the performance of a bitset) or have one GIANT BitSet for each range. If you check the size of each BitSet before you'll see some are as big as 0x10FFFF (depending on the largest inserted value).

martin-core/src/resources/fonts/mod.rs
```diff
@@ -123,25 +122,25 @@ pub struct FontSources {
...
 }
 
 impl Default for FontSources {
     fn default() -> Self {
         let mut masks = Vec::with_capacity(MAX_UNICODE_CP_RANGE_ID + 1);
 
-        let mut bs = BitSet::with_capacity(CP_RANGE_SIZE);
+        let mut set = BTreeSet::new();
         for v in 0..=MAX_UNICODE_CP {
-            bs.insert(v);
```
Auspicus, line 135: I think the core issue is that although we allocate a BitSet::with_capacity(256) the value for v here might be 0x10FFFF causing a MASSIVE reallocation to fit values 0-0x10FFFF in the BitSet. From what I understand, the values must be contiguous without holes so each BitSet could be as large as (0x10FFFF / 32) (bits per u32) * 4 (bytes per u32) and then each of these is inserted for each range up to 4352.

nyurik:
one thing we could try without any other changes:

```diff
 fn get_available_codepoints(face: &mut Face) -> Option<GetGlyphInfo> {
-    let mut codepoints = BitSet::with_capacity(MAX_UNICODE_CP);
+    let mut codepoints = BitSet::new();  // dynamic memory growth
 ...
     for cp in 0..=MAX_UNICODE_CP {
     ...
+      if count >= face.num_glyphs() { break; } // we know we won't find any more glyphs in this font
```

martin-core/src/resources/fonts/mod.rs
```diff
@@ -53,7 +52,7 @@ const RADIUS: usize = 8;
 const CUTOFF: f64 = 0.25_f64;
 /// Maximum Unicode codepoint range ID.
 ///
-/// Each range is 256 codepoints long, so the highest range ID is 0xFFFF / 256 = 255.
+/// Each range is 256 codepoints long, so the highest range ID is 0x10FFFF / 256 = 255.
```
Auspicus: Should fix this number which is now 0x10FFFF/256 = 4352 (rounded up)

nyurik:
P.S.  Oh, silly me -- we no longer need to even do that -- I added character iteration to the freetype lib 2 years ago -- https://github.com/PistonDevelopers/freetype-rs/pull/257

so should just be able to call `for ch in face.chars() { ... }`

Auspicus:
So maybe no masks? I was wondering how expensive it would be to iterate vs difference the set in the first place

martin-core/src/resources/fonts/mod.rs
nyurik: my understanding is that each file would use a 136KB of RAM (`136*1024*8 = 1_114_112 = 0x11_0000` codepoint presence bits) - not that much memory. If we follow the optimization I proposed below, most fonts won't  even get that high, but it will keep the complexity to the minimal.

nyurik:
freetype-rs doesn't support iteration from arbitrary location (PRs welcome i guess), so i think it might still be useful to have a bitmask for perf reasons, esp since it is much simpler now.  Since you ran all the memory usage tests, could you try this implementation?

```rust
fn get_available_codepoints(face: &mut Face) -> Option<GetGlyphInfo> {
    let mut codepoints = BitSet::new();
    let mut spans = Vec::new();
    let mut first: Option<usize> = None;
    let mut last = 0;

    for (cp, _) in face.chars() {
        codepoints.insert(cp);
        if let Some(start) = first {
            if cp != last + 1 {
                spans.push((start, last));
                first = Some(cp);
            }
        } else {
            first = Some(cp);
        }
        last = cp;
    }

    if let Some(first) = first {
        spans.push((first, last));
        let start = spans[0].0;
        let end = spans[spans.len() - 1].1;
        Some((codepoints, usize::try_from(face.num_glyphs()).unwrap(), spans, start, end))
    } else {
        None
    }
}
```

nyurik:
(didn't test, but had to update the above a few times)

nyurik:
interestingly enough, my original implementation seemed to have had a bug - if a font file had a codepoint at the `MAX_UNICODE_CP-1` position (last allowed), the last span would not be included in the result.

martin-core/src/resources/fonts/mod.rs
```diff
@@ -63,36 +62,43 @@ mod cache;
...
+        spans.push((first, last));
         let start = spans[0].0;
         let end = spans[spans.len() - 1].1;
-        Some((codepoints, count, spans, start, end))
+        Some((
+            codepoints,
+            usize::try_from(face.num_glyphs()).unwrap(),
+            spans,
+            start,
+            end,
```
nyurik: you actually don't need to compute `end` here

```suggestion
            last,
```

martin-core/src/resources/fonts/mod.rs
Auspicus: The issue is less with each font source since it's only a single BitSet of 34816 u32 (139KB). The issue is with the pre-computed masks. Each range (of which there are 4352) carries a masking BitSet of up to `0x11_0000` (34816 u32 [4 bytes] in the backing BitVec). Assuming every range is this size (inaccurately) we get 600MB (`34816*4*4352` bytes). More accurately we get masking BitSets with sizes growing from 8-34816 u32 and more likely around 300MB total. Prior to adjusting the max codepoint size we had smaller BitSets (8-2048 u32) and less of them 256 (0xFFFF/256). Resulting in at most ~2MB (`2048*4*256` bytes) for masks but likely less due to not all BitSets being 2048 size.

nyurik:
(note to self - github editor is NOT that great for ... editing code)

@Auspicus have you had a chance to memory profile it?

martin-core/src/resources/fonts/mod.rs
Auspicus: See images attached with debugger running on `main` branch in the `default` for `FontSources`:

Screenshot_20260123_083203
Screenshot_20260123_083257

Auspicus:
I am just starting my workday now but I'll review again after work today. I will also post the dhat file in the viewer as it really highlights the BitSet allocations.

martin-core/src/resources/fonts/mod.rs
Auspicus: Is this so bad? I checked the md5sum of the output compared to main and it seems the same for the test font. Skips masking entirely.

```suggestion
            for i in start..=end {
                if let Ok(g) = render_sdf_glyph(&face, i as u32, BUFFER_SIZE, RADIUS, CUTOFF) {
                    stack.glyphs.push(g);
                }
            }
```

Auspicus, line 149: Should this function also require that `start` and `end` are less than `0x110000`?

nyurik: yes, I think so

martin-core/src/resources/fonts/mod.rs
```diff
@@ -63,36 +59,37 @@ mod cache;
...
 /// Glyph information: (codepoints, count, ranges, first, last).
-type GetGlyphInfo = (BitSet, usize, Vec<(usize, usize)>, usize, usize);
+type GetGlyphInfo = (BitSet, u32, Vec<(usize, usize)>, usize, usize);
 
 /// Extracts available codepoints from a font face.
 ///
 /// Returns `None` if the font contains no usable glyphs.
 fn get_available_codepoints(face: &mut Face) -> Option<GetGlyphInfo> {
-    let mut codepoints = BitSet::with_capacity(MAX_UNICODE_CP);
+    let mut codepoints = BitSet::new();
```
Auspicus, line 68: Without explicit capacity there will be some required resizing but it allows for this BitSet to be as small as 8 bytes (256 bits) when the font only provides codepoints within the range 0-255. If the font provides even a single codepoint at the end of the range this will need to be resized to 139KB. So we're trading off a small amount of start up time here for a lower memory footprint per font.

.github/files/lambda.yaml
```diff
@@ -17,5 +17,5 @@ Resources:
         - Ref: MartinLayer
       CodeUri: lambda-function/
       Handler: config.yaml
-      MemorySize: 350
-      Timeout: 10
+      MemorySize: 128
```
Auspicus, line 20: These are back to the default values, we'll see if the Lambda can run within these constraints now.

Auspicus:
Before:
```
git checkout main && \
cargo b --release --no-default-features --features fonts && \
valgrind --tool=dhat ./target/release/martin -C 0 -W 16 --font ./tests/fixtures/fonts

Total:     5,770,215,231 bytes in 90,450 blocks
At t-gmax: 4,857,098,381 bytes in 71,914 blocks
At t-end:  4,857,098,381 bytes in 71,914 blocks
Reads:     5,241,483,644 bytes
Writes:    5,475,459,168 bytes
```

After:
```
git checkout fix/2518/fonts-memory-usage && \
cargo b --release --no-default-features --features fonts && \
valgrind --tool=dhat ./target/release/martin -C 0 -W 16 --font ./tests/fixtures/fonts

Total:     8,947,164 bytes in 16,304 blocks
At t-gmax: 1,674,288 bytes in 4,653 blocks
At t-end:  96,933 bytes in 191 blocks
Reads:     4,826,227 bytes
Writes:    8,527,250 bytes
```

Auspicus:
I'm happy with where this is at now, let me know if you think we need to change anything @CommanderStorm / @nyurik

CommanderStorm (approved):
Thank you 👍

nyurik:
oops, we should have added to the subject / changelog that this also fixes incorrect font behavior - now the font glyph count will be reported correctly

PR #2521, opened 2026-01-26
refactor(martin-core)!: remove the configration from the martin-core crate

CommanderStorm:
I don't think the location this is located quite makes sense since this is not very general purpose usefull likely.


nyurik (approved):

nyurik:
@CommanderStorm seems like there is a hickup with precommit - you may want to force merge when ready

PR #2538, opened 2026-02-03
feat(geojson): serve MVT tiles from a GeoJSON file

thomfuhrmann:
## Description
This is a first proof-of-concept for _GeoJSON_ integration into Martin tile server. It depends on the _geozero_ crate for MVT tile creation and on _geo-index_ for the spatial Hilbert R-Tree. The clipping depends on the _i_overlay_ crate.

Here is a summary of the processing pipeline:

Steps to pre-process _GeoJSON_ features and create the data source:

1. Convert from WGS84 EPSG:4326 to Web Mercator EPSG:3857
2. Create spatial index using a packed Hilbert R-Tree

The following processing steps will be done when a tile request arrives:

1. Search for geometries that overlap with a given tile bounding box using the R-Tree
2. Clip geometries with tile bounding box (and optional buffer)
3. Transform into tile coordinate space and convert to MVT binary format

## Related Issue
Addresses #2054

## Limitations and open questions
The _geozero_ crate currently does not have an implementation for GeozeroDatasource for the GeoJson datatype. The processing steps are copied from there to avoid unnecessary string conversions.
Currently the buffer is created with a default size of 256 - should we make this configurable?
Testing is done only rudimentary.
Geometries are not cleaned before or after processing, instead it relies on correct geometries.


CommanderStorm:
Interestingly, your approach is a bit different from the one in 
- https://github.com/maplibre/martin/pull/2098

I am not sure which is better (I would need advice on this).
I am sure that there are a few things to steal from that PR (docs).
In terms of tests, this needs to be integrted into `test.sh` and a few more unit tests, for me to really trust that the code is battleproof.

I am fine with marking it as not-battleproof (`geojson` -> `unstable-geojson`) similar to rendering or COG.

martin-tile-utils/src/lib.rs
```diff
@@ -421,7 +421,7 @@ pub fn xyz_to_bbox(zoom: u8, min_x: u32, min_y: u32, max_x: u32, max_y: u32) ->
 }
 
 #[expect(clippy::cast_lossless)]
-fn tile_bbox(x: u32, y: u32, tile_length: f64) -> [f64; 4] {
+pub fn tile_bbox(x: u32, y: u32, tile_length: f64) -> [f64; 4] {
```
CommanderStorm, line 452:
```suggestion
#[must_use]
pub fn tile_bbox(x: u32, y: u32, tile_length: f64) -> [f64; 4] {
```

docs/src/sources-files.md
```diff
@@ -13,6 +13,16 @@ it and use it with `--config my-config.yaml` option.
...
+>
+> The difference between MBTiles and PMTiles is that:
+>
+> - **MBTiles** require the entire archive to be on the same machine. **PMTiles** can utilise a remote HTTP-Range request supporting server or a local file.
+> - Performance wise, **MBTiles** is slightly faster than **PMTiles**, but with caching this is negligible.
+> - Disk size wise, **MBTiles** is slightly (10-15%) higher than **PMTiles**.
+> - **PMTiles** requires less memory in extreme cases as sqlite has a small in-memory cache.
+>
+> The choice depends on your specific usecase and requirements.
+
```
CommanderStorm:
```suggestion

```

CommanderStorm:
So I added this to our benchmarking harness and CLI.
I also added a few tests.

One of the tests discovered that this:
> Testing geojson_fc1_0_0_0.pbf from http://localhost:3111/feature_collection_1/0/0/0
> Error in layer 0 in feature 0 in geometry 0: First ring isn't an outer ring
> Error in layer 0 in feature 1 in geometry 0: First ring isn't an outer ring
> Error in layer 0 in feature 2 in geometry 0: First ring isn't an outer ring

That looks like a real error and not related to how the tests work.

docs/src/sources-files.md
```diff
@@ -13,7 +13,15 @@ it and use it with `--config my-config.yaml` option.
...
+>
+> The difference between MBTiles and PMTiles is that:
+>
+> - **MBTiles** require the entire archive to be on the same machine. **PMTiles** can utilise a remote HTTP-Range request supporting server or a local file.
+> - Performance wise, **MBTiles** is slightly faster than **PMTiles**, but with caching this is negligible.
+> - Disk size wise, **MBTiles** is slightly (10-15%) higher than **PMTiles**.
+> - **PMTiles** requires less memory in extreme cases as sqlite has a small in-memory cache.
+>
+> The choice depends on your specific usecase and requirements.
 ### Autodiscovery
```
CommanderStorm:
```suggestion
> The choice depends on your specific usecase and requirements.

### Autodiscovery
```

CommanderStorm (commented):
Minor linting issue that I overlooked

thomfuhrmann:
Thank you @CommanderStorm for setting up the test cases! I reverted the point order because MVT requires it that way.
I still need to check for invalid geometries.

thomfuhrmann:
I have been working on geometry validation using the validation trait from the geo crate. It works for simple geometries. However, for more complex geometries it still serves invalid geometries too. Also, the response times take quite long (it may be worth to parallelize to speed it up). I still want to test and compare the validation routines from GEOS as this seems to have the most mature algorithms.

martin-core/src/tiles/geojson/source.rs
```diff
@@ -0,0 +1,554 @@
...
+            .map_err(|err| GeoJsonError::IoError(err, path))?;
+        let geojson = geojson_str
+            .parse::<GeoJson>()
+            .map_err(|err| GeoJsonError::GeoJsonError(Box::new(err)))?;
+
+        let (geojson, rtree) = preprocess_geojson(geojson);
+
+        let tilejson = tilejson::tilejson! {
+            tiles: vec![],
+        };
```
CommanderStorm: We likely should have the bounds be specified here.
We already do BBoxing in the rtree, so this likely makes sense.

These are two things that we likely can do, but don't need to do in this PR
- We might be able to populate `vector_layers`, but that would be some effort.
- `minzoom` and `maxzoom` are likely helpfull to provide at the config level

docs/src/installation.md
```diff
@@ -86,3 +86,22 @@ If you [install Rust](https://www.rust-lang.org/tools/install), you can build ma
...
+#### Platform-Specific Build Notes
+
+##### Windows
+
+When building from source on Windows, please note the following feature limitations:
+
+- **`unstable-rendering`**: This feature is **not available on Windows**. It requires `maplibre_native` which currently only supports MacOS and Linux. For updates, see [`maplibre/maplibre-native-rs`](https://github.com/maplibre/maplibre-native-rs).
+
+- **`geojson`**: This feature works on Windows but requires **C++ build tools** (CMake and a C++ compiler) because the `geos` dependency is compiled from source. Install [Visual Studio Build Tools](https://visualstudio.microsoft.com/downloads/) with C++ support before building:
+  ```bash
```
CommanderStorm:
```suggestion
- **`geojson`**: This feature works on Windows but requires **C++ build tools** (CMake and a C++ compiler) because the `geos` dependency is compiled from source. Install [Visual Studio Build Tools](https://visualstudio.microsoft.com/downloads/) with C++ support before building:

  ```bash
```

tests/fixtures/geojson/items.json
CommanderStorm, line 1: We need to find something a bit more minimal than 500k lines / 11 MB.

This curiously does not even seem to be used, so might just be an artefact

justfile
```diff
@@ -106,8 +108,21 @@ build-release target:
...
+            cargo build --release --target {{target}} --package martin --locked \
+                --no-default-features \
+                --features fonts,lambda,mbtiles,metrics,pmtiles,postgres,sprites,styles,webui
+        elif [[ "{{target}}" =~ .+-apple-darwin ]]; then
+            cargo build --release --target {{target}} --package martin --locked \
+                --no-default-features \
+                --features fonts,lambda,mbtiles,metrics,pmtiles,postgres,sprites,styles,webui
+        else
+            cargo build --release --target {{target}} --package martin --locked
+        fi
```
CommanderStorm: This is a really ugly hack and we likely should not merge it as-is.
Geos is just a bit hard to install it seems.

I think making geojson an non-default feature and enabling it in the build here in CI:
- we are building for musl (aka statically)
- if packagecfg finds the library

This way, we have a much simpler life and our contributors don't neeed to deal with geos if they want to contribute to martin.

Other oppinions?

CommanderStorm:
I think it should now be integrated into our CI, with most issues fixed. Lets see.

tests/fixtures/geojson/items.json
thomfuhrmann: Of course we can remove it, sorry for readding it. I just wanted to test it as an edge case.

justfile
thomfuhrmann: Yes, I also found it hard to build on my Windows machine in the office. At first I wanted to use the geo crate for geometry validation, which would not require a dependency on C++ code. Unfortunately, the geo crate validation routine seems to be much slower than GEOS.

thomfuhrmann:
> I think it should now be integrated into our CI, with most issues fixed. Lets see.

Thank you @CommanderStorm! Sorry for my late reply, I was a little busy at work.

thomfuhrmann:
I reverted back to geometry validation with the geo crate to simplify the build process. I think the performance is still acceptable even though worse than with GEOS.

Also, I found out that there is an open issue in the geojson crate that would allow much faster parsing and conversion to geo-types geometries. I am planning to integrate this once it will be released:
https://github.com/georust/geojson/issues/263

tests/expected/geojson/save_config.yaml
```diff
@@ -0,0 +1,12 @@
...
+mbtiles: tests/fixtures/geojson
+geojson:
+  paths: tests/fixtures/geojson
+  sources:
+    countries: tests/fixtures/geojson/countries.geojson
+    feature_1: tests/fixtures/geojson/feature_1.geojson
+    feature_collection_1: tests/fixtures/geojson/feature_collection_1.geojson
+    feature_collection_2: tests/fixtures/geojson/feature_collection_2.geojson
+    feature_collection_3: tests/fixtures/geojson/feature_collection_3.json
+    items: tests/fixtures/geojson/items.json
```
CommanderStorm:
```suggestion
```

martin-core/README.md
```diff
@@ -14,6 +14,7 @@ The library providing the basic building blocks for [Martin](https://maplibre.or
   - [x] mbtiles
   - [x] pmtiles
   - [x] cog
+  - [x] geojson (requires C++ build tools or `libgeos-dev` being installed)
```
CommanderStorm:
```suggestion
  - [x] geojson
```

docs/content/sources-geojson.md
```diff
@@ -0,0 +1,55 @@
...
+# Martin will watch this directory and reflect any *.geojson changes live
+martin /path/to/geojson/directory
+```
+
+Or via config file:
+
+```yaml
+geojson:
+  paths:
+    - /path/to/geojson/directory
```
nyurik, line 42: should we add `sources: ...` subsection?

Possible TODO for the future - we may want to allow fetching geojson from the web, caching it, and serving tiles from it

CommanderStorm, line 42:
```suggestion
  paths:
    - /path/to/geojson/directory
  sources:
    foo: /path/to/file.geojson
```

nyurik (approved):
TODO: eagerly cut-up into tiles should take into account the expected zoom levels  - i.e. we may want to set min/max zoom for geojson source, and based on that eagerly pre-cut for all of them, or pre-cut per zoom on request, etc

martin-core/tests/geojson_test.rs
```diff
@@ -1,31 +1,20 @@
...
 use std::io::Write as _;
+use std::num::NonZeroU32;
 
-/// MVT layer extent and clip buffer, mirroring the private `rect::{EXTENT, BUFFER_SIZE}`.
-/// Clipped coordinates land within `[-BUFFER, EXTENT + BUFFER]`, give or take one unit of
-/// `transform_to_tile_coordinates`' `.floor()` rounding at the buffered edge.
-const EXTENT: i64 = 4096;
-const BUFFER: i64 = 256;
-const FLOOR_SLACK: i64 = 1;
+const 1: i64 = 1;
```
CommanderStorm:
```suggestion
```

thomfuhrmann:
Really happy to see this merged! 🎉

CommanderStorm:
Thank you for all of this work 😉

PR #2541, opened 2026-02-06
fix: Make sure that `route-prefix` does not break the UI

albing:
Using `--route-prefix` breaks the UI.

If I start Martin with `--route-prefix /martin`, I can't use the UI because index.html is always going to reach out to `/_/assets/*` for its static assets.

Steps to reproduce from the release:
1. Download and unzip the 1.3.0 release
2. Run `./martin --webui enable-for-all --route-prefix /martin my.mbtiles`
3. Visit http://localhost:3000/martin/catalog or http://localhost:3000/martin/health : works as expected
4. Visit http://localhost:3000/martin/ with the inspector open : 404 on
    * `/_/assets/index-Cw7_pX04.js`
    * `/_/assets/maplibre-CAAeTMl5.js`

Testing the changes:
1. In the main branch, `just run` and see that http://localhost:3000 works
2. `just run --webui enable-for-all --route-prefix /martin` and see that it breaks.  You might need to set `VITE_MARTIN_BASE=http://localhost:3000/martin` in `.env.local` if you have that set.
3. Check out this branch and run commands 1 and 2 again and see that they both work (making sure to adjust .env.local)

Full disclosure, vite (and frontend in general) isn't my main thing, so I'm not sure if this has greater implications than I realize.  Feedback welcome!

This may relate to https://github.com/maplibre/martin/discussions/2207


CommanderStorm (changes requested):
I don't think this will work as expected, at least it does not work this way when I run the production bundle:

```bash
just run --webui enable-for-all --route-prefix /martin ./tests/fixtures/pmtiles/stamen_toner__raster_CC-BY+ODbL_z3.pmtiles
```

We don't use vite in prod, so these URLs need to be relative indeed.
Not quite sure how to best handle this.

CommanderStorm (approved):
Seems to work now mostly.

We still don't support serving the UI at `/martin` but we support `/martin/` now, which is progress.

I also tested the regular base URLs.

PR #2546, opened 2026-02-10
refactor: centralized cache config

CommanderStorm:
This centralises the cache config which allows us to better extend it with for example TTL/..

This is PR is backwars compatible and tested to be so, but will raise warnings in all previous users, unless they did not configure the cache.

Specifically, this resolves blockers for https://github.com/maplibre/martin/pull/2511


.github/files/lambda-function/config.yaml
```diff
@@ -3,4 +3,5 @@ pmtiles:
     webp2: ./webp2.pmtiles
  # if lambdas are added/removed for one request,
  # caching needs to be done externally
-cache_size_mb: 0
+cache:
+  size_mb: 0
```
CommanderStorm:
```suggestion
cache: false
```

docs/src/config.yaml
```diff
@@ -22,28 +22,54 @@ base_path: null
...
+  # - Sprites: `cache.sprites.size`
+  size: 512MB
+
+  tiles:
+    # Allows overriding the size of the tile cache.
+    # Defaults to `cache.size` / 2
+    size: 256MB
+
+  pmtile_directorys:
+    # Size of the pmtile-directorys cache
```
nyurik:
```suggestion
  pmtile_directories:
    # Size of the pmtile-directories cache
```

nyurik (commented):
i don't think we want to replicate per tile source type structure under cache.  We would want to preserve the config structure for that as before, and see how we can introduce only overall caching configuration to the common `cache:` top element.

CommanderStorm:
That message is a bit cryptic.

I currently have

```yaml
cache:
  size: 512MB          # Overall cache budget
  tiles:
    size: 256MB
    #expiry: 20h
    #idle_timeout: 1h
  pmtile_directories:
    size: 128MB
  sprites:
    size: 64MB
  fonts:
    size: 64MB
```

Would you prefer this?

```yaml
cache:
  size: 512MB          # Only overall/common cache settings
  #expiry: 20h
  #idle_timeout: 1h

pmtiles:
  cache:
    size: 128MB        # PMTiles-specific cache config
  # other pmtiles config...

sprites:
  cache:
    size: 64MB         # Sprite-specific cache config
  # other sprites config...

fonts:
  cache:
    size: 64MB         # Font-specific cache config  
  # other fonts config...
```

CommanderStorm:
the second system has better coherency, when just having one group of things, but I think that is not the common case.

It also has:
- a bit more confusing disabling support -> no `cache: false`
- where does the tile cache size configuration now live?
- not sure if it is harder or simpler to reason about the caching header adjustments that you want to add or how that should be done. Likely this is possible under both, maybe more ergonomic under your version

nyurik:
isn't how we have it now? (the second variant)

nyurik:
I think this should work:

```yaml
cache: false  # disable

---

cache:
  enabled: false  # disable, and this way users can keep the existing
                  # settings in the config file, i.e. to try things
  total_size: 15GB
  tiles:
    size: 10GB  # I am not certain if this should override total_size, or if total_size should not exist
    ttl: 1h
    expiry-header: 30m

pmtiles:
  cache:
    tile-size: 2GB  # This overrides cache.tiles.size
    meta-size: 1GB  # Name TBD. Store directories data
```

CommanderStorm:
Just so that we are on the same page, could you add `pmtiles` to your example and where those cache keys are?

And how should this case be handled?

```
cache: false
pmtiles:
   cache:
     size: 20MB

nyurik:
I updated the example, although there are still uncertainties of how to handle tile cache vs non-tiles cache, and how to have "total" cache - if we even want to go there.  I.e. if its possible to have sub-groups inside one big cache size, with each group having smaller sub-max size.

As for your example, if the global cache is false, all other cache settings are disabled.  Note that we may want to NOT have `cache: false` because it doesn't make sense if we control headers AND memory size in the same section.

CommanderStorm:
Pmtiles.cache.tile-size is a bit weird since would imply that you can disable the cache on a per-source level. That is currently not the case.

So how would mbtiles.cache.size interact with this? A bit undefined

PR #2563, opened 2026-02-12
chore: more cfg magic instead of #[allow(unused_variables)]

CommanderStorm:
this is more of a discussion which way we want to go down:

- cfg_attr with allowing unused variables for the function
- cfg the attributes

### DIFF
--- a/martin/src/srv/server.rs
+++ b/martin/src/srv/server.rs
@@ -14,12 +14,7 @@ use tracing_actix_web::TracingLogger;
 
 #[cfg(all(feature = "webui", not(docsrs)))]
 use crate::config::args::WebUiMode;
-#[cfg(any(
-    feature = "_tiles",
-    feature = "sprites",
-    feature = "fonts",
-    feature = "styles"
-))]
+#[cfg(feature = "_catalog")]
 use crate::config::file::ServerState;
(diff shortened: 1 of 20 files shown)


martin/src/srv/server.rs
```diff
@@ -221,7 +257,9 @@ pub fn new_server(config: SrvConfig, state: ServerState) -> MartinResult<(Server
 
         app.wrap(TracingLogger::default())
             .wrap(NormalizePath::new(TrailingSlash::MergeOnly))
-            .configure(|c| router(c, &config))
+            .configure(|c| {
+                router(c, &config);
+            })
```
CommanderStorm:
```suggestion
            .configure(|c| router(c, &config))
```

martin/tests/pg_server_test.rs
```diff
@@ -1136,7 +1136,9 @@ tables:
             ))
             .app_data(actix_web::web::Data::new(state.tiles))
             .app_data(actix_web::web::Data::new(SrvConfig::default()))
-            .configure(|c| ::martin::srv::router(c, &SrvConfig::default())),
+            .configure(|c| {
+                ::martin::srv::router(c, &SrvConfig::default());
+            }),
```
CommanderStorm:
```suggestion
            .configure(|c| ::martin::srv::router(c, &SrvConfig::default())),
```

nyurik (approved):

PR #2571, opened 2026-02-22
feat: add POST /_/reload endpoint for dynamic catalog updates (#288)

uucokgis:
feat: add POST /_/reload endpoint for dynamic catalog updates

Add reload endpoint infrastructure to enable dynamic catalog updates without server restarts, addressing issue #288.

Changes:
- Add POST /_/reload endpoint in srv/admin.rs
- Register endpoint in srv/server.rs
- Add upsert_sources() method to TileSources for adding/updating sources
- Add remove_sources() method to TileSources for cleanup
- Methods leverage existing DashMap for thread-safe concurrent updates

The infrastructure is now in place for dynamic source updates:
- DashMap provides thread-safe concurrent access
- upsert_sources() can add or update sources atomically
- remove_sources() can clean up stale sources

### DIFF
--- a/martin/src/config/file/main.rs
+++ b/martin/src/config/file/main.rs
@@ -68,6 +68,9 @@ pub struct ServerState {
 
     #[cfg(feature = "styles")]
     pub styles: martin_core::styles::StyleSources,
+
+    #[cfg(feature = "postgres")]
+    pub postgres_configs: OptOneMany<super::postgres::PostgresConfig>,
 }
 
 #[serde_with::skip_serializing_none]
@@ -274,6 +277,9 @@ impl Config {
 
             #[cfg(feature = "styles")]
             styles: self.styles.resolve()?,
+
+            #[cfg(feature = "postgres")]
+            postgres_configs: self.postgres.clone(),
         })
     }
 
--- a/martin/src/source.rs
+++ b/martin/src/source.rs
@@ -3,25 +3,28 @@ use dashmap::DashMap;
 use martin_core::tiles::catalog::TileCatalog;
 use martin_core::tiles::{BoxedSource, Source};
 use martin_tile_utils::TileInfo;
+use std::sync::Arc;
 use tracing::debug;
 
 /// Thread-safe registry of tile sources indexed by ID.
 ///
-/// Uses a [`DashMap`] for concurrent access without explicit locking.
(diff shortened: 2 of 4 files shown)


CommanderStorm:
Have you looked at https://github.com/maplibre/martin/issues/288#issuecomment-3707778253

Our internal architecture is not currently very well optimised to allowing for the endpoint that you want to add. (as you are seeing)
I think we would need to refactor this a bit before this can be possible.

To prevent you from wasting your time, I would suggest to break parts of that refactoring off and work toawards this larger goal this way. Would this be something that you want to do?

uucokgis:
> Have you looked at [#288 (comment)](https://github.com/maplibre/martin/issues/288#issuecomment-3707778253)
> 
> Our internal architecture is not currently very well optimised to allowing for the endpoint that you want to add. (as you are seeing) I think we would need to refactor this a bit before this can be possible.
> 
> To prevent you from wasting your time, I would suggest to break parts of that refactoring off and work toawards this larger goal this way. Would this be something that you want to do?

Thanks for patience and leading, that is my first PR on OpenSource :)
I'm happy to iterate, but I'd like to clarify the scope:
1. Should I implement just the HTTP endpoint with proper diffing/cache invalidation?
2. Or do you want the full background actor architecture with polling?
3. Would you accept a simpler Phase 1 (HTTP endpoint + proper diffing) as a stepping stone, or do you want the full RFC implemented?"

And I'd like to ask this: This is a lot of code for a feature that's been "wanted" for 3+ years. We can maybe merge with a "flag"? 
Because current users may satisfy with "reload" endpoint instead of auto-discovery/polling mechanism.

CommanderStorm:
Thanks again for working on this - especially as your first OSS PR 🙌

Unfortunately, this isn’t something we can safely "hack around" or merge behind a flag. The current architecture doesn't support this cleanly, and layering the endpoint on top would introduce technical debt we’d likely have to undo later.

We really need to address the underlying architectural constraints first. If you’re open to breaking that refactor proposed in https://github.com/maplibre/martin/issues/288#issuecomment-3707778253 into smaller, reviewable steps, I'd be happy to guide you through it.

CommanderStorm:
As stated above, this issue needs a bit of deeper work to get done.
I think this should be a new PR (if you want to contribute towards this goal).
Lets close this PR to start fresh

PR #2572, opened 2026-02-24
feat: Add retry mechanism on locked/busy mbtiles files

### LINKED ISSUE #1591: Retry locked mbtiles files
Currently Martin stops in case of corrupt tile file or when tile file is in use (being created by tippecanoe). 
It would be better to console log this as a warning, ignore the tile and continue running the tile server.

sdemeesterde:
Fix #1591 

Hello ! 
This is my first-ever PR to an open-source project.

Currently, martin crashes when trying to fetch the metadata of a busy mbtile file. It was proposed in the open issue: https://github.com/maplibre/martin/issues/1591 to be more lenient and perform retries when the file was locked.

### Technical note:

The `SQLITE_BUSY` error is hardcoded as the sqlite wrapper used in martin (sqlx) doesn't offer an error enum match for specific sqlite error codes  (unlike rusqlite).

### Retry strategy:
The retrial strategy is an exponential backoff with the following delays: 50ms, 0.1s, 0.2s, 0.4s, 0.8s, 1.6s, 3.2s, and finally 6.4s.

### I would happily take feedback on:
- Is limiting the retry behavior strictly to SQLITE_BUSY errors the intended approach ?
- Are the exponential backoff delays acceptable for this use case ?
- Is the warning message too verbose for the logs ?
- Is there anything else that I might have missed ? :)

Thanks !

Screenshot from 2026-02-24 14-09-21


CommanderStorm (changes requested):
The main feedback I have is to NEVER `std::thread::sleep` in an async context.
That is what `tokio`'s sleep is for.
Since this is likely not avaliable under this feature already, you need to add this to Cargo.toml for the mbtiles feature.
Here is a resource that you might enjoy to learn more https://youtu.be/o2ob8zkeq2s

Also, could we use something like https://docs.rs/backon/1.6.0/backon/
This way we can have a simpler impl, but also a better impl (-> jitter+fibonacci).

sdemeesterde:
Thanks for the precious feedback ! I really appreciate that as I'm eager to learn more !
I actually had saved this video on my watch list, will watch it sooner I guess :)

sdemeesterde:
Here is a version implementing your suggestion to use the backon crate.
Indeed, using Fibonacci with jitter is less aggressive while being robust !

(The "Decrusting the tokio crate" video was very interesting)

martin-core/src/tiles/mbtiles/source.rs
```diff
@@ -41,12 +57,34 @@ impl MbtSource {
             .map_err(|e| io::Error::other(format!("{e:?}: Cannot open file {}", path.display())))
             .map_err(|e| MbtilesError::IoError(e, path.clone()))?;
 
-        let meta = mbt
-            .get_metadata()
+        // Attempt to fetch metadata with Fibonacci backoff & jitter
+        // if the database is busy
```
CommanderStorm:
```suggestion
        // Attempt to fetch metadata
```

martin-core/src/tiles/mbtiles/source.rs
```diff
@@ -41,12 +57,34 @@ impl MbtSource {
...
+                );
+            })
             .await
-            .map_err(|e| MbtilesError::InvalidMetadata(e.to_string(), path.clone()))?;
+            .map_err(|_err| {
+                MbtilesError::InvalidMetadata(
+                    format!("SQLite still busy after {max_attempts} attempts"),
+                    path.clone(),
+                )
+            })?;
```
CommanderStorm: not nessesarily that it is locked.
Might also be another error

```suggestion
            .map_err(|e| MbtilesError::InvalidMetadata(e.to_string(), path.clone()))?;
```

CommanderStorm (approved):
Thank you!

martin-core/src/tiles/mbtiles/source.rs
```diff
@@ -41,12 +57,29 @@ impl MbtSource {
             .map_err(|e| io::Error::other(format!("{e:?}: Cannot open file {}", path.display())))
             .map_err(|e| MbtilesError::IoError(e, path.clone()))?;
 
-        let meta = mbt
-            .get_metadata()
+        // Attempt to fetch metadata with Fibonacci backoff & jitter
+        // Attempt to fetch metadata
```
CommanderStorm:
```suggestion
        // Attempt to fetch metadata with backoff
```

PR #2574, opened 2026-02-26
fix: restrict zooming and panning on data inspector

### LINKED ISSUE #2325: Restrict zooming and panning on the data inspector

manbhav234:
This PR fixes #2325

The TileJson is being fetched again to set the bounds and zoom values properly. The main map component has been shifted to a new component, wrapping it inside suspense to fetch the tileJson asynchronously. The tiles now load in the restricted area if the data is local.

One peculiar thing I noticed:
When setting the bounds, if the first value for west is -180, it leads to the following error:
Screenshot 2026-02-26 at 7 58 32 PM

If the value is less than -180 or you set it to -179 or less if it is -180, the error vanishes. I haven't looked too deep as to why this is happening. Maybe it's because of how setMaxBounds internally manipulates these values that lead to a NaN. But not too sure.

### DIFF
--- a/martin/martin-ui/src/components/dialogs/tile-inspect-map.tsx
+++ b/martin/martin-ui/src/components/dialogs/tile-inspect-map.tsx
@@ -1,0 +1,179 @@
+"use client";
+
+import { type ErrorInfo, useCallback, useId, useRef, useEffect } from "react";
+import { useAsyncOperation } from "@/hooks/use-async-operation";
+import { Layer, Map as MapLibreMap, Source } from "@vis.gl/react-maplibre";
+import { buildMartinUrl } from "@/lib/api";
+import type { MapRef } from "@vis.gl/react-maplibre";
+import MaplibreInspect from "@maplibre/maplibre-gl-inspect";
+import { Popup } from "maplibre-gl";
+import { TileSource } from "@/lib/types";
+import { ErrorBoundary } from "../error/error-boundary";
+import { Toaster } from "@/components/ui/toaster";
+import { useToast } from "@/hooks/use-toast";
+
+interface TileInspectDialogMapProps {
+  name: string;
+  source: TileSource;
+}
+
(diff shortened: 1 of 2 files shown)


martin/martin-ui/src/components/dialogs/tile-inspect-map.tsx
```diff
@@ -0,0 +1,155 @@
...
+      console.error('Map not found despite being initialized, this cannot happen');
+      return;
+    }
+    const map = mapRef.current.getMap();
+    const tileJson: TileJson = tileJsonOperation.data;
+    //Error when first value set to -180. it gets resolved on setting to -179 or lower. Other values work fine.
+    map.setMaxBounds([
+      [tileJson.bounds[0] === -180 ? -179 : tileJson.bounds[0], tileJson.bounds[1]],
+      [tileJson.bounds[2], tileJson.bounds[3]],
+    ]);
```
CommanderStorm: lets be a bit more fuzzy about this 😉

```suggestion
    if (tileJson.bounds) {
      const [west, south, east, north] = tileJson.bounds;
      const isWorld =
        west <= -179 &&
        east >= 179;
      if (!isWorld) {
          map.setMaxBounds([
            [west, south],
            [east, north],
          ]);
      }
    }
```

CommanderStorm (approved):
Nice work, thank you ❤️

PR #2576, opened 2026-02-28
feat: Migrate mdbooks to zensical

### LINKED ISSUE #2573: Migrating mdbook -> zenzical
Our docs currently are fairly barebones and without much visual appeal.

Comparing this with some of the other projects under the maplibre umbrella, I think we can do a lot better and make our live simpler as well:

| zensical | mdbook |
|--------|--------|
| Image | Image | 
| https://maplibre.org/maplibre-tile-spec/ | https://maplibre.org/martin/ | 

Migrating this should be fairly simple:
- change the config file from mdbook -> zensical
- fix file imports `#include`
- fix tabs
- fix admoniations (those `[!NOTE]`-things)

One of the things that we would need to figure out on this migration is if we can keep at least redirects to the new content.
If not, that would be unfortunate, since this means that old stackoverflow posts don't link well.
Since the zensical comes with a much better linking and search subsystem, I am not super concearned, but @nyurik likely will be a lot more picky on that one.

Maybe redirecting is easy and we can add something like this, but I would also be open to doing a breaking docs change to have much improved docs.

```html
<!DOCTYPE html>
<html>
  <head>
    <meta http-equiv="refresh" content="0;url=/new/" />
    <link rel="canonical" href="/new/" />
  </head>
</html>
```

manbhav234:
This pr resolves #2573 

This PR attempts to migrate the martin documentation from mdbooks to zensical. Following tasks needs to be completed for a successful migration : 

- [x] Fixing Navigation
- [x] Fixing File includes
- [x] Fixing Tabs
- [x] Fixing Admonitions
- [x] Fixing External Links by Redirection
- [x] Building and CI/CD changes
- [x] Remove book.toml
- [x] Docs README Update

With zensical's directory based navigation structure, the external links will break. I tested the doc links present in repo's README.md and they won't work. However, generating HTML snippets like the one mentioned in #2573 does lead to successful redirection.  We will have to do this for every file though. Or maybe there's a better way to achieve this, will have to see. 

For building and workflow changes, zensical is available as a python package and needs to be installed via `pip` and then built using `zensical build` command. Zensical generates a default `.github/workflows/docs.yml` but that is for hosting on github pages. I haven't commited it yet for the same reason.


CommanderStorm:
> For building and workflow changes, zensical is available as a python package and needs to be installed via pip and then built using `zensical build` command

I think `uvx zensical ...` (bacause of taking less time + `uv` is avaliable via `cargo-binstall` -> also fast) or a docker based apprach are sligntly better for contributors.
pip is just a little slow to do things via `pip` ^^

Here is how to do the docker based approach:

https://github.com/maplibre/maplibre-tile-spec/blob/04b83986715417f53ae2c7d4c96329a7961c8b61/justfile#L56-L60

CommanderStorm:
And stating the obvous: Nice work!

manbhav234:
I went with the docker based approach as it seemed more convenient to work with. Other tasks are also completed.

For redirects, I found a better way to deal with it. Zensical allows you to override pages it generates. Adding a javascript snippet for redirecting to `/` in  place of `.html` on the 404.html page does the trick.

CommanderStorm (approved):
Nice work reworking this 👍🏻 

I did minor changes (adding a favicon, ...) but the core was very solid!

Usually, these LGTM merges on larger refactorings rarely happen by newer-contributors. Very impressive.

PR #2584, opened 2026-03-03
docs: zensical and content updates

todtb:
- Adds `mbtiles/sql` snippets to base path
- Clarifies table sources to include views
- Fixes `edit_uri` for "Edit this page"


CommanderStorm (approved):
I am going to accept this for the spelling fixes and the edit_uri, but I am going to go with symlinking the files instead of the woraround you came up.. doing non-root files is likely a bit confusing in the long run

PR #2585, opened 2026-03-03
docs: change the way mbtiles schemas are included to sylink instead of config driven inclusion

CommanderStorm:
This does not quite work as I would expect. Not sure why it does not throw an error.

### DIFF
--- a/docs/content/mbtiles-schema.md
+++ b/docs/content/mbtiles-schema.md
@@ -7,7 +7,7 @@ The `mbtiles` tool builds on top of the original [MBTiles specification](https:/
 Flat schema is the closest to the original MBTiles specification. It stores all tiles in a single table. This schema is the most efficient when the tileset contains no duplicate tiles.
 
 ```sql
---8<-- "mbtiles/sql/init-flat.sql"
+--8<-- "files/init-flat.sql"
 ```
 
 ## flat-with-hash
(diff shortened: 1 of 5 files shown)


nyurik (approved):

docs/content/mbtiles-schema.md
```diff
@@ -7,23 +7,27 @@ The `mbtiles` tool builds on top of the original [MBTiles specification](https:/
...
---8<-- "init-flat-with-hash.sql"
+--8<-- "files/init-flat-with-hash.sql"
 ```
 
 ## normalized
 
 Normalized schema is the most efficient when the tileset contains duplicate tiles. It stores all tile blobs in the `images` table, and stores the tile Z,X,Y coordinates in a `map` table. The `map` table contains a `tile_id` column that is a foreign key to the `images` table. The `tile_id` column is a hash of the `tile_data` column, making it possible to both validate each individual tile like in the `flat-with-hash` schema, and also to optimize storage by storing each unique tile only once.
 
+```sql
+--8<-- "files/init-normalized.sql"
```
CommanderStorm:
```suggestion
--8<-- "files/init-normalized.sql:0:28"
```

docs/content/mbtiles-schema.md
```diff
@@ -7,23 +7,27 @@ The `mbtiles` tool builds on top of the original [MBTiles specification](https:/
...
+```
+
 Optionally, `.mbtiles` files with `normalized` schema can include a `tiles_with_hash` view. All `normalized` files created by the `mbtiles` tool will contain this view.
 
 ```sql
---8<-- "init-normalized.sql"
+--8<-- "files/init-normalized.sql:26:"
```
CommanderStorm:
```suggestion
--8<-- "files/init-normalized.sql:30:39"
```

PR #2599, opened 2026-03-13
fix: redirect ignoring `--route-prefix` for .pbf tile requests

### LINKED ISSUE #2596: 301 redirect ignores --route-prefix, breaking requests behind a reverse proxy
When using `--route-prefix` and a client requests a tile URL with a `.pbf` suffix, Martin returns a 301 redirect whose `Location` header does not include the route prefix. This causes the redirect to point to a broken URL when Martin is behind a reverse proxy.

**Version:** v1.3.1 (Docker)

**Steps to reproduce:**

1. Start Martin with `--route-prefix /geotile` behind a reverse proxy
2. Request a tile URL with a `.pbf` suffix, e.g. `/geotile/my_source/0/0/0.pbf`
3. Martin responds with a 301 redirect
4. The `Location `header points to `/my_source/0/0/0` — missing the `/geotile` prefix

**Expected behavior:** The `Location `header should include the prefix, e.g.` /geotile/my_source/0/0/0`

**Root cause (suspected):** The redirect logic introduced in #2528 does not account for `--route-prefix`.

**Workaround:** Avoid URLs with a `.pbf` suffix.

Paritoshcodes:
Fixes #2596

Ensure redirect URLs include the configured route prefix when handling
`.pbf` tile requests behind a reverse proxy.

Previously the redirect logic constructed URLs without the route prefix,
causing incorrect redirects when Martin was deployed behind a reverse proxy.

This change prepends the configured route prefix to the redirect
Location header.

A regression test was added to verify correct behavior.

### DIFF
--- a/martin/src/srv/tiles/content.rs
+++ b/martin/src/srv/tiles/content.rs
@@ -86,7 +86,15 @@ pub async fn redirect_tile_ext(req: HttpRequest, path: Path<RedirectTileRequest>
         })
         .await;
 
-    redirect_tile_with_query(ids, *z, *x, *y, req.query_string())
+    redirect_tile_with_query(
+        ids,
+        *z,
+        *x,
+        *y,
+        req.query_string(),
+        req.app_data::<Data<SrvConfig>>()
+            .and_then(|cfg| cfg.route_prefix.as_deref()),
+    )
 }
 
 /// Redirect `/tiles/{source_ids}/{z}/{x}/{y}` to `/{source_ids}/{z}/{x}/{y}` (HTTP 301)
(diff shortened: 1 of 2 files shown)


CommanderStorm (approved):

CommanderStorm (approved):
Thanks for the bugfix! 🎉

c-meier:
Shouldn't the same thing be done for the pluralization redirect handlers (styles→style, sprites→sprite, etc.) created by #2528 ?

And, if I understand the doc correctly, the `base_path` config parameter can also change the `Location` of the redirect for the tile api.

Paritoshcodes:
Hello @c-meier, that is a good point but my fix currently only handles the `.pbf` redirect path, but I agree the same logic may need to be applied to the pluralization redirect handlers as well. I can take a look at that after this PR is merged if needed.

PR #2603, opened 2026-03-14
chore(deps): Upgrade PBF font tools

### LINKED ISSUE #2591: martin fails to build on freebsd : pbf_font_tools v3.0.0
**SYSTEM**
FreeBSD dev_tile.jail 15.0-RELEASE-p4 FreeBSD 15.0-RELEASE-p4 GENERIC amd64

**REPRO**
```
cargo install cargo-binstall
cargo binstall -y martin
```

**ERROR**
```
error: failed to run custom build command for `pbf_font_tools v3.0.0`

Caused by:
  process didn't exit successfully: `/tmp/cargo-installidshM9/release/build/pbf_font_tools-3f0e8ef049b99022/build-script-build` (exit status: 101)
  --- stderr

  thread 'main' (101059) panicked at /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/pbf_font_tools-3.0.0/build.rs:16:48:
  Unable to find protoc!: Error { os: "freebsd", arch: "x86_64" }
warning: build failed, waiting for other jobs to finish...
error: failed to compile `martin v1.3.1`, intermediate artifacts can be found at `/tmp/cargo-installidshM9`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.
ERROR Cargo errored! ExitStatus(unix_wait_status(25856))
ERROR Fatal error:
  \x1b[31m×\x1b[0m For crate martin: subprocess /usr/local/bin/cargo install martin --version 1.3.1 --root /root/.cargo errored with exit status: 101
\x1b[31m  ╰─▶ \x1b[0msubprocess /usr/local/bin/cargo install martin --version 1.3.1 --root /root/.cargo errored with exit status: 101
```

ianthetechie:
This upgrades the pbf_font_tools dependency to v3.1.0. It also adds a FreeBSD-specific check to `just validate-tools`. The latest version of the crate will fall back to the system protoc if a vendored binary is unavailable. This should be fine for FreeBSD, since ports there are generally much faster to update (more like a homebrew than a Debian).

Other notes:

* I considered adding a feature flag to enable protoc-from-src, but that seems unnecessary and would add more complexity short-term (martin does not currently have any feature flags and I'm assuming you'd like to keep it that way).
* I couldn't find where GNU sed is required but the macOS deps specifically list that. If you DO really depend on GNU sed features, we should probably add that under FreeBSD too (the package name is `gsed`), since both FreeBSD and macOS (everything but Linux) use BSD sed which has different flags.
* The justfile probably needs some more... improvements later if we want to have wider platform support. Requiring docker-compose is a non-starter on many platforms (including FreeBSD).

Fixes #2591

### DIFF
--- a/justfile
+++ b/justfile
@@ -558,6 +558,15 @@ validate-tools:
         fi
     fi
 
+    # Check FreeBSD-specific tools
+    if [[ "$OSTYPE" == "freebsd"* ]]; then
+        # This should eventually go away if the upstream pbf_glyph_tools can vendor the source artifacts
+        # (no more need for protoc). Other platforms automatically install a vendored binary.
+        if ! command -v protoc >/dev/null 2>&1; then
+            missing_tools+=("protoc")
+        fi
+    fi
+
     # Report results
     if [[ ${#missing_tools[@]} -eq 0 ]]; then
         echo "✓ All required tools are installed"
@@ -565,6 +574,7 @@ validate-tools:
         echo "✗ Missing tools: ${missing_tools[*]}"
         echo "  Ubuntu/Debian: sudo apt install -y jq file curl grep sqlite3-tools gdal-bin"
         echo "  macOS: brew install jq file curl grep sqlite gdal gsed"
+        echo "  FreeBSD: pkg install jq curl sqlite3 gdal protobuf"
         echo ""
         exit 1
     fi
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2609,9 +2609,9 @@ checksum = "e6d5a32815ae3f33302d95fdcb2ce17862f8c65363dcfd29360480ba1001fc9c"
 
 [[package]]
 name = "futures"
-version = "0.3.31"
+version = "0.3.32"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "65bc07b1a8bc7c85c5f2e110c476c7389b4554ba72af57d8445ea63a576b0876"
+checksum = "8b147ee9d1f6d097cef9ce628cd2ee62288d963e16fb287bd9286455b241382d"
 dependencies = [
  "futures-channel",
  "futures-core",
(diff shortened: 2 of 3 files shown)


justfile
```diff
@@ -558,13 +558,23 @@ validate-tools:
         fi
     fi
 
+    # Check FreeBSD-specific tools
```
nyurik, line 561: note that this only affects local developers. This script is not used by cargo install

nyurik (approved):

justfile
ianthetechie: Yeah that's fine I think. If you've got a better idea/other places to document it that's cool; this seemed like the easiest at a glance. I think a proper port is also pretty easy, so local dev is probably the most important thing to optimize for at the moment.

ianthetechie: For posterity, if you don't run `just validate-tools and you try building without protoc around and obviously in the path you'll get the following message. This is a big improvement over the reported issue, as it contains a clear resolution path:

```
error: failed to run custom build command for `pbf_font_tools v3.1.0`

Caused by:
  process didn't exit successfully: `/usr/home/ianthetechie/martin/target/debug/build/pbf_font_tools-2bef1ad720127077/build-script-build` (exit status: 1)
  --- stderr
  Protobuf features: protoc-from-src: false, protoc-vendored: true
  Failed to find vendored protoc binary; trying to fall back to protoc in $PATH...
  Error: Custom { kind: NotFound, error: "Could not find `protoc`. If `protoc` is installed, try setting the `PROTOC` environment variable to the path of the `protoc` binary. Try installing `protobuf-compiler` or `protobuf` using your package manager. It is also available at https://github.com/protocolbuffers/protobuf/releases  For more information: https://docs.rs/prost-build/#sourcing-protoc" }
  ```

AFAIK the `cargo install` from crates.io output is the same as `cargo build` (and cargo binstall will fall back to plain install since there's no FreeBSD binary for Martin).

PR #2619, opened 2026-03-15
ci(perf): don't test pg twice

CommanderStorm:
Legacy, was not removed when we should have.
Clogs up the pipeline a bit making it slower than it needs to


martin/tests/pg_server_test.rs
CommanderStorm:
```suggestion
#![cfg(all(feature = "test-pg", not(feature = "fonts"), not(feature = "sprites"), not(feature = "styles")))]
```

nyurik (approved):

Issue #2614, opened 2026-03-15
Docs: Add font name escaping note to `run-with-nginx.md`


Emanuelmsjv:
### Problem
Users serving fonts through NGINX with Martin can run into HTTP 400 errors because font names contain spaces (e.g. `Open Sans Regular`) that NGINX does not escape by default. This is a known pain point (see #1115).

While the fonts page briefly mentions URL-escaping font names, `run-with-nginx.md` has no reference to this issue. Users following the NGINX setup guide have no indication this can happen or how to fix it.

### Proposed fix
Add a short troubleshooting note to `run-with-nginx.md` pointing to the font name escaping requirement, so users hitting 400 errors have a clear path to resolution without having to dig through discussions.

### Would like to work on this
Happy to submit a PR for this if the maintainers are okay with the approach.

CommanderStorm:
That discussion is talking about how the nginx.conf needs to look like.
I don't have good experience in this aspect.
If the nginx.conf is indeed still incorrect,  I would prefer to just fix the .conf instead of adding a note

Emanuelmsjv:
> That discussion is talking about how the nginx.conf needs to look like.
> I don't have good experience in this aspect.
> If the nginx.conf is indeed still incorrect,  I would prefer to just fix the .conf instead of adding a note

Thanks for the clarification! I will fix `.conf` directly other than adding a note

Looking at the current `run-with-nginx.md`, there's no font location block at all. I'd like to add one that handles font name escaping. Would a `location ~ /font/` block with proper URL encoding be the right direction, or do you have a preferred approach?

Emanuelmsjv:
Hey @CommanderStorm! Just wanted to follow up — I was thinking of adding a "Serving Fonts" section right after the "Rewriting URLs" block, with a location ~ /font/ proxy block similar to the tiles one. Does that sound like the right direction to you, or did you have something else in mind?

CommanderStorm:
I don't think that should be nessesary. Please check in ngimx what the actual minimal fix formthis should ne be

Emanuelmsjv:
"Thanks for the feedback! To keep the fix a little minimal, would it work for you if I added a brief sub-section under Rewriting URLs?

```nginx

location ~ /font/(?<fwd_path>.*) {
    proxy_pass http://martin:3000/font/$fwd_path$is_args$args;
}
```
This keeps the fix surgical without needing a major new section in the docs. Let me know if this looks good to you, and I'll update the PR!"

[...]

PR #2626, opened 2026-03-18
docs: add nginx location block for font name encoding

Emanuelmsjv:
**Fixes** #2614 
Font names with spaces (e.g. Open Sans Regular) cause HTTP 400 errors when proxied through NGINX, because NGINX decodes %20 back to a literal space before forwarding to Martin. 
- [x] This adds a minimal location ~ /font/ block to handle font requests correctly.

### DIFF
--- a/docs/content/run-with-nginx.md
+++ b/docs/content/run-with-nginx.md
@@ -23,6 +23,7 @@ location ~ /tiles/(?<fwd_path>.*) {
     proxy_pass        http://martin:3000/$fwd_path$is_args$args;
 }
 ```
+
 ## Serving Fonts
 
 If your font names contain spaces (e.g. `Open Sans Regular`), NGINX may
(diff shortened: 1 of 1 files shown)


CommanderStorm:
How is this different from the section above to where you are documenting? The regex and the proxy directive are the same..

Emanuelmsjv:
The key difference is the proxy_pass path — the tiles block passes `$fwd_path` directly, while this one prefixes `font/` to ensure font requests are routed correctly. Without it, font names with spaces cause HTTP 400 errors.

Do you want me to change any thing?

CommanderStorm:
Ok so why do you think this is the case?
Why would `/tiles/font/Foo Bar` cause an issue, but your version not?

They have the exact same `fwd_path` and `proxy_pass` directive..

Have you tested your assumptions that this actually fixes an issue that exists?
If yes, how?

Emanuelmsjv:
> Ok so why do you think this is the case?
> Why would `/tiles/font/Foo Bar` cause an issue, but your version not?
> 
> They have the exact same `fwd_path` and `proxy_pass` directive..
> 
> Have you tested your assumptions that this actually fixes an issue that exists?
> If yes, how?

I haven't tested this locally. I assumed a separate location block would handle font path routing differently, but I can see now the proxy logic is identical.

 Thus I'm sorry that I'm wasting your time without actual proof, ready to be guided if this issue still has any way to resolve.

PR #2629, opened 2026-03-18
feat: package martin as a freebsd-package

ToddG:
Create a FreeBSD Package

* make.sh generates a package versioned the same as martin, e.g

```
martin-VERSION.pkg
```

* the martin package can be installed with FreeBSD package manager

```
freebsd # pkg install martin-VERSION.pkg
```

* installing the martin package installs both the martin binaries and service scripts to start and stop martin

```
service martin start
service martin stop
```

Related issues:
* https://github.com/maplibre/martin/issues/2591#issuecomment-4059583718


freebsd/martin/gleam.toml
```diff
@@ -0,0 +1,24 @@
+name = "martin"
+version = "1.0.0"
+
+# Fill out these fields if you intend to generate HTML documentation or publish
+# your project to the Hex package manager.
+#
+# description = ""
+# licences = ["Apache-2.0"]
+# repository = { type = "github", user = "", repo = "" }
+# links = [{ title = "Website", href = "" }]
```
CommanderStorm:
```suggestion
description = "Blazing fast and lightweight PostGIS, MBtiles and PMtiles tile server, tile generation, and mbtiles tooling"
licences = ["Apache-2.0", "MIT"]
repository = { type = "github", user = "maplibre", repo = "martin" }
links = [{ title = "Website", href = "https://maplibre.org/martin" }]
```

freebsd/scripts/freebsd/+MANIFEST
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "martin",
+  "version": "1.0.0",
+  "origin": "devel/martin",
+  "comment": "vector tile server integrated with postgres",
+  "www": "https://github.com/ToddG/martin/tree/main",
```
CommanderStorm:
```suggestion
  "www": "https://github.com/maplibre/martin/tree/main",
```

freebsd/scripts/freebsd/+MANIFEST
```diff
@@ -0,0 +1,19 @@
...
+    "post-install": "PKG_NAME=\"martin\"\nCONFIG_DIR=\"/usr/local/etc/martin.d\"\nCONFIG_FILE=\"/usr/local/etc/martin.d/martin.env\"\n\n\nPKG_USER=\"martin\"\n\nif [ -n \"${PKG_ROOTDIR}\" ] && [ \"${PKG_ROOTDIR}\" != \"/\" ]; then\n  PW=\"/usr/sbin/pw -R ${PKG_ROOTDIR}\"\nelse\n  PW=/usr/sbin/pw\nfi\n\necho \"===> Creating user.\"\nif ! ${PW} groupshow ${PKG_USER} >/dev/null 2>&1; then\n  echo \"Group: '${PKG_USER}'.\"\n  ${PW} groupadd ${PKG_USER}\nelse\n  echo \"Using existing group: '${PKG_USER}'.\"\nfi\n\nif ! ${PW} usershow ${PKG_USER} >/dev/null 2>&1; then\n  echo \"User: '${PKG_USER}'.\"\n  ${PW} useradd ${PKG_USER} -g ${PKG_USER} -c \"${PKG_NAME} user\" -d /nonexistent -s /usr/sbin/nologin\nelse\n  echo \"Using existing user: '${PKG_USER}'.\"\nfi\n\n\nif [ ! -f $CONFIG_FILE ]\nthen\n  echo \"===> Creating config in ${CONFIG_FILE}\"\n  echo \"# martin config file\" > $CONFIG_FILE\n  echo 'FOO=\"bar\"' >> $CONFIG_FILE\n  chmod 0444 $CONFIG_FILE\nfi\n"
+  },
+  "deps": {}
```
CommanderStorm, line 14: might not be entirely correct.. not sure what dependencys we have if build the way we are build below

freebsd/build.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
+set -x
+ARTIFACT_DIR=/data/builds/martin
+
+echo "build martin binaries"
+cd ..
+cargo install martin --locked --no-default-features --features mbtiles,sprites,metrics,pmtiles,postgres,styles,webui
```
CommanderStorm: If building from source is common in FreeBSD, I would recomend the following.
This effectively allows rust to use the full CPU, but drops any cross-system portability guarantees.

```suggestion
RUSTFLAGS="-Ctarget-cpu=native" cargo install martin --locked --no-default-features --features mbtiles,sprites,metrics,pmtiles,postgres,styles,webui
```

freebsd/martin/.github/workflows/test.yml
CommanderStorm, line 1: I don't think CI being at this level works.

-> Please move this two levels up into the `.github` folder, similar how rust dependency linting is prioritised.

CommanderStorm:
> Here's what I need to move this forward:
>
> - instructions on how to build a rust project in freebsd ports
> - instructions on how to build pbf_font_tools,https://github.com/maplibre/martin/issues/2591#issuecomment-4059583718

Maybe @ianthetechie can comment on the first aspect.

The second aspect will resolve itsself once we bump our depenencys to include https://github.com/stadiamaps/sdf_font_tools/pull/30 (thanks @nyurik).

freebsd/martin/gleam.toml
```diff
@@ -0,0 +1,24 @@
+name = "martin"
+version = "1.0.0"
```
CommanderStorm: please register this right here, to take advantage of not having to manually bump this version

https://github.com/maplibre/martin/blob/a99f7a978c8ede00cb20d6002ded4ca86f82b112/.pre-commit-config.yaml#L51

(also applies in the other place)

nyurik:
@CommanderStorm we need to bump the lock file to make sure latest dep that does not need protoc is used

CommanderStorm:
> will resolve itsself once we bump our depenencys

I am aware.

freebsd/martin/.github/workflows/test.yml
ToddG: That was cruft from how I created the package metadata. Deleted.

freebsd/scripts/freebsd/+MANIFEST
ToddG: `deps` here are for runtime dependencies.

Let me know what runtime dependencies martin has and I'll update. However,  I thought it was self contained?

freebsd/martin/gleam.toml
ToddG: Thx! I've removed some cruft and updated the `make.sh` script to label the generated freebsd packages with the version from martin/Cargo.toml.

freebsd/scripts/freebsd/+MANIFEST
```diff
@@ -0,0 +1,19 @@
...
+    "post-install": "PKG_NAME=\"martin\"\nCONFIG_DIR=\"/usr/local/etc/martin.d\"\nCONFIG_FILE=\"/usr/local/etc/martin.d/martin.env\"\n\n\nPKG_USER=\"martin\"\n\nif [ -n \"${PKG_ROOTDIR}\" ] && [ \"${PKG_ROOTDIR}\" != \"/\" ]; then\n  PW=\"/usr/sbin/pw -R ${PKG_ROOTDIR}\"\nelse\n  PW=/usr/sbin/pw\nfi\n\necho \"===> Creating user.\"\nif ! ${PW} groupshow ${PKG_USER} >/dev/null 2>&1; then\n  echo \"Group: '${PKG_USER}'.\"\n  ${PW} groupadd ${PKG_USER}\nelse\n  echo \"Using existing group: '${PKG_USER}'.\"\nfi\n\nif ! ${PW} usershow ${PKG_USER} >/dev/null 2>&1; then\n  echo \"User: '${PKG_USER}'.\"\n  ${PW} useradd ${PKG_USER} -g ${PKG_USER} -c \"${PKG_NAME} user\" -d /nonexistent -s /usr/sbin/nologin\nelse\n  echo \"Using existing user: '${PKG_USER}'.\"\nfi\n\n\nif [ ! -f $CONFIG_FILE ]\nthen\n  echo \"===> Creating config in ${CONFIG_FILE}\"\n  echo \"# martin config file\" > $CONFIG_FILE\n  echo 'FOO=\"bar\"' >> $CONFIG_FILE\n  chmod 0444 $CONFIG_FILE\nfi\n"
```
CommanderStorm, line 12: is this idomatic? It seems hard to maintain..

CommanderStorm (approved):
If this unblocks you I would be happy to merge, would appreciate a review by @ianthetechie though if there are some things that are not so best practice

freebsd/scripts/freebsd/+MANIFEST
nyurik: `$CONFIG_FILE\n  echo 'FOO=\"bar\"'`  does not look very idiomatic tbh :)

ToddG: Yes, the foo=bar should be removed :-).

As for the +MANIFEST file itself... there's two ways to go:
1. Generate the file instead of storing it in the source tree. This means bringing in the tooling that generates it. I'll create a separate branch and show you what that looks like. 
2. Add the manifest to the repo more-or-less as-is (minus foo=bar)

fwiw I prefer option 1

nyurik: I agree - this is non-maintainable, and thus unmergable.  I'm ok to have this as a standalone file that gets converted during packaging.  The simplest method is actually to use jq commands in the justfile and three files:

```
packages/
    debian/   <--  we should probably move the top level debian dir here
    freebsd/
        manifest.json   <-- same content as MANIFEST but without the scripts
        pre-deinstall  <-- content of the individual script
        post-install
        ...
```

in justfile:

```justfile
package-freebsd:
  jq \
  --rawfile pre_deinstall packages/freebsd/pre-deinstall \
  --rawfile post_install packages/freebsd/post-install \
  '. + {
    "pre-deinstall": $pre_deinstall,
    "post-install": $post_install
  }' packages/freebsd/manifest.json > +MANIFEST
```

freebsd/scripts/freebsd/+MANIFEST
nyurik: openssl is still in the lock file -- so despite some efforts to migrate to RusTLS, it might still be needed

ianthetechie (commented):
Hey, thanks for the tag!

I guess my only question is whether having all of this is the Martin source tree really makes things easier. There isn't anything particularly special required to compile Martin anymore besides a Rust compiler and (probably?) freetype2.

To me it feels like most of these files _probably_ make more sense in the FreeBSD ports tree long term, and this is mostly just reinventing the existing ports tree architecture. I could take a swing at a "real" port next week.

I want to emphasize that this is just my own loosely held opinion. And it's mostly based on the fact that FreeBSD users prefer ports + packages from the main tree whenever possible. It's cleaner, works with most jail management tools seamlessly, has an update mechanism, etc... Ports also has conventions for including sample configs, and I think that's where _most_ FreeBSD users would expect to find stuff (the martin docs can just reference that).

The main argument _for_ including packaging tooling here is that it would be faster to update than submitting a PR to ports I guess ;) I also know of a few other projects that publish package on their own (e.g. continuwuity, a Matrix server), but this is almost always regarded as a temporary measure from what I've seen.

Finally, I was about to write that CI was another reason to not do this, as it'd just get stale without one of the maintainers running it on FreeBSD (@nyurik joked to me the other day that if it's not exercised by CI, it doesn't exist)... but then I decided to look into this a bit more and it turns out that my presupposition was wrong 😂 You _can_ run FreeBSD CI jobs on GitHub Actions! It's obviously a bit slower since it runs in a QEMU VM, but it's quite serviceable. I just took a swing at it for the SDF crate in https://github.com/stadiamaps/sdf_font_tools/pull/31. I'd be happy to whip up something similar for Martin once you all decide on how much should actually be in the source tree.

TL;DR, I do think build instructions and/or justfile recipes to ensure the FreeBSD prerequisites are installed. But I think the configuration, package manifest, etc. are probably best to put in a "proper" port.

freebsd/scripts/freebsd/+MANIFEST
ianthetechie: Fun fact: this is in the base system in FreeBSD. No external dependency needed :)

ToddG:
@ianthetechie Putting all this in the FreeBSD ports tree sounds like a great idea. Please tag me on that so I can learn from the process. I have several other projects I'd like to put in the ports tree.

ToddG:
Closing, per comments above.

PR #2642, opened 2026-03-22
fix: center style previews according to center/zoom from style JSON

### LINKED ISSUE #2569: Style previews in the Web UI not Centered according to `center` or `zoom` propertys in the style
The style previes in the styles section of the web UI do not follow the center: or zoom: properties given from the style json. However the style editor (maputnik) does.

mvanhorn:
## Summary

Style previews in the Styles Catalog now respect the `center` and `zoom`
properties from the style JSON, matching the behavior of the style editor
(Maputnik).

## Changes

**`martin/martin-ui/src/components/catalogs/styles.tsx`**

Extracted a `StylePreviewMap` component that manages its own view state
per card. On initial render, no explicit latitude/longitude/zoom is passed
to `MapLibreMap`, so react-maplibre uses whatever center/zoom the style
JSON specifies. After the user interacts with a map preview, the component
tracks that specific map's position independently.

Previously, all previews shared a single hardcoded `viewState`
(`latitude: 53, longitude: 9, zoom: 2`) which overrode the style JSON.

- Type checks pass (`tsc --noEmit`)
- Linting passes (`biome check`)

Fixes #2569

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/martin/martin-ui/src/components/catalogs/styles.tsx
+++ b/martin/martin-ui/src/components/catalogs/styles.tsx
@@ -10,31 +10,13 @@ import { TooltipCopyText } from '@/components/ui/tooltip-copy-text';
 import { buildMartinUrl } from '@/lib/api';
 import type { Style } from '@/lib/types';
 import 'maplibre-gl/dist/maplibre-gl.css';
-import {
-  FullscreenControl,
-  Map as MapLibreMap,
-  type ViewStateChangeEvent,
-} from '@vis.gl/react-maplibre';
-import { useCallback, useState } from 'react';
+import { FullscreenControl, Map as MapLibreMap } from '@vis.gl/react-maplibre';
 
 function StylePreviewMap({ styleName }: { styleName: string }) {
-  const [viewState, setViewState] = useState<{
-    latitude: number;
-    longitude: number;
-    zoom: number;
-  } | null>(null);
-
-  const handleMove = useCallback((evt: ViewStateChangeEvent) => {
-    const { latitude, longitude, zoom } = evt.viewState;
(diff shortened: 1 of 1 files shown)


martin/martin-ui/src/components/catalogs/styles.tsx
```diff
@@ -10,8 +10,43 @@ import { TooltipCopyText } from '@/components/ui/tooltip-copy-text';
...
+
+  const handleMove = useCallback((evt: ViewStateChangeEvent) => {
+    const { latitude, longitude, zoom } = evt.viewState;
+    setViewState({ latitude, longitude, zoom });
+  }, []);
+
+  return (
+    <MapLibreMap
+      reuseMaps
+      {...(viewState ?? {})}
```
CommanderStorm: why id you add the `handleMove` callback and the `viewState`?

This seems like a very strange design decision..
Yes, removing the useState syncing ALL maps is likely a good idea, but then lets do that..

CommanderStorm:
@mvanhorn friendly ping ^^

martin/martin-ui/src/components/catalogs/styles.tsx
mvanhorn: You're right, that was overcomplicating it. Removed both in a2246a5. The component is now a pass-through to MapLibreMap so each map manages its own view state. Verified: styles.test.tsx (22 tests) pass, tsc + biome clean, vite build green.

mvanhorn:
Sorry for the delay on this. Dropped the handleMove/viewState in a2246a5, simpler approach as you suggested.

CommanderStorm (approved):
Thank you!

Issue #1684, opened 2025-02-10
Martin serving multiple database connections


darrenpierre90:
Hi, I'm attempting to use Martin to serve from multiple Postgres databases. To do that I'm trying to setup Martin config where we can set up various 2 but potentially more Postgres databases 

what have I tried so far? 

I've tried adding two connection strings and it seems the second connection string overwrites the first one.

I've put a connection string in an array, but that also doesn't work. 

in the Martin docs , it says 

> # Database configuration. This can also be a list of PG configs.

But confused about how to set it up . 

Any Help would be appreciative.

CommanderStorm:
This might be badly documented or we might have a bug.

We mean this like here:

https://github.com/maplibre/martin/blob/2d3e4d0f37bab0409c37c3c49d611284d5758526/martin/src/pg/config.rs#L196-L198

=> a list of `postgres`-configurations, not a list of `connection_string`s

Could you recheck if you have 2 configs? 

If the issue is what I think it is, how would you have understood the docs how We meant it:
Would the following be better
```yaml
# Database configuration.
# 
# This can also be a list of postgres configurations (i.e. a list of the dicutionary shown below)
```

darrenpierre90:
<img width="1387" alt="Image" src="https://github.com/user-attachments/assets/497aa6e2-6ff4-4e6c-bb65-d03dff2367e2" />

I've also tried that variant and I got the following error : 
<img width="1432" alt="Image" src="https://github.com/user-attachments/assets/13bb006e-5bd7-470c-815f-614405757098" />

darrenpierre90:
I think if there was just an example shown , it would make the documentation better

nyurik:
this is not valid yaml.

```yaml
listen_addresses: '0.0.0.0:3000'
postgres:
  - connection_string: 'postgresql://server1'
    default_srid: 4326
    auto_bounds: skip
  - connection_string: 'postgresql://server2'
    default_srid: 4326
    auto_bounds: skip
```

nyurik:
Any documentation improvement PRs are welcome ;)

darrenpierre90:
Great thanks it works now thank you @nyurik @CommanderStorm

Issue #1289, opened 2024-04-08
HowTo Provide Multiple PG Connections in Config


michaelmink:
Hey, that is probably a quick one: I have the requirement on my side to handle table source from multiple db connections. 

Is this supported? If yes, then how can I add those into my config.yaml? And please add some hints in the example config file.

Thanks a lot in advance!

Michael

sharkAndshark:
Hi @veschberwilli 
```yaml
keep_alive: 75

listen_addresses: '0.0.0.0:3000'

worker_processes: 8

cache_size_mb: 8
# Database configuration. This can also be a list of PG configs.
postgres:
  - connection_string:  postgres://postgres:postgres@localhost:5411/db
    default_srid: 4326
  - connection_string:  postgres://postgres:postgres@localhost:5422/db
    default_srid: 4326
```

nyurik:
@veschberwilli there are two answers to your question:
* if you have **different** data in two different databases, martin can publish them as two separate sources, just like @sharkAndshark showed above (thx!)
* if you have **identical** data in two different databases, i.e. mirrors / replicas, and simply want to speed up tile serving because of performance or stability concerns, then no, martin does not support load balancing between multiple databases.  You can still get the same result if you configure one of the standard high availability load balancers, and point Martin to it instead of going directly to the database. A quick googling gave me [this blog post](https://severalnines.com/blog/comparing-load-balancers-postgresql/) that mentions multiple load balancers.

sharkAndshark:
> @veschberwilli there are two answers to your question:
> 
> * if you have **different** data in two different databases, martin can publish them as two separate sources, just like @sharkAndshark showed above (thx!)
> 
> * if you have **identical** data in two different databases, i.e. mirrors / replicas, and simply want to speed up tile serving because of performance or stability concerns, then no, martin does not support load balancing between multiple databases.  You can still get the same result if you configure one of the standard high availability load balancers, and point Martin to it instead of going directly to the database. A quick googling gave me [this blog post](https://severalnines.com/blog/comparing-load-balancers-postgresql/) that mentions multiple load balancers.

Maybe we could document this under the [3.1 PostgreSQL Connection](https://maplibre.org/martin/pg-connections.html) though user could find a clue in the description about configuration file.

nyurik:
Sure, would be great, thanks!

michaelmink:
Hey @sharkAndshark and @nyurik ,
thanks a lot for the quick response and for proviing the code snippet for the config.

[...]

PR #2644, opened 2026-03-24
docs: document multiple databases

maltaesousa:
I was wondering how to list multiple databases and found that it was already asked here #1289 and here #1684


nyurik (approved):
thanks!

PR #2648, opened 2026-03-29
test: make sure our unit tests run under macos too

Weixing-Zhang:
Use libm for `sinh` and `atan` in `webmercator_to_wgs84` conversion, since transcendental functions can produce slightly different results across platforms.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,13 +36,20 @@ jobs:
       - run: just ui::test
       - run: just ui::type-check
   unit-test-rust:
-    name: Unit test the server
+    name: Unit test the server (${{ matrix.os }})
     permissions:
       contents: read
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        # 1. Linux-only for PRs (faster feedback)
+        # 2. Run on all three OS platforms when merging to main 
+        os: ${{ fromJSON('["ubuntu-latest", "macos-latest"]') || (github.ref == 'refs/heads/main' && fromJSON('["ubuntu-latest", "macos-latest", "windows-latest"]')) }}
     steps:
       - run: rustup update stable && rustup default stable
       - uses: BRAINSia/free-disk-space@7048ffbf50819342ac964ef3998a51c2564a8a75 # v2.1.3
+        if: runner.os == 'Linux'
         with:
           tool-cache: false
           mandb:   true
@@ -55,6 +62,10 @@ jobs:
       - name: Checkout sources
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with: { persist-credentials: false }
+      - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
+        if: github.event_name != 'release' && github.event_name != 'workflow_dispatch'
+        with:
+          key: ${{ matrix.os }}
       - uses: taiki-e/install-action@a57ddfbcd928cf9406a78dcc300dacc5d367a547 # v2.68.31
         with: { tool: 'just' }
       - name: Run cargo test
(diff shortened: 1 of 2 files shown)


CommanderStorm (commented):
How did you come across this? How are you triggering this?

I am specifically asking because internally rust is also using libm for all supported platforms -> https://doc.rust-lang.org/src/std/sys/cmath.rs.html#3

Weixing-Zhang:
@CommanderStorm 

From what I understand, Rust std does not always use a pure Rust implementation for transcendental functions. The code in Rust standard library shows these are declared as foreign C functions, typically resolved to the system’s math library - libm.

I ran into this when running tests on macOS — I can consistently reproduce (this [commit](https://github.com/Weixing-Zhang/martin/commit/7e6e0436f238abab82add9622a240150357d6bf6)) the mismatch locally and in GitHub CI on macOS:
https://github.com/Weixing-Zhang/martin/actions/runs/23728260972/job/69116176976#:~:text=1230-,%2D%2D%2D%2D%20tiles%3A%3Acog%3A%3Asource%3A%3Atests%3A%3Acan_generate_tilejson_from_source%3A%3Acase_2%20stdout%20%2D%2D%2D%2D,right%3A%20%22%2D121.34948730468746%2C41.96357478222515%2C%2D121.34399414062497%2C41.97174336327965%22,-1243

So it seems the platform math implementation can lead to slightly different results across environments.

CommanderStorm:
Okay..
Lets do the following:

- add macos and windows to CI as a matrix for our unit tests
- try if we can just relax the requirement on this math operation to `2 * f32::EPSILON` (which is still very tiny) is enough

If relaxing the requirement does not work we can overrride the compilers choice and use the fallback library directly (rust `libm` is only used if no system level special functions are avaliable it seems).

nyurik:
there must be a reason why Rust stdlib chose not to use rust based code, and most likely its performance.. floating point precision on the other hand is one of those hard to chase aspects... Should we perhaps just make tests less exact?

Weixing-Zhang:
I totally agree. I have considered relaxing the precision, since keeping 16-digit precision isn’t necessary. On the other hand, I also think making a very small trade-off in performance for determinism could be worthwhile.

Based on the suggestions above, I will do the following:
- Relax the matching precision
- Make all unit tests on Linux, macOS, and Windows for every merge to main

.github/workflows/ci.yml
```diff
@@ -36,13 +36,20 @@ jobs:
...
     permissions:
       contents: read
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        # 1. Linux-only for PRs (faster feedback)
+        # 2. Run on all three OS platforms when merging to main 
+        os: ${{ fromJSON('["ubuntu-latest", "macos-latest"]') || (github.ref == 'refs/heads/main' && fromJSON('["ubuntu-latest", "macos-latest", "windows-latest"]')) }}
```
CommanderStorm, line 50: fmt fix for pre-commit

```suggestion
    runs-on: ${{ matrix.os }}
    strategy:
      fail-fast: false
      matrix:
        # 1. Linux-only for PRs (faster feedback)
        # 2. Run on all three OS platforms when merging to main
        os: ${{ fromJSON('["ubuntu-latest", "macos-latest"]') || (github.ref == 'refs/heads/main' && fromJSON('["ubuntu-latest", "macos-latest", "windows-latest"]')) }}
```

.github/workflows/ci.yml
```diff
@@ -36,13 +36,20 @@ jobs:
       - run: just ui::test
       - run: just ui::type-check
   unit-test-rust:
-    name: Unit test the server
+    name: Unit test the server (${{ matrix.os }})
     permissions:
       contents: read
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
```
CommanderStorm, line 42: Since we added the benchmarking job, we need to push this one level down

```suggestion
    runs-on: ${{ matrix.os }}
    needs:
      - lint-rust # not because it is needed, but because we want less concurrent jobs competing
```

CommanderStorm (approved):

PR #2649, opened 2026-03-30
docs: update outdated links and commands

Apolo151:
Two minor documentation fixes I discovered while setting up my development env on ubuntu.

### Updates

- Fix Debian release link: the [old link](https://github.com/maplibre/martin/releases/latest/download/martin-Debian-x86_64.deb) returns 404. 
    - I think It would be better to name the debian package in upcoming release `martin-x86_64-debian.deb` so that it is consistent with other pkgs names. I can look into the packaging setup and work on it if that sounds fine.
- Replace the probably outdated `just book` command with the new `just docs` to run docs locally.
    - currently running `just book` returns this error `error: Justfile does not contain recipe book`

### DIFF
--- a/docs/content/development/index.md
+++ b/docs/content/development/index.md
@@ -155,7 +155,7 @@ just run         # start Martin server
 just test        # run all tests
 just fmt         # format code
 just clippy      # lint code
-just book        # build documentation
+just docs        # serve documentation preview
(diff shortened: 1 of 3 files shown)


CommanderStorm (approved):
Thank you

PR #2650, opened 2026-03-31
feat: sprite SDF preview

### LINKED ISSUE #1984: Add a button to showcase sdf based sprites
We currently in the frontend have these two views:

ImageImage

I think we should have an button to switch to sdf based sprites and to showcase the effect of `halo` + `size` + (maybe) `color` parameters.

Example of the tiny-sdf library:
https://mapbox.github.io/tiny-sdf/

Implementation wise, I think this should be done like this:
- [ ] make the size of sprites on the preview page adjustable via a slider
- [ ] add the [`tiny-sdf`](https://github.com/mapbox/tiny-sdf) library
- [ ] add support for displaying sdf sprites via switching between sdf and non-sdf sprites
- [ ] (flex-goal) add the `halo` parameter to sdfs and maybe too for regular sprites
- [ ] (flex-goal) add the ability to runtime-recolor them

gambhirsharma:
Fixes: #1984 

**Summary**

Add an SDF preview mode to the sprite dialog that renders icons via MapLibre GL, allowing users to visualize SDF sprites with dynamic icon color, halo color/width/blur, and scale matching real map renderer behavior.
Introduce a toolbar with a PNG/SDF toggle, size/scale sliders, and color pickers so users can interactively inspect sprites without leaving the Martin UI.

Parameters added for SDF preview: 
- [x] Icon color
- [x] Icon size
- [x] Halo color
- [x] Halo width
- [x] Halo blur

**ScreenShot** 

https://github.com/user-attachments/assets/c0a21084-75c9-4efe-b230-9f64d0781e79

### DIFF
--- a/martin/martin-ui/src/components/dialogs/sprite-preview.tsx
+++ b/martin/martin-ui/src/components/dialogs/sprite-preview.tsx
@@ -1,6 +1,7 @@
 import { Download } from 'lucide-react';
-import { Suspense } from 'react';
+import { Suspense, useId, useState } from 'react';
 import { LoadingSpinner } from '@/components/loading/loading-spinner';
+import { Badge } from '@/components/ui/badge';
 import { Button } from '@/components/ui/button';
 import {
   Dialog,
@@ -9,7 +10,9 @@ import {
   DialogHeader,
   DialogTitle,
 } from '@/components/ui/dialog';
+import { Switch } from '@/components/ui/switch';
 import type { SpriteCollection } from '@/lib/types';
+import { cn } from '@/lib/utils';
 import { SpritePreview } from '../sprite/SpritePreview';
 
 interface SpritePreviewDialogProps {
@@ -19,18 +22,43 @@ interface SpritePreviewDialogProps {
   onDownloadAction: (sprite: SpriteCollection) => void;
 }
 
+const SIZE_MIN = 16;
+const SIZE_MAX = 128;
+const SIZE_DEFAULT = 80;
+
+const SDF_SCALE_MIN = 0.5;
+const SDF_SCALE_MAX = 5;
+const SDF_SCALE_DEFAULT = 2;
+
+const HALO_DEFAULT = 0;
+const HALO_MAX = 10;
+const HALO_BLUR_DEFAULT = 0;
+const HALO_BLUR_MAX = 10;
+
 export function SpritePreviewDialog({
   name,
   sprite,
   onDownloadAction,
   onCloseAction,
 }: SpritePreviewDialogProps) {
+  const uid = useId();
+  const iconColorId = `${uid}-icon-color`;
+  const haloColorId = `${uid}-halo-color`;
+
+  const [sdfMode, setSdfMode] = useState(false);
+  const [displaySize, setDisplaySize] = useState(SIZE_DEFAULT);
+  const [sdfScale, setSdfScale] = useState(SDF_SCALE_DEFAULT);
+  const [iconColor, setIconColor] = useState('#1a1a2e');
+  const [haloColor, setHaloColor] = useState('#ffffff');
+  const [haloWidth, setHaloWidth] = useState(HALO_DEFAULT);
+  const [haloBlur, setHaloBlur] = useState(HALO_BLUR_DEFAULT);
+
   return (
     <Dialog onOpenChange={(v) => !v && onCloseAction()} open={true}>
       <DialogContent className="max-w-4xl w-full p-6 max-h-[80vh] overflow-auto">
         {sprite && (
           <>
-            <DialogHeader className="mb-6 truncate">
+            <DialogHeader className="mb-4 truncate">
               <DialogTitle className="text-2xl flex gap-4">{name}</DialogTitle>
               <DialogDescription>
                 <span>Preview the selected sprite.</span>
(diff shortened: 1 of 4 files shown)


gambhirsharma:
Hi @CommanderStorm, it's ready for review. Thanks

martin/martin-ui/src/components/sprite/SdfMapPreview.tsx
```diff
@@ -0,0 +1,215 @@
...
+    const name = queryIconAtPoint(map, point);
+    container.style.cursor = name ? 'pointer' : '';
+  }, []);
+
+  useEffect(() => {
+    if (!containerRef.current || spriteIds.length === 0) return;
+
+    const sdfSpriteUrl = buildMartinUrl(spriteUrl.replace('/sprite/', '/sdf_sprite/'));
+
+    const map = new maplibregl.Map({
```
CommanderStorm, line 117: lets talk about the basic approach first:
- why maplibre and not the tinysdf library?
- why maplibre and not the react wrapper?

martin/martin-ui/src/components/dialogs/sprite-preview.tsx
```diff
@@ -41,7 +69,161 @@ export function SpritePreviewDialog({
                 </Button>
               </DialogDescription>
             </DialogHeader>
-            <div className="pace-y-4 bg-gray-50 rounded-lg text-gray-900">
+
+            {/* Toolbar */}
```
CommanderStorm, line 73: at the point where we need section headers, please move this to a new component

martin/martin-ui/src/components/dialogs/sprite-preview.tsx
```diff
@@ -41,7 +69,161 @@ export function SpritePreviewDialog({
...
+            <div className="flex flex-nowrap items-center gap-4 mb-4 p-3 rounded-lg border bg-muted/40 overflow-x-auto">
+              <div className="flex shrink-0 items-center gap-2">
+                {!sdfMode ? (
```
CommanderStorm, line 76: please make only the things dynamic that need to be dynamic.
Otherwise this code is harder to parse than nessesary.

martin/martin-ui/src/components/dialogs/sprite-preview.tsx
CommanderStorm, line 73: if we need section headers that is usually a good indicator that the split into components needs to be improved

martin/martin-ui/src/components/dialogs/sprite-preview.tsx
```diff
@@ -41,7 +69,161 @@ export function SpritePreviewDialog({
...
+                ) : (
+                  <span className="text-sm font-medium text-muted-foreground select-none px-2 py-0.5">
+                    PNG
+                  </span>
+                )}
+                <Switch
+                  aria-label="Toggle SDF mode"
+                  checked={sdfMode}
+                  onCheckedChange={setSdfMode}
+                />
```
CommanderStorm, line 92: this seems to be a "poor mans" `<Tab />` -> lets use https://ui.shadcn.com/docs/components/radix/tabs instead, or?

martin/martin-ui/src/components/dialogs/sprite-preview.tsx
```diff
@@ -41,7 +69,161 @@ export function SpritePreviewDialog({
...
+                ) : (
+                  <span className="text-sm font-medium text-muted-foreground select-none px-2 py-0.5">
+                    SDF
+                  </span>
+                )}
+              </div>
+
+              <div className="h-5 w-px bg-border shrink-0" />
+
+              <div className="flex shrink-0 items-center gap-2 min-w-[160px]">
```
CommanderStorm, line 109: currently, you need to scroll to see all options.
I don't like this, since it is not really nessesary.

Lets flex-cpllumn instead for each of the settings

martin/martin-ui/src/components/dialogs/sprite-preview.tsx
```diff
@@ -41,7 +69,161 @@ export function SpritePreviewDialog({
...
+                {sdfMode ? (
```
CommanderStorm, line 110: please factor the entire sdf vs png settings out to different components instead of doing it like this.

The current apprach is not entirely obvious in what it does..

martin/martin-ui/src/components/sprite/SdfMapPreview.tsx
CommanderStorm, line 1: I am not sure if this is the "best" approach as noted earlier today.
I would like to know your rational why you chose this

martin/martin-ui/src/components/dialogs/sprite-preview.tsx
```diff
@@ -51,6 +233,13 @@ export function SpritePreviewDialog({
               >
                 <SpritePreview
                   className="w-full grid grid-cols-2 sm:grid-cols-4 md:grid-cols-6 gap-4"
+                  displaySize={displaySize}
+                  haloBlur={haloBlur}
+                  haloColor={haloColor}
+                  haloWidth={haloWidth}
+                  iconColor={iconColor}
+                  iconSize={sdfScale}
+                  sdfMode={sdfMode}
```
CommanderStorm, line 242: I don't think this looks like good design.
Since we only the sdf code here, lets move this to a different component instead of `<SpritePreview`

martin/martin-ui/src/components/sprite/SpriteCanvas.tsx
```diff
@@ -9,44 +9,73 @@ type SpriteCanvasProps = {
...
+}: SpriteCanvasProps) => {
   const canvasRef = useRef<HTMLCanvasElement>(null);
-  // not using copied since on-click the tooltip closes
   const { copy } = useCopyToClipboard({
     successMessage: `Sprite ID "${label}" copied to clipboard`,
   });
 
   const handleClick = () => copy(label);
 
+  const dpr = typeof window !== 'undefined' ? window.devicePixelRatio || 1 : 1;
```
CommanderStorm, line 29: please use this instead

```suggestion
  const dpr = window?.devicePixelRatio ?? 1;
```

CommanderStorm (changes requested):
Thanks for the PR.

here is the requested review that goes above the already voiced questions above

martin/martin-ui/src/components/sprite/SdfMapPreview.tsx
gambhirsharma: > why maplibre and not the tinysdf library?

TinySDF is not designed to generate SDF sprite icons. It can generate SDF glyphs/fonts. 

> TinySDF is a tiny and fast JavaScript library for generating SDF (signed distance field) from system fonts on the browser using Canvas 2D and [Felzenszwalb/Huttenlocher distance transform](https://cs.brown.edu/~pff/papers/dt-final.pdf). This is very useful for [rendering text with WebGL](https://www.mapbox.com/blog/text-signed-distance-fields/).

The Project description didn't say it supports generating sprite icons. Also, I got this issue that states that they don't support Sprite. [Tiny image-based SDFs](https://github.com/mapbox/tiny-sdf/issues/5)

In https://mapbox.github.io/tiny-sdf/, the SDF font (glyphs) rendering is done using the HTML <canvas> element. [tiny-SDF:259](https://github.com/mapbox/tiny-sdf/blob/51c177e226f1d6f35eeab86e2789cb693fbc6816/index.html#L259)

So, I concluded that `tiny-sdf` is a library used to generate SDF from the system font. It's not used to generate PNG/SVG Sprite into SDF, and it's also not for rendering SDF too.  


> why maplibre and not the react wrapper?

Now talking about this, I hadn’t really explored its React wrapper or checked whether it supported all the features and options I need. Instead, I went with the vanilla MapLibre lib since it gave me more direct control over core function like disabling interactivity and other low-level behaviours.

My basic Idea was to: 
Get the SDF image from `/sdf_sprite/{spriteID}` --> Serve it WebGL using Maplibre ( pass on the change properties). 

Thanks for the review and for making the PR Draft. I will look into Maplibre React Rapper component and make a better decision on what to choose. 

Also, a few additional improvements came to mind while going through the review. I will work on those as well.

Thanks

CommanderStorm:
PR seems a bit dead.
Lets close it.

I would love to see this pushed over the line if you want to continue this work

PR #2661, opened 2026-04-06
chore: introduce `TileSourceManager` and `ReloadAdvisory`

Auspicus:
Lay the foundation for live-reload / decentralized-reload by replacing the two separate actix-web Data<> registrations (TileSources + OptTileCache) with a single, cheaply-cloneable TileSourceManager, and introducing ReloadAdvisory as the type-safe diff carrier.

**TileSourceManager** (`martin/src/tile_source_manager.rs`)
- Owns a shared `DashMap<String, BoxedSource>` and `TileCache` behind inner Arcs — clone cost is O(1)
- `from_sources()` constructor for bulk loading at startup
- `apply_changes(advisory)` applies updates first (cache-invalidate then swap), then additions, then removals — ordering ensures time-critical refreshes land before new sources appear and stale entries are GC'd last
- `tile_sources()` returns a `TileSources` view for read-only serving

**ReloadAdvisory** (`martin/src/reload.rs`)
- `NewSource` / `DeletedSource` structs for describing catalog mutations
- `from_sets()` for unversioned sources (Postgres, PMTiles) — computes additions and removals from two BTreeSets
- `from_maps()` for versioned sources (MBTiles, COG) — additionally detects updates when a version value changes

- `TileSources` now wraps `Arc<DashMap<...>>` so the manager and the server share the same underlying map without copying
- `TileCache` builder enables `support_invalidation_closures()`, required by moka for `invalidate_entries_if()` to work
- `ServerState` drops the `tiles` + `tile_cache` pair in favour of a single `tile_manager` field
- `server.rs` registers one `Data<TileSourceManager>` instead of two
- All tile-serving handlers (`get_tile`, `get_source_info`, `DynTileSource::new`) updated to accept the manager and extract sources / cache internally
- `Catalog::new()` populates static fields (sprites/fonts/styles) at startup with tiles defaulting to empty; the `get_catalog` handler merges in the live tile catalog from `TileSourceManager` per-request
- All 6 integration test suites updated to register `state.tile_manager`

Related to #288


martin/src/tile_source_manager.rs
```diff
@@ -0,0 +1,248 @@
...
+                cache.invalidate_source(&new_source.id);
+            }
+            self.tile_sources.insert(new_source.id.clone(), src);
+            info!("Updated source: {:?}", new_source.id);
+        }
+
+        // 2. Additions: make new sources available
+        for new_source in advisory.additions {
+            let Some(src) = new_source.source else {
+                tracing::warn!(
```
Auspicus: Not sure if we should make this a `Result`, not add the source to the advisory at all OR base the behavior of the `on_invalid` config.

martin/src/srv/admin.rs
```diff
@@ -39,7 +44,19 @@ impl Catalog {
...
+    #[cfg(feature = "_tiles")]
+    let catalog = {
+        let mut catalog = catalog.as_ref().clone();
+        catalog.tiles = tile_manager.tile_sources().get_catalog();
+        catalog
+    };
+    #[cfg(not(feature = "_tiles"))]
+    let catalog = catalog.as_ref();
+
     HttpResponse::Ok().json(catalog)
```
CommanderStorm, line 61: wonder if there is a better way to do this. this is a bit weird architecturally, but no biggie

martin/src/reload.rs
```diff
@@ -0,0 +1,243 @@
...
+    async fn from_sets_mixed() {
+        let prev: BTreeSet<String> = ["a", "b", "c"].into_iter().map(String::from).collect();
+        let next: BTreeSet<String> = ["b", "c", "d"].into_iter().map(String::from).collect();
+        let advisory = ReloadAdvisory::from_sets(&prev, &next, make_source).await;
+        assert_eq!(advisory.additions.len(), 1);
+        assert_eq!(advisory.additions[0].id, "d");
+        assert_eq!(advisory.removals.len(), 1);
+        assert!(advisory.removals.contains(&DeletedSource {
+            id: "a".to_string(),
+        }));
```
CommanderStorm: for cases like this where we inherently are asserting the internal structure, can you use `insta` and just assert the debug output of `advisory` instead? I feel this way it is easier to maintain than these assertuions.
This applies to the entirety of the file

Yes, they likely won't change, but they might.

martin/src/source.rs
```diff
@@ -9,19 +11,25 @@ use tracing::debug;
 ///
 /// Uses a [`DashMap`] for concurrent access without explicit locking.
 #[derive(Default, Clone)]
-pub struct TileSources(DashMap<String, BoxedSource>);
+pub struct TileSources(Arc<DashMap<String, BoxedSource>>);
```
CommanderStorm, line 14: To be on the same page/context: why the `Arc`, is `DashMap` not already cheaply clonable?

martin/src/tile_source_manager.rs
```diff
@@ -0,0 +1,248 @@
...
+    pub fn apply_changes(&self, advisory: ReloadAdvisory) {
+        if advisory.is_empty() {
+            return;
+        }
+
+        // 1. Updates: time-critical, invalidate cache then swap
+        for new_source in advisory.updates {
+            let Some(src) = new_source.source else {
+                tracing::warn!(
+                    "Skipping update for {:?}: source failed to initialize",
```
CommanderStorm:
```suggestion
                    "Skipping update for {:?}: source failed because it disapeared during an update. This likely indicates a bug and we would like a report if you can give us details what you are using",
```

martin/src/reload.rs
```diff
@@ -0,0 +1,243 @@
...
+
+use martin_core::tiles::BoxedSource;
+
+/// A source to be added or updated in the [`TileSourceManager`](super::TileSourceManager).
+#[derive(Debug)]
+pub struct NewSource {
+    /// Resolved source ID.
+    pub id: String,
+    /// The tile source implementation.
+    pub source: Option<BoxedSource>,
```
CommanderStorm: where it is used, we always seem to warn in the case where this is `None`.
It might be worth looking into if we can just remove the Option or why I/you thought this is nesseary

martin/src/tile_source_manager.rs
```diff
@@ -0,0 +1,248 @@
...
+                    new_source.id
+                );
+                continue;
+            };
+            if let Some(cache) = &self.tile_cache {
+                cache.invalidate_source(&new_source.id);
+            }
+            self.tile_sources.insert(new_source.id.clone(), src);
+            info!("Updated source: {:?}", new_source.id);
```
CommanderStorm: is this info or debug? What do you think?

martin/src/tile_source_manager.rs
```diff
@@ -0,0 +1,248 @@
...
+    #[test]
+    fn apply_additions() {
+        let mgr = make_manager();
+        let advisory = ReloadAdvisory {
+            additions: vec![new_source("src_a"), new_source("src_b")],
+            ..Default::default()
+        };
+        mgr.apply_changes(advisory);
+        let sources = mgr.tile_sources();
+        assert_eq!(sources.source_names().len(), 2);
```
CommanderStorm: Lets not just assert length, but just use `insta` and assert the debug output.
Maybe with sorting, not sure.

Any case, just asserting length is a bit weak of a test.
Applies below too

CommanderStorm (changes requested):
Nice work, that should get us a lot closer to that goal
Could you run the perf benchmark that you updated in this PR to make sure that there are no weird things here?

I have a few comments below, but they are all nits basically

martin/src/srv/tiles/content.rs
```diff
@@ -176,16 +174,17 @@ impl<'a> DynTileSource<'a> {
     #[expect(clippy::too_many_arguments)]
```
```suggestion
```

martin/src/srv/admin.rs
Auspicus: Yeah this will probably require a further refactor to have it make more sense. I kept this minimally invasive for now.

martin/src/reload.rs
Auspicus: This is related to my comment here: https://github.com/maplibre/martin/pull/2661#discussion_r3039273217

Because `MbtSource` and other source creation returns a `MartinResult<T>` we need to decide how best to handle failures to initialize a source during reload. The current setup is `Option<BoxedSource>` where `None` means there was a failure. I think this would make more sense as a `Result<T>` but then what should the reader do with the `Err`? Should we use the `on_invalid` configuration to inform on whether we `abort` or `warn`?

martin/src/source.rs
Auspicus: I believe the intended way to share a `DashMap` across threads is to wrap it in an `Arc<T>`. See docs here:

> This allows you to put a DashMap in an Arc<T> and share it between threads while being able to modify it.

~ https://docs.rs/dashmap/latest/dashmap/struct.DashMap.html

We want this same `DashMap` to be shared across all worker threads and cloned inside the `TileSourceManager` by reference rather than a copy.

martin/src/reload.rs
CommanderStorm: likely, yes

CommanderStorm: idk what makes more sense, i,e, where confug already exists

martin/src/tile_source_manager.rs
Auspicus: I think if we change to using a `Result<T>` for the advisory sources then the log level here should depend on the `on_invalid` config.

Auspicus:
I'll take another pass at this tomorrow, using `on_invalid` to inform how to process failed source init.

martin/src/reload.rs
CommanderStorm: Wait a minute:
I think that a newly failed source should act as a Removal operation, or?
Adding something unhealthy or updating to it does not seem like a good idea.
It is much cleander to just remove it, or?

If the underlying data changed in the way that we cannot work with it, there might be a point to warning or aborting as you said in the other comment

martin/src/tile_source_manager.rs
```diff
@@ -0,0 +1,263 @@
...
+                    new_source.id
+                );
+                continue;
+            };
+            if let Some(cache) = &self.tile_cache {
+                cache.invalidate_source(&new_source.id);
+            }
+            self.tile_sources.insert(new_source.id.clone(), src);
+            info!("Updated source: {:?}", new_source.id);
+        }
```
CommanderStorm: @nyurik found something after reviewing my proposal - We should do this like this.

Maybe with the `self.tile_cache.run_maintenance().await;` in a tokio task (run_blocking or run?), not sure.
Depending on how expensive they are in relation to each other, this might also make sense outside of this fn, not sure.

```suggestion
            // insert, then invalidate to make sure that no interleving is possible for stale data reads
            self.tile_sources.insert(new_source.id.clone(), src);
            if let Some(cache) = &self.tile_cache {
                cache.invalidate_source(&new_source.id);
            }
            info!("Updated source: {:?}", new_source.id);
        }
        // Remove stale data from the cache immediately, not on the next interval
        if updates.is_empty() {
           self.tile_cache.run_maintenance().await;
        }
```

CommanderStorm:
🤞🏻 lets see if hotpath does like to sumbit its feedback about perf now.

martin/src/config/file/main.rs
```diff
@@ -258,9 +254,15 @@ impl Config {
...
-            tiles,
-            #[cfg(feature = "_tiles")]
-            tile_cache: cache_config.create_tile_cache(),
+            // TODO: hard-coded to Warn so that a single source failure during
+            // reload doesn't abort the running server. This should be revisited
+            // once we decide whether on_invalid_reload deserves its own config
+            // knob (some operators may prefer aborting over serving degraded).
+            tile_manager: TileSourceManager::from_sources(
+                cache_config.create_tile_cache(),
+                OnInvalid::Warn,
```
Auspicus: For now this seems like the safe option, always warning never aborting, but I'm not sure what way we want to go with this one. It could be its own individual configuration option.

CommanderStorm: I don't think we should have an intalisation vs reload split.
The first load is basically a reload too and should be implemented this way once done 😉.

Lets handle this the same way and reuse the config option

Auspicus: My only hesitation is that it's defaulted to abort which would mean a reload event by default will crash the server.

This might be surprising and also might be unexpected if the error is transient (ie. file is copied into folder watched by martin and then eventually becomes a fully formed source).

Maybe we can just resolve that problem later though, once we start adding reloader implementations. I think they are best added behind an unstable feature flag to start with.

CommanderStorm: sure, that works for me.

Auspicus: I've set this back to use the `on_invalid` config.

CommanderStorm (approved):
👍🏻 Thanks a ton

Auspicus:
@CommanderStorm thanks for work-shopping this one with me. Next one to add MBTilesReloader behind flag will be coming soon =)

PR #2665, opened 2026-04-07
ci: fix hotpath not commeting due to overly constrictive validation

CommanderStorm:


mbtiles/Cargo.toml
```diff
@@ -37,6 +37,7 @@ clap = { workspace = true, optional = true }
 enum-display.workspace = true
 env_logger = { workspace = true, optional = true }
 flume.workspace = true
+hotpath.workspace = true
```
CommanderStorm:
```suggestion
```

PR #2669, opened 2026-04-08
fix: Use HOTPATH_ALLOC_SELF for benchmarks

pawurb:
Hi, https://github.com/pawurb/hotpath-rs author here.

I've noticed you're using CI integration. Default alloc tracking mode produces invalid results for recursive functions as you can see e.g. https://github.com/maplibre/martin/pull/2668 . 

This config should produce better results, I'm planning to make it default in next release.

BTW I'm actively working on the lib, so any feedback/bugs will be welcome.

### DIFF
--- a/.github/workflows/hotpath-profile.yml
+++ b/.github/workflows/hotpath-profile.yml
@@ -65,6 +65,7 @@ jobs:
         env:
           HOTPATH_OUTPUT_FORMAT: json
           HOTPATH_OUTPUT_PATH: /tmp/metrics/head_timing.json
+          HOTPATH_ALLOC_SELF: true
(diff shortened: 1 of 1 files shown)


CommanderStorm (approved):
Your work is awsome, thank you for all of this work you are putting into this ❤️

It really helped us tune some bugs in our MLT work and is a joy to use. Cannot speak higher of your work.

pawurb:
hi @CommanderStorm I'm starting work on a better hotpath CI integration that produces significantly less noise than the current workflow. Just added a waitlist on https://hotpath.rs/ , would be great to get your feedback once it launches.

CommanderStorm:
would love that 😉 
Signed up

PR #2672, opened 2026-04-09
docs: Update Docker image version in installation guide

musicguoke:
`ghcr.io/maplibre/martin:1.5.0` This version does not exist. The latest version is 1.4.0.


musicguoke:
Or you can directly release version `1.5.0`  :tada: .

CommanderStorm:
Sorry, human error. I forgot to sign of on the last part. I should really automate this better, this is tedious at best to have to do two clicks for every releases (they are created as drafts so that all artefacts can be uploaded..)

Issue #1137, opened 2024-01-17
Make the tile source cache configuration per source configurable


nyurik:
Each tile source cache should be configurable, at least to be able to selectively enable/disable cache, but possibly more advanced.

Steps:
* [ ] Add a boolean or a tri-state enum to each tile source config to allow `cache=enable|disable|auto`.  The `auto` setting (default) should be the same as `enable` for now, but may become more complex in the future.
* [ ] Make cache configurable per source type (postgres vs mbtiles vs pmtiles)
* [ ] For postgres, make cache configurable for all auto-discoverable tables / functions

saosebastiao:
I would like to add that a common use case for caching with tile servers is to have Z-level specific caching. For example, large datasets at low zoom levels are very expensive, but at high zoom levels they are really cheap, so it would be nice to be able to say something like : "for data_source_a I want to cache zoom levels from 0-5 only, but for data_source_b I want to cache zoom from 0-9"

saosebastiao:
Additionally, cache invalidation rules...we should be able to specify a TTL at the very least. 

An ideal-world scenario, but obviously far more complex is cache invalidation that happens in sync with the database via LISTEN/NOTIFY updates via trigger functions attached to tables.

bernatfortet:
> I would like to add that a common use case for caching with tile servers is to have Z-level specific caching. For example, large datasets at low zoom levels are very expensive, but at high zoom levels they are really cheap, so it would be nice to be able to say something like : "for data_source_a I want to cache zoom levels from 0-5 only, but for data_source_b I want to cache zoom from 0-9"

This would be amazing to have. In my case, users need to update specific geometries so when that happens the cache needs to be busted. Right now I need to choose between all or northing.

johncarmack1984:
Went through the checklist against current main, most of this has shipped since:

- Per-source: every file source and every declared postgres table/function takes `cache:` with `minzoom`/`maxzoom` bounds, and `cache: disable` turns it off for that source (#2673 for the zoom bounds).
- Server-wide defaults: `cache.minzoom` / `cache.maxzoom`, plus the whole sizing block (`cache.size_mb`, `cache.tile_size_mb`, per-cache splits).
- The TTL asked for in the comments: `cache.expiry` / `cache.idle_timeout` with tile-specific overrides (#2691).
- Targeted invalidation on top of that is #1707's thread (`DELETE /cache/{source_id}` in #3194, ~~LISTEN/NOTIFY under discussion there~~).

[...]

PR #2673, opened 2026-04-09
feat: support per-source cache zoom level

carderne:
### Overview
Implements this issue: https://github.com/maplibre/martin/issues/1137

Allows global and per-source cache zoom configuration. Requests outside the range will not be cached.

Most obvious use is to not cache anything above eg zoom 8.

### Testing
Not yet tested with actual data, but it's a very simple change.

### Performance impact
Adds two `.is_none()` checks to the `get_tile_content` hot path.

### Config
Adds two new config options at the top level:
```yml
# global settings
default_cache_minzoom: 0   # default null
default_cache_maxzoom: 9  # default null
```

And per-source settings:
```yml
postgre:
  my_table:
    cache_minzoom: 0
    cache_maxzoom: 5
```


docs/content/files/config.yaml
```diff
@@ -45,6 +45,18 @@ cache_size_mb: 512
...
+# Tiles below this zoom will bypass the cache entirely.
+# Can be overridden per-source (e.g. cache_minzoom on an individual source).
+# default: null (no lower bound, all zoom levels cached)
+default_cache_minzoom: null
+
+# Default maximum zoom level (inclusive) for tile caching.
+# Tiles above this zoom will bypass the cache entirely.
+# Can be overridden per-source.
+# default: null (no upper bound, all zoom levels cached)
+default_cache_maxzoom: null
```
CommanderStorm: We would like to longer term add more configurabiltiy to the cache.
Lets start going down this path here, since this seems fairly simple to add.

I would prefer this to be configured this way instead.
This way, this is compatible with the `ttl`, `idle_timeout`, `size`.

```suggestion
# Defaults for how to configure the cache
cache:
  # Default minimum zoom level (inclusive) for tile caching.
  # Tiles below this zoom will bypass the cache entirely.
  # Can be overridden per-source (e.g. cache_minzoom on an individual source).
  # default: null (no lower bound, all zoom levels cached)
  minzoom: null

  # Default maximum zoom level (inclusive) for tile caching.
  # Tiles above this zoom will bypass the cache entirely.
  # Can be overridden per-source.
  # default: null (no upper bound, all zoom levels cached)
  maxzoom: null
```

martin-core/src/tiles/source.rs
```diff
@@ -45,6 +45,18 @@ pub trait Source: Send + Sync + Debug {
...
+    /// `None` means no lower bound. Default: None.
+    fn cache_minzoom(&self) -> Option<u8> {
+        None
+    }
+
+    /// Maximum zoom level (inclusive) at which tiles from this source should be cached.
+    /// `None` means no upper bound. Default: None.
+    fn cache_maxzoom(&self) -> Option<u8> {
+        None
+    }
```
CommanderStorm: All of the sources implement this, so lets not add the default. This would be confusing otherwise since this is never hit

```suggestion
    fn cache_minzoom(&self) -> Option<u8>;

    /// Maximum zoom level (inclusive) at which tiles from this source should be cached.
    fn cache_maxzoom(&self) -> Option<u8>;
```

docs/content/files/config.yaml
```diff
@@ -292,6 +314,11 @@ pmtiles:
     pm-src1: /path/to/pmt.pmtiles
     # A named source to a web server with a PMTiles file that supports range requests
     pm-web2: https://example.org/path/tiles.pmtiles
+    # A source with per-source cache zoom range
+    pm-with-cache:
+      path: /path/to/pmt2.pmtiles
+      cache_minzoom: 0
+      cache_maxzoom: 6
```
CommanderStorm: on the one hand yes, this makes sense. on the other, we won't be able to support TTL, expiry or size limiting at the per-source level.

How about the following way of configuring this? The code is essentialyl the same, just how we configure this is different.

```suggestion
      # You can enable (`cache: enable`) or disable (`cache: disable`) a cache for a source.
      # If not specfied, datasource or global defaults apply.
      cache:
        minzoom: 0
        maxzoom: 6
```

CommanderStorm (changes requested):
Shure, that seems like a nice feature.
I would like this configured sligtly differently ("more future proof").

Could you also add an integration test right here for every level of cache-config we add -> https://github.com/maplibre/martin/blob/main/tests/test.sh

docs/content/files/config.yaml
carderne: done in https://github.com/maplibre/martin/pull/2673/changes/f8c5c16f0dae697fb542e660d6d6ef4ff5d41f9e

Ultimately it would make sense for the `cache_size_mb` to go in there as well, but obviously not touching that here...

martin-core/src/tiles/source.rs
carderne: done in https://github.com/maplibre/martin/pull/2673/changes/f8c5c16f0dae697fb542e660d6d6ef4ff5d41f9e

docs/content/files/config.yaml
carderne: (Mostly) done in https://github.com/maplibre/martin/pull/2673/changes/f8c5c16f0dae697fb542e660d6d6ef4ff5d41f9e

I didn't add the `cache: enable/disable` as I'm not sure about the semantics of the semantics of allowing a string or a struct there...? Could add a field `cache.enable: enable/disable` or `bool`... Or document that it can be disabled by just setting minzoom/maxzoom to 0/0?

carderne:
Responded to all comments and added some e2e test config as suggested.

I named the new config struct `CachePolicy`. Briefly thought of using the existing `CacheConfig` but that's root-level only, so decided not to...

docs/content/files/config.yaml
CommanderStorm: Oh, they are just aliases.

`cache: enable` means

```yaml
cache:
  minzoom: null
  maxzoom: null
```

`cache: disable` means

```yaml
cache:
  minzoom: 1
  maxzoom: 0
```

docs/content/files/config.yaml
```diff
@@ -45,6 +45,20 @@ cache_size_mb: 512
 # Defaults to `cache_size_mb` / 2
 tile_cache_size_mb: 256
 
+# Defaults for how to configure the cache
+cache:
```
CommanderStorm, line 27: the code currently does name this `default_cache`
@nyurik after your vacation, do you think we should name this `default_cache` or `cache`

martin/src/config/file/file_config.rs
```diff
@@ -499,6 +535,32 @@ fn parse_url(is_enabled: bool, path: &Path) -> Result<Option<Url>, ConfigFileErr
...
+/// Zoom-level bounds for tile caching. Used at the top level (as a global default),
+/// at backend level, and per-source to control which zoom levels are cached.
+#[serde_with::skip_serializing_none]
+#[derive(Clone, Copy, Debug, Default, PartialEq, Serialize, Deserialize)]
+pub struct CachePolicy {
+    pub minzoom: Option<u8>,
+    pub maxzoom: Option<u8>,
+}
+
+impl CachePolicy {
```
CommanderStorm: Lets rewrite this a bit differently to not churn too much on this code.

Lets have `CachePolicy` be the generic thing that is at the root and source level, but move your code into `CacheZoomRange`.
This cleans up the code upstream of this signifcantly ^^

This has the advantage of making this easier to pass around which is something that we should do.

```suggestion
#[serde_with::skip_serializing_none]
#[derive(Clone, Copy, Debug, Default, PartialEq, Serialize, Deserialize)]
struct CachePolicy {
    #[serde(flatten, default)]
    zoom: CacheZoomRange,
}


/// Zoom-level bounds for tile caching. Used at the top level (as a global default),
/// at backend level, and per-source to control which zoom levels are cached.
#[serde_with::skip_serializing_none]
#[derive(Clone, Copy, Debug, Default, PartialEq, Serialize, Deserialize)]
pub struct CacheZoomRange {
    pub minzoom: Option<u8>,
    pub maxzoom: Option<u8>,
}

impl CacheZoomRange {
```

docs/content/files/config.yaml
```diff
@@ -45,6 +45,20 @@ cache_size_mb: 512
...
+  # Default minimum zoom level (inclusive) for tile caching.
+  # Tiles below this zoom will bypass the cache entirely.
+  # Can be overridden per-source (e.g. cache_minzoom on an individual source).
```
CommanderStorm:
```suggestion
  # Can be overridden per-source (e.g. cache.minzoom on a type of source or an individual source).
```

docs/content/files/config.yaml
```diff
@@ -45,6 +45,20 @@ cache_size_mb: 512
...
+  # default: null (no lower bound, all zoom levels cached)
+  minzoom: null
+
+  # Default maximum zoom level (inclusive) for tile caching.
+  # Tiles above this zoom will bypass the cache entirely.
```
CommanderStorm: also applies in other cases, but please don't talk about above or below.
Non-native speakers will have trouble visualising this.

Talk about further zoomed out/in instead.

martin/src/config/file/main.rs
```diff
@@ -83,6 +83,11 @@ pub struct Config {
     /// Overrides [`cache_size_mb`](Self::cache_size_mb)
     pub tile_cache_size_mb: Option<u64>,
 
+    /// Default zoom-level bounds for tile caching.
+    /// Applied as a fallback to any tile source that does not configure its own value.
+    #[serde(default)]
+    pub cache: Option<CachePolicy>,
```
CommanderStorm: Since this is already `serde(default)` and implements `Default`, this is more ergonomic.

Has downstream implications

```suggestion
    /// Defaults for how our cache works
    #[serde(default)]
    pub cache: CachePolicy,
```

CommanderStorm (changes requested):
Much better and future facing. lets clean up the internal interface to this and get this merged.

carderne:
@CommanderStorm I can spend some time on this tomorrow to finalise it, if useful, but happy to hand off also. Let me know what's useful, and what's still to be done.

martin/src/config/file/main.rs
```diff
@@ -491,8 +485,88 @@ where
...
+    };
+    let mut current = &mut *mapping;
+    for &segment in parents {
+        if !current.contains_key(segment) {
+            current.insert(
+                serde_yaml::Value::String(segment.into()),
+                serde_yaml::Value::Mapping(serde_yaml::Mapping::default()),
+            );
+        }
+        let Some(nested) = current.get_mut(&segment).and_then(|v| v.as_mapping_mut()) else {
```
CommanderStorm:
```suggestion
        let Some(nested) = current.get_mut(segment).and_then(|v| v.as_mapping_mut()) else {
```

martin/src/config/file/main.rs
```diff
@@ -491,8 +485,88 @@ where
...
+            warn!(
+                "deprecated config: `{old_key}` is ignored because `{segment}` is already set. \
+                 Please remove `{old_key}` from your configuration"
+            );
+            return;
+        };
+        current = nested;
+    }
+
+    if current.contains_key(&leaf) {
```
CommanderStorm:
```suggestion
    if current.contains_key(leaf) {
```

martin/src/config/file/file_config.rs
```diff
@@ -499,6 +536,233 @@ fn parse_url(is_enabled: bool, path: &Path) -> Result<Option<Url>, ConfigFileErr
...
+    pub fn zoom(self) -> CacheZoomRange {
+        self.zoom
+    }
+
+    /// Returns `true` if no cache bounds are configured.
+    #[must_use]
+    #[expect(
+        clippy::trivially_copy_pass_by_ref,
+        reason = "serde skip_serializing_if requires &self"
+    )]
```
CommanderStorm, line 584:
```suggestion
```

martin/src/config/file/main.rs
```diff
@@ -534,4 +622,137 @@ mod tests {
         assert!(parse_base_path("").is_err());
         assert!(parse_base_path("foo/bar").is_err());
     }
+
+    #[test]
+    fn migrate_cache_size_mb_to_cache_size_mb() {
+        let config = parse_yaml("cache_size_mb: 512");
+        assert_eq!(config.cache.size_mb, Some(512));
+    }
```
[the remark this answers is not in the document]
CommanderStorm:
```suggestion
    #[test]
    fn cache_migrates_old_to_new_cache_config_key() {
        let config = parse_yaml("cache_size_mb: 512");
        assert_eq!(config.cache.size_mb, Some(512));
    }
```

CommanderStorm (approved):

carderne:
Nice one thank you! Looking forward to using this.

PR #2681, opened 2026-04-12
feat(mbtiles): Support planetilers' normalised schema

CommanderStorm:
Resolves #2630


martin/src/srv/tiles/content.rs
```suggestion
```

PR #2682, opened 2026-04-12
feat(mbtiles): Add a transcoder API

CommanderStorm:
this PR adds the ability to transcode schemas to the mbtiles crate


mbtiles/src/transcoder.rs
```diff
@@ -0,0 +1,869 @@
...
+        dst_conn: &mut SqliteConnection,
+        dst_type: MbtType,
+    ) -> MbtResult<TranscodeStats> {
+        let schema = src_type
+            .normalized_schema()
+            .expect("run_normalized_path called with non-normalized src_type");
+        let content_table = schema.content_table();
+        let tile_id_col = schema.tile_id_column();
+        let src_map = schema.map_table();
+
```
nyurik, line 245: shouldn't this be used in both paths?  Please dedup this code thoroughly!

nyurik (approved):
i don't think you need the temp table - instead use the standard sqlite optimizations for bulk inserts:
* transaction
* capture and delete any indexes and recreate after
* some pragma magic
* whatever else ai suggests :)

Other things:
* use proper "hash" value for deduping the images in a normalized schema
* dedup normalized and de-norm paths - feels like you have tons of dup code
* see if existing sql can be reused instead of adding new

martin-core/Cargo.toml
```diff
@@ -80,7 +80,7 @@ image = { workspace = true, optional = true }
 itertools = { workspace = true, optional = true }
 martin-tile-utils.workspace = true
 mbtiles = { workspace = true, optional = true }
-moka.workspace = true
+moka = { workspace = true, features = ["future"] }
```
CommanderStorm:
```suggestion
moka = { workspace = true, features = ["future", "quanta"] }
```

martin/src/srv/tiles/content.rs
```diff
@@ -19,7 +19,10 @@ use tracing::warn;
 use crate::config::args::PreferredEncoding;
 use crate::config::file::srv::SrvConfig;
 use crate::srv::server::{DebouncedWarning, map_internal_error};
+<<<<<<< transcoder
+=======
 use crate::tile_source_manager::TileSourceManager;
+>>>>>>> main
```
```suggestion
use crate::tile_source_manager::TileSourceManager;
```

CommanderStorm:
@nyurik I am going to skip work on this, because while this might matter to someone, for our usecase these optimisations really don't and I have already spend way too much time on this tiny part.

Lets remember that wrtiting the files is actually the least amount of time on this, not the most.
I cannot justify spending soo much time on the last 1% while there are actual gains to be made in the first 99%...

I won't implement:
- index recreation
- more than the current insert optimisation or 
- acutally deduplicate the sql here (it is not in the other parts, finding good ORMthat is not a sieve is not something that I currently have the time for..)

Yes, a bit of technical dept, but I own this and I am chosing to move faster over doing it right

PR #2683, opened 2026-04-13
feat: impl pre-processing encodings

### LINKED ISSUE #2689: RFC Post-processing (compression, encoding, conversion, pruning)
Here is the RFC for how processing should work:

Martin can postprocess tiles before serving them. The `process` block controls two optional steps:

- **MLT conversion** — convert MVT tiles to [MapLibre Tiles](using-guides/mlt.md) format on the fly
- **Compression** — override the compression format preference order

The `process` block can appear at three levels. The most specific level wins entirely (no merging between levels):

1. **Global** — applies to all sources
2. **Source-type** — applies to all sources of that type (e.g. all PMTiles sources)
3. **Per-source** — applies to a single source

```yaml
# Global: convert all MVT sources to MLT
process:
  mlt: auto

postgres:
  connection_string: postgresql://localhost/mydb
  # Source-type: override for all PG sources (disables MLT, forces brotli)
  process:
    compress: [br, gzip]
  tables:
    my_table:
      # Per-source: this table gets MLT conversion
      process:
        mlt: auto
```

See the [MLT usage guide](using-guides/mlt.md) for encoder tuning options.

I thought about these alternatives:
- using `mlt: auto` at the root level seems too chaotic given the cascading nature
- using a pipelined approach (the oder is fixed -> simpler and really does not make sense to make)

CommanderStorm:
Adds an MVT -> MLT post-processing pipeline

## Bench numbers

Local run on a single postgres source (`table_source/1/1/0`, GEOMETRY/SRID 4326, 1204 B raw MVT -> 735 B raw MLT for this fixture).

| Scenario                 | MVT (`Accept: application/x-protobuf`) | MLT (`Accept: application/vnd.maplibre-tile`) | Notes                                                                           |
| ------------------------ | -------------------------------------- | --------------------------------------------- | ------------------------------------------------------------------------------- |
| **Cache-hit, steady**    | 109,268 req/s, p50 0.3 ms, p99 3.2 ms  | 123,050 req/s, p50 0.3 ms, p99 2.7 ms         | MLT entry is uncompressed and smaller than the gzipped MVT entry -> ~12% faster |
| **No-cache, every req converts** | 9,170 req/s, p50 5.2 ms, p99 9.1 ms    | 7,672 req/s, p50 6.3 ms, p99 10.7 ms          | ~1.1 ms encode cost per tile, ~16% throughput loss at zero cache hit            |
| **Wire size (raw)**      | 1,204 B                                | 735 B                                         | ~39% smaller for this fixture                                                   |

Repro:

```bash
just bench-server &
sleep 20
just bench-http 50k 50k function_zxy_query
```


martin/src/config/file/main.rs
```diff
@@ -1,3 +1,4 @@
+use std::collections::HashMap;
```
CommanderStorm, line 2:
```suggestion
```

PR #2690, opened 2026-04-14
ci: FreeBSD CI

### LINKED ISSUE #2680: FreeBSD CI
To ensure against future breakage. Just adding a task for myself. I already did this on the SDF Font tools repo and we should be able to basically copy that.

ianthetechie:
It looks like there are a few somewhat disjoint ways of executing tests, so I went with something that looked reasonable. This is a separate stage since it's quite a bit different from the other (e.g. multi-os) builds as it needs this qemu action to wrap the logic.

Closes #2680.

Notes:

- I added a test-freebsd just recipe to codify what we're testing.
- This does NOT exercise the integration test suite. This shouldn't be an issue as it is already run on many other platforms. The reason we can't run these is because they rely on Docker being available directly within the test environment. To make this more cross-platform, there might be a way to attach "services" to GitHub Actions (I don't know it that well but some other CI platforms do). Another option would be to switch to Podman which isn't so Linux-specific. But it looks like this is a _very_ deep rabbit hole between direct docker invocations in the justfile and the `testcontainers` package trying to spin up postgres directly in benchmark and integration tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -571,6 +571,31 @@ jobs:
             target/test_logs/*
           retention-days: 5
 
+  test-freebsd:
+    name: Unit test on FreeBSD
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    needs:
+      - lint-rust
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with: { persist-credentials: false }
+      - name: Test in FreeBSD
+        id: test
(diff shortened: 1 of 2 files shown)


CommanderStorm (approved):

ianthetechie:
Haha let's make sure it builds first, but I appreciate the vote of confidence ;) Can't really test this CI stuff locally.

CommanderStorm:
I mean yea, if it fails CI this won't automerge ^^
If it passes, this LGTM...

ianthetechie:
The tests seem to make a few assumptions about the host environment 😂

ianthetechie:
Ok; I assume it will work this time (actually tested on my local box). I'll document the decisions in the PR description.

PR #2691, opened 2026-04-14
feat: make internal cache expiry configurable

### LINKED ISSUE #2505: Making internal cache expiry more configurable

CommanderStorm:
Resolves #2505


martin-core/src/resources/styles/snapshots/martin_core__resources__styles__tests__src2_maptiler_basic_0_0_0.snap.png
CommanderStorm, line 1: interesting how the US changed their name 😉

PR #2703, opened 2026-04-17
chore: impl the accept header

CommanderStorm:
as discussed, the sanatised accept header needs to be a cache key.

This now means that the cache key is not a unique mapping anymore


martin-core/src/tiles/cache.rs
```diff
@@ -41,14 +41,15 @@ impl TileCache {
...
+        format: Option<Format>,
         compute: F,
     ) -> Result<Tile, Arc<E>>
     where
         F: FnOnce() -> Fut,
         Fut: Future<Output = Result<Tile, E>>,
         E: Send + Sync + 'static,
     {
-        let key = TileCacheKey::new(source_id, xyz, query);
+        let key = TileCacheKey::new(source_id, xyz, query, format);
```
nyurik, line 52: why not just pass the `TileCacheKey` as the param into `get_or_insert`?

CommanderStorm: legacy? Also, was this way -> different PR

martin-tile-utils/src/lib.rs
```diff
@@ -129,6 +132,22 @@ impl Format {
...
 
+    /// Parse a content type string back to a `Format`.
+    #[must_use]
+    pub fn from_content_type(supertype: &str, subtype: &str) -> Option<Self> {
+        Some(match (supertype, subtype) {
+            ("image", "gif") => Self::Gif,
+            ("image", "jpeg") => Self::Jpeg,
+            ("application", "json") => Self::Json,
+            ("application", "x-protobuf" | "vnd.mapbox-vector-tile") => Self::Mvt,
+            ("application", "vnd.maplibre-vector-tile") => Self::Mlt,
```
nyurik: the only reference to this string is in https://github.com/maplibre/maplibre-tile-spec/issues/536 - and I don't think the word `vector` should be in that - it makes the string longer, wasting extra bytes, and it may not be true in the future.

nyurik:
```suggestion
            ("application", "vnd.maplibre-tile") => Self::Mlt,
```

nyurik (approved):

martin-tile-utils/src/lib.rs
```diff
@@ -122,13 +125,29 @@ impl Format {
             Self::Jpeg => "image/jpeg",
             Self::Json => "application/json",
             Self::Mvt => "application/x-protobuf",
-            Self::Mlt => "application/vnd.maplibre-vector-tile",
+            Self::Mlt => "application/vnd.maplibre-tile",
```
pnorman, line 128: This is not a valid MIME type yet so should not be merged until it is registered.

The same applies to other occurrences of the MIME type

CommanderStorm: we should likey get to registering this, but that is a bit of daunting work that I want to have the entire toolchain ready for. They likely requrire a toolchain to be ready and production use, which we currently I think don't have

louwers:
The TileJSON spec is very clear 

> All endpoints MUST return the same content for the same URL.

I don't think Martin can claim to implement TileJSON with this change.

pnorman:
1. Most deployments should be assuming http/2 these day and not using multiple tile URLs, so this doesn't apply in most cases as there's not a different endpoint to be inconsistent with.

2. I don't see that the tilejson spec prohibits content-negotiation, largely because I don't think anyone gave it any thought when the spec was written. HTTP servers will **always** return different results depending on the headers.

louwers:
It says "all endpoints must return the same content for the same URL". So the way I read it, if there's only one endpoint, and it returns completely different contents for the same URL (depending on what headers are provided), that would not be compliant.

I think you're right though, content negotiation was not considered when this was written.

Issue #2180, opened 2025-09-15
Allow a `.pmtiles` "directory" located in S3, not just a local path


gabeschine:
In short, this would allow the following config snippet:

```
pmtiles:
  paths:
    - s3://my-bucket/dir_with_many_pmtiles_files/
```

This same functionality could apply to COGs and anything else that expands a directory into a set of entities (is there another application other than COGs?).

I'm happy to work on a PR once a discussion here converges on an approach.

gabeschine:
@nyurik and I had a discussion in Slack and here was the important bit, included for completeness here:

martin uses pmtiles crate to serve it - https://github.com/maplibre/martin/blob/main/martin/src/pmtiles/mod.rs

[from Slack](https://osmus.slack.com/archives/C04EQRCKK6W/p1754602373609359?thread_ts=1754599125.762699&cid=C04EQRCKK6W)
> the config stuff is in the same dir
> basically if its a path, it tries to enumerate files locally
> if its a dir
> otherwise it simply uses the pmtiles crate
> for s3, it does the same thing
> https://github.com/maplibre/martin/blob/bfab917b4804242e246fc25719e79121b652b4da/martin/src/file_config.rs#L373
> in short - we would need to implement s3 support directly in martin to support file listing
> i think its a good feature to have - perhaps create an issue so we can discuss how to do it, and maybe even a PR?

nyurik:
Note that pmtiles already uses object store which seems to support enumeration:  https://docs.rs/object_store/latest/object_store/#list-objects -- the question is how we can reuse it. One option might be to modify pmtiles to re-export object_store - this way it could allow externally-provided configuration for s3 access if needed, and also allow martin to use object_store directly to enumerate content before telling pmtiles which specific objects to use... tbd.  CC: @lseelenbinder

CommanderStorm:
Migrating to `object_store` would also have other advantages like supporting google/aws/.. as a S3 alternative.

I think that is a reasonable choice and so before v1.0 is also a good thing to do.

The design that I added to `pmtiles` should allow for what we are after. No modification needed.
I think splitting this into two parts (migrating to `object_store`, adding directory listing) is a sane choice implementation wise.

CommanderStorm:
Of note:
I currently have [two refactorings open](https://github.com/maplibre/martin/pulls?q=is%3Apr+is%3Aopen+pmtiles+author%3ACommanderStorm).

If you want to work on it before those are in `main`, please rebase any work on 
[...]

PR #2704, opened 2026-04-17
feat: support remote pmtiles directory auto-discovery

carderne:
**NB:** Still a work-in-progress, but tested to be working with a GCS bucket I control.

Resolves this issue: https://github.com/maplibre/martin/issues/2180

In theory quite simple, but there are a bunch of design choices I'm not really sure about...

### Sequential load
It has to do sequential GETs for each pmtiles. This will be very slow for large collections. Tried pointing it at a bucket with 5k+ pmtiles :P Would be super slow + memory issue I'm sure. A single failing header open inside a prefix aborts the whole prefix (same as the existing single-URL behavior). Could be softened to per-file warnings without much work if maintainers prefer.

### Path vs file detection
Just does "URL path ends with .pmtiles" -> single file, else list as prefix

### New URL creation
Listed object's URL is rebuilt via child.set_path("/{location}"). Works for s3://, gs://, az://, file:// because in all of them the URL path component is the object_store key. Not tested across backends... http:// listing is rare and may behave unevenly, error path should handle.

Did not refactor the TileSourceConfiguration trait to pass (store, path) pairs instead of URLs, even though that would avoid the URL reconstruction step. Kept the trait-shape change minimal (one default method).

### No generic HTTP paths
Eg https://host/prefix/ will obviously fail unless it has the appropriate APIs.

### Testing
Unit tests cover expand_url against file:// (tempdir) for the three main cases (prefix-with-matches, direct file URL, empty prefix). No integration test hits real S3... the existing test.sh still uses a single-file S3 URL.                                                  

## Other
Sorry about the `Arc` moving around, I ran `just fmt`.

### DIFF
--- a/martin/src/config/file/file_config.rs
+++ b/martin/src/config/file/file_config.rs
@@ -80,6 +80,20 @@ pub trait TileSourceConfiguration: ConfigurationLivecycleHooks {
         url: Url,
         cache: CachePolicy,
     ) -> impl Future<Output = MartinResult<BoxedSource>> + Send;
+
+    /// Expand a URL that may refer to a remote "directory" (prefix) into one URL per object.
+    ///
+    /// The default implementation treats the URL as a single object and returns it unchanged.
+    /// Source types that support remote listing (e.g. `PMTiles` via `object_store`) may override
+    /// this to enumerate objects matching `allowed_extension` under a prefix.
+    #[allow(unused_variables)]
+    fn expand_url(
+        &self,
+        url: Url,
+        allowed_extension: &[&str],
+    ) -> impl Future<Output = MartinResult<Vec<Url>>> + Send {
+        async move { Ok(vec![url]) }
+    }
 }
 
 #[derive(Clone, Debug, Default, PartialEq, Serialize, Deserialize)]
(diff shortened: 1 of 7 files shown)


martin/src/config/file/file_config.rs
```diff
@@ -419,29 +433,54 @@ async fn resolve_one_path_int<T: TileSourceConfiguration>(
...
+        // scans it on reload, mirroring the local-directory branch below.
+        let was_prefix = expanded.len() != 1 || expanded[0] != url;
+        if was_prefix {
+            directories.push(path.clone());
+        }
+        if expanded.is_empty() {
+            warn!(
+                "No files matching {extension:?} found under {}",
+                sanitize_url(&url)
+            );
```
CommanderStorm, line 448: this likely should trigger our on-invalid setting

https://github.com/maplibre/martin/blob/7071f1244d8ba584e3180ebcdb6ea10c2a0d3c69/martin/src/config/file/main.rs#L477

martin/src/config/file/file_config.rs
```diff
@@ -80,6 +80,20 @@ pub trait TileSourceConfiguration: ConfigurationLivecycleHooks {
...
         cache: CachePolicy,
     ) -> impl Future<Output = MartinResult<BoxedSource>> + Send;
+
+    /// Expand a URL that may refer to a remote "directory" (prefix) into one URL per object.
+    ///
+    /// The default implementation treats the URL as a single object and returns it unchanged.
+    /// Source types that support remote listing (e.g. `PMTiles` via `object_store`) may override
+    /// this to enumerate objects matching `allowed_extension` under a prefix.
+    #[allow(unused_variables)]
+    fn expand_url(
```
CommanderStorm, line 90: I am not entriely a fan of this achritecture, since this means maintaining it.

We will need to refactor this to this sometime in the future.
If this PR comes with good tests, we can still go forward with this, but it needs to come with tests that this behaviour works as intended.

https://github.com/maplibre/martin/blob/7071f1244d8ba584e3180ebcdb6ea10c2a0d3c69/martin/src/reload.rs#L29

martin/src/config/file/tiles/pmtiles.rs
```diff
@@ -256,4 +257,99 @@ impl TileSourceConfiguration for PmtConfig {
...
+            let mut child = url.clone();
+            child.set_path(&format!("/{loc}"));
+            results.push(child);
+        }
+        Ok(results)
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
CommanderStorm, line 296: I would like to see a test using a (`cfg-scoped` test like the `postgres` feature) feature.
I.e. testing the s3 support.

I think this should work via either [minio](https://docs.rs/testcontainers-modules/latest/testcontainers_modules/minio/struct.MinIO.html), [localstack](https://docs.rs/testcontainers-modules/latest/testcontainers_modules/localstack/struct.LocalStack.html) or some other version.

CommanderStorm (changes requested):
I am fine with merging this code (even though we want to move away from this concrete way of implementing init and towards reloadable init).
I would like a testcase though for the path you are adding.
The test you added is great, but I would reaaaly like a test for the "actual path" you care about.

This is important because this way we can also catch problems in the path you care about

carderne:
@CommanderStorm we don't need this feature anymore, so if you're not convinced this is useful/necessary I won't be sad if you don't merge it. I'm also not likely to have the time in the next week or two to get this to a mergable state.

About the test: I tried with the existing S3 bucket you have in the test cases, but there weren't many files and I wasn't completely sure how to do it without breaking the existing S3 test. The files I tested it with are not public unfortunately.

CommanderStorm:
We now (because of my and @Auspicus) work have two other impls of this.
I feel they are more complete, lets regroup on that.

CommanderStorm:
Thanks for the work, even though this did not yield in an immediate merge 🙇🏻

PR #2711, opened 2026-04-19
ci: maybe fix OOM in freebsd build??

CommanderStorm:
CC @ianthetechie


CommanderStorm:
@nyurik if Freebsd succeeds, please rerun that job specifically and approve it.
If that succeeds twice in a row, that has increased the likelyhood that I actually found the issue

ianthetechie:
Oh interesting. Maybe a limitation of the Actions runners and/or the weird QEMU setup.

PR #2712, opened 2026-04-19
feat(mbtiles): add --strict flag to use STRICT SQLite tables

### LINKED ISSUE #904: Use STRICT SQLite tables in MBTiles
Since `2021-11-27 (3.37.0)`, SQLite supports [STRICT tables](https://www.sqlite.org/stricttables.html) -- where the table column's type actually means that the entire table has that type in that column, and not "preserve whatever you inserted there as is, but use column type for comparison and other magical stuff".  Sadly this is a relatively recent addition (~2 years), so when sqlite is not statically linked, it is possible for the `libsqlite` to not support them.  For now, we should add a non-default `--strict` (?) flag for `mbtiles` that will use this. The strict mode will be used on all commands that create a new table, e.g. `mbtiles copy`

mvanhorn:
## Summary

Closes #904. Adds a non-default `--strict` flag to `mbtiles copy` that emits [STRICT SQLite tables](https://www.sqlite.org/stricttables.html) when creating the destination schema. When the flag is omitted, behavior is unchanged.

## Why this matters

Per your issue: SQLite has supported STRICT tables since 3.37 (2021), and opting in for `mbtiles` destinations is a straightforward type-safety win for anyone running against a recent libsqlite. The default stays off because older statically linked libsqlite variants may not have STRICT support yet.

The schema helpers in `mbtiles/src/queries.rs` are the canonical CREATE TABLE surface for `mbtiles copy`:
- `create_metadata_table` (line 190)
- `create_flat_tables` (line ~205)
- `create_flat_with_hash_tables` (line ~225)
- `create_bsdiffraw_tables` (bsdiff patch tables)

Each of these had a non-STRICT `CREATE TABLE ... (...);` that now conditionally closes with `) STRICT;` when the caller opts in.

## Changes

- `mbtiles/src/queries.rs`: the four `create_*` helpers now take `strict: bool` and swap the closing `);` for `) STRICT;` when true. Views are untouched - STRICT is table-only.
- `mbtiles/src/copier.rs`: `MbtilesCopier` gains a `strict: bool` field (default false). The option is threaded to every helper call site. A small `maybe_make_table_strict` helper also patches CREATE TABLE text captured from `sqlite_master` so schema-clone paths honour the flag too.
- `mbtiles/src/bindiff.rs`: passes the `strict` option through when building bsdiff patch tables.
- `mbtiles/src/bin/mbtiles.rs`: `--strict` flag on the `copy` subcommand, piped into `MbtilesCopier.strict`.

## Testing

- `cargo build -p mbtiles` - clean.
- `cargo test -p mbtiles --lib` - 45 passed (includes the new `copier::tests::diff_with_bindiff_uses_strict_patch_tables_when_requested` exercising the strict plumbing through the bsdiff path).
- Running `mbtiles copy --strict src.mbtiles dst.mbtiles` produces a destination where `sqlite_master` rows for `tiles`, `tiles_with_hash`, and `metadata` have `STRICT` in their SQL; omitting the flag produces byte-identical CREATE TABLE text to pre-PR behaviour.

## Scope (for review)

Intentionally narrow first cut:
- `mbtiles/src/transcoder.rs` and `mbtiles/src/metadata.rs` still emit their own raw CREATE TABLE SQL that bypasses the `create_*` helpers. A follow-up can route those through the same helpers so the flag applies there too.
- No runtime sqlite-version check for 3.37+. If the user runs against an older libsqlite the CREATE TABLE will error with a clear SQLite syntax message; a friendly pre-flight check can be a separate PR if you'd prefer that.

Happy to expand either here or in a follow-up, whichever matches your preference.

Closes #904

_This contribution was developed with AI assistance (Codex)._

### DIFF
--- a/mbtiles/src/bin/mbtiles.rs
+++ b/mbtiles/src/bin/mbtiles.rs
@@ -172,6 +172,9 @@ pub struct SharedCopyOpts {
     /// When copying tiles only, the agg_tiles_hash will still be updated unless --skip-agg-tiles-hash is set.
     #[arg(long, value_name = "TYPE", default_value_t=CopyType::default())]
     copy: CopyType,
+    /// Use SQLite STRICT tables when creating a new destination file.
+    #[arg(long)]
+    strict: bool,
     /// Output format of the destination file, ignored if the file exists. If not specified, defaults to the type of source
     #[arg(long, alias = "dst-type", alias = "dst_type", value_name = "SCHEMA")]
     mbtiles_type: Option<MbtTypeCli>,
@@ -227,6 +230,7 @@ impl SharedCopyOpts {
             skip_agg_tiles_hash: self.skip_agg_tiles_hash,
             force: self.force,
             validate: self.validate,
+            strict: self.strict,
             // Constants
             dst_type: None, // Taken from dst_type_cli
         }
@@ -423,6 +427,25 @@ mod tests {
                     options: SharedCopyOpts {
                         min_zoom: Some(1),
                         max_zoom: Some(100),
+                        ..Default::default()
+                    },
+                    ..Default::default()
+                })
+            }
+        );
+    }
+
+    #[test]
+    fn test_copy_strict_argument() {
+        assert_eq!(
+            Args::parse_from(["mbtiles", "copy", "src_file", "dst_file", "--strict"]),
+            Args {
+                verbose: false,
+                command: Copy(CopyArgs {
+                    src_file: PathBuf::from("src_file"),
+                    dst_file: PathBuf::from("dst_file"),
+                    options: SharedCopyOpts {
+                        strict: true,
                         ..Default::default()
                     },
                     ..Default::default()
--- a/mbtiles/src/bin/mbtiles.rs
+++ b/mbtiles/src/bin/mbtiles.rs
@@ -167,7 +167,7 @@ pub struct DiffArgs {
     reason = "for command line arguments, formatting `TileJSON` is awkward"
 )]
 #[derive(Clone, Default, PartialEq, Debug, clap::Args)]
-#[allow(clippy::struct_excessive_bools)]
+#[expect(clippy::struct_excessive_bools, reason = "CLI interface")]
 pub struct SharedCopyOpts {
     /// Limit what gets copied.
     /// When copying tiles only, the agg_tiles_hash will still be updated unless --skip-agg-tiles-hash is set.
--- a/mbtiles/src/bin/mbtiles.rs
+++ b/mbtiles/src/bin/mbtiles.rs
@@ -167,12 +167,13 @@ pub struct DiffArgs {
     reason = "for command line arguments, formatting `TileJSON` is awkward"
 )]
 #[derive(Clone, Default, PartialEq, Debug, clap::Args)]
+#[allow(clippy::struct_excessive_bools)]
 pub struct SharedCopyOpts {
     /// Limit what gets copied.
     /// When copying tiles only, the agg_tiles_hash will still be updated unless --skip-agg-tiles-hash is set.
     #[arg(long, value_name = "TYPE", default_value_t=CopyType::default())]
     copy: CopyType,
-    /// Use SQLite STRICT tables when creating a new destination file.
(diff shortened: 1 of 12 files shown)


mbtiles/src/bin/mbtiles.rs
```diff
@@ -167,11 +167,15 @@ pub struct DiffArgs {
     reason = "for command line arguments, formatting `TileJSON` is awkward"
 )]
 #[derive(Clone, Default, PartialEq, Debug, clap::Args)]
+#[allow(clippy::struct_excessive_bools)]
```
CommanderStorm: to document this better

```suggestion
#[expect(clippy::struct_excessive_bools, reason = "CLI interface")]
```

mbtiles/src/queries.rs
```diff
@@ -187,54 +187,98 @@ where
     Ok(sql.fetch_one(&mut *conn).await?.is_valid == 1)
 }
 
+#[must_use]
+pub(crate) fn maybe_make_table_strict(sql: &str, strict: bool) -> String {
```
CommanderStorm: this seems like a weird choice how to implemnt this, especially how this is guardiing a case that cannot happen.

Please inline this and use a `format!("..{s};", s = if strict {" strict"} else { "" })` to make the query more readable instead

mbtiles/src/queries.rs
```diff
@@ -187,54 +187,98 @@ where
...
+}
+
 pub async fn create_metadata_table<T>(conn: &mut T) -> MbtResult<()>
+where
+    for<'e> &'e mut T: SqliteExecutor<'e>,
+{
+    create_metadata_table_with_strict(conn, false).await
+}
+
+pub(crate) async fn create_metadata_table_with_strict<T>(
```
CommanderStorm: Lets break the API here instead. It is less code and likely the api that we should have had in the first place.
It is what is likely wanted and this is sort of an internal-ish thing.

This also affects the fn's below that were changed in a similar way

```suggestion
pub async fn create_metadata_table<T>(
```

mbtiles/src/copier.rs
```diff
@@ -1072,6 +1097,105 @@ mod tests {
...
+        assert!(
+            get_table_sql(&mut dst_conn, "tiles")
+                .await
+                .contains(" STRICT")
+        );
+        assert!(
+            get_table_sql(&mut dst_conn, "bsdiffraw")
+                .await
+                .contains(" STRICT")
+        );
```
CommanderStorm: I would prefer we assert these things using `insta::assert_snapshot!(.., @"..")`

CommanderStorm (changes requested):
Mostly LGTM.
There are a few cases which seem "a bit vibed" (the level you care about backwards-compat in a non-core api), but nothing outrageous.

Should be simple to clean up and get merged

mvanhorn:
@CommanderStorm addressed all four in c8746df0:

- mbtiles.rs:170 - `expect(clippy::struct_excessive_bools, reason = "CLI interface")`.
- queries.rs:191 - dropped `maybe_make_table_strict`. Each DDL inlines `if strict { " STRICT" } else { "" }` via format! as you suggested.
- queries.rs:211 - broke the API. `create_metadata_table`, `create_flat_tables`, `create_flat_with_hash_tables`, `create_bsdiffraw_tables`, `create_normalized_tables`, and `init_mbtiles_schema` all take `strict: bool` directly. The `_with_strict` aliases are gone. All in-repo callers updated to pass false where strict does not apply.
- copier.rs:1196 - converted the three contains-check asserts in `diff_with_bindiff_uses_strict_patch_tables_when_requested` to inline `insta::assert_snapshot!`. Did the matching rewrite for `copy_same_type_uses_strict_tables_when_requested` and `copy_same_type_keeps_non_strict_tables_by_default` so the three related tests use the same style.

cargo test --workspace, cargo clippy --all-targets -D warnings, and cargo fmt --check all green locally.

On the "a bit vibed" bit - appreciate the honesty. The wrapper-and-alias pattern was pure AI reflex (keep the old signature, add a new one) rather than a real backcompat need. Gone now.

CommanderStorm (approved):
thank you

PR #2717, opened 2026-04-20
feat: Add MBTilesReloader (Tile Reload Phase 2)

Auspicus:
todo:

- [x] cleanup
- [x] `tests/test.sh` end to end test of moving files in and out
- [x] update docs/...

### DIFF
--- a/docs/content/sources-files.md
+++ b/docs/content/sources-files.md
@@ -62,16 +62,16 @@ MBTiles and PMTiles are both formats for storing tiled geospatial data, but they
 
 ### Key Differences
 
-* **Deployment model**
+- **Deployment model**
   **MBTiles** archives must be stored locally on the same machine as the tile server.
   **PMTiles** archives can be accessed either locally or remotely via HTTP range requests, e.g. from an object storage like S3.
   **PMTiles** allows simpler production deployment with Kubernetes, as it allows the large data file to reside in S3 and shared by multiple pods, but restricted from direct user access.
-* **Performance**
+- **Performance**
   **MBTiles** typically provide slightly lower latency due to local access and SQLite indexing.
   **PMTiles** may introduce additional latency when accessed remotely, but this is usually mitigated by HTTP caching and CDN usage.
-* **Storage efficiency**
+- **Storage efficiency**
   **PMTiles** archives are generally more space-efficient, typically ~10–15% smaller than equivalent **MBTiles** archives.
-* **Memory usage**
+- **Memory usage**
   **MBTiles** relies on SQLite, which maintains an internal page cache and may increase memory usage under load.
   **PMTiles** can, in some cases, operate with lower memory overhead, depending on access patterns and caching configuration.
 
@@ -79,21 +79,21 @@ MBTiles and PMTiles are both formats for storing tiled geospatial data, but they
 
 PMTiles archives can be served directly from HTTP range–capable storage without a dedicated tile server. This approach has several limitations:
 
-* **Unrestricted access risk**
+- **Unrestricted access risk**
   Without proper access controls, clients may download large portions (or all) of an archive, leading to significant egress costs.
   A tile server restricts access to tile requests, but bulk extraction remains possible via many requests, which are generally easier to detect and block.
-* **Over-fetching**
+- **Over-fetching**
   PMTiles may fetch more data than strictly required per tile request to minimize the number of HTTP requests.
-* **Lack of source composition**
+- **Lack of source composition**
   Direct serving does not support combining PMTiles with dynamic data sources (e.g., PostGIS) into a unified tile service.
   A tile server (e.g, Martin) is required for this.
-* **Caching behavior**
+- **Caching behavior**
   Cache efficiency may be reduced compared to setups with a dedicated tile server that can optimize request patterns.
 
 The choice between MBTiles and PMTiles depends on system requirements:
 
-* Use **MBTiles** for local, self-contained deployments with minimal external dependencies.
-* Use **PMTiles** for cloud-native or distributed setups where remote access, CDN integration, or object storage is preferred.
+- Use **MBTiles** for local, self-contained deployments with minimal external dependencies.
+- Use **PMTiles** for cloud-native or distributed setups where remote access, CDN integration, or object storage is preferred.
 
 ### Serving PMTiles from local file systems, http or Object Storage
 
--- a/docs/content/sources-files.md
+++ b/docs/content/sources-files.md
@@ -48,7 +48,10 @@ The following events are handled automatically:
 
 - **File added** — the new source appears in the catalog.
 - **File modified** — the source is reloaded and its tile cache is invalidated.
-- **File removed** — the source is removed from the catalog.
+- **File added** - the new source appears in the catalog.
+- **File modified** - the source is reloaded and its tile cache is invalidated.
+  Not avaliable on windows due to OS-limtations (SQLite not allowing `FILE_SHARE_DELETE`).
+- **File removed** - the source is removed from the catalog.
 
 !!! note
     Hot reload applies to directories configured under `mbtiles.paths` (or passed on the CLI). Named sources listed under `mbtiles.sources` are snapshotted at startup and are not watched for changes.
--- a/docs/content/sources-files.md
+++ b/docs/content/sources-files.md
@@ -27,8 +27,34 @@ martin  /path/to/directory
     For remote PMTiles, we don't currently support auto-discovery.
     If you want to implement this feature, please see <https://github.com/maplibre/martin/issues/2180>
 
-    We also don't currently support refreshing the catalog at runtime.
-    If you want to implement this feature, please see <https://github.com/maplibre/martin/issues/288> instead.
+### MBTiles Hot Reload
+
+Martin watches directories configured under `mbtiles` for changes at runtime. When `.mbtiles` files are added, modified, or removed from a watched directory, Martin automatically updates the tile catalog — no restart required.
+
+```bash
+# Martin will watch this directory and reflect any *.mbtiles changes live
+martin  /path/to/mbtiles/directory
+```
+
+Or via config file:
+
+```yaml
+mbtiles:
+  paths:
+    - /path/to/mbtiles/directory
+```
+
+The following events are handled automatically:
+
+- **File added** — the new source appears in the catalog.
+- **File modified** — the source is reloaded and its tile cache is invalidated.
+- **File removed** — the source is removed from the catalog.
+
+!!! note
+    Hot reload applies to directories configured under `mbtiles.paths` (or passed on the CLI). Named sources listed under `mbtiles.sources` are snapshotted at startup and are not watched for changes.
+
+!!! note
+    PMTiles hot reload is not yet supported. If you want to help implement it, see <https://github.com/maplibre/martin/issues/2180>.
 
 ## MBTiles vs PMTiles
 
--- a/docs/content/sources-files.md
+++ b/docs/content/sources-files.md
@@ -46,8 +46,6 @@ mbtiles:
 
 The following events are handled automatically:
 
-- **File added** — the new source appears in the catalog.
-- **File modified** — the source is reloaded and its tile cache is invalidated.
 - **File added** - the new source appears in the catalog.
 - **File modified** - the source is reloaded and its tile cache is invalidated.
   Not avaliable on windows due to OS-limtations (SQLite not allowing `FILE_SHARE_DELETE`).
--- a/martin/src/bin/martin.rs
+++ b/martin/src/bin/martin.rs
@@ -45,13 +45,11 @@ async fn start(args: Args) -> MartinResult<()> {
         if let FileConfigEnum::Config(cfg) = config.mbtiles.clone() {
             let mgr = sources.tile_manager.clone();
             let reloader = MBTilesReloader::new(mgr.id_resolver(), BTreeMap::new());
-            tokio::spawn(async move {
-                if let Err(e) = reloader
-                    .watch(&mgr, cfg.paths.clone().into_iter().to_owned().collect())
-                {
-                    warn!("failed to stop MBTilesReloader {e:?}")
-                }
-            });
+            if let Err(e) = reloader
+                .watch(mgr, cfg.paths.clone().into_iter().to_owned().collect())
+            {
+                warn!("failed to stop MBTilesReloader {e:?}")
+            }
         }
     }
 
--- a/martin/src/bin/martin.rs
+++ b/martin/src/bin/martin.rs
@@ -47,10 +47,12 @@ async fn start(args: Args) -> MartinResult<()> {
 
     #[cfg(feature = "mbtiles")]
     {
-        let reloader = martin::config::file::reload::mbtiles::MBTilesReloader::new(mgr, resolver, &config.mbtiles);
-        if let Err(e) =
-            reloader.start()
-        {
+        let reloader = martin::config::file::reload::mbtiles::MBTilesReloader::new(
+            mgr,
+            resolver,
+            &config.mbtiles,
+        );
+        if let Err(e) = reloader.start() {
             tracing::warn!("failed to start MBTilesReloader {e:?}")
         }
     }
--- a/martin/src/bin/martin.rs
+++ b/martin/src/bin/martin.rs
@@ -39,8 +39,8 @@ async fn start(args: Args) -> MartinResult<()> {
     #[cfg(feature = "mbtiles")]
     {
         use martin::config::file::{FileConfigEnum, reload::mbtiles::MBTilesReloader};
-        use std::{collections::BTreeMap, path::PathBuf};
         use std::time::UNIX_EPOCH;
+        use std::{collections::BTreeMap, path::PathBuf};
         use tracing::warn;
 
         let mut watch_paths: Vec<PathBuf> = vec![];
--- a/martin/src/bin/martin.rs
+++ b/martin/src/bin/martin.rs
@@ -36,6 +36,25 @@ async fn start(args: Args) -> MartinResult<()> {
     #[cfg(feature = "_catalog")]
     let sources = config.resolve().await?;
 
+    #[cfg(feature = "mbtiles")]
+    {
+        use martin::config::file::{FileConfigEnum, reload::mbtiles::MBTilesReloader};
+        use std::collections::BTreeMap;
+        use tracing::warn;
+
+        if let FileConfigEnum::Config(cfg) = config.mbtiles.clone() {
+            let mgr = sources.tile_manager.clone();
+            let reloader = MBTilesReloader::new(mgr.id_resolver(), BTreeMap::new());
+            tokio::spawn(async move {
+                if let Err(e) = reloader
+                    .watch(&mgr, cfg.paths.clone().into_iter().to_owned().collect())
+                {
+                    warn!("failed to stop MBTilesReloader {e:?}")
+                }
+            });
+        }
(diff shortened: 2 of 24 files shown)


martin/src/bin/martin.rs
```diff
@@ -33,8 +37,71 @@ async fn start(args: Args) -> MartinResult<()> {
...
+    #[cfg(feature = "_tiles")]
+    let resolver = IdResolver::new(RESERVED_KEYWORDS);
+
     #[cfg(feature = "_catalog")]
-    let sources = config.resolve().await?;
+    let sources = config.resolve(&resolver).await?;
+    let mgr = sources.tile_manager.clone();
+
+    #[cfg(feature = "mbtiles")]
+    {
```
CommanderStorm, line 49: can we move all of this logic to `MBTilesReloader::new`?
This method is alredy fairly complex, so this way we can keep complexity in check

martin/src/bin/martin.rs
```diff
@@ -33,8 +37,71 @@ async fn start(args: Args) -> MartinResult<()> {
...
+                        m.insert(
+                            id.clone(),
+                            (path.clone(), unix_epoch_delta.as_millis() as u64),
+                        );
+                    }
+                }
+                m
+            }
+            _ => BTreeMap::new(),
+        };
```
CommanderStorm, line 94: I think this collecting from the FS will be a relatively common pattern, at least for COG and PMTiles we will need to use something like this.

Lets extract the usefull part (checking one path? Very open to oppinions) to a method.

martin/src/bin/martin.rs
```diff
@@ -33,8 +37,71 @@ async fn start(args: Args) -> MartinResult<()> {
...
+                        let path = src.get_path();
+                        let Ok(metadata) = path.metadata() else {
+                            continue;
+                        };
+                        let Ok(modified) = metadata.modified() else {
+                            continue;
+                        };
+                        let Ok(unix_epoch_delta) = modified.duration_since(UNIX_EPOCH) else {
+                            continue;
+                        };
```
CommanderStorm, line 83: I am not sure if this is the right handling here.
At least we should log here that we are skipping this path and not registering it.

Otherwise a typo would be quite easy to trip you up.
Think about the assumption that if I configure `martin /foo/bar` that it would watch that directory.

We need to at least log about
- that we ignore it because it does not exist
- but will continue watching that path for changes.

martin/src/config/file/tiles/reload/mbtiles.rs
```diff
@@ -0,0 +1,157 @@
...
+        .map_err(|e| MartinError::DirectoryWatchError(e.kind))?;
+
+        for dir in directories
+            .iter()
+            .map(|d| d.canonicalize())
+            .flatten()
+            .filter(|d| d.is_absolute())
+        {
+            watcher
+                .watch(&dir.clone(), notify::RecursiveMode::NonRecursive)
```
CommanderStorm, line 49: why `NonRecursive`?
I somewhat think that this would be a more "expected" default, or?

```suggestion
                .watch(&dir.clone(), notify::RecursiveMode::Recursive)
```

martin/src/config/file/tiles/reload/mbtiles.rs
```diff
@@ -0,0 +1,157 @@
...
+            sources: initial_sources,
+        }
+    }
+
+    pub fn watch(mut self, tsm: TileSourceManager, directories: Vec<PathBuf>) -> MartinResult<()> {
+        let (tx, mut rx) = mpsc::channel::<Event>(256);
+        let mut watcher = RecommendedWatcher::new(
+            move |result: notify::Result<Event>| {
+                if let Ok(event) = result {
+                    let _ = tx.blocking_send(event);
```
CommanderStorm, line 35: why `blocking_send` instead of `send`?
I somewat feel like both are resonable here..

martin/src/config/file/tiles/reload/mbtiles.rs
```diff
@@ -0,0 +1,157 @@
...
+            .map(|d| d.canonicalize())
+            .flatten()
+            .filter(|d| d.is_absolute())
+        {
+            watcher
+                .watch(&dir.clone(), notify::RecursiveMode::NonRecursive)
+                .map_err(|e| MartinError::DirectoryWatchError(e.kind))?;
+        }
+
+        tokio::spawn(async move {
```
CommanderStorm, line 53: plese move this to 1-2 internal functions.

This fn is fairly long otherwise

martin/src/config/file/tiles/reload/mbtiles.rs
```diff
@@ -0,0 +1,157 @@
...
+                    };
+                    let Ok(modified) = metadata.modified() else {
+                        continue;
+                    };
+                    let Ok(unix_epoch_delta) = modified.duration_since(UNIX_EPOCH) else {
+                        continue;
+                    };
+
+                    let id = self.id_resolver.resolve(stem, path_str.clone());
+                    out.insert(id, (path, unix_epoch_delta.as_millis() as u64));
```
CommanderStorm, line 150: instead of `as_millis() as u64`, lets do `as_secs()` or stick with `as_millis` and change this to `u128`.
Both are fine by me, but truncating like this sounds longer-term buggy.

martin/src/bin/martin.rs
```diff
@@ -33,8 +37,71 @@ async fn start(args: Args) -> MartinResult<()> {
...
+                }
+                m
+            }
+            _ => BTreeMap::new(),
+        };
+
+        if !watch_paths.is_empty() {
+            let reloader = MBTilesReloader::new(resolver, init);
+            if let Err(e) =
+                reloader.watch(mgr, watch_paths.clone().into_iter().to_owned().collect())
```
CommanderStorm, line 99: Does this cancel properly or do we need an `tokio_util::sync::CancellationToken` properly?

CommanderStorm (changes requested):
Very clean. Good job.

I have a few comments (a bit more nit-picky than usual, but we are going to model the rest of this so I hope this is OK with you).

The biggest ask that I have:
- Please add a test in `tests/test.sh` that auto-discovery works as intended with add, edit, move.
- Please document this behaviour in our docs

Also, I don't nessarily think that we need an `unstable-reloader` feature..
LGTM if it works properly..

martin/src/bin/martin.rs
Auspicus: 100% agree

martin/src/config/file/tiles/reload/mbtiles.rs
Auspicus: Does the Martin tile resolution work recursively? I remember looking and I thought it didn't so I kept this as NonRecursive to match that behaviour but I could be wrong.

Auspicus:
Thanks for the swift feedback @CommanderStorm! I'll get on this again some time this week.

martin/src/config/file/tiles/reload/mbtiles.rs
CommanderStorm: Current version might not be recursive, but maybe it should? Not sure..
Likely being recursive is what you would expect, or?

Lets ask on slack..

CommanderStorm:
```suggestion
                // FIXME: find a naming scheme for paths that makes sense under recursive and enable it
                .watch(&dir.clone(), notify::RecursiveMode::NonRecursive)
```

martin/src/config/file/tiles/reload/mbtiles.rs
```diff
@@ -0,0 +1,205 @@
...
+
+        MBTilesReloader {
+            id_resolver,
+            sources,
+            directories,
+        }
+    }
+
+    pub fn watch(mut self, tsm: TileSourceManager) -> MartinResult<()> {
+        let (tx, mut rx) = mpsc::channel::<Event>(256);
```
Auspicus, line 108: Picked a random power of 2 here that seemed large enough to absorb temporary bursts. Could be tweaked.

martin/src/config/file/tiles/reload/mbtiles.rs
Auspicus: I also gawked at this at first, but I think it's OK. This is a closure passed to the watcher so I'm not sure what making it async (calling `send`) would mean since it's passed to the underlying file-system watcher implementation outside of Tokio. It appears `inotify` starts a thread running an event loop blocked on `mio` to run this function. https://github.com/notify-rs/notify/blob/main/notify/src/inotify.rs#L201-L224

We're also using the cross-platform file-system watcher here. There's an argument to just support this on Linux since file-system watching is so different on each OS. I think using it only as a signal to discover sources within watched directories is simple enough that it won't run into cross platform quirks but another thing to consider.

martin/src/bin/martin.rs
Auspicus: I've moved almost all of this into the `MBTilesReloader::new` now. Opting to only pass the `&config.mbtiles`

martin/src/config/file/tiles/reload/mbtiles.rs
Auspicus: Created a separate function `process_event` to handle the reading of the directories and diffing against previous list of sources and versions.

martin/src/config/file/tiles/reload/mbtiles.rs
```diff
@@ -0,0 +1,206 @@
...
+        let adv = ReloadAdvisory::from_maps(
+            &prev,
+            &next,
+            async move |id| -> MartinResult<BoxedSource> {
+                let p = sources_clone.get(&id).ok_or(
+                    MartinError::DirectoryWatchError(notify::ErrorKind::Generic(
+                        format!("Source {id} not found in discovered sources"),
+                    )),
+                )?;
+                let src = MbtSource::new(id, p.0.clone(), CacheZoomRange::default()).await?;
```
Auspicus: todo: need to pipe this cache config in from configuration.

martin/src/error.rs
```diff
@@ -72,9 +72,6 @@ pub enum MartinError {
     #[error(transparent)]
     SpriteError(#[from] martin_core::sprites::SpriteError),
 
-    #[error(transparent)]
-    WebError(#[from] actix_web::Error),
```
Auspicus, line 76: Removed this so that `MartinError` can be `Send + Sync`. As far as I can tell it was unused, however it affects `martin-cp` (see related changes there). From my perspective, it's strange that `martin-cp` a CLI would contain references to Actix since it doesn't start an Actix web server.

martin/src/bin/martin.rs
Auspicus: This has been changed slightly (renamed to `start` and no longer accepts any arguments) but all this does is setup the file system watcher and it's not async, it returns once `watch` has been called on all the configured directories. If the `reloader` is dropped the event handler function will stop being called. With that, I'm not sure we need a `CancellationToken` but maybe I'm missing something.

martin/src/config/file/tiles/reload/mbtiles.rs
```diff
@@ -0,0 +1,219 @@
...
+        let next: BTreeMap<String, u128> =
+            sources.iter().map(|(k, v)| (k.clone(), v.1)).collect::<_>();
+        let sources_clone = sources.clone();
+
+        let adv =
+            ReloadAdvisory::from_maps(&prev, &next, async move |id| -> MartinResult<BoxedSource> {
+                let p = sources_clone
+                    .get(&id)
+                    .ok_or(MartinError::DirectoryWatchError(
+                        notify::ErrorKind::Generic(format!(
```
Auspicus: This shouldn't use the `notify::ErrorKind::Generic`. This should never fail but needs to convert to a `MartinResult`.

martin/src/bin/martin.rs
Auspicus: All (rare) skip errors are logged as warnings now.

docs/content/sources-files.md
```diff
@@ -27,47 +27,73 @@ martin  /path/to/directory
...
+```
+
+The following events are handled automatically:
+
+- **File added** — the new source appears in the catalog.
+- **File modified** — the source is reloaded and its tile cache is invalidated.
+- **File removed** — the source is removed from the catalog.
+
+!!! note
+    Hot reload applies to directories configured under `mbtiles.paths` (or passed on the CLI). Named sources listed under `mbtiles.sources` are snapshotted at startup and are not watched for changes.
```
CommanderStorm, line 55: 🤔 likely we can get rid of this by reshuffling some code.
 Lets see if we can get rid of this.

Alas, I am fine with pushing this to a future PR or until someone has time to do this redhuffeling

martin/src/bin/martin.rs
```diff
@@ -33,8 +37,29 @@ async fn start(args: Args) -> MartinResult<()> {
...
+        )
+        .await?;
+    #[cfg(feature = "mbtiles")]
+    let mgr = sources.tile_manager.clone();
+
+    #[cfg(feature = "mbtiles")]
+    {
+        use martin::config::file::reload::mbtiles::MBTilesReloader;
+
+        let reloader = MBTilesReloader::new(mgr, resolver, &config.mbtiles);
```
CommanderStorm:
```suggestion
        let reloader = MBTilesReloader::new(mgr, resolver, &config.mbtiles);
```

martin/src/bin/martin.rs
```diff
@@ -6,8 +6,12 @@ use martin::config::args::Args;
 #[cfg(all(feature = "webui", not(docsrs)))]
 use martin::config::args::WebUiMode;
 use martin::config::file::{Config, read_config};
+#[cfg(feature = "_tiles")]
+use martin::config::primitives::IdResolver;
 use martin::config::primitives::env::OsEnv;
 use martin::logging::{ensure_martin_core_log_level_matches, init_tracing};
+#[cfg(feature = "_tiles")]
+use martin::srv::RESERVED_KEYWORDS;
```
CommanderStorm, line 16:
```suggestion
use martin::srv::RESERVED_KEYWORDS;
#[cfg(feature = "mbtiles")]
use martin::config::file::reload::mbtiles::MBTilesReloader;
```

martin/src/config/file/tiles/reload/mbtiles.rs
```diff
@@ -0,0 +1,219 @@
...
+use crate::config::primitives::{IdResolver, OptOneMany};
+use crate::{MartinError, MartinResult, ReloadAdvisory, TileSourceManager};
+
+pub struct MBTilesReloader {
+    /// ID resolver that ensures a unique ID is assigned to each source.
+    id_resolver: IdResolver,
+    /// Tile Source Manager to which we should send `ReloadAdvisory` messages.
+    tile_source_manager: TileSourceManager,
+    /// Map of Source ID => (path, modified timestamp, cache policy).
+    sources: BTreeMap<String, (PathBuf, u128, CachePolicy)>,
```
CommanderStorm, line 23: nit: I usually prefer a struct for these things. Makes it clearer what the u128 is refering to for others, but I am fine as-is ^^

CommanderStorm (approved):
Impressive work.
I have a few passing remarks, but let's merge

CommanderStorm:
CI needs to be made happy though..

tests/test.sh
```diff
@@ -960,6 +1003,50 @@ else
...
+
+>&2 echo "Test reload: updating an MBTiles file triggers source update"
+touch "$RELOAD_WATCH_DIR/world_cities.mbtiles"
+wait_for_log_str "$LOG_FILE" 'Updated source: "world_cities"'
+test_jsn reload_catalog_updated catalog
+
+>&2 echo "Test reload: removing an MBTiles file triggers source removal"
+rm "$RELOAD_WATCH_DIR/world_cities.mbtiles"
+wait_for_catalog_source_removed "world_cities"
+$CURL "$MARTIN_URL/catalog" | jq --sort-keys > "$TEST_OUT_DIR/catalog_after_remove.json"
```
CommanderStorm: I think this should solve it

```suggestion
>&2 echo "Test reload: removing an MBTiles file triggers source removal"

if [[ "$OSTYPE" == cygwin* || "$OSTYPE" == msys* || "$OSTYPE" == win32* ]]; then
  # Sqlite asks windows to not deltete files while it is reading it -> this fails on windows
  # We cannot get around this
  # So for windows, we need to kill martin and re-create it, thus "faking it"
  kill_process "$MARTIN_PROC_ID" Martin
  rm "$RELOAD_WATCH_DIR/world_cities.mbtiles"
  MARTIN_BIN "${ARG[@]}" 2>&1 | tee "$LOG_FILE" &
  MARTIN_PROC_ID=$(jobs -p | tail -n 1)
else
  rm "$RELOAD_WATCH_DIR/world_cities.mbtiles"
fi

wait_for_catalog_source_removed "world_cities"
$CURL "$MARTIN_URL/catalog" | jq --sort-keys > "$TEST_OUT_DIR/catalog_after_remove.json"
```

docs/content/sources-files.md
Auspicus: Yeah, my main hesitation to doing this now was that I don't know what the _right_ way to determine the directories for watching should be. Right now, it's reasonable to assume if you configure a directory for Martin to gather sources from that you probably want Martin to also notice when new sources become available in that directory. With named sources, I still have the following questions about the _right_ directories to watch:
- should we watch directories of the named sources?
- should we only watch the immediate parent of the source?
- if we watch more than the immediate parent, how far above the source should we watch?

martin/src/config/file/tiles/reload/mbtiles.rs
Auspicus: Maybe can be addressed in the follow up to add support for the other local file tile sources.

Auspicus:
@CommanderStorm Windows is passing now (using a dirty hack). I wish we could simulate a real delete but it looks like SQLite has no way to add a FILE_SHARE_DELETE permission so faking the behaviour feels wrong but accurate since it _shouldn't_ be possible on Windows.

docs/content/sources-files.md
```diff
@@ -27,47 +27,73 @@ martin  /path/to/directory
...
+mbtiles:
+  paths:
+    - /path/to/mbtiles/directory
+```
+
+The following events are handled automatically:
+
+- **File added** — the new source appears in the catalog.
+- **File modified** — the source is reloaded and its tile cache is invalidated.
+- **File removed** — the source is removed from the catalog.
```
CommanderStorm:
```suggestion
- **File added** - the new source appears in the catalog.
- **File modified** - the source is reloaded and its tile cache is invalidated.
  Not avaliable on windows due to OS-limtations (SQLite not allowing `FILE_SHARE_DELETE`).
- **File removed** - the source is removed from the catalog.
```

docs/content/sources-files.md
```diff
@@ -27,47 +27,76 @@ martin  /path/to/directory
...
+```
+
+The following events are handled automatically:
+
+- **File added** — the new source appears in the catalog.
+- **File modified** — the source is reloaded and its tile cache is invalidated.
+- **File added** - the new source appears in the catalog.
+- **File modified** - the source is reloaded and its tile cache is invalidated.
+  Not avaliable on windows due to OS-limtations (SQLite not allowing `FILE_SHARE_DELETE`).
+- **File removed** - the source is removed from the catalog.
```
CommanderStorm:
```suggestion
- **File added** - the new source appears in the catalog.
- **File modified** - the source is reloaded and its tile cache is invalidated.
  Not avaliable on windows due to OS-limtations (SQLite not allowing `FILE_SHARE_DELETE`).
- **File removed** - the source is removed from the catalog.
```

PR #2726, opened 2026-04-22
feat: enable rendering feature by default on Linux

CommanderStorm:
mostly to see what happens to our build times and our binary size when run wiht this in the default set in CI.

Docker size seems to go up by 50MB -> 500MB.
Not sure if this is breaking for some..


martin-core/src/resources/styles/mod.rs
```diff
@@ -207,37 +207,144 @@ mod tests {
...
+                std::fs::write(&reference_path, &rendered_bytes).unwrap();
+                panic!(
+                    "Created new reference image at {reference_path:?}. Commit this file and re-run the test."
+                );
+            }
+            panic!(
+                "Reference image not found at {reference_path:?}: {e}. \
+                 Run with BLESS_RENDERING=1 to generate it."
+            );
+        });
```
CommanderStorm:
```suggestion
        // Reference image MUST exist - if missing, the test fails and stores it on disk
        let reference_bytes = std::fs::read(&reference_path).unwrap_or_else(|e| {
                std::fs::create_dir_all(reference_path.parent().unwrap()).unwrap();
                // Sanity check: refuse to bless tiny or blank images
                assert!(
                    rendered_bytes.len() > 1000,
                    "Refusing to bless suspiciously small image ({} bytes)",
                    rendered_bytes.len()
                );
                std::fs::write(&reference_path, &rendered_bytes).unwrap();
                panic!(
                    "Created new reference image at {reference_path:?}. Commit this file and re-run the test."
                );
        });
```

martin-core/src/resources/styles/mod.rs
```diff
@@ -207,37 +207,137 @@ mod tests {
...
             z,
             x,
             y
         );
-        insta::assert_binary_snapshot!(&snapshot_name, img_buffer.into_inner());
+        let reference_path =
+            Path::new("../tests/fixtures/rendering_references").join(&reference_name);
+
+        // Reference image MUST exist - if missing, the test fails and stores it on disk
+        let reference_bytes = std::fs::read(&reference_path).unwrap_or_else(|e| {
```
CommanderStorm:
```suggestion
        let reference_bytes = std::fs::read(&reference_path).unwrap_or_else(|_| {
```

nyurik:
docker image now install dependencies like ` build-essential` -- which is large, but unlikely to be needed by the users. If they are needed in the docker container itself, we should use multi-pass docker image and copy results, or we may want to remove `build-essential` and possibly other

CommanderStorm:
there is no reference to `build-essential` in this PR.

Yes, `ubuntu` is a bit larger than `alpine`, but it comes with propper glibc, which seems to be required to run the gnu build MLN.
I think this is likely fine.

We can have a slim variant that does not include MLN, which is the large part to why this docker image is 400M.
As we asked on slack that does not seem to be an issue, so lets go without that complexity until someone asks for a slim variant.

nyurik:
ah, you are right, install-dependencies is only called from CI, not when forming the docker

Issue #852, opened 2023-08-30
Include Maputnik style editor into the Martin binary


nyurik:
[maputnik](https://github.com/maplibre/maputnik) is an excellent style editor (now part of maplibre), but it is often complicated to combine maputnik with martin due to CORS. It is possible to combine them via docker compose and an nginx reverse proxy, but that's also combersome and complicated for novice users.

I think we should use [actix_web_static_files](https://docs.rs/actix-web-static-files/latest/actix_web_static_files/) or similar to build+package maputnik as part of the `build.rs` script, and include it inside the martin binary.

sharkAndshark:
[maputnik](https://maputnik.github.io/) is awesome, I like it. But what's the reason to include it in `martin`?  
What's the link / integration between two parts?  Is something  in future like this?
* The included maputnik can show all sources in martin by default?
* The inclded maputnik can upload styles && assests to martin?

nyurik:
I think it would be awesome if maputnik would be able to show the catalog content, including at some point the list of sprites and once implemented - fonts.

I am not sure if maputnik should upload anything. I still see Maputnik as somewhat of an elaborate playground for martin data - e.g. you can try out what data is available, try to style it differently, experiment with it, but later you would still save the style to your own site... TBD i guess

NH-RedAnt:
I've been exploring both of these tools for a project I'm working on.

I'm having one odd problem using these tools together that I am about to generate an issue for.  Honestly I don't know which side has the problem.

My goto toolset is python.  To work around the CORS issue I created a tiny Flask app that really just serves up a static folder tree.  From it I serve a MapLibreGL view and the style.json files.  Enabling CORS lets me use everything locally to 3 different ports. 

By the way after alots of looking at various MBTile servers, martin really just worked out of the box.  Great job!  Nothing else works as well for me.

nyurik:
FYI, Maputnik has been moved to [maplibre](https://github.com/maplibre/maputnik), and so is the [maplibre-gl-inspect](https://github.com/maplibre/maplibre-gl-inspect), with lots of work being done on them.

mzeedijk:
Especially now that Martin is able to serve style files it would be a nice addition to include Maputnik so that Maputnik can list and load (and maybe even store in the future if styles can be POSTed to Martin) styles from Martin.

CommanderStorm:
In `v0.18.0` we will release a basic integration with Maputnik.
Said integration is quite basic and I think we should improve upon this.

[...]

PR #2731, opened 2026-04-23
feat(ui): Add Tile URLs TileJSON and XYZ Tiles URLs

jo-chemla:
See micro discussion here https://github.com/maplibre/martin/issues/852#issuecomment-4296464995

Adding TileJSON and XYZ urls and copy buttons, quickly drafted with Claude Sonnet 4.6

image

### DIFF
--- a/martin/martin-ui/src/components/dialogs/tile-inspect.tsx
+++ b/martin/martin-ui/src/components/dialogs/tile-inspect.tsx
@@ -9,11 +9,11 @@ import {
 } from '@/components/ui/dialog';
 import type { TileSource } from '@/lib/types';
 import '@maplibre/maplibre-gl-inspect/dist/maplibre-gl-inspect.css';
-import { buildMartinUrl } from '@/lib/api';
 import { Database, Link } from 'lucide-react';
+import { CopyableUrl } from '@/components/ui/copyable-url';
+import { buildMartinUrl } from '@/lib/api';
 import { LoadingSpinner } from '../loading/loading-spinner';
 import { TileInspectDialogMap } from './tile-inspect-map';
-import { CopyableUrl } from '@/components/ui/copyable-url';
 
 interface TileInspectDialogProps {
   name: string;
--- a/martin/martin-ui/src/components/dialogs/tile-inspect.tsx
+++ b/martin/martin-ui/src/components/dialogs/tile-inspect.tsx
@@ -44,6 +44,7 @@ function CopyableUrl({ label, url }: { label: string; url: string }) {
       <span className="flex items-center gap-2 mt-1">
         <code className="text-xs break-all flex-1">{url}</code>
         <button
+          type="button"
           onClick={handleCopy}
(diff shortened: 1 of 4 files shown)


martin/martin-ui/src/components/dialogs/tile-inspect.tsx
```diff
@@ -28,7 +28,41 @@ function TileMapLoading() {
...
+          title={`Copy ${label}`}
+        >
+          {copied ? (
+            <Check className="w-3.5 h-3.5 text-green-500" />
+          ) : (
+            <Copy className="w-3.5 h-3.5" />
+          )}
+        </button>
+      </span>
+    </p>
```
CommanderStorm: Can we do this via our existing utility functions instead?
Seems rather duplicated otherwise.

https://github.com/maplibre/martin/blob/01231d0a28e42a9be6f702d46046e3d1aa1566ef/martin/martin-ui/src/components/ui/copy-link-button.tsx#L25

Similarly, is there a reason why this is not usiing our `Button` element?

martin/martin-ui/src/components/dialogs/tile-inspect.tsx
```diff
@@ -28,7 +28,41 @@ function TileMapLoading() {
   );
 }
 
+function CopyableUrl({ label, url }: { label: string; url: string }) {
```
CommanderStorm: lets move this component to a different file, since I assume this will be somethin we would like to use elsewere as well

martin/martin-ui/src/components/dialogs/tile-inspect.tsx
jo-chemla: thanks for the comment, will fix, I did not know these components existed

CommanderStorm (approved):
Design and functionality LGTM
Code is a bit duplciated, but as long as we dedup this a bit LGTM as well.

jo-chemla:
Alright, addressed the comments and action errors, should be mergeable! 
Note there were 2 `<Toaster>`, one in each of these files, therefore the `URL Copied` message appears twice. 
 
```
martin-ui\src\components\dialogs\tile-inspect-map.tsx
martin-ui\src\components\dashboard-content.tsx
```

Do you think one should be removed? If so, I can do it as part of this PR

Edit: 
~~There is just a vi.mock error for the ExternalLink, not sure why and how to pass the test.~~ addressed

CommanderStorm:
> Note there were 2 <Toaster>, one in each of these files, therefore the URL Copied message appears twice.
> Do you think one should be removed? If so, I can do it as part of this PR

Yes, that would be appreciated ^^
If that is hard, we can also create an issue about this

jo-chemla:
Alright just did remove the extra Toaster, and `Lint and Unit Test the frontend` github-action did succeed, ready to merge from my point-of-view!

CommanderStorm:
Thanks for this work (I enabled automerge already earlier, no change in status)

PR #2733, opened 2026-04-24
refactor: Replace manual Debug impls with derive_debug::Dbg

CommanderStorm:
## Summary
- Replace hand-written `impl Debug for` blocks in `PmtilesSource` and `MbtSource` with `#[derive(Dbg)]` from the [`derive-debug`](https://crates.io/crates/derive-debug) crate
- Uses `#[dbg(skip)]` to omit fields and `#[dbg(alias)]` to rename fields, producing the exact same Debug output as before
- `SpriteCache` and `TileXyz` still use manual impls as they rely on computed values / custom formatting that can't be expressed with derive attributes

## Test plan
- [x] `cargo check --features mbtiles,pmtiles` passes
- [ ] CI passes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -791,7 +791,7 @@ jobs:
       attestations: write
       # for writing release artifacts
       contents: write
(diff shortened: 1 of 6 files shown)


nyurik (approved):

nyurik:
Note that you can use computed values with this crate too by using a helper function

PR #2735, opened 2026-04-25
feat: add a cache hit rate metric

CommanderStorm:
image
image

Should be fairly self-explainatory and nice to track


martin-core/src/resources/fonts/cache.rs
```diff
@@ -61,9 +61,15 @@ impl FontCache {
             .await?;
 
         if entry.is_fresh() {
+            crate::metrics::CACHE_REQUESTS_TOTAL
```
nyurik:
```suggestion
            CACHE_REQUESTS_TOTAL
```

nyurik (approved):
use the `use`, otherwise seems good

PR #2752, opened 2026-04-28
fix(styles): default the optional `rendering` field so configs without it parse

### LINKED ISSUE #2751: Config format change?
My config file hasn't changed. Yet for some reason I keep getting this when trying to use latest. Pinning to 1.6.0 works however

CommanderStorm:
## Summary

- Fixes #2751: a `styles:` block that does not explicitly set `rendering:` now parses on builds with the rendering feature enabled.
- Root cause: since #2726 flipped the rendering feature on by default on Linux, every published build compiles in the `rendering: OptBoolObj<RendererConfig>` field on `InnerStyleConfig`. The field was missing `#[serde(default)]`, so when `rendering` was absent from the YAML, all `FileConfigEnum` variants failed and serde reported `data did not match any variant of untagged enum FileConfigEnum`.
- The fix adds `#[serde(default, skip_serializing_if = "OptBoolObj::is_none")]` to the field — the same pattern already used for every other `OptBoolObj` in the codebase (e.g. `postgres/config.rs:71,90,92`).

## Reproduction (before the fix)

```bash
cat > /tmp/martin.yaml <<'YAML'
listen_addresses: '0.0.0.0:3000'
mbtiles:
  paths:
    - /data
styles:
  paths:
    - /data
YAML
docker run --rm -v /tmp/martin.yaml:/etc/martin/config.yaml \
  ghcr.io/maplibre/martin:latest --config /etc/martin/config.yaml
# ERROR martin: Unable to parse config file /etc/martin/config.yaml:
#   data did not match any variant of untagged enum FileConfigEnum
```

After the fix the same config parses and Martin proceeds to source resolution as expected.

## Test plan

- [x] Added a `serde_yaml::from_str` regression test (`test_styles_parse_paths_only_without_rendering_field`) for the bare `styles: { paths: [...] }` shape, so CI's rendering-enabled Linux test job will catch any future regression.
- [x] `cargo test --no-default-features --features fonts,lambda,mbtiles,metrics,pmtiles,postgres,sprites,styles,webui --package martin` — all green.
- [x] `cargo clippy` on the same feature set — clean.
- [ ] CI to verify with the rendering feature enabled on Linux (I cannot build maplibre_native locally in this environment).

### DIFF
--- a/martin/src/config/file/resources/styles.rs
+++ b/martin/src/config/file/resources/styles.rs
@@ -197,8 +197,6 @@ mod tests {
     use super::*;
     use crate::config::file::FileConfigSrc;
 
-    /// `styles: { paths: [...] }` without an explicit `rendering:` must parse, even when the
-    /// rendering feature compiles in the optional `rendering` field on `InnerStyleConfig`.
     #[test]
     fn test_styles_parse_paths_only_without_rendering_field() {
         let yaml = indoc! {"
--- a/martin/src/config/file/resources/styles.rs
+++ b/martin/src/config/file/resources/styles.rs
@@ -21,6 +21,7 @@ pub struct InnerStyleConfig {
     /// We are not currently happy with the performance of this endpoint and intend to improve this in the future
     /// Marking this experimental means that we are not stuck with single threaded performance as a default until v2.0
     #[cfg(all(feature = "rendering", target_os = "linux"))]
+    #[serde(default, skip_serializing_if = "OptBoolObj::is_none")]
     pub rendering: OptBoolObj<RendererConfig>,
 
     #[serde(flatten, skip_serializing)]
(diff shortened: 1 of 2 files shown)


martin/src/config/file/resources/styles.rs
```diff
@@ -191,9 +192,28 @@ fn is_hidden(entry: &walkdir::DirEntry) -> bool {
 
 #[cfg(test)]
 mod tests {
+    use indoc::indoc;
+
     use super::*;
     use crate::config::file::FileConfigSrc;
 
+    /// `styles: { paths: [...] }` without an explicit `rendering:` must parse, even when the
+    /// rendering feature compiles in the optional `rendering` field on `InnerStyleConfig`.
```
CommanderStorm:
```suggestion
```

CommanderStorm:
tagging a few more people to get this merged faster.
Was an unintentional breaking change

nyurik (approved):
thx

nyurik (approved):

PR #2755, opened 2026-04-28
refactor(mbtiles): migrate from log/env_logger to tracing

CommanderStorm:
log -> tracing for the mbtiles crate


nyurik (approved):
as before, performance is my only concern with all this

PR #2759, opened 2026-04-29
refactor: add #[tracing::instrument] to hot-path entry points

CommanderStorm:
## Summary

Adds function-level tracing spans (`#[tracing::instrument]`) to hot-path entry points across `martin` and `martin-core`. Field names follow the dotted hierarchical naming convention from #2757 (`source.id`, `tile.{z,x,y}`, `font.fontstack`, `font.range.{start,end}`, `style.id`, `sprite.{sdf,pixel_ratio}`, etc.) so the spans line up with the structured log fields once that PR lands.


nyurik:
first pass was +9%, second within +1.4% performance -- are you certain this approach won't significantly affect perf?

nyurik (commented):
i'm ok with this, just make sure there is no performance cost

nyurik (approved):
i'm ok with this, just make sure there is no performance cost

martin-core/src/resources/fonts/mod.rs
```diff
@@ -141,6 +141,16 @@ impl FontSources {
     /// Combines multiple fonts (comma-separated) with later fonts filling gaps.
     /// Range must be exactly 256 characters (e.g., 0-255, 256-511).
     #[expect(clippy::cast_possible_truncation)]
+    #[instrument(
+        level = "debug",
```
CommanderStorm, line 145: @nyurik they are all debug-only instrumentation, so will be skipped by tracings' cache at nearly no cost even at info level.

first run was with this at info, second with debug

nyurik: debug=only - is that log level or build option?

PR #2773, opened 2026-05-03
chore: add `AutoOption` for having something as auto, disabled or explicit

CommanderStorm:
This is the config struct to be able to aut, disable or explicitly confgure something.

Split from the others because this way we can unblock me working on the MVT converter

### DIFF
--- a/martin/src/config/primitives/auto_option.rs
+++ b/martin/src/config/primitives/auto_option.rs
@@ -76,7 +76,7 @@ impl<'de, T: Deserialize<'de>> Visitor<'de> for AutoOptionVisitor<T> {
 
     fn visit_str<E: de::Error>(self, v: &str) -> Result<Self::Value, E> {
         match v {
-            "auto" | "default" => Ok(AutoOption::Auto),
+            "auto" | "default" | "true" => Ok(AutoOption::Auto),
             "disabled" | "false" | "off" | "no" => Ok(AutoOption::Disabled),
             _ => Err(E::invalid_value(Unexpected::Str(v), &self)),
         }
@@ -107,7 +107,7 @@ impl<T: schemars::JsonSchema> schemars::JsonSchema for AutoOption<T> {
             "oneOf": [
                 {
                     "type": "string",
-                    "enum": ["auto", "default"],
+                    "enum": ["auto", "default", "true"],
                     "description": "Use the feature with default settings."
                 },
                 {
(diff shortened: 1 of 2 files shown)


martin/src/config/primitives/auto_option.rs
```diff
@@ -0,0 +1,191 @@
...
+        Ok(if v {
+            AutoOption::Auto
+        } else {
+            AutoOption::Disabled
+        })
+    }
+
+    fn visit_str<E: de::Error>(self, v: &str) -> Result<Self::Value, E> {
+        match v {
+            "auto" | "default" => Ok(AutoOption::Auto),
```
nyurik: should this also have "true"?

nyurik (approved):

martin/src/config/primitives/auto_option.rs
```diff
@@ -0,0 +1,191 @@
...
+                "Auto/disabled/explicit configuration.\n\n\
+                 - `\"auto\"` or `\"default\"` - use defaults\n\
+                 - `\"disabled\"`, `\"false\"`, `\"off\"`, `\"no\"` - disable\n\
+                 - An object - explicit {} settings",
+                T::schema_name()
+            ),
+            "oneOf": [
+                {
+                    "type": "string",
+                    "enum": ["auto", "default"],
```
CommanderStorm:
```suggestion
                    "enum": ["auto", "default", "true"],
```

martin/src/config/primitives/auto_option.rs
```diff
@@ -0,0 +1,191 @@
...
+
+    #[rstest]
+    #[case("auto", AutoOption::Auto)]
+    #[case("default", AutoOption::Auto)]
+    #[case("true", AutoOption::Auto)]
+    #[case("disabled", AutoOption::Disabled)]
+    #[case("off", AutoOption::Disabled)]
+    #[case("no", AutoOption::Disabled)]
+    #[case("false", AutoOption::Disabled)]
+    #[case("\"false\"", AutoOption::Disabled)]
```
CommanderStorm, line 151:
```suggestion
    #[case("\"false\"", AutoOption::Disabled)]
    #[case("\"true\"", AutoOption::Auto)]
```

martin/src/config/primitives/auto_option.rs
CommanderStorm:
```suggestion
            "auto" | "default" | "true" => Ok(AutoOption::Auto),
```

CommanderStorm: I guess.. the actual boolean is already handled. true as a string in addion I guess..

Issue #1364, opened 2024-06-09
Add png-image support when generating sprite


exotfboy:
As [martin](https://maplibre.org/martin/sources-sprites.html#sprite-sources) said:

> Given a directory with SVG images, Martin will generate a sprite – a JSON index and a PNG image, for both low and high resolution displays. 

Martin only support svg images for generating a sprite.  How about other image format like png,jpg?

mzeedijk:
As a temporary workaround you could convert a raster image to a svg file in a kind of brute force method.
Use png2svg for this. Use only small images because it creates rectangles for every square area of one color inside the raster image. Maybe use ImageMagick's convert first to convert and resize any raster image to a small sized PNG file first:

Example:
convert src.jpg -resize 320x320 dest.png
png2svg dest.png > dest.svg

CommanderStorm:
We use https://docs.rs/spreet/latest/spreet/ for building the sprites (they don't support pngs).

The problem with supporting pngs that I see is that they could not really be supported via png inputs..
Not shure how to handle these two interacting if we were to support this.

CommanderStorm:
I found this issue
- https://github.com/flother/spreet/issues/60


pngs are supported if they are written like this:

```svg
<svg xmlns="http://www.w3.org/2000/svg">
  <image href="./flag.png" />
</svg>
```

The dependency decided that they won't support this directly.

> Given that
> - vector input gives the best results,
> - there doesn't seem to be any demand for raster input in the original spritezero project (in fact they moved from raster to vector input in 2015 with v2.0.0),
> - my concerns about scaling above, and
> - support via SVG's <image> element, 

We could hallucinate said svg file.
Should be a fairly simple PR.

johncarmack1984:
spreet decided against raster input (flother/spreet#60), and the `<image href>` wrapper Frank showed above is the supported route. Nothing left to do on the Martin side, so I would close this.

CommanderStorm:
does not interact well with sdfs anyhow..

Issue #1336, opened 2024-05-11
Unable to generate spritesheet


gkinsman:
Hi there!

Kicking the tyres on this as an alternative to tileserver-gl-light and was pretty quick to get going which was great!

I'm having some issues getting sprites working, though my config is pretty simple:

```yaml
# Publish MBTiles files
mbtiles:
  sources:
    # named source matching source name to a single file
    greater-london: ../tiles/London/greater-london.mbtiles

# Sprite configuration
sprites:
  paths:
    - ./osm-liberty

```

Tiles load fine with this config, but attempts to load sprites cause this error in the martin console:

```
[2024-05-11T21:34:26Z ERROR martin::srv::server] Unable to generate spritesheet
```
when trying to navigate to `http://localhost:3000/sprite/osm-liberty.json`

I found where this happens in the source code [here](https://github.com/maplibre/martin/blob/01c165c368841403a75fd885a8a010130d6456ab/martin/src/sprites/mod.rs#L216) but there don't appear to be logs from spreet and no verbose mode.

Does anyone have any ideas what I might be doing wrong?

CommanderStorm:
@gkinsman
Sorry for the long wait, but could you please share the exact files you used?
When I try the `osm-liberty` or related `maki` svgs in that folder it works..

CommanderStorm:
Closing as not reproducible.

If someone runs into this, we would still like to look into why this happens, but without a clue where it might be, we cannot help here. 🤷🏻‍♂️

zhe-CUNY:
I seem to have ran into the issue. 

https://github.com/maputnik/osm-liberty/tree/gh-pages/sprites

It doesn't work with `osm-liberty` or any other sprite library for me. I am running on Windows.

on the ui end
`Error: Failed to load sprite image: .../sprite/osm-liberty@2x.png`

on the endpoint .../sprite/osm-liberty.png
`Unable to generate spritesheet`

nyurik:
@zhe-CUNY those are pre-generated sprites - martin does not support those. Instead, Martin supports if you give it a bunch of [svg images](https://maplibre.org/martin/sources-sprites.html).

nyurik:
P.S. Perhaps we should support these btw, but it would require us to validate the sprites somehow - shouldn't be very difficult - just ensure the image is a valid PNG, and that the image size is more than the max coordinate in the json files.  Feel free to create a feature request for this, describing the configuration and the logic of the feature.

zhe-CUNY:
[...]

PR #2793, opened 2026-05-09
fix(sprites): warn early and explain `.svg`-only requirement

### LINKED ISSUE #2136: Improve warnings in case of a bad sprite dir data

idovmamane:
## Summary

Issue #2136 asks for a clearer warning when a user points Martin at a sprite directory that has no `.svg` files. The original report (#1336, closed) involved a user pointing Martin at a pre-generated spritesheet (`sprite.png` + `sprite.json`). Martin returned `500 No sprite SVG files found in <path>` with no explanation of what files Martin does want.

This PR improves that signal *without* changing what Martin supports, actually consuming pre-generated sprites is tracked separately in #1364 and is out of scope here.

* emits a `WARN` from `SpriteSources::add_source` when the directory exists but contains no `.svg` files, so users see the problem at startup rather than at request time
* expands the `NoSpriteFilesFound` error to call out that Martin generates spritesheets from source `.svg` files
* adds a unit test covering both the variant and the wording

## Verification

Reproduced locally with a directory containing only `sprite.png` and `sprite.json`, saw the new warning at startup and the improved 500 body at request time.

- [x] `cargo test -p martin-core --features sprites` -> 40/40 passing
- [x] `cargo fmt --check`
- [x] `cargo clippy -p martin-core --features sprites --tests -- -D warnings`

Closes #2136.

### DIFF
--- a/martin-core/src/resources/sprites/error.rs
+++ b/martin-core/src/resources/sprites/error.rs
@@ -26,7 +26,10 @@ pub enum SpriteError {
     InvalidSpriteFilePath(String, PathBuf),
 
     /// No SVG files found in directory.
-    #[error("No sprite SVG files found in {0}")]
+    #[error(
+        "No sprite SVG files found in {0}. Martin generates spritesheets from source `.svg` files; \
+         a directory containing only pre-generated `sprite.png` / `sprite.json` cannot be used."
+    )]
     NoSpriteFilesFound(PathBuf),
 
     /// Failed to read sprite file.
--- a/martin-core/src/resources/sprites/error.rs
+++ b/martin-core/src/resources/sprites/error.rs
@@ -26,10 +26,7 @@ pub enum SpriteError {
     InvalidSpriteFilePath(String, PathBuf),
 
     /// No SVG files found in directory.
-    #[error(
-        "No sprite SVG files found in {0}. Martin generates spritesheets from source `.svg` files; \
-         a directory containing only pre-generated `sprite.png` / `sprite.json` cannot be used."
-    )]
+    #[error("No sprite SVG files found in {0} to generate spritesheets from")]
     NoSpriteFilesFound(PathBuf),
 
     /// Failed to read sprite file.
--- a/martin-core/src/resources/sprites/mod.rs
+++ b/martin-core/src/resources/sprites/mod.rs
@@ -108,10 +108,8 @@ impl SpriteSources {
                         warn!(
                             source.id = %v.key(),
                             sprite.path = %disp_path,
-                            "Sprite source directory contains no `.svg` files. \
-                             Martin generates spritesheets from source SVG files; \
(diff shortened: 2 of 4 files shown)


martin-core/src/resources/sprites/error.rs
```diff
@@ -26,7 +26,10 @@ pub enum SpriteError {
     InvalidSpriteFilePath(String, PathBuf),
 
     /// No SVG files found in directory.
-    #[error("No sprite SVG files found in {0}")]
+    #[error(
+        "No sprite SVG files found in {0}. Martin generates spritesheets from source `.svg` files; \
+         a directory containing only pre-generated `sprite.png` / `sprite.json` cannot be used."
+    )]
```
CommanderStorm:
```suggestion
    #[error("No sprite SVG files found in {0} to generate spritesheets from")]
```

martin-core/src/resources/sprites/mod.rs
```diff
@@ -102,6 +102,18 @@ impl SpriteSources {
...
+                    if let Ok(paths) = get_svg_input_paths(&path, true)
+                        && paths.is_empty()
+                    {
+                        warn!(
+                            source.id = %v.key(),
+                            sprite.path = %disp_path,
+                            "Sprite source directory contains no `.svg` files. \
+                             Martin generates spritesheets from source SVG files; \
+                             pre-generated `sprite.png` / `sprite.json` files are not used. \
+                             Sprite requests for this source will fail."
```
CommanderStorm: Lets focus the error message on why and what happens

```suggestion
                            "No sprite SVG files found in directory to generate spritesheets from. \
                             Sprite requests for this source will fail, until at least one svg is present."
```

martin-core/src/resources/sprites/mod.rs
```diff
@@ -253,6 +265,26 @@ mod tests {
...
+
+        let source = sprites.get("bad").expect("source registered");
+        let Err(err) = get_spritesheet([source].iter(), 1, false).await else {
+            panic!("expected NoSpriteFilesFound, got Ok");
+        };
+
+        assert!(matches!(err, SpriteError::NoSpriteFilesFound(ref p) if p == tmp.path()));
+        let msg = err.to_string();
+        assert!(msg.contains(".svg"), "error message lacks `.svg`: {msg}");
+        assert!(msg.contains(&tmp.path().display().to_string()));
```
CommanderStorm: not really an actionable test otherwise ^^

```suggestion
        assert!(matches!(err, SpriteError::NoSpriteFilesFound(_)));
```

CommanderStorm (approved):
Thanks. ❤️ 

I think we can shorten the docs a bit.
Wording is hard

PR #2794, opened 2026-05-09
feat(rendering): add static map overlay endpoint

CommanderStorm:
Adds two new HTTP endpoints under our rendering feature:
- `POST /style/{id}/static/{camera}/{wxh.fmt}` -> render a static map image at a center point, bounding box, or auto-fit to overlays.
- MLN generates the map


martin/src/srv/static_overlay/mod.rs
```diff
@@ -0,0 +1,61 @@
+//! Static-map overlay primitives: parse query parameters, project lon/lat to
+//! pixels, compute bounds, and draw paths/markers via tiny-skia.
+//!
+//! Used by `srv::styles_static` to add overlays on top of the base map image
+//! that `martin-core`'s static renderer produces.
```
CommanderStorm:
```suggestion
//! Static images - draw paths/markers via tiny-skia over a map
```

martin/src/srv/static_overlay/parse.rs
```diff
@@ -0,0 +1,260 @@
...
+        let coords = polyline::decode_polyline(encoded, 5).ok()?.0;
+        return (!coords.is_empty()).then_some(coords);
+    }
+
+    s.split('|')
+        .map(|part| {
+            let (a, b) = part.split_once(',')?;
+            let a: f64 = a.trim().parse().ok()?;
+            let b: f64 = b.trim().parse().ok()?;
+            Some(if latlng {
```
nyurik: what is this magic?

martin/src/srv/static_overlay/parse.rs
```diff
@@ -0,0 +1,260 @@
...
+pub fn parse_marker(s: &str, latlng: bool) -> Option<MarkerOverlay> {
+    let mut parts = s.splitn(3, '|');
+    let coord_str = parts.next()?;
+    let icon_spec = parts.next()?;
+    if icon_spec.is_empty() {
+        return None;
+    }
+    let icon = parse_icon_ref(icon_spec);
+
+    let (a, b) = coord_str.split_once(',')?;
```
nyurik: duplicated here?

martin/src/srv/static_overlay/parse.rs
```diff
@@ -0,0 +1,260 @@
...
+    if source_id.is_empty() || icon_name.is_empty() {
+        return None;
+    }
+    // sprite_id is a configured-source key — never contains slashes.
+    if source_id.contains('/') || source_id.contains('\\') || source_id.contains("..") {
+        return None;
+    }
+    // icon_name may be nested (e.g., `markers/red`) but must stay relative
+    // and not climb out of the source dir.
+    if icon_name.starts_with('/')
```
nyurik: don't manually resolve paths - this was the cause of lots of security bugs in the early web servers.  Check if there are any path-resolution approaches using built-in path handling / canocicalizations / relative path computation.

martin/src/srv/static_overlay/project.rs
```diff
@@ -0,0 +1,63 @@
...
+    let px_x = world_x - center_world_x + f64::from(img_width) / 2.0;
+    let px_y = world_y - center_world_y + f64::from(img_height) / 2.0;
+
+    (px_x, px_y)
+}
+
+/// Convert pixel position back to geographic coordinates.
+#[cfg(test)]
+#[must_use]
+pub(super) fn pixel_to_geo(
```
nyurik: we must not have any tests outside of the `mod tests{}`.  Also, seems like a dup with above. Also, please review AI-genned code

martin/src/srv/styles_static.rs
```diff
@@ -0,0 +1,747 @@
...
+
+/// Parse `static_params` into a `StaticType`.
+///
+/// Formats:
+/// - `lon,lat,zoom` — center
+/// - `lon,lat,zoom@bearing` — center with bearing
+/// - `lon,lat,zoom@bearing,pitch` — center with bearing and pitch
+/// - `minx,miny,maxx,maxy` — bounding box
+/// - `auto` — auto-fit from overlays
+fn parse_static_params(s: &str) -> Option<StaticType> {
```
nyurik: shouldn't this be `FromStr` impl?

martin/src/srv/styles_static.rs
```diff
@@ -0,0 +1,747 @@
...
+                coord! { x: min_lon, y: min_lat },
+                coord! { x: max_lon, y: max_lat },
+            )))
+        }
+        _ => None,
+    }
+}
+
+/// Parse `size_fmt` into `SizeFormat`: `"800x600.png"`, `"800x600@2x.jpeg"`.
+fn parse_size_format(s: &str) -> Option<SizeFormat> {
```
nyurik: same FromStr

martin/src/srv/static_overlay/parse.rs
```diff
@@ -0,0 +1,260 @@
...
+        stroke,
+        fill,
+        width,
+    })
+}
+
+/// Parse coordinates from string: raw `lng,lat|lng,lat` or `enc:polyline`.
+fn parse_coords(s: &str, latlng: bool) -> Option<Vec<Coord>> {
+    if let Some(encoded) = s.strip_prefix("enc:") {
+        let coords = polyline::decode_polyline(encoded, 5).ok()?.0;
```
nyurik: do we really want to use google polyline here? I agree that WKT would be larger, but it is also much more standard, would be compressed, would allow much more flexibility (essentially geojson rather than coordinate seq), and plus this is not often used -- these are one-offs, so not a significant extra cost...  Feel like an overoptimization tbh

nyurik (commented):
this is huge!  And unlike experimental MLT, this is for a production system... can we somehow break it into smaller chunks?

Bigger issues:
* lots of dup code - ai was going all out on that
* is there a full doc on all functionality?  I'm not yet convinced  google polyline is the right approach
* should we support geojson with styling approach as used in https://geojson.io/next -- this would be far cleaner imo, rather than reinventing this language

Also, is this all still part of experimental, or is this built by default?

martin/src/srv/static_overlay/parse.rs
CommanderStorm: polylines are fairly stnandard in routing and these sort of spaces, i.e. where you are getting the line from.

I have not seen WKT come out of one of these serices..
Lets not add WKT

CommanderStorm:
> - I'm not yet convinced google polyline is the right approach
> - should we support geojson with styling approach as used in https://geojson.io/next -- this would be far cleaner imo, rather than reinventing this language

It is not reinventing a language.
This is what tileserver-gl does -> https://tileserver.readthedocs.io/en/latest/endpoints.html#static-images

The entire API is modeled off what seems to work already.

I have not heard of the simplestyle apprach, that is one alternative of doing it, but an editor comes later for this.
If this is what helps me get this merged we can first do geojson and then the polyline support.
Polylines are attractive given that valhalla & co output that natively...

docs/content/sources-styles.md
```diff
@@ -90,3 +98,192 @@ The `{camera}` segment chooses what the image looks at:
...
+- **Paint / layout values are literals only.** Data-driven expressions
+  (`["interpolate", …]`, `["get", "name"]`, etc.) are rejected.
+
+Supported paint and layout properties:
+
+| Layer type | Paint properties                                                                                                  | Layout properties           |
+| ---------- | ----------------------------------------------------------------------------------------------------------------- | --------------------------- |
+| `fill`     | `fill-color`, `fill-opacity`, `fill-outline-color`                                                                | —                           |
+| `line`     | `line-color`, `line-opacity`, `line-width`                                                                        | `line-cap`, `line-join`     |
+| `circle`   | `circle-color`, `circle-opacity`, `circle-radius`, `circle-stroke-color`, `circle-stroke-opacity`, `circle-stroke-width` | —                           |
```
CommanderStorm:
```suggestion
| Layer type | Paint properties                                                                                                         | Layout properties       |
|------------|--------------------------------------------------------------------------------------------------------------------------|-------------------------|
| `fill`     | `fill-color`, `fill-opacity`, `fill-outline-color`                                                                       | —                       |
| `line`     | `line-color`, `line-opacity`, `line-width`                                                                               | `line-cap`, `line-join` |
| `circle`   | `circle-color`, `circle-opacity`, `circle-radius`, `circle-stroke-color`, `circle-stroke-opacity`, `circle-stroke-width` | —                       |
```

nyurik (approved):
niiice

PR #2795, opened 2026-05-09
feat: add if rendering is active to the catalog

CommanderStorm:
For some of the more global things, such as if rendering is active, we need to add this to the catalog.
This will allow me to pipe this through to the frontend in a nice way.

There are a few other settings which are interesting for the frontend, such as MLT/MVT conversion, but that is a differernt PR.
This PR just adds the basic primitives


martin/src/srv/admin.rs
```diff
@@ -33,6 +33,26 @@ pub struct Catalog {
...
+/// corresponding capability is compiled in.
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+#[cfg_attr(
+    feature = "unstable-schemas",
+    derive(schemars::JsonSchema, utoipa::ToSchema)
+)]
+pub struct CatalogSettings {
+    /// Whether server-side style rendering endpoints (e.g. `/style/{id}/{z}/{x}/{y}.png`) are
+    /// enabled. Only present in builds where the `rendering` feature is compiled in and the
+    /// host platform is Linux.
```
CommanderStorm:
```suggestion
/// Server-wide capability flags
#[cfg_attr(
    feature = "unstable-schemas",
    derive(schemars::JsonSchema, utoipa::ToSchema)
)]
pub struct CatalogSettings {
    /// Whether server-side style rendering endpoints are enabled.
```

nyurik (approved):
i'm ok with this, but i think it will quickly become an unmanagable mess with the proposed testing framework - depending on which features are enabled/disabled, you will have 2^n variants of test cases. Can you think of a different way to test it?  Please address this before merging.

Should we use a flag on the styles instead of a flag collection?   Would there be a case when we want to enable only a subset of styles? (possibly not).

```json
  "styles": {
    "maplibre_demo": {
      "path": "path/to/maplibre_demo.json",
      "render": true,
    },
  },
```

Another question:  how do you see styles and backend data to be tied together? Should there be any kind of `N to M` association with a set of data sources, or is this fully client-side responsibility?

CommanderStorm:
> Should we use a flag on the styles instead of a flag collection? Would there be a case when we want to enable only a subset of styles? (possibly not).

Currently, this is a global setting.
You cannot currently enable this at the style level, so this is done this way.
If this were a style-level setting, this would make sense, but it is not.

CommanderStorm:
> how do you see styles and backend data to be tied together?
> Should there be any kind of N to M association with a set of data sources, or is this fully client-side responsibility?

Not sure I get what you are angling for.
Styles currently entirely independent of the vector data backend we offer.
They may use the data backends, and likely are, but we are not optimising off this.

This can be opimised by adding my thesis' work into marin (-20% .. -50% transfer size..), But that is a different thing and would need to be opt-in and require a lot of rearchitecting for which I don't currently see the review time in this project, so I am deliberately not proposing this.

CommanderStorm:
> I think it will quickly become an unmanagable mess with the proposed testing framework - depending on which features are enabled/disabled, you will have 2^n variants of test cases. Can you think of a different way to test it? Please address this before merging.

I am not sure how to fix this.
Testing platform-cfg-dependent functionality requires the features to be cfg-ed.

Maybe lets remove the no-cfg test??
This way we are only testing the happy-path??
So only one variation??

martin/tests/styles_server_test.rs
```diff
@@ -73,6 +73,42 @@ async fn catalog_multiple_styles() {
...
+
+    let req = test_get("/catalog").to_request();
+    let response = call_service(&app, req).await;
+    let response = assert_response(response).await;
+    let body: Value = read_body_json(response).await;
+
+    insta::with_settings!({sort_maps => true}, {
+        assert_json_snapshot!(body["settings"], @r#"{}"#);
+    });
+}
```
CommanderStorm:
```suggestion
```

martin/tests/styles_server_test.rs
```diff
@@ -1,11 +1,13 @@
-#![cfg(feature = "styles")]
+#[cfg(all(feature = "styles", feature = "rendering", target_os = "linux"))]
```
CommanderStorm:
```suggestion
#![cfg(all(feature = "styles", feature = "rendering", target_os = "linux"))]
```

martin/tests/styles_server_test.rs
```diff
@@ -73,6 +68,25 @@ async fn catalog_multiple_styles() {
...
+    let response = assert_response(response).await;
+    let body: Value = read_body_json(response).await;
+
+    insta::with_settings!({sort_maps => true}, {
+        assert_json_snapshot!(body["settings"], @r#"
+        {
+          "rendering": true
+        }
+        "#);
+    });
```
CommanderStorm, line 87:
```suggestion
    });
}
```

PR #2801, opened 2026-05-13
feat: Expand relative URLs in style response

michaelkirk:
Addresses https://github.com/maplibre/martin/discussions/1870

This is working great for my use case: blue/green deployments, promoting with DNS.

Note, I don't personally use the base_path or route_prefix options. So please apply extra scrutiny to that code. I'm not entirely sure I understand the difference.

### DIFF
--- a/martin/src/maplibre_style.rs
+++ b/martin/src/maplibre_style.rs
@@ -12,6 +12,7 @@ use std::collections::BTreeMap;
 
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
+use url::Url;
 
 /// A partially-typed `MapLibre` style document.
 #[serde_with::skip_serializing_none]
@@ -116,8 +117,8 @@ fn expand_if_relative_url(url: &mut String, base_url: &str) {
     if url.starts_with("//") {
         return;
     }
-    // Has a scheme (https:, data:, mailto:, mapbox:, mbtiles:, ...) → leave alone.
-    if has_uri_scheme(url) {
+    // Already a valid absolute URL → leave alone.
+    if Url::parse(url).is_ok() {
         return;
     }
     // Ensure exactly one '/' between the base and the path, so a relative path
@@ -127,28 +128,6 @@ fn expand_if_relative_url(url: &mut String, base_url: &str) {
         url.insert(0, '/');
     }
     url.insert_str(0, base_url);
-}
-
-/// True if `url` begins with an RFC 3986 `scheme:` prefix.
-///
-/// `scheme = ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )` followed by `:`.
-fn has_uri_scheme(url: &str) -> bool {
-    let mut chars = url.chars();
-    let Some(first) = chars.next() else {
-        return false;
-    };
-    if !first.is_ascii_alphabetic() {
-        return false;
-    }
-    for c in chars {
-        if c == ':' {
-            return true;
-        }
-        if !(c.is_ascii_alphanumeric() || c == '+' || c == '-' || c == '.') {
-            return false;
-        }
-    }
-    false
 }
 
 #[cfg(test)]
--- a/martin/src/maplibre_style.rs
+++ b/martin/src/maplibre_style.rs
@@ -1,0 +1,387 @@
+//! Partial [`MapLibre` style spec][spec] types.
+//!
+//! Only the fields needed for URL rewriting are modeled; everything else
+//! round-trips through `serde_json::Value` via `#[serde(flatten)]`.
+//!
+//! This module is intentionally self-contained and depends only on
+//! `serde` / `serde_json`, so it could be lifted into a standalone crate later.
+//!
+//! [spec]: https://maplibre.org/maplibre-style-spec/
+
+use std::collections::BTreeMap;
+
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+
+/// A partially-typed `MapLibre` style document.
+#[serde_with::skip_serializing_none]
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+pub struct Style {
+    pub glyphs: Option<String>,
+
+    pub sprite: Option<Sprite>,
+
+    #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
+    pub sources: BTreeMap<String, Source>,
+
+    #[serde(flatten)]
+    pub other: BTreeMap<String, Value>,
+}
+
+/// The `sprite` field is either a single URL or a list of `{id, url}` entries.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum Sprite {
+    Single(String),
+    Multi(Vec<SpriteEntry>),
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct SpriteEntry {
+    pub id: String,
+    pub url: String,
+    #[serde(flatten)]
+    pub other: BTreeMap<String, Value>,
+}
+
+#[serde_with::skip_serializing_none]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Source {
+    pub url: Option<String>,
+
+    pub tiles: Option<Vec<String>>,
+
+    /// May be a URL string (for remote `GeoJSON`) or inline `GeoJSON`.
+    pub data: Option<Value>,
+
+    #[serde(flatten)]
+    pub other: BTreeMap<String, Value>,
+}
+
+impl Style {
+    /// Rewrite any URL field that lacks a scheme (does not contain `://`)
+    /// by prepending `base_url`.
+    ///
+    /// Lets a style.json on disk use protocol-less URLs like
+    /// `"/font/{fontstack}/{range}"`, which the `MapLibre` style spec doesn't
+    /// allow, while still serving spec-compliant absolute URLs to clients.
+    ///
+    /// Fields rewritten: top-level `glyphs`, `sprite` (both string and
+    /// `[{id, url}]` forms), and per-source `url`, `tiles[]`, and `data`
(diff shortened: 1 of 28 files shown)


martin/src/srv/styles.rs
```diff
@@ -68,6 +91,28 @@ pub async fn get_style_json(path: Path<StyleRequest>, styles: Data<StyleSources>
...
+    if let Some(prefix) = srv_config.public_path_prefix() {
+        prefix.to_string()
+    } else {
+        req.headers()
+            .get("X-Forwarded-Prefix")
+            .and_then(|v| v.to_str().ok())
+            .and_then(|v| v.parse::<Uri>().ok())
+            .map(|v| v.path().trim_end_matches('/').to_string())
+            .unwrap_or_default()
+    }
```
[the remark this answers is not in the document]
michaelkirk: This is a little baroque, but more or less copied from the tile metadata logic. 

Would you like me to make this and the tile metadata code more robust to malformed input?

TBH I'm a little surprised that we were expecting this "prefix" field to hold a URI, I'd assume just a path prefix, but I don't think I've ever used this header in the wild.

martin/src/maplibre_style.rs
```diff
@@ -0,0 +1,279 @@
...
+    }
+}
+
+fn expand_if_relative_url(url: &mut String, base_url: &str) {
+    // Already absolute (has any scheme like https://, http://, mapbox://, mbtiles://, ...).
+    if url.contains("://") {
+        return;
+    }
+    url.insert_str(0, base_url);
+}
```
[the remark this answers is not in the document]
michaelkirk: Protocol relative URLs and data (maybe?) URLs might be plausible. I've added some more robust "absolute URI" detection.

martin/src/srv/styles.rs
```diff
@@ -86,3 +131,62 @@ pub(crate) async fn redirect_styles(path: Path<StyleRequest>) -> HttpResponse {
         .insert_header((LOCATION, format!("/style/{style_id}")))
         .finish()
 }
+
+#[cfg(test)]
```
CommanderStorm, line 135: nit:
instead of writitng an unit test against `path_prefix`, can you add an integration test?

This way, this is tested against the public api instead of an private one.
So more stable.

Feel free to push back if you don't feel like doing that, this is likely fine too.

martin/src/maplibre_style.rs
```diff
@@ -0,0 +1,390 @@
...
+use serde_json::Value;
+
+/// A partially-typed `MapLibre` style document.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Style {
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub glyphs: Option<String>,
+
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub sprite: Option<Sprite>,
```
CommanderStorm: please use 

https://docs.rs/serde_with/latest/serde_with/struct.OneOrMany.html

this removes the need for that wrapper class ^^

martin/src/maplibre_style.rs
```diff
@@ -0,0 +1,390 @@
...
+//!
+//! [spec]: https://maplibre.org/maplibre-style-spec/
+
+use std::collections::BTreeMap;
+
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+
+/// A partially-typed `MapLibre` style document.
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
CommanderStorm: nit:
instead of `Option::is_none`, please use the `serde_with` way of doing this for consistency and because this way this is less likely to be forgottonen

CommanderStorm (commented):
I need to review this tomorrow (it is currently 2am), but here is a bit of feedback now.

I mostly need to look into my "research grade" codegen-impl of the maplbre style spec and see if there are parts that we should steal now instead of when nessary 😉

martin/src/srv/styles.rs
michaelkirk: By integration tests, do you mean whats run in `tests/test.sh` or something else?

martin/src/maplibre_style.rs
michaelkirk: Neat - I've never used serde_with before.

My understanding is we could use that if the grammer were:
```
sprite: None | Some(URI) | Some([URI...])
```

But it's actually:
```
sprite: None | Some(URI) | Some([{id: String, url: URI}])
```

I don't see how we can easily capture that with generic tools like serde_with::OneOrMany, since the "multi" case introduces fields (`id`, `url`) which don't exist in the "just one" case. Or am I missing something?

CommanderStorm: You are right, did not think far enough. Sorry

martin/src/srv/styles.rs
CommanderStorm: Likely yes.

Over the long term I want to get rid of that in favour of rust-style integration tests, but not now.

martin/src/maplibre_style.rs
```diff
@@ -0,0 +1,387 @@
...
+    for c in chars {
+        if c == ':' {
+            return true;
+        }
+        if !(c.is_ascii_alphanumeric() || c == '+' || c == '-' || c == '.') {
+            return false;
+        }
+    }
+    false
+}
```
CommanderStorm: a bit weird way of doing url parsing (usually, I would just use the `url`-crate).
Likely fine too to do it this way, just a bit weird.

martin/src/srv/styles.rs
michaelkirk: Ok, I've added some integration tests. I also included some changes to get these tests running on macos. If you'd prefer, I could open up a separate PR with those changes.

martin/src/maplibre_style.rs
michaelkirk: Done!

martin/src/srv/styles.rs
michaelkirk: I've left the unit tests - since I think they are more understandable and it's easier to enumerate many edge cases, while keeping the number of integration tests fewer - testing only the primary flows. That's how I typically do things, but let me know if you'd prefer something different

tests/expected/configured/style_maplibre_demo.1.json.headers
```diff
@@ -1,5 +1,5 @@
 content-encoding: br
 content-type: application/json
```
michaelkirk, line 2: Should we strip etags from the header content? It seems brittle and not useful.

CommanderStorm: They SHOULD only change if the content changes.
very strange..

I assume this pr outputs different json ordering and thus is "different", but after formatting is actually the same.
So can be ignored

CommanderStorm (approved):

martin/src/srv/styles.rs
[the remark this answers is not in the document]
CommanderStorm: We have support for `x-rewrite-url` which APPARENTLY, IIS might be using.

We can remove `X-Forwarded-Prefix`, honestly looking at that PR this might be an AI hallucination that I did not catch 😅 
It is also not documented or tested.

martin/src/srv/styles.rs
```diff
@@ -68,6 +91,28 @@ pub async fn get_style_json(path: Path<StyleRequest>, styles: Data<StyleSources>
...
+/// Note: `X-Rewrite-URL` is intentionally not honored here. Unlike the
+/// `TileJSON` case where the header's full path can be used directly, for
+/// styles the header would contain the full style request path
+/// (e.g. `/tiles/style/foo/style.json`), which isn't a usable prefix.
+fn path_prefix(req: &HttpRequest, srv_config: &SrvConfig) -> String {
+    if let Some(prefix) = srv_config.public_path_prefix() {
+        prefix.to_string()
+    } else {
+        req.headers()
+            .get("X-Forwarded-Prefix")
```
CommanderStorm, line 108: on a cursory read of the literature I think we MIGHT want to do this, but I am fine with current change too.

```suggestion
            .get("X-Rewrite-URL")
            .or(req.headers().get("X-Original-URL"))
```

martin/src/maplibre_style.rs
michaelkirk: Simplified! Note, I still have the special handling for protocol relative URLs, which the URL crate doesn't handle. I don't actually know if any clients actually even handle protocol relative URLs.

martin/src/srv/styles.rs
michaelkirk: I don't think we want `X-Rewrite-URL` or `X-Original-URL` here. 

It works in the case of tile metadata because the tile metadata URL is a prefix of the tile render request URL.

e.g. the user requests the tile info:
```
https://tiles.com/tileserver/my-tile-source
```

And the corresponding tile URLs are
```
https://tiles.com/tileserver/my-tile-source/{z}/{x}/{y}
```

Whereas in this case, the user requests:
```
https://tiles.com/tileserver/styles/basic
```

But we don't want to render the style.json with URLs like
```
https://tiles.com/tileserver/styles/basic/glyphs/my-glyphs.pbf
https://tiles.com/tileserver/styles/basic/my-tile-source
```

Instead, we want something like:
```
https://tiles.com/tileserver/glyphs/my-glyphs.pbf
https://tiles.com/tileserver/my-tile-source
```

I think `X-Forwarded-Prefix` could be used to resolve this? But I'm not personally using it.

michaelkirk:
Screenshot 2026-05-14 at 13 57 31

CI failure seems unrelated?

CommanderStorm:
Yea, one seems like a flake and one seems like good motivation for killing x86 macos support as the things we package.
I don't think people are using that, so likely fine.

CommanderStorm:
I will merge+release this once GHA has fixed `apt update`..
I assume that will be noticed by someone and won't investigate further right now 😉

PR #2804, opened 2026-05-14
feat: static rendering core

CommanderStorm:
This PR is the core implemntation of static rendering.

Martin can render a still image of a style at a chosen camera.
The endpoint is:

```text
GET /style/{style_id}/static/{camera}/{size}.{ext}
```

#### Camera

The `{camera}` segment chooses what the image looks at:

| Form                            | Meaning                                              |
| ------------------------------- | ---------------------------------------------------- |
| `lon,lat,zoom`                  | Center at `(lon, lat)` and `zoom` (north up, flat).  |
| `lon,lat,zoom@bearing`          | Center + bearing in degrees (clockwise from north).  |
| `lon,lat,zoom@bearing,pitch`    | Center + bearing + pitch in degrees.                 |
| `minLon,minLat,maxLon,maxLat`   | Fit the given bounding box to the requested size.    |

> [!IMPORTANT]
> The image is always rendered INSIDE of the requested size.
> So if the bbox is `[-10°,-1°,10°,1°]` and size `500x500` is requested, the image will be centered on 0,0 with a 500x500 box that is fully inside the bbox, so the left-top most pixel is approximately at `1°,-1°`.
> We will not expand the image outside the bbox.

#### Size and format

`{size}.{ext}` follows the `WIDTHxHEIGHT[@{scale}x].{ext}` pattern, e.g.
`800x600.png`, `400x300@2x.jpeg`.
Allowed extensions are `png`, `jpeg`, and `webp`.
Width and height are capped at 2048 px each; scale is capped at `4x`.


docs/content/sources-styles.md
```diff
@@ -45,7 +45,35 @@ styles:
...
+| ------------------------------- | ---------------------------------------------------- |
+| `lon,lat,zoom`                  | Center at `(lon, lat)` and `zoom` (north up, flat).  |
+| `lon,lat,zoom@bearing`          | Center + bearing in degrees (clockwise from north).  |
+| `lon,lat,zoom@bearing,pitch`    | Center + bearing + pitch in degrees.                 |
+| `minLon,minLat,maxLon,maxLat`   | Fit the given bounding box to the requested size.    |
+
+#### Size and format
+
+`{size}.{ext}` follows the `WIDTHxHEIGHT[@{scale}x].{ext}` pattern, e.g.
+`800x600.png`, `400x300@2x.jpeg`. Allowed extensions are `png`, `jpg`, `jpeg`.
```
nyurik: lets not support both jpg and jpeg - pick one. Possibly redirect, but don't document that we support both here -- force users to always use the same one.

martin-core/src/resources/styles/render_pool.rs
```diff
@@ -0,0 +1,270 @@
...
+    pub zoom: f64,
+    /// Logical output width in pixels.
+    pub width: u32,
+    /// Logical output height in pixels.
+    pub height: u32,
+    /// Pixel density ratio (1.0 = standard, 2.0 = retina). Multiplies the
+    /// renderer's internal output by `pixel_ratio`.
+    pub pixel_ratio: f32,
+    /// Bearing in degrees, clockwise from north (0 = north-up).
+    pub bearing: f64,
```
nyurik: does bearing and pitch need to be f64?

martin-core/src/resources/styles/mod.rs
```diff
@@ -188,6 +200,23 @@ impl StyleSources {
     }
 }
 
+/// Center longitude/latitude of an XYZ tile.
+///
+/// `maplibre-native`'s `setCamera` expects the camera target in WGS84 degrees;
+/// at integer zoom `z` we frame tile `(x, y)` by aiming the camera at its
+/// geometric center.
+#[cfg(all(feature = "rendering", target_os = "linux"))]
+fn tile_center_lon_lat(z: u8, x: u32, y: u32) -> (f64, f64) {
```
nyurik: this fn should have some test cases, and also double check if we don't have something like this in some other code already.  Seems like mln-rs may have a helper for this?

martin/src/srv/server.rs
```diff
@@ -150,6 +152,11 @@ fn register_services(
     cfg.service(styles::get_style_json)
         .service(styles::redirect_styles);
 
+    // `static` literal must win over the tile route's `{z}` wildcard, and
+    // actix matches in registration order - so `static` registers first.
+    #[cfg(all(feature = "rendering", target_os = "linux"))]
+    cfg.service(styles_static::get_style_static);
```
nyurik: naming is a bit confusing - technically both are "rendering" - one is an image, and the other is a tile rendering.  Should we keep that naming for consistency everywhere?

martin-core/src/resources/styles/render_pool.rs
```diff
@@ -0,0 +1,270 @@
...
+/// same worker — `maplibre-native` isn't safe to drive from two renderers
+/// concurrently, and a single pool removes the need for a cross-pool lock.
+///
+/// Cloning is cheap (Arc-clone of the shared channel/handle).
+#[derive(Debug, Clone)]
+pub struct RenderPool {
+    inner: Arc<Inner>,
+}
+
+impl RenderPool {
```
nyurik, line 117: i thought we already had a pool for tile rendering. Why do we introduce one for image rendering?

martin-core/src/resources/styles/render_pool.rs
```diff
@@ -0,0 +1,270 @@
...
+    pub width: u32,
+    /// Logical output height in pixels.
+    pub height: u32,
+    /// Pixel density ratio (1.0 = standard, 2.0 = retina). Multiplies the
+    /// renderer's internal output by `pixel_ratio`.
+    pub pixel_ratio: f32,
+    /// Bearing in degrees, clockwise from north (0 = north-up).
+    pub bearing: f64,
+    /// Pitch in degrees away from straight-down (0 = flat top-down view).
+    pub pitch: f64,
```
CommanderStorm:
```suggestion
    /// Path to the style JSON file.
    style_path: PathBuf,
    /// Camera target latitude in degrees (WGS84).
    lat: f64,
    /// Camera target longitude in degrees (WGS84).
    lon: f64,
    /// Map zoom level.
    zoom: f64,
    /// Logical output width in pixels.
    width: u32,
    /// Logical output height in pixels.
    height: u32,
    /// Pixel density ratio (1.0 = standard, 2.0 = retina). Multiplies the
    /// renderer's internal output by `pixel_ratio`.
    pixel_ratio: f32,
    /// Bearing in degrees, clockwise from north (0 = north-up).
    bearing: f64,
    /// Pitch in degrees away from straight-down (0 = flat top-down view).
    pitch: f64,
```

martin-core/src/resources/styles/render_pool.rs
CommanderStorm: what would you do the bearing and pitch otherwise?
This seems to be the type in MLN..

martin-core/src/resources/styles/render_pool.rs
CommanderStorm: Once I am sure this is 100% the design we need I will change upstream impl to what we have here.
I want to avoid the extra latency of having to get another PR approved if I can avoid it.

This fixes the style+resolution switching bug and handles the case if the pool needs to be retired.

nyurik: what do we use in the main?  do we rely on some upstream pool?  or do we just crash?  TBH, I would much rather introduce this code into upstream - as we can quickly release it, and its not stable yet anyway.  Not a blocker though - if you feel strongly that it should live here... just feels weird.

martin-core/src/resources/styles/render_pool.rs
nyurik: meh, just wasn't sure if f64 is needed rather than f32,  but no biggie

martin-core/src/resources/styles/render_pool.rs
CommanderStorm: the MLN crate will have a few iterations, this (pooling) seems a fairly application specific thing, so likely best kept here.
We will need access to this anyhow once I integrate the request-precaching optimisation.

docs/content/sources-styles.md
CommanderStorm: I guess, I adjusted the tile endpoint as well.
Will be a breaking change, but likely not one forcing us to v2.

martin-core/src/resources/styles/mod.rs
CommanderStorm: I guess.. it is fairly specificly coupled to how MLN does tile math and the oddities related to that.

nyurik (approved):

nyurik (approved):
Ok to merge, but please don't release just yet

PR #2810, opened 2026-05-16
refactor: share code accross the resource caches

CommanderStorm:


martin-core/src/cache.rs
```diff
@@ -0,0 +1,334 @@
...
+
+/// The `"hit"` / `"miss"` metric label, shared by [`CacheKey::record_outcome`]
+/// implementations.
+#[must_use]
+pub const fn hit_miss_label(hit: bool) -> &'static str {
+    if hit { "hit" } else { "miss" }
+}
+
+#[cfg(test)]
+mod tests {
```
CommanderStorm, line 170: larger in terms of LOC because it also comes with unit tests

martin-core/src/resources/sprites/cache.rs
```diff
@@ -1,150 +1,66 @@
...
+            hotpath::gauge!("sprite_cache_hits").inc(1.0);
+        } else {
+            hotpath::gauge!("sprite_cache_misses").inc(1.0);
+        }
+    }
+}
+
+impl Cacheable for Bytes {
+    fn weight(&self) -> u32 {
+        self.len().try_into().unwrap_or(u32::MAX)
```
nyurik, line 64: not a blocker, we may want to introduce a helper extension method `try_into_or(u32::MAX)` - seems a common enough pattern?

nyurik (approved):
looks good, thx

PR #2819, opened 2026-05-23
chore: update deps, cleanup justfile

nyurik:
I am not sure these apply, so I removed them:

```bash
cargo update --precise 1.44.3 insta
cargo update --precise 1.24.0 libdeflater
cargo update --precise 1.24.0 libdeflate-sys
```

### DIFF
--- a/justfile
+++ b/justfile
@@ -624,6 +624,10 @@ test-ssl-cert: start-ssl-cert
 update:
     cargo +nightly -Z unstable-options update --breaking
     cargo update
+    # Make sure that 'evil' dependencies are at the last compatible version
+    # below needs to be synced with deny.toml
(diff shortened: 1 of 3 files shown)


CommanderStorm (approved):

justfile
```diff
@@ -628,11 +619,6 @@ test-ssl-cert: start-ssl-cert
 update:
     cargo +nightly -Z unstable-options update --breaking
     cargo update
```
CommanderStorm, line 626: insta is fixed, libdeflater not so much.

```suggestion
    cargo update
    # Make sure that 'evil' dependencies are at the last compatible version
    # below needs to be synced with deny.toml
    cargo update --precise 1.24.0 libdeflater
    cargo update --precise 1.24.0 libdeflate-sys
```

PR #2828, opened 2026-05-24
test: cache render fixtures

CommanderStorm:
Currently, our rendering fixtures reach out via the network to demotiles & co.
If we avoid this, our tests are more stable.

I chose mitmproxy and the cassete trick for doing this.


.github/workflows/coverage.yml
```diff
@@ -48,6 +48,7 @@ jobs:
         run: sudo apt-get update && sudo apt-get install -y gdal-bin sqlite3-tools postgresql-client
       - name: Install rendering dependencies
         run: just install-dependencies
+      - uses: astral-sh/setup-uv@bd01e18f51369d5a26f1651c3cb451d3417e3bba # v7.1.1
```
CommanderStorm:
```suggestion
      - uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
```

martin-core/tests/rendering_test.rs
```diff
@@ -0,0 +1,134 @@
...
+
+    // Bless on first run; commit the file and re-run.
+    let reference_bytes = std::fs::read(&reference_path).unwrap_or_else(|_| {
+        std::fs::create_dir_all(reference_path.parent().unwrap()).unwrap();
+        assert!(
+            rendered_bytes.len() > 1000,
+            "Refusing to bless suspiciously small image ({} bytes)",
+            rendered_bytes.len()
+        );
+        std::fs::write(&reference_path, &rendered_bytes).unwrap();
```
nyurik, line 106: make sure to minimize paths - i cleaned it up a bit in the next commit, but we should not let AI use full paths everywhere - makes code less readable (for humans)

nyurik (approved):

martin-core/tests/rendering_test.rs
CommanderStorm: We SHOULD have a clippy lint against this even. not sure why it did not trigger.

nyurik: i'm not sure what a clippy's lint logic would be - there are tons of cases when full path is needed or at least preferred

CommanderStorm: - https://doc.rust-lang.org/clippy/lint_configuration.html#absolute-paths-max-segments
- https://doc.rust-lang.org/clippy/lint_configuration.html#absolute-paths-allowed-crates

Currently, we are at

https://github.com/maplibre/martin/blob/df59cee2f28b4ea7b3ac1a4da96f542eb0433784/clippy.toml#L4

CommanderStorm: IMO 3 is exactly reasonable because the `tokio::fs::write` vs `std::fs::write` case.
We can tighten this to `2`, but that is a lot of churn and pain for contributors, so I would rather not.

Issue #2264, opened 2025-09-28
Support for DuckDB as a backend


ggaans:
Now that DuckDB 1.4.0 supports ST_AsMVT and ST_AsMVTGeom is would be great to have Martin connect to it like is does for PostGIS.  This would provide a pathway for Parquet -> DuckDB -> Martin.

CommanderStorm:
This should be much simpler to do, given that we now have refactored our interals to the `martin-core`-Architecture.

We already have an issue for geoparquet, so lets continue tracking this there.
- #1693 

Would be open to a PR

AbelVM:
I'd love to see this working regardless of geoparquet, just for the sake of DuckDB. So it can be a first-class Martin citizen like Postgres. Geoparquet support would be just a side effect of this issue, indeed.

So, IMHO, this issue is not a clean duplicate of #1693 

For reference, two implementations of this DuckDB feature to serve tiles:
* Flask: https://gist.github.com/Maxxen/37e4a9f8595ea5e6a20c0c8fbbefe955
* R: https://gist.github.com/walkerke/c90ab6b8f403169e615eabeb0339b15b

CommanderStorm:
which features of duckdb would you like to see explicitely?
All the non-parquet file formats that duckdb supports (https://duckdb.org/docs/stable/guides/performance/file_formats) are not ideal for generating tiles from.

AbelVM:
I second that, but we might keep it agnostic to future-proofing the efforts.  I mean, once the logic for DuckDB is implemented, the actual source of the data is mostly only at the `FROM` statement, regardless it is a geoparquet, an avro file, or just a db file.

Regarding the features... just mimicking the functionality Martin provides for Postgres would be great

nyurik:
sure, but please keep in mind - every feature we add must be "useful" rather than just mimicking - otherwise we raise the cost of maintenance without benefits... grumble grumble grumble :)

AbelVM:
My point is that once DuckDB (via spatial extension) supports MVT-related functions, it might be an in-place replacement in a stack where Martin plays as the tiles endpoint. 

As an example, right now I have datasets with sizes in the hundreds of millions of features, and I need them to be available for :

1. Martin, to serve the info as tiles
2. Services that respond to both simple and complex search queries
3. Services that aggregate information on the fly

First requirement is fulfilled by Postgres, but the 2nd and 3rd would be better served by DuckDB. 

[...]

PR #2831, opened 2026-05-25
feat: duckdb-source implementation at martin-core

manbhav234:
#2264 

This PR introduces `DuckDBSource` at `martin-core` for supporting sources added by DuckDB. A brief on implementation details:

1. `DuckDBSource` is a single top level source type that implements the `Source` trait for all the sources generated/queried via DuckDB
2. The distinction between different sources is mainly at `DuckDBSqlInfo` since the difference is only at `FROM ...` in the SQL generated for tile query.
3. `DuckDBPool` stores the connection pool for that particular source. It also differentiates between local `.duckdb` files, local geoparquet files and remote geoparquet files. For local `.duckdb`, the pool contains connections opened with specified file path and only the spatial extension is loaded. For local and remote geoparquet files, in memory connections are created (httpfs extension loaded only for remote geoparquet)

Details on `DuckDBPool`  and how thread and async architecture is handled: 

1. It is a `deadpool` async pool that hands off synchronous `duckdb::Connection`. Using `r2d2` does not make sense here as waiting for a connection (if all are handed out) here, is a blocking process. With `deadpool`, the connection checkout is async and then handed over to `tokio::task::spawn_blocking()` for query execution.
2. New connection creation is also a blocking task and needs to be handled by `tokio::task::spawn_blocking()`. This is handled by `DuckDBPoolManager` which creates connection as per appropriate target (local .duckdb vs local/remote geoparquet) at `open_ready_connection()`. This also handles setting thread and memory limit.
3. `async fn generate_tile` checks out a connection asynchronously and hands over the `execute_tile_query` callback to generate tile. The used connection is then recycled back to the pool.

This async and threading structure seemed most convenient to me both logically and code wise. Would love your thoughts on this.

Some things I need for clarification on:

1. Do we need `url_query_params` with any of the sources that this adds ? For now I have excluded but I don't have much clarity on this.
2. there are no upper bound checks right now on `pool_size`. Since duckdb parallelizes query execution internally, this likely should not be a big value (max 4 or something). 
3. Currently, there is no limit on how many tasks can wait for a connection. Is this something that we should do since they are waiting for blocking threads to finish execution ? 
4.  There are no tests right now. How should i go about testing this and prepare sample data for testing ?


martin-core/Cargo.toml
```diff
@@ -57,6 +58,7 @@ sprites = [
 styles = ["tokio/fs", "dep:dashmap", "dep:walkdir"]
 mbtiles = ["dep:backon", "dep:mbtiles", "dep:tokio", "_tiles"]
 pmtiles = ["dep:pmtiles", "dep:object_store", "_tiles"]
+duckdb = ["dep:duckdb", "dep:deadpool", "dep:tokio", "_tiles"]
```
CommanderStorm: I might be wrong, but I think I remember that the `rt` feature was required to spawn tasks.

Cargo features are additive, so someone in the dependency tree needing `rt` would enable this already, so this might not be needed.
Since we do `spawn_blocking`, this seems like a good call nevertheless.

```suggestion
duckdb = ["dep:duckdb", "dep:deadpool", "dep:tokio", "tokio/rt", "_tiles"]
```

martin-core/src/tiles/duckdb/errors.rs
```diff
@@ -0,0 +1,84 @@
...
+    #[error("Unable to get a DuckDB connection from pool {1}: {0}")]
+    DuckDBPoolConnError(#[source] PoolError<DuckDBPoolManagerError>, String),
+
+    /// The blocking task used by `generate_tile()` failed.
+    #[error("DuckDB blocking task failed while {1}: {0}")]
+    DuckDBTaskJoinError(#[source] JoinError, &'static str),
+
+    /// Query preparation failed while serving a tile.
+    #[error("Error preparing DuckDB query for tile source '{1}' ({2}): {3} {0}")]
+    PrepareQueryError(#[source] DuckdbError, String, String, String),
```
CommanderStorm: nit: for errors like this where the type is not really that meaningful what the 2nd vs 3rd vs 4th string is, consider using the struct way of doing enums.

This way, you know what the second string is without looking this up.
This also applies to the other sources.

martin-core/src/tiles/duckdb/pool.rs
```diff
@@ -0,0 +1,270 @@
...
+            Self::GeoParquetRemote { url } => url.clone(),
+        }
+    }
+}
+
+// Deadpool manager responsible for opening and bootstrapping DuckDB connections.
+#[derive(Clone, Debug)]
+struct DuckDBPoolManager {
+    target: DuckDBPoolTarget,
+    threads_per_query: Option<usize>,
```
CommanderStorm: I think `Option<NonZeroUsize>` is what we want.
This way, we specify exactly the invariant that is possible for this field on a type-level.

martin-core/src/tiles/duckdb/pool.rs
```diff
@@ -0,0 +1,270 @@
...
+    memory_limit: Option<String>,
```
CommanderStorm: `String` is a bit weird for memory sizes.
It makes sense and I see how you got there, but I think it would be simpler to use a similar `Option<NonZeroUsize>` as above, to encode the correctness information.

Also in terms of naming, please suffix this with a unit.
It is otherwise a bit unclear if this is Bits, Bytes, ... 😉

martin-core/src/tiles/duckdb/source.rs
```diff
@@ -0,0 +1,154 @@
...
+        self.tile_info
+    }
+
+    fn clone_source(&self) -> BoxedSource {
+        Box::new(self.clone())
+    }
+
+    fn support_url_query(&self) -> bool {
+        false
+    }
```
CommanderStorm, line 65: > Do we need url_query_params with any of the sources that this adds ?
> For now I have excluded but I don't have much clarity on this.

Eventually, I think this might be nice, but lets leave this as a "future" thing.
It is not actually hard, just a bit of plumbing that needs to be done 😉

martin-core/src/tiles/duckdb/pool.rs
```diff
@@ -0,0 +1,270 @@
...
+    fn build(
+        id: String,
+        target: DuckDBPoolTarget,
+        pool_size: usize,
+        threads_per_query: Option<usize>,
+        memory_limit: Option<String>,
+    ) -> DuckDBResult<Self> {
+        let manager = DuckDBPoolManager::new(target, threads_per_query, memory_limit.clone());
+        let pool = Pool::builder(manager)
+            .max_size(pool_size)
```
CommanderStorm, line 86: > there are no upper bound checks right now on pool_size. Since duckdb parallelizes query execution internally, this likely should not be a big value (max 4 or something).

For now, lets leave this up to the layer above.
I think this would benefit from benchmarking 😉

If the source is remote for example, we might want a higher allowed pool size than if it is local.

No clue, same as you what is a good default.

CommanderStorm, line 86: 🤔 I think timeouts might be one of these fields that are important to control, or at least set a default is timeouts

https://docs.rs/deadpool/latest/deadpool/managed/struct.PoolConfig.html#structfield.timeouts

Currently, there are none, which means that If I have an incedibly slow query, this might loop forever.
I have no idea what is a good default here.
10s?
I doubt that a reasonable query in duckdb can hit that, but maybe I am wrong.

martin-core/src/tiles/duckdb/pool.rs
```diff
@@ -0,0 +1,270 @@
...
+    pub fn get_id(&self) -> &str {
+        &self.id
+    }
+}
+
+#[derive(Clone, Debug)]
+enum DuckDBPoolTarget {
+    DatabaseFile { path: PathBuf },
+    GeoParquetLocal { path: PathBuf },
+    GeoParquetRemote { url: String },
```
CommanderStorm: If possible, I would prefer stronger types.
Here for example we could use `url::Url` insead, which gives the API a stricter invariant.

Modeling things out via the type system prevents an entire class of bugs, even though it might seem tedious 😉

martin-core/src/tiles/duckdb/pool.rs
```diff
@@ -0,0 +1,270 @@
...
+                    location: self.target.get_path(),
+                })?
+            }
+            DuckDBPoolTarget::GeoParquetLocal { .. }
+            | DuckDBPoolTarget::GeoParquetRemote { .. } => {
+                Connection::open_in_memory().map_err(|source| Open {
+                    source,
+                    location: self.target.get_path(),
+                })?
+            }
```
CommanderStorm, line 216: This does not look quite right.
Should we not need to use the parameters here to attach something?

martin-core/src/tiles/duckdb/pool.rs
```diff
@@ -0,0 +1,270 @@
...
+        };
+
+        self.load_extension(&conn, "spatial")?;
+        if matches!(self.target, DuckDBPoolTarget::GeoParquetRemote { .. }) {
+            self.load_extension(&conn, "httpfs")?;
+        }
```
CommanderStorm, line 222: so spatial is apparently not autoloadable.
We thus need that `load_extension` call.

`httpfs` is autoloadable, though.
Should we enable this in `Config` instead?
Both variants seem valid to me..

martin-core/src/tiles/duckdb/pool.rs
```diff
@@ -0,0 +1,270 @@
...
+
+        if let Some(threads_per_query) = self.threads_per_query {
+            self.set_threads_per_query(&conn, threads_per_query)?;
+        }
+        if let Some(memory_limit) = &self.memory_limit {
+            self.set_memory_limit(&conn, memory_limit)?;
+        }
+
```
CommanderStorm: lets configure `custom_user_agent` -> so that we and the version we are running is clear to a calee

martin-core/src/tiles/duckdb/source.rs
```diff
@@ -0,0 +1,154 @@
...
+
+    fn benefits_from_concurrent_scraping(&self) -> bool {
+        //duckdb parallelizes queries internally on a fixed number of threads so having too many concurrent queries is not beneficial. but can be set to say 4 instead of 20+ for pg
+        true
+    }
```
CommanderStorm, line 70: did you mean this?

```suggestion
    fn benefits_from_concurrent_scraping(&self) -> bool {
        //duckdb parallelizes queries internally on a fixed number of threads so having too many concurrent queries is not beneficial. but can be set to say 4 instead of 20+ for pg
        false
    }
```

martin-core/src/tiles/duckdb/source.rs
```diff
@@ -0,0 +1,154 @@
...
+    let mut stmt = conn.prepare_cached(sql).map_err(|e| {
+        PrepareQueryError(
+            e,
+            source_id.to_string(),
+            info.signature.clone(),
+            info.sql_query.clone(),
+        )
+    })?;
+
+    debug!("SQL: {sql} [{xyz}]");
```
CommanderStorm: nit: Let's log sql like this, if at all

```suggestion
    trace!(%sql, %xyz, "duckdb tile query");
```

martin-core/src/tiles/duckdb/source.rs
```diff
@@ -0,0 +1,154 @@
...
+#[derive(Clone, Debug)]
+/// SQL query information for `DuckDB` tile sources.
+pub struct DuckDBSqlInfo {
+    /// SQL query string.
+    pub sql_query: String,
+    /// Whether the query uses URL query parameters.
+    pub use_url_query: bool,
+    /// Signature of the query.
+    pub signature: String,
+}
```
CommanderStorm, line 112: This seems a bit of a weird architecture.
This means that the "how do we do the query" gets lifted into the `martin` crate.

I think it would be simpler/more testable/higher locality if we were to keep this "knowledge" how to actually query right here...

martin-core/src/tiles/duckdb/pool.rs
```diff
@@ -0,0 +1,270 @@
...
+impl DuckDBPoolTarget {
+    fn get_path(&self) -> String {
+        match self {
+            Self::DatabaseFile { path } | Self::GeoParquetLocal { path } => {
+                path.display().to_string()
+            }
+            Self::GeoParquetRemote { url } => url.clone(),
+        }
+    }
+}
```
CommanderStorm: in general: lets `impl Display`.. instead and simplify our live in cases like this

In this specific case, we can remove this though by making this enum `pub` and add it to the Errors instead, which seems to be the only usage

martin-core/src/tiles/mod.rs
```diff
@@ -28,6 +28,10 @@ pub mod cog;
 /// Implementation of `PostgreSQL`' [`Source`].
 pub mod postgres;
 
+#[cfg(feature = "duckdb")]
+/// Implementation of `DuckDB`' [`Source`].
```
CommanderStorm, line 32: please also sync the respective `martin-core` Readme that this is supported

martin-core/src/tiles/duckdb/pool.rs
```diff
@@ -0,0 +1,270 @@
...
+        if matches!(self.target, DuckDBPoolTarget::GeoParquetRemote { .. }) {
+            self.load_extension(&conn, "httpfs")?;
+        }
+
+        if let Some(threads_per_query) = self.threads_per_query {
+            self.set_threads_per_query(&conn, threads_per_query)?;
+        }
+        if let Some(memory_limit) = &self.memory_limit {
+            self.set_memory_limit(&conn, memory_limit)?;
+        }
```
CommanderStorm: can we use [`Config`](https://docs.rs/duckdb/latest/duckdb/struct.Config.html) instead?
Seems simpler and more straightforward

martin-core/src/tiles/duckdb/pool.rs
```diff
@@ -0,0 +1,270 @@
...
+                .map_err(DuckDBPoolManagerError::from)?
+        }
+    }
+
+    fn recycle(
+        &self,
+        _obj: &mut Self::Type,
+        _metrics: &Metrics,
+    ) -> impl Future<Output = RecycleResult<Self::Error>> + Send {
+        async { Ok(()) }
```
CommanderStorm: I think we need to implement this, or?

I think below might work, but a test case for this specific case would be very nice to have.

```suggestion
    fn recycle(
        &self,
        conn: &mut Self::Type,
        _metrics: &Metrics,
    ) -> impl Future<Output = RecycleResult<Self::Error>> + Send {
        // When a connection returns to the pool and is later handed out again, deadpool calls recycle to decide if it's still good
        // We need to healthcheck to prevent a deleted and in place recreteated file leaving the pool in a broken state
        async move {
          tokio::task::block_in_place(|| conn.execute_batch("SELECT 1"))
              .map_err(...)?;æ
          Ok(())
      }
```

Likewise, I also don't know if we want to recreate based on `Metrics`.

CommanderStorm (commented):
Nice progress.
I reviewed inline and did answer some of your questins questions inline as well.

As said on slack, this is likely better than what I would have done as a first attempt 👍🏻 

I think to be mergable, this needs:
- integration tests that the public api works as expected
- possibly unit tests if you think something might benefit from that
- pass clippy 😉

I also have a few comments inline, but nothing major, as far as I can remember.
Otherwise, good work and LGTM.

> Currently, there is no limit on how many tasks can wait for a connection.
Is this something that we should do since they are waiting for blocking threads to finish execution ?

I think if we add a timeout to our `deadpool`, we can effectively fix this.
There might be a better option, but I am struggling to come up with one.
Good question.

> There are no tests right now.
How should I go about testing this and prepare sample data for testing?

Under the current paradigm of having the sql live outside likely an integration test.
Integration tests have the advantage of forcing you to test against the public API, but are a bit slower, usually.
Testing against the public api is better, because more stable.

If we change the paradigm and make this actually use `ST_AsMVT`/`ST_AsMVTGeom`, a bit more complicated, but likely fine.


```rust
#![cfg(feature="duckdb")]
use martin_core::duckdb::*;

async fn fixture_with_tiles() -> (tempfile::TempDir, std::path::PathBuf) {
    let dir = tempfile::tempdir().unwrap();
    let path = dir.path().join("tiles.duckdb");
    let conn = duckdb::Connection::open(&path).unwrap();
    conn.execute_batch(
        "CREATE TABLE tiles(z INTEGER, x INTEGER, y INTEGER, tile_data BLOB);
        INSERT INTO tiles VALUES
            (0,0,0, '\\x1a2e'::BLOB);",
    ).unwrap();
    (dir, path)   // keep `dir` alive for the test's duration
}

#[tokio::test]
async fn serves_known_tile() {
    let (_dir, path) = fixture_with_tiles().await;
    let pool = DuckDBPool::new_database_file("t".into(), path, 2, None, None).await.unwrap();
    let info = DuckDBSqlInfo::new(
        "SELECT tile_data FROM tiles WHERE z = ? AND x = ? AND y = ?".into(),
        false, "sig".into());
    let src = DuckDBSource::new("t".into(), info, TileJSON::default(), pool, tile_info, zoom);

    let tile = src.get_tile(TileCoord { z: 0, x: 0, y: 0 }, None).await.unwrap();
    assert_eq!(tile, b"\\x1a2e");
}

// test that file bricking and restoring does not permanently break the pool

// does the rest of the public api work as expected
```

manbhav234:
I'll mark this as draft for now and continue working on the following :

- [x] Going through and resolving inline comments
- [x] Writing tests
- [x] Fixing remaining clippy warnings

martin-core/src/tiles/duckdb/pool.rs
manbhav234: What all information should we specify in the `custom_user_agent` value ? DuckDB specifies this as the default value : `duckdb/v1.4.4(osx_arm64) cli 6ddac802ff`

martin-core/src/tiles/duckdb/source.rs
manbhav234: Shall I work on how this new design of having the query knowledge in martin-core for this would look like in this PR or open a new PR and get the design and approach approved there ?

manbhav234:
I resolved most of the comments. just need to configure timeouts and `custom_user_agent` for which I left a comment above. 

I added tests for what I thought is good to test. They aren't a lot. Can add more tests for specific behaviors or remove any existing ones if they seem not too beneficial. 

As for the architecture of lifting the query information to `martin`, i can work out how that would look like. Again, added a comment above for this. I haven't included tests for local/remote geoparquet. I think that is more testable once the logic for querying is implemented either at `martin` or `martin-core`

Also, `just check` seems to fail in the CI but passes locally. I couldn't really figure out the issue.

CommanderStorm:
> Also, just check seems to fail in the CI but passes locally. I couldn't really figure out the issue.

CI just complains about the until now dead code. I pushed a fix, should likely pass CI

CommanderStorm (approved):
should have +1-ed earlier

PR #2832, opened 2026-05-26
refactor: make sure that error types with multiple identical types have names

CommanderStorm:
- noticed in https://github.com/maplibre/martin/pull/2831#discussion_r3298959225


nyurik (approved):
tbh, i'm not a big fan of non-tuple values because they explode in linecount, but they are a bit more readable, so fine

PR #2841, opened 2026-05-27
feat(pg): enable Postgres reloading

CommanderStorm:


martin/src/config/args/postgres.rs
```diff
@@ -8,7 +8,9 @@ use tracing::{info, warn};
 use super::connections::Arguments;
 use super::connections::State::{Ignore, Take};
 use crate::config::file::UnrecognizedValues;
-use crate::config::file::postgres::{POOL_SIZE_DEFAULT, PostgresConfig, PostgresSslCerts};
+use crate::config::file::postgres::{
+    DEFAULT_RELOAD_INTERVAL, POOL_SIZE_DEFAULT, PostgresConfig, PostgresSslCerts,
```
gabeschine: nit: do we care that there are two patterns for placing `DEFAULT` in the name of the constants? (at beginning, at end)

martin/src/config/file/tiles/discovery/postgres.rs
```diff
@@ -0,0 +1,228 @@
...
+
+use crate::config::file::postgres::{PostgresAutoDiscoveryBuilder, PostgresConfig, SourceSpec};
+use crate::config::file::tiles::discovery::{Discovery, Version};
+use crate::config::file::{CachePolicy, ProcessConfig};
+use crate::config::primitives::IdResolver;
+use crate::{MartinError, MartinResult};
+
+/// A [`Discovery`] over one `PostgreSQL` connection.
+///
+////Entries are versioned by their [`SourceSpec::fingerprint`], so an in-place
```
gabeschine:
```suggestion
/// Entries are versioned by their [`SourceSpec::fingerprint`], so an in-place
```

gabeschine: What exactly does go into the fingerprint? That might be useful to have here instead of only one example of what isn't.

martin/src/config/file/tiles/postgres/config.rs
```diff
@@ -134,6 +161,30 @@ pub struct PostgresConfig {
...
+            tables: None,
+            functions: None,
+            #[cfg(all(feature = "mlt", feature = "_tiles"))]
+            convert_to_mlt: None,
+            #[cfg(all(feature = "mlt", feature = "_tiles"))]
+            convert_to_mvt: None,
+            unrecognized: UnrecognizedValues::default(),
+        }
+    }
+}
```
gabeschine, line 186: I'm not sure if this'll work, but shorter and more robust to future changes might be to use the struct update syntax:

```rust
fn default() -> Self {
        Self {
            reload_interval: DEFAULT_RELOAD_INTERVAL,
            ..Default::default()
        }
    }
```

I haven't tested it.

gabeschine (approved):
nice!

martin/src/config/file/tiles/postgres/config.rs
CommanderStorm: I like to be explicit because this way I can't "accidentally" change this default.
`..Default::default()` has the effect that the code gets less verbosee, but also a lot less defensive

https://corrode.dev/blog/defensive-programming/#code-smell-lazy-use-of-default

martin/src/config/args/postgres.rs
CommanderStorm: good catch, fixed it

martin/src/config/file/tiles/discovery/postgres.rs
CommanderStorm: > What exactly does go into the fingerprint

it is a bit of a longer list. See https://github.com/maplibre/martin/blob/a6722f2831159b1fe94fa5755aab7cfe07644c8d/martin/src/config/file/tiles/postgres/spec.rs#L19-L73

The gist is that it is "anything that materially changes bytes or metadata"

PR #2840, opened 2026-05-26
refactor: migrate Pmtiles to `ReloadDriver` and `Discovery` (pg-reload-7)

CommanderStorm:
migrates pmtiles to the new arch

PR #2844, opened 2026-05-31
refactor: extract `Sink` trait for the tile-source manager

CommanderStorm:
Pulls the `Sink` trait from
- #2840

### DIFF
--- a/martin/src/config/file/tiles/driver/sink.rs
+++ b/martin/src/config/file/tiles/driver/sink.rs
@@ -1,0 +1,14 @@
+use async_trait::async_trait;
+
+use crate::MartinResult;
+use crate::reload::ReloadAdvisory;
+
+/// Where a reload driver applies advisories.
+///
+/// The production implementation is [`TileSourceManager`](crate::TileSourceManager);
+/// tests use a recording spy.
+#[async_trait]
+pub trait Sink: Send + 'static {
+    /// Applies a [`ReloadAdvisory`] to the live source set.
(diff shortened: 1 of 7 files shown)


martin/src/config/file/tiles/driver/sink.rs
```diff
@@ -0,0 +1,14 @@
+use async_trait::async_trait;
+
+use crate::MartinResult;
+use crate::reload::ReloadAdvisory;
+
+/// Where a reload driver applies advisories.
+///
+/// The production implementation is [`TileSourceManager`](crate::TileSourceManager);
+/// tests use a recording spy.
+#[async_trait]
```
nyurik, line 10: you don't need async_trait here - I removed it in a subsequent commit

nyurik (approved):

CommanderStorm:
`async_trait` is needed, otherwise this cannot be Send+Sync..

martin/src/config/file/tiles/driver/sink.rs
CommanderStorm: we can re-add it in the next PR once I need it I guess..

nyurik (approved):

nyurik:
@CommanderStorm could you clarify?  It compiles without it

CommanderStorm:
See next PR, does need `Send` I think

https://smallcultfollowing.com/babysteps/blog/2019/10/26/async-fn-in-traits-are-hard/

CommanderStorm:
I see what you did, maybe sub-trait-ing to `Send + 'static` works too, not sure.

PR #2848, opened 2026-06-01
ci: add test-packages justfile recipe and use it in CI

### LINKED ISSUE #2492: Simplify CI by moving some steps to the justfile
One justfile benefit is to be able to run the same CI commands locally - i.e. CI should call `just <command>` as much as possible.

I think at least this code should be moved to a justfile recipe (recipe name is TBD). Note that the `tr` usage can be done by the justfile's case conversion [functions](https://github.com/casey/just?tab=readme-ov-file#case-conversion).

All the `mv` commands should also be its own recipe - because it is being used by another CI step as well.

```yaml
      - run: rustup target add ${{ matrix.target }}
      - name: Build ${{ matrix.target }}
        run: |
          export "CARGO_TARGET_$(echo ${{ matrix.target }} | tr 'a-z-' 'A-Z_')_RUSTFLAGS"='-C strip=debuginfo'
          cargo zigbuild --release --target ${{ matrix.target }} --workspace --locked

      - name: Move the artefacts to the expected place
        run: |
          mkdir -p target_releases/
          mv target/${{ matrix.target }}/release/martin target_releases/
          mv target/${{ matrix.target }}/release/martin-cp target_releases/
          mv target/${{ matrix.target }}/release/mbtiles target_releases/
```

chrisdebian:
Closes #2492.

Consolidates the per-package `cargo test` block in `ci.yml` into a new `test-packages` justfile recipe, consistent with how `build-binary.yml` already delegates to `just build-release-musl` and `just move-artifacts`.

The `AWS_SKIP_CREDENTIALS` and `AWS_REGION` env vars are removed from the CI step as they are already exported by the justfile.

The `[linux]`-only `with-render-cache` call is preserved via a `uname` check inside the recipe's bash shebang block, matching the original `$RUNNER_OS` conditional in the workflow.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,23 +77,8 @@ jobs:
         if: runner.os == 'Linux'
         uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
       - name: Run cargo test
-        run: |
-          set -x
-          cargo test --package martin-tile-utils
-          cargo test --package mbtiles --no-default-features
-          cargo test --package mbtiles
-          cargo test --package martin-core
-          # style rendering requires upstream resources
-          # we use a cassete recording to not have random flakes
-          if [ "$RUNNER_OS" = Linux ]; then
-            just with-render-cache 'cargo test --package martin'
-          else
-            cargo test --package martin
-          fi
-        env:
-          AWS_SKIP_CREDENTIALS: 1
-          AWS_REGION: eu-central-1
-      - run: cargo test --doc
+        run: just test-packages
+      - run: just test-doc
   lint-rust:
     name: Lint the Rust code
     permissions:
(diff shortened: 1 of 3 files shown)


justfile
```diff
@@ -594,6 +594,20 @@ test-minio:
 test-cargo *args:
     MLN_PRECOMPILE=1 cargo test {{args}}
 
+# Run unit tests for each package in dependency order (used by CI)
+test-packages:
```
CommanderStorm: Please suffix all jobs intended for ci use with `-ci` vor consistency sake


```suggestion
# Run unit tests for each package in dependency order
test-packages-ci:
```

.github/workflows/ci.yml
```diff
@@ -77,23 +77,8 @@ jobs:
...
-          if [ "$RUNNER_OS" = Linux ]; then
-            just with-render-cache 'cargo test --package martin'
-          else
-            cargo test --package martin
-          fi
-        env:
-          AWS_SKIP_CREDENTIALS: 1
-          AWS_REGION: eu-central-1
-      - run: cargo test --doc
+        run: just test-packages
```
```suggestion
        run: just test-packages-ci
```

CommanderStorm (approved):
Looks got to me but please match the naming sceme

PR #2850, opened 2026-06-02
feat: Auto-refresh pmtiles source on source data change

### LINKED ISSUE #2713: Reload PMTiles sources if their underlying PMTiles file data changes
Involves upgrading `pmtiles-ts` to the latest version to include https://github.com/stadiamaps/pmtiles-rs/pull/117, and then handling the new error enum to:

- Reload PMTiles sources
- Clear relevant caches
- Retry failed requests

gabeschine:
What's left:

- [ ] Invalidate the local tile cache for this source
- [ ] Tests for invalidation

Fixes [issue #2713](https://github.com/maplibre/martin/issues/2713)

### DIFF
--- a/martin-core/src/tiles/pmtiles/source.rs
+++ b/martin-core/src/tiles/pmtiles/source.rs
@@ -182,8 +182,7 @@ impl Source for PmtilesSource {
         xyz: TileCoord,
         _url_query: Option<&UrlQuery>,
     ) -> MartinCoreResult<TileData> {
-        let coord =
-            pmtiles::TileCoord::new(xyz.z, xyz.x, xyz.y).map_err(PmtilesError::PmtError)?;
+        let coord = pmtiles::TileCoord::new(xyz.z, xyz.x, xyz.y).map_err(PmtilesError::PmtError)?;
 
         let result = {
             let reader = self.pmtiles.read().await;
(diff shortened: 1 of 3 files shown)


martin-core/src/tiles/pmtiles/source.rs
```diff
@@ -20,7 +21,15 @@ use crate::tiles::{BoxedSource, MartinCoreResult, Source, UrlQuery};
 pub struct PmtilesSource {
     id: String,
     #[dbg(skip)]
-    pmtiles: Arc<AsyncPmTilesReader<ObjectStoreBackend, PmtCacheInstance>>,
+    pmtiles: Arc<RwLock<AsyncPmTilesReader<ObjectStoreBackend, PmtCacheInstance>>>,
```
CommanderStorm, line 24: Likely all of this locking is fine, but for local sources on high performance SSD storage maybe not ideal.

Have you experimented with just returning an error from the tile-getting call that then triggers a re-adding of a new source and then re-issues the call against that source?
Why did you do it the current way? 🤔
This way, we automatically would have a consistent cache state and many things would also not "internally tangle" as much, perhaps.

gabeschine: ^ I will give it a shot. I started on this PR before PMTilesReloader was born, so I probably anchored on the approach and didn't zoom out enough.

CommanderStorm: Fair ^^

PR #2852, opened 2026-06-03
refactor: adopt `ReloadDriver`/`Discovery` for Mbtiles (pg-reload-5)

CommanderStorm:
Looks like numerous additions, but that is just new tests that are now simple to write.


martin/src/config/file/tiles/discovery/mod.rs
```diff
@@ -0,0 +1,9 @@
+//! The `Discovery` trait and its implementations: `FsDiscovery` for the file-backed kinds.
+
+mod r#trait;
```
nyurik: lets not use keywords as filenames

martin/src/config/file/tiles/reload/cog.rs
```diff
@@ -1,23 +1,18 @@
...
-use crate::config::file::tiles::reload::{discover_sources_by_ext, path_modified_ms};
-use crate::config::file::{CachePolicy, FileConfigEnum};
-use crate::config::primitives::{IdResolver, OptOneMany};
-use crate::{MartinError, MartinResult, ReloadAdvisory, TileSourceManager};
+use crate::config::file::tiles::discovery::{FsDiscovery, FsSourceBuilder};
+use crate::config::file::tiles::driver::{NotifyTrigger, ReloadDriver};
+use crate::config::primitives::IdResolver;
+use crate::{MartinResult, TileSourceManager};
 
+/// Watches configured directories for `.tif`/`.tiff` changes.
```
nyurik, line 12: perhaps a separate PR - the styling rules for abbreviation is to lower case all letters after the first - i.e. `CogReloader`.  Note that it is `FsDiscovery`, not `FSDiscovery`

nyurik (approved):
looks ok. Do we have any CLI flags or other ways of specifying watch vs no-watch?  Which is the default?

CommanderStorm:
> Do we have any CLI flags or other ways of specifying watch vs no-watch? Which is the default?

We watch by default.
It cannot be disabled for cog, mbtiles, it is very cheap because it is just listening to fs events.

For PG, PMT it can be configured, but only because there the polling rate is interesting.
If there were a cheap api, it would also be automatic.
You can enable it for these two.

docs/content/architecture.md
```diff
@@ -424,6 +425,50 @@ Read it when you're curious **why** certain choices were made.
...
+    The traits are:
+
+    - **`Discovery`** - the cheap **"what should exist now"** read (a directory scan or object-store listing).
+      Returns one `(Version, Args)` per source id.
+      `Version` is a `u128` change token (e.g. mtime millis); a changed value triggers an in-place update.
+      `Args` is the kind-specific payload used to build the source on demand.
+    - **`Trigger`** - decides **when to reconcile**.
+      `NotifyTrigger` fires on filesystem events; `PollTrigger` fires on a fixed interval.
+    - **`Sink`** - where the diff is **applied**.
+      `TileSourceManager`, which holds the live catalog, implements it directly via `apply`.
```
CommanderStorm:
```suggestion
      `TileSourceManager`, which holds the live catalog, implements it directly via `apply_changes`.
```

PR #2853, opened 2026-06-03
refactor: migrate `MBTilesReloader` to be based on `ReloadDriver` and `Discovery` (pg-reload-6)

CommanderStorm:


martin/src/config/file/tiles/reload/mbtiles.rs
```diff
@@ -1,53 +1,31 @@
...
-use crate::config::file::{CachePolicy, FileConfigEnum};
-use crate::config::primitives::{IdResolver, OptOneMany};
-use crate::{MartinError, MartinResult, ReloadAdvisory, TileSourceManager};
+use crate::config::file::tiles::discovery::{FsDiscovery, FsSourceBuilder};
+use crate::config::file::tiles::driver::{NotifyTrigger, ReloadDriver};
+use crate::config::primitives::IdResolver;
+use crate::{MartinResult, TileSourceManager};
 
+/// Watches configured directories for `.mbtiles` changes.
 pub struct MBTilesReloader {
```
CommanderStorm: @nyurik so what is the upper/lowercaseness here?

`MbtilesReloader` to be consistent with `CogReloader`?

martin/src/config/file/tiles/reload/mbtiles.rs
```diff
@@ -65,153 +43,28 @@ impl MBTilesReloader {
...
-        }
-
-        directories.sort();
-        directories.dedup();
+        let build: FsSourceBuilder = Box::new(|id, path, policy| {
+            Box::pin(async move {
+                let src = MbtSource::new(id, path, policy.zoom()).await?;
+                Ok(Box::new(src) as BoxedSource)
+            })
+        });
```
CommanderStorm: Yes, all of the boxes have a point ^^

martin/src/config/file/tiles/reload/mbtiles.rs
nyurik: `MbTilesReloader` or `MbtilesReloader` - technically the first, but i'm fine with either

martin/src/config/file/tiles/reload/mbtiles.rs
nyurik: if you have to write this, you have to add a long comment explaining WHY you need them

nyurik (approved):
looks good. One minor thing - perhaps we should standardize on `Mbtiles` as the name?  naming is hard

PR #2858, opened 2026-06-03
feat: Auto-refresh pmtiles source on source data change

### LINKED ISSUE #2713: Reload PMTiles sources if their underlying PMTiles file data changes

gabeschine:
Reloads a tile source and retries the tile request for a source if it returns a `MartinCoreError::SourceNeedsReload`.

Teaches the `PMTiles` source to return this if the `pmtiles` create indicates the PMTiles file/source object has changed.

Fixes https://github.com/maplibre/martin/issues/2713


gabeschine:
@CommanderStorm this is still draft, but ready for some feedback.

What I like:

- The "source needs reload" signal and the reload-and-replace mechanics are now generic to any `Source`, and uses the `ReloadAdvisory` mechanism.

What I don't like:

- Reloading the source is implemented using a `Source::try_reload()` method that I had to add to the trait.
- Unsure of if the unit test coverage of the generic reload path is sufficient, as it uses a mock source that simulates a self-reload.

I'd prefer to reuse the existing `PmtConfig` path to generate a refreshed source, but that did not seem possible without some rather more ugly wiring and breaking of abstraction boundaries.

Got any ideas or feedback?

martin-core/src/tiles/source.rs
```diff
@@ -84,6 +84,17 @@ pub trait Source: Send + Sync + Debug {
...
+    /// Attempts to create a fresh instance of this source.
+    ///
+    /// The default implementation returns `None`, meaning the source cannot self-reload and
+    /// the error will be surfaced to the caller.
+    ///
+    /// Sources that return a MartinCoreError::SourceNeedReload from get_tile() must also
+    /// implement this method.
+    async fn try_reload(&self) -> MartinCoreResult<BoxedSource> {
+        unreachable!()
+    }
```
CommanderStorm, line 95:
```suggestion
    /// Attempts to create a fresh instance of this source.
    async fn try_reload(&self) -> MartinCoreResult<BoxedSource> {
        unreachable!("only sources that return MartinCoreError::SourceNeedReload from get_tile must implement this method")
    }
```

martin/src/srv/tiles/content.rs
```diff
@@ -381,38 +385,91 @@ impl<'a> DynTileSource<'a> {
...
+                } else {
+                    compute().await.map_err(Arc::new)
+                }
+            }
+        };
+
+        let tile = do_fetch(s).await;
+
+        // On SourceNeedsReload, rebuild the source and retry once.
+        let tile = if matches!(&tile, Err(e) if matches!(e.as_ref(), MartinCoreError::SourceNeedsReload))
```
CommanderStorm: 2x match inside..
I am assuming this does not work?

```suggestion
        let tile = if matches!(&tile, Err(MartinCoreError::SourceNeedsReload))
```

martin/src/srv/tiles/content.rs
```diff
@@ -381,38 +385,91 @@ impl<'a> DynTileSource<'a> {
...
+        not(all(feature = "mlt", feature = "_tiles")),
+        expect(unused_variables)
+    )]
+    async fn get_tile_content_from_one_source(
+        &self,
+        s: &BoxedSource,
+        pc: &ProcessConfig,
+        xyz: TileCoord,
+    ) -> ActixResult<Tile> {
+        let do_fetch = |src: &BoxedSource| {
```
CommanderStorm: The code quality that you found existing is not ideal, lets try to clean it a bit.

Lets promote this closure to a method and name it `fetch_with_cache`.

This makes this method to something like

```rust
match self.fetch_with_cache(s, pc, xyz).await {
    Err(MartinCoreError::SourceNeedsReload) => self.reload_source_and_retry(s, pc, xyz).await,
    result => result.map_err(|e| map_internal_error(e.as_ref())),
}
```

martin/src/srv/tiles/mod.rs
```diff
@@ -49,6 +49,71 @@ pub mod tests {
...
+    impl SourceNeedsReloadTestSource {
+        pub fn new(id: &'static str, data: TileData) -> Self {
+            Self {
+                id,
+                tilejson: tilejson! { tiles: vec![] },
+                data,
+                call_count: 0,
+            }
+        }
+    }
```
CommanderStorm: nit: all fields pub and a constructor is a bit weird combination, but it is a test 😉

martin/src/srv/tiles/content.rs
```diff
@@ -381,38 +385,91 @@ impl<'a> DynTileSource<'a> {
...
+
+    #[cfg_attr(
+        not(all(feature = "mlt", feature = "_tiles")),
+        expect(unused_variables)
+    )]
+    async fn get_tile_content_from_one_source(
+        &self,
+        s: &BoxedSource,
+        pc: &ProcessConfig,
+        xyz: TileCoord,
```
CommanderStorm, line 398: nit: the shorter the scope the shorter the variables, but the reverse also applies.

-> name this a bit more explicitly

martin/src/srv/tiles/mod.rs
```diff
@@ -49,6 +49,71 @@ pub mod tests {
         }
     }
 
+    /// A test source that returns [`MartinCoreError::SourceNeedsReload`] on the first `get_tile`
+    /// call and real data on every subsequent call.
+    ///
+    /// Used to verify that the tile-serving layer detects `SourceNeedsReload`, reloads the
+    /// source, and retries the request rather than propagating the error to the caller.
+    #[derive(Debug, Clone)]
+    pub struct SourceNeedsReloadTestSource {
```
CommanderStorm, line 58: (different PR)

We likely should move all of these doubles to a separate module.
They are sort of growing...

martin-core/src/tiles/pmtiles/cache.rs
```diff
@@ -67,6 +70,21 @@ impl PmtCacheInstance {
...
     }
 
+    /// Creates a new instance with a globally unique ID.
+    #[must_use]
+    pub fn new_auto_id(cache: PmtCache) -> Self {
+        Self {
+            id: NEXT_PMT_CACHE_ID.fetch_add(1, Ordering::Relaxed),
+            cache,
+        }
+    }
```
CommanderStorm, line 77: 🤔 
Interesting that this works. I thought it would not for some reason I guess...
`AtomicUsize::new` is const since 1.24, not sure why I wrote it this way a year ago.

nit: I would prefer to keep the `static` a little closer to its usage:

```suggestion
    pub fn new_auto_id(cache: PmtCache) -> Self {
        static NEXT_PMT_CACHE_ID: AtomicUsize = AtomicUsize::new(0);

        let auto_id = NEXT_PMT_CACHE_ID.fetch_add(1, Ordering::Relaxed);
        Self::new(auto_id, cache)
    }
```

CommanderStorm (approved):
Yea, that is about what I would have done.
that `Source` keeps getting "random" (locally sensible) methods is somwhat calling for a bit of a rearchitecting of that trait, but that is out of scope for this PR.

I am a bit concearned about how the additions to `DynTileSource` make the already not that clear part of the codebase gets a bit less clear.
If we reduce the nesting there, I think it looks a hole lot better already.

Overall, good job!

martin-core/src/tiles/source.rs
gabeschine: i thought about this more: the "needs reload" condition will be hit at seemingly random times post-startup, and this code would crash the whole `martin` proc, right? that seems undesirable. a more graceful degradation would be better: log an error, and fail the request?

i can't think of a way to couple / assert at build, test, or during startup, that a `Source` impl _might_ return this new error enum, and that it has implemented this method. is there some rust trait magic that could pull this off?

martin/src/srv/tiles/content.rs
gabeschine: 🤷 seemed to work but also kind of unreadable. will update.

martin/src/srv/tiles/mod.rs
gabeschine: hm i don't actually think the fields need to be pub. i'll remove.

martin-core/src/tiles/source.rs
CommanderStorm: > the "needs reload" condition will be hit at seemingly random times post-startup, and this code would crash the whole martin proc, right? that seems undesirable

How? The code that calls this would need to throw `MartinCoreError::SourceNeedReload`.
I think crashing if we do such a logic mistake is fair.
Hard fails for these kinds of logic bugs that are unlikely to make it past review are preferable imo.

CommanderStorm:
Also CI is failing, otherwise please don't hesitate to merge once you are happy with this 😉

martin-core/src/tiles/source.rs
gabeschine: yeah i suppose the mistake would be on the developer. i'll add a comment on the error enum about implementing this method if that error is ever raised -- that should help tie the contract together in both directions.

martin/src/srv/tiles/content.rs
gabeschine: `xyz` is the param name used everywhere else. not sure if you were referring to something else but opting for consistency.

gabeschine:
> Also CI is failing, otherwise please don't hesitate to merge once you are happy with this 😉

@CommanderStorm the only CI failure is [CI / Test on windows-latest (pull_request)](https://github.com/maplibre/martin/actions/runs/27032341837/job/79790450909?pr=2858). I'm unable to tell it to restart. As far as I can tell it's an unrelated failure -- would you validate this and merge if ok?

gabeschine:
> > Also CI is failing, otherwise please don't hesitate to merge once you are happy with this 😉
> 
> @CommanderStorm the only CI failure is [CI / Test on windows-latest (pull_request)](https://github.com/maplibre/martin/actions/runs/27032341837/job/79790450909?pr=2858). I'm unable to tell it to restart. As far as I can tell it's an unrelated failure -- would you validate this and merge if ok?

For some reason the second time I merged main it kicked off new CI runs, and this time of course it passed :P.

CommanderStorm:
Yea, installing PG has been very unstable on windows the last few days.. like 50% flake rate

PR #2859, opened 2026-06-05
chore: update to maplibre_native@8.1

CommanderStorm:
a bit of a rewrite of the backend, but this will simplify the overlay PR


martin-core/src/resources/styles/render_pool.rs
```diff
@@ -182,127 +237,241 @@ fn default_worker_count() -> NonZeroUsize {
...
+            pixel_ratio,
+            loaded_style: None,
+        }
+    }
+
+    /// Whether this renderer's build-time geometry matches `params`.
+    fn matches(&self, params: &RenderParams) -> bool {
+        self.width == params.width
+            && self.height == params.height
+            && (self.pixel_ratio - params.pixel_ratio).abs() <= 0.01
```
michaelkirk, line 369: Why do we need slop here? Is it expected that it would shift only slightly? I'd expect it would be 1.0, 2.0, or 3.0

It's not a big deal, but it made me question if pixel_ratio meant what I thought it did.

justfile
```diff
@@ -82,21 +82,20 @@ gen-schemas:
     set -euo pipefail
     mkdir -p schemas
     # Include `rendering` on Linux (the only platform it compiles on) so the
-    # spec covers all routes a release build serves. `MLN_PRECOMPILE=1` skips
```
michaelkirk, line 85: What happened to `MLN_PRECOMPILE`?

.github/workflows/ci.yml
```diff
@@ -794,7 +794,16 @@ jobs:
       attestations: write
       # for writing release artifacts
       contents: write
-    needs: [ lint-unit-test-js, unit-test-rust, lint-rust, lint-rust-dependencies, validate-schemas, test-multi-os, test-with-svc, test-aws-lambda, test-publish, build-aarch64-apple-darwin, build-aarch64-unknown-linux-gnu, build-aarch64-unknown-linux-musl ]
+    # Only the build-binary jobs whose artifacts package downloads. Tests,
```
michaelkirk, line 797: Is the intent here that you will continue to package binaries even though tests have failed?

michaelkirk (approved):
Mostly I'm just following along to learn the code base, but LGTM

justfile
CommanderStorm: It unfortunately does not compile in v0.8 anymore.
Needs an upstream fix becausee of nasty cpp errors and missing functions (? don't quite remember what the exact error is, but this is the "average").

.github/workflows/ci.yml
CommanderStorm: no, this just shifts the way we block CI.
All of those are now under `done`.

It is a bit britte because keeping CI under 30m is a bit harder than it sounds, so every time I do a "major" change regarding how long some parts take, some rebalancing needs to happen.

The goal is to keep all 10x2 cores fed and the 10GB cache fully utilised, but not oversubscribed.

martin-core/src/resources/styles/render_pool.rs
CommanderStorm: I did this because these are two floats that are user provided.

They might legitimately be a bit 'different' without actually being functionally different.
This being a f32 is a bit of a wrong modelling, but that is the one that MLN uses internally...

CommanderStorm: I also have no idea why this is a float in MLN...

michaelkirk: I guess pixel ratio can be partial?

e.g. I think https://developer.android.com/training/multiscreen/screendensities shows HDPI as 1.5x pixel ratio.

Today I learned!

PR #2868, opened 2026-06-10
fix: migrate env-var substitution to serde-saphyr properties

### LINKED ISSUE #2851: Martin tries to substitute environment variables in config comments
Martin version 1.9.1

I have a config file containing the following:
```
...

# Database configuration. This can also be a list of PG configs.
postgres:
  # Database connection string. You can use env vars too, for example:
  #   $DATABASE_URL
  #   ${DATABASE_URL:-postgresql://postgres@localhost/db}
  connection_string: 'postgres://postgres:postgres@db:5432/ehrenamtskarte'

  functions:
    # Calls the `physical_stores` DB function
    physical_stores:
      schema: public
      function: physical_stores

...
```

Apparently martin tries to substitute environment variables in comments and errors, even though the configuration is actually correct:

```
2026-06-03T08:09:53.591718Z ERROR martin: martin::config::substitution (https://maplibre.org/martin/config-file/)
  × Unable to substitute environment variables in config file /app/
  │ config.yaml: No such variable: $DATABASE_URL
    ╭─[/app/config.yaml:16:8]
 15 │   # Database connection string. You can use env vars too, for example:
 16 │   #   $DATABASE_URL
    ·        ──────┬─────
    ·              ╰── No such variable: $DATABASE_URL
 17 │   #   ${DATABASE_URL:-postgresql://postgres@localhost/db}
    ╰────
  help: Make sure every ${VAR} reference resolves to an environment variable,
        or supply a default with `${VAR:-fallback}`.
```

CommanderStorm:
Fixes #2851.

## Summary
- Replaces `subst::substitute`'s text-level preprocessing — which scanned comments and quoted strings — with `serde-saphyr`'s `properties` feature, which only interpolates inside plain YAML scalar values. Comments containing `${VAR}` no longer abort startup.
- Adopts `PropertySyntax::BracedOrBare` so existing configs using `${VAR}`, `${VAR:-default}`, and bare `$VAR` keep working unchanged.
- Drops `subst` from the workspace. Pins `serde-saphyr` to the master rev `261ccbe6…` containing the four PRs we contributed upstream (#122 defaults, #125 modifiers, #128 error-on-missing, #129 bare syntax). Allowed via `deny.toml`; both have `TODO`s to drop once 0.0.28 is published.
- Substitution failures now flow through `ConfigFileError::YamlParseError` with a wrapper that emits `martin::config::substitution` as the diagnostic `code` (preserved from the old dedicated variant) when the underlying saphyr error is property-related, otherwise `martin::config::yaml`. Inline insta snapshots across the failing-deserializer tests were updated to match the consistent code/help/url footer.

Behavior change to flag in release notes: substitution no longer happens inside *quoted* or *block* scalars. A YAML like `connection_string: \"\${DATABASE_URL}\"` now keeps the literal `\${DATABASE_URL}` instead of expanding. Drop the quotes (or use `connection_string: \${DATABASE_URL}`) to get the previous behavior.

## Test plan
- [x] `cargo test -p martin --lib --no-default-features --features postgres,pmtiles` — 212/212 pass
- [x] Added regression test `substitution_ignores_dollar_tokens_in_comments` using the exact YAML from #2851
- [x] `cargo fmt --all --check` clean
- [ ] CI: full feature matrix via cargo-hack
- [ ] CI: cargo-deny (the new `allow-git` entry is the load-bearing change)
- [ ] CI: clippy `-D warnings` workspace-wide

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2901,9 +2901,9 @@ dependencies = [
 
 [[package]]
 name = "granit-parser"
-version = "0.0.3"
+version = "0.0.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f50ba32164f9e098d5da618776a32afbb32270adcbe3d3d006107dae11e37c91"
+checksum = "599a28c6c3171864dbf7fdb3a666f7ac301fab54c2b6e71534f413959cd37b49"
 dependencies = [
  "arraydeque",
  "smallvec",
@@ -4360,7 +4360,6 @@ dependencies = [
  "serde_with",
  "serde_yaml",
  "static-files 0.2.5",
- "subst",
  "tempfile",
  "testcontainers-modules",
  "thiserror 2.0.18",
@@ -6759,9 +6758,8 @@ dependencies = [
 
 [[package]]
 name = "serde-saphyr"
-version = "0.0.27"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5897b4c3faadadd35fdb6689f015641f3bc481d5adaaac56231ea15aeb243db3"
+version = "0.0.28"
+source = "git+https://github.com/bourumir-wyngs/serde-saphyr?rev=261ccbe62a8368fcffbd780dd923647edfd2f927#261ccbe62a8368fcffbd780dd923647edfd2f927"
 dependencies = [
  "ahash 0.8.12",
  "annotate-snippets",
@@ -7548,16 +7546,6 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn 2.0.117",
-]
-
-[[package]]
-name = "subst"
-version = "0.3.8"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0a9a86e5144f63c2d18334698269a8bfae6eece345c70b64821ea5b35054ec99"
-dependencies = [
- "memchr",
- "unicode-width 0.1.14",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -126,7 +126,13 @@ rustls-pemfile = "2"
 schemars = { version = "1.2", default-features = false, features = ["chrono04", "derive", "std"] }
 semver = "1"
 serde = { version = "1", features = ["derive"] }
-serde-saphyr = { version = "0.0.27", features = ["miette"] }
+# TODO: drop the `git` pin once a release containing PRs #122, #125, #128, #129 is published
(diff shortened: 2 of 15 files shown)


CommanderStorm:
## Update — integration tests surfaced a subst-specific syntax gap

Three CI groups are red. Two have the same root cause; one is the cargo-publish dry-run which is expected.

### `Test on ubuntu-latest`, `Test postgis:18-3.6 sslmode=require`, `Build and test docker images`

All three fail in the same place:

```
ERROR Failed to create postgres pool: Unable to parse connection string
${DATABASE_URL:postgres://postgres@localhost:5432/db}: invalid connection string
```

The fixtures (`tests/config.yaml:48`, `tests/config-process.yaml:12`) use **subst's single-colon default syntax** inside single quotes:

```yaml
connection_string: '${DATABASE_URL:postgres://postgres@localhost:5432/db}'
```

`subst` accepts this — its docs say:

> Variables have the form `$NAME`, `${NAME}` or `${NAME:default}`.
> ([subst-0.3.8 lib.rs:103](https://docs.rs/subst/0.3.8/subst/))

POSIX shell, docker-compose, and saphyr's `properties` all use `${VAR:-default}` (with the `-`). Saphyr would normally raise `InvalidPropertyName` for the single-colon form, but in these fixtures the string is **single-quoted**, and saphyr only interpolates plain scalars — so the placeholder reaches `tokio-postgres` verbatim and fails as an invalid connection string. Two behaviour deltas from `subst`, stacked.

Three options to consider:

1. **Upstream saphyr** — add `${NAME:default}` as an alias for `${NAME:-default}` (the docker-compose form). Easiest on downstreams migrating off `subst`, but unblocking this PR depends on an external review/release.
2. **Rewrite the fixtures + a release note** — change every `'${VAR:default}'` to unquoted `${VAR:-default}`. Breaking for users with the old syntax in their configs.
3. **Tiny pre-processor in Martin** — before handing the YAML to saphyr, rewrite `${VAR:default}` → `${VAR:-default}` in plain *and* quoted scalars. Keeps user configs compatible, no upstream dependency.

I lean (1) → (3) → (2). Want me to open the saphyr issue/PR? I already drafted the body, but filing an issue under your identity on an external repo needs your sign-off.

### `Test publishing via cargo-publish`

Expected to be red until 0.0.28 is on crates.io — `cargo publish --dry-run` can't verify a git dep. Same `TODO` already tracked in `Cargo.toml` and `deny.toml`.

### Pushed: clippy fix

`fc766f37` wraps the bare issue URL in the new regression test in `<…>` so `clippy::doc_markdown` passes.

CommanderStorm:
ups, left claude unatented for a bit 👀

PR #2873, opened 2026-06-11
refactor: split config/file/main.rs into focused modules

CommanderStorm:
Convert main.rs into a directory module.
Production code splits like this:
- the Config struct/types in config.rs,
- impl Config in lifecycle.rs, and
- the read_config / parse_config / deprecated-key migration pipeline in parse.rs.

purely mechanical code moves


CommanderStorm:
needs more splitting

PR #2874, opened 2026-06-11
refactor: extract impl Config to lifecycle module (confgi-1)

CommanderStorm:
Move main.rs into a directory module and pull the `impl Config` block into a sibling main/lifecycle.rs


michaelkirk (approved):
Assuming this is just cut and paste and updated imports.

CommanderStorm:
it is 😉

PR #2877, opened 2026-06-11
refactor: migrate YAML serialization to serde-saphyr, drop serde_yaml (config-5)

CommanderStorm:
simpifies the serde serialisation to strongly prefer using serde-saphyr.


tests/expected/auto/save_config.yaml
```diff
@@ -6,29 +6,29 @@ postgres:
...
+      schema: "\"Quotes' \\ \\' \\\" and Space.Dot."
       table: . Points" \ \' \" 'quote
       srid: 4326
       geometry_column: . ' \ \' \" "Geom"
       bounds:
-      - -170.94984639 # truncated to 10 digits
-      - -84.2002558073 # truncated to 10 digits
-      - 167.7089285828 # truncated to 10 digits
-      - 74.2357328475 # truncated to 10 digits
+  -  -170.94984639 # truncated to 10 digits
```
CommanderStorm: - https://github.com/bourumir-wyngs/serde-saphyr/pull/133

martin/src/config/file/main/parse.rs
```diff
@@ -66,54 +66,51 @@ fn needs_deprecated_migration(yaml: &str) -> bool {
 
 /// Migrates deprecated cache configuration keys in raw YAML before deserialization.
 ///
-/// This runs on the `serde_yaml::Value` directly, so the `Config` struct
+/// This runs on the `serde_json::Value` directly, so the `Config` struct
 /// never needs to know about deprecated field names.
-fn migrate_deprecated_config(value: &mut serde_yaml::Value) {
-    let Some(root) = value.as_mapping_mut() else {
+fn migrate_deprecated_config(value: &mut serde_json::Value) {
```
michaelkirk, line 71: I'm not familiar with this - but does it make sense to update the *migration* code? Presumably this is for existing configs which are in yaml, and now we want to parse them as json, surely that can't work. What am I missing?

Maybe it's just that we're using the JSON types as the target to deserialize yaml to these more familiar types, rather than the built in saphyr types?

tests/expected/auto/save_config.yaml
michaelkirk: Looks like we still need to regenerate these to take take advantage of the upstream fix? Is this PR blocked on that?

martin/src/config/file/main/parse.rs
CommanderStorm: No, saphyr does not have "catch-all" types as an intermediate datastructure.
This has pros and cons, the pro is that there are no allocations regarding them.

The con is, that I am using `serde_json::Value` as exactly this catch-all type.

michaelkirk (approved):
My understanding is we still need to wait on integrating https://github.com/bourumir-wyngs/serde-saphyr/pull/133 in order to produce valid yaml, but other than that LGTM

CommanderStorm:
and potentially
- https://github.com/bourumir-wyngs/serde-saphyr/pull/141

But arguably, I don't think your keys should be longer than 1024 Unicode chars…
Gotta love fuzzers 😉

CommanderStorm:
🎉 seems like all blockers are resolved now.
Just need an upstream release.

PR #2878, opened 2026-06-12
fix(unstable-cog): Allow for 0.1% (up to 3m of error at z0) in COGs from a matching zoom level in WebMercatorQuad

Auspicus:
- allow for up to 0.001 meters of error
- round to the nearest grid instead of flooring


CommanderStorm (approved):

martin-core/src/tiles/cog/source.rs
Auspicus: Maybe we should make this a fraction of the expected value rather than an absolute?

At the moment an absolute value of 0.001 is used which works well for lower zooms but around z25 it starts to get quite tight, especially at higher tile sizes.

CommanderStorm: whatever you think is best.
z25 is fairly close to 

https://github.com/maplibre/martin/blob/4bbe3af2115e4548b355e8dc0cdb86b39318ed58/martin-tile-utils/src/lib.rs#L18

I am not entirely worried, especially given that people are unlikely to have this detailed data..
Z20 is already in the range where you are measuring in cm per pixel and every zoom.

z25 is in the sub-centimeter range.. lets be honest, we will not have rasters this detailed

martin-core/src/tiles/cog/README.md
```diff
@@ -21,6 +21,47 @@ See [wiki.openstreetmap.org](https://wiki.openstreetmap.org/wiki/Slippy_map_tile
...
+### Resolution Error Tolerance
+
+When matching a COG image's resolution to a WebMercatorQuad zoom level, Martin applies
+`min(3.0m, resolution × 0.1%)` as the allowed error. The absolute cap of 3m dominates at low
+zoom levels (z0–z5 for 256px tiles, z0–z4 for 512px tiles); the 0.1% relative threshold takes
+over above that, keeping tolerance proportional to the pixel size.
+
+| Zoom | Resolution 256px (m/px) | Tolerance 256px (m) | Resolution 512px (m/px) | Tolerance 512px (m) |
+|------|-------------------------|---------------------|-------------------------|---------------------|
+| 0    | 156,543.034             | 3.0 (abs cap)       | 78,271.517              | 3.0 (abs cap)       |
```
CommanderStorm, line 33: just to be sure.. was this table done by you or claude? (aka do we need to double check it or not)

martin-core/src/tiles/cog/README.md
```diff
@@ -21,6 +21,47 @@ See [wiki.openstreetmap.org](https://wiki.openstreetmap.org/wiki/Slippy_map_tile
 | 3   | 256        | 1    | 78271.52                 |
 | 4   | 256        | 0    | 156543.03                |
 
+### Resolution Error Tolerance
+
+When matching a COG image's resolution to a WebMercatorQuad zoom level, Martin applies
```
CommanderStorm:
```suggestion
When matching a COG image's resolution to a `WebMercatorQuad` zoom level, Martin applies
```

CommanderStorm (approved):
LGTM still, assuming you clarify if the table content is trusted (=calculated by you) or needs to be verified (=calculated by claude)

martin-core/src/tiles/cog/README.md
Auspicus: Claude did it using a Python script. It's worth double checking. I eyeballed it and it seems about right. The math for the zoom level resolution is based on the EARTH_CIRCUMFERENCE in meters.

PR #2893, opened 2026-06-15
chore: improve logging of out-of-range zoom

### LINKED ISSUE #2888: Lack of information to analyze error message
Hello,

Since a migration to the 1.9.1 version, we have new error messages similar to the following:

{"timestamp":"2026-06-15T09:11:11.407652Z","level":"ERROR","fields":{"error":"\"No valid sources found\""},"target":"martin::srv::tiles::content"}
{"timestamp":"2026-06-15T09:11:11.407674Z","level":"WARN","fields":{"message":"Error encountered while processing the incoming HTTP request: \"No valid sources found\""},"target":"tracing_actix_web::middleware"}

We don't see any behavior issues from the user point of view but we would like to solve this error message, in case there is a real issue underneath.

Is it possible to change something in the configuration in order to have more information on the incoming request?

Regards
Alexandre

CommanderStorm:


tests/test.sh
```diff
@@ -778,6 +778,12 @@ test_accept_header pmt/0/0/0 "application/x-protobuf" 406
 test_accept_header pmt/0/0/0 "application/vnd.maplibre-vector-tile" 406
 test_accept_header pmt/0/0/0 "application/vnd.maplibre-tile" 406
 
+>&2 echo "***** Test out-of-zoom-range tiles return 404 (issue #2888) *****"
```
```suggestion
>&2 echo "***** Test out-of-zoom-range tiles return 404 *****"
```

Auspicus (approved):
Looks much better. A more Rust-like (sympathetic) error message 👌

I presume CI is failing due to log diff?

PR #2894, opened 2026-06-15
feat: duckdb-config-parsing

manbhav234:
#2264

This PR is in draft state right now. It implements config parsing for duckdb at `martin`. 

The core config parsing logic has been implemented. `builder` and `resolver` implementations need to be written. This is mostly similar to postgres with some adjustments based on the config structure for duckdb (which I am attaching below for reference):

```
duckdb:
  pool_size: 4
  auto_bounds: quick
  threads_per_query: 4
  memory_limit_mb: 2048
  sources:
    - database: /data/tiles.duckdb
      auto_publish:
        tables:
          from_schemas: autodetect
          source_id_format: "{table}"
          id_columns: [id, gid]
          extent: 4096
          buffer: 64
          clip_geom: true
        macros:
          from_schemas: autodetect
          source_id_format: "{macro}"

      tables:
        roads:
          schema: main
          table: roads
          geometry_column: geom
          srid: 4326
          minzoom: 0
          maxzoom: 14
          properties:
            id: int4
            name: varchar

      macros:
        custom:
          schema: main
          macro: get_custom_tile

    - geoparquet: /data/buildings.parquet
      layer_id: buildings
      geometry_column: geom
      srid: 4326
      minzoom: 0
      maxzoom: 14
      extent: 4096
      buffer: 64
```

### DIFF
--- a/martin/Cargo.toml
+++ b/martin/Cargo.toml
@@ -101,6 +101,7 @@ required-features = ["postgres"]
 
 [features]
 default = [
+    "duckdb",
     "fonts",
     "lambda",
     "mbtiles",
@@ -115,6 +116,7 @@ default = [
 ]
 unstable-cog = ["martin-core/unstable-cog", "_tiles"]
 rendering = ["styles", "martin-core/rendering", "dep:image"]
+duckdb = ["martin-core/duckdb", "_tiles"]
 fonts = ["martin-core/fonts", "_catalog"]
 lambda = ["dep:lambda-web"]
 mbtiles = ["martin-core/mbtiles", "dep:mbtiles", "_tiles"]
--- a/martin/src/config/file/tiles/duckdb/builder.rs
+++ b/martin/src/config/file/tiles/duckdb/builder.rs
@@ -1,0 +1,5 @@
+//! DuckDB source auto-discovery and instantiation.
+
+/// Builder for [`DuckDBSource`](martin_core::tiles::duckdb::DuckDBSource) auto-discovery.
+#[derive(Debug, Default)]
+pub struct DuckDbAutoDiscoveryBuilder;
--- a/martin/src/config/file/tiles/duckdb/config_geoparquet.rs
+++ b/martin/src/config/file/tiles/duckdb/config_geoparquet.rs
@@ -1,0 +1,251 @@
+use std::collections::BTreeMap;
+use std::fmt::{self, Display, Formatter};
+use std::path::{Path, PathBuf};
+
+use martin_tile_utils::{Encoding, Format, TileInfo};
+use serde::{Deserialize, Serialize};
+use tilejson::{Bounds, TileJSON, VectorLayer};
+use url::Url;
+
+use super::DuckDbInfo;
+use crate::config::args::BoundsCalcType;
+#[cfg(feature = "unstable-schemas")]
+use crate::config::file::duckdb::config_table::bounds_world_example;
+use crate::config::file::file_config::is_remote_url;
+use crate::config::file::{CachePolicy, ConfigFileError, ConfigFileResult, UnrecognizedValues};
+#[cfg(all(feature = "mlt", feature = "_tiles"))]
+use crate::config::file::{MltProcessConfig, MvtProcessConfig};
+
+/// Resolved GeoParquet location after config finalization.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum GeoParquetTarget {
+    /// Local `.parquet` file served via an in-memory DuckDB connection.
+    Local(PathBuf),
+    /// Remote parquet object served via DuckDB `httpfs`.
+    Remote(Url),
+}
+
+impl Display for GeoParquetTarget {
+    fn fmt(&self, formatter: &mut Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Local(path) => Display::fmt(&path.display(), formatter),
+            Self::Remote(url) => Display::fmt(url, formatter),
+        }
+    }
+}
+
+/// Parse a configured `geoparquet:` value into the target type expected by `martin-core`.
+pub fn parse_geoparquet_target(raw: &str) -> ConfigFileResult<GeoParquetTarget> {
+    let path = Path::new(raw);
+    if is_remote_url(path) {
+        let url = Url::parse(raw)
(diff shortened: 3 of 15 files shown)


manbhav234:
I think it's better to get this reviewed before proceeding. 
I'll get the CI checks to pass, resolve clippy errors etc.

martin/Cargo.toml
```diff
@@ -115,6 +116,7 @@ default = [
 ]
 unstable-cog = ["martin-core/unstable-cog", "_tiles"]
 rendering = ["styles", "martin-core/rendering", "dep:image"]
+duckdb = ["martin-core/duckdb", "_tiles"]
```
CommanderStorm, line 119: please name it `unstable-` while we are still developing it.
This way, we can ship releases even with this half-done.

```suggestion
unstable-duckdb = ["martin-core/duckdb", "_tiles"]
```

martin/src/config/file/tiles/duckdb/builder.rs
```diff
@@ -0,0 +1,5 @@
+//! DuckDB source auto-discovery and instantiation.
+
+/// Builder for [`DuckDBSource`](martin_core::tiles::duckdb::DuckDBSource) auto-discovery.
+#[derive(Debug, Default)]
+pub struct DuckDbAutoDiscoveryBuilder;
```
CommanderStorm, line 5: do note that the postgres implementation is largely a product of how it was implemented.

If it is simpler to add the code for this to discovery or name it differently, that is fine too.

martin/src/config/file/tiles/duckdb/config_geoparquet.rs
CommanderStorm, line 1: instead of having four modules named `config*`, please have one `mod` named `config`, which has different modules.
Much cleaner overall.

martin/src/config/file/tiles/duckdb/spec.rs
```diff
@@ -0,0 +1,68 @@
...
+    /// A macro source. SQL may already be known from catalog discovery.
+    Macro(MacroInfo, DuckDBSqlInfo),
+    /// A single-layer GeoParquet source.
+    GeoParquet(DuckDbGeoParquetSourceConfig),
+}
+
+impl SourceSpec {
+    /// Content hash over fields that affect served tile bytes or metadata.
+    #[must_use]
+    pub fn fingerprint(&self) -> u128 {
```
CommanderStorm, line 23: Please move this code to `TableInfo` or better desctructure each `info` value here.
This way, it is more obvious if a field gets added and which fields get ignored.

This way, we get this code to be defensive against fields getting added.
See https://corrode.dev/blog/defensive-programming/ for a more in-depth argument why this style is better.

martin/src/config/file/tiles/duckdb/utils.rs
CommanderStorm, line 1: Instead of copying postgres code 1:1, please extract sharable helpers.

https://github.com/maplibre/martin/blob/8fe9ee45fe51387235f503ec3d8b8726ef411f30/martin/src/config/file/tiles/postgres/utils.rs

martin/src/config/file/tiles/duckdb/config_geoparquet.rs
```diff
@@ -0,0 +1,251 @@
...
+    /// Zoom-level bounds for tile caching (overrides top-level cache).
+    #[cfg_attr(
+        feature = "unstable-schemas",
+        schemars(with = "Option<crate::config::file::CachePolicyShape>")
+    )]
+    pub cache: Option<CachePolicy>,
+
+    /// Override wrapper-level pool size for this source entry.
+    #[cfg_attr(feature = "unstable-schemas", schemars(example = &4usize))]
+    pub pool_size: Option<usize>,
```
CommanderStorm, line 116: Please do it like this and remove the error.
If we move code to types, we make our life much simpler.

```suggestion
    pub pool_size: Option<NonZeroUsize>,
```

martin/src/config/file/tiles/duckdb/config_geoparquet.rs
```diff
@@ -0,0 +1,251 @@
...
+        tilejson.bounds = self.bounds;
+        tilejson.vector_layers = Some(vec![VectorLayer {
+            id: layer_id,
+            fields: BTreeMap::default(),
+            description: None,
+            maxzoom: None,
+            minzoom: None,
+            other: BTreeMap::default(),
+        }]);
+        tilejson
```
CommanderStorm, line 201: nit: instead of the mutating approach, please use the conveniences of the `tilejson!` macro.

I think that would be the following, +- syntax issues


```suggestion
        tilejson::tilejson! {
            tiles: vec![],
            name: source_id,
            description: self.format_id(),
            minzoom: self.minzoom,
            maxzoom: self.maxzoom,
            bounds: self.bounds,
            vector_layers: Some(vec![VectorLayer {
              id: layer_id,
              fields: BTreeMap::default(),
              description: None,
              maxzoom: None,
              minzoom: None,
              other: BTreeMap::default(),
            }]),
        }])
        }
```

CommanderStorm (changes requested):
Sorry, but 1.6k is a loot of code.
Yes, it is plumbing, but it is not plumbed into anything which makes it harder to review.
It would be much simpler to review if this were plumbed up (so if you were going in vertical instead of horizontal slices)
=> Can you split this into multiple (stacked) PRs?

So for example
- the bounds and utils refactoring,
- adding a basic (non-functional, behind an unstable feature) base to this part of the config system and
- then each of the different config sources (geoparquet, macros, tables) as we implement each of them?

This way, the diff stays smaller and I can actually keep in my head what is happeinig. for 1.6k lines, I cannot.

manbhav234:
> Sorry, but 1.6k is a loot of code. Yes, it is plumbing, but it is not plumbed into anything which makes it harder to review. It would be much simpler to review if this were plumbed up (so if you were going in vertical instead of horizontal slices) => Can you split this into multiple (stacked) PRs?
> 
> So for example
> 
> * the bounds and utils refactoring,
> * adding a basic (non-functional, behind an unstable feature) base to this part of the config system and
> * then each of the different config sources (geoparquet, macros, tables) as we implement each of them?
> 
> This way, the diff stays smaller and I can actually keep in my head what is happeinig. for 1.6k lines, I cannot.

Yup, felt the same. sorry for the inconvenience. will refactor this properly and do small incremental stacked PR's so that it's easy to track and review

Issue #2895, opened 2026-06-15
Martin potentially leaks sensitive passowrd information when printing BadConnectionString/PostgresError


planted-sam:
Noticed when a coworker was setting up a new server and entered an incorrectly formatted connection string on our server. 

```
2026-06-15T19:13:09.861445Z  INFO martin: Starting Martin v1.10.1
2026-06-15T19:13:09.861460Z  INFO martin: Using /config/martin-config.yaml
2026-06-15T19:13:09.862482Z ERROR resolve:resolve_tile_sources: martin::config::file::main: error=ConfigFileError(PostgresPoolCreationFailed(BadConnectionString(Error { kind: ConfigParse, cause: Some("unterminated parameter") }, "postgres://postgres:testpassword@host.docke???WQD?wq/dqdr.internal:5432/database")))
2026-06-15T19:13:09.862498Z ERROR resolve: martin::config::file::main: error=ConfigFileError(PostgresPoolCreationFailed(BadConnectionString(Error { kind: ConfigParse, cause: Some("unterminated parameter") }, "postgres://postgres:testpassword@host.docke???WQD?wq/dqdr.internal:5432/database")))
2026-06-15T19:13:09.862507Z ERROR martin: Failed to create postgres pool: Unable to parse connection string postgres://postgres:testpassword@host.docke???WQD?wq/dqdr.internal:5432/database: invalid connection string
```
(I added a bunch of gibberish to the conn string here, making it invalid, but my `testpassword` still gets logged sadly)

We noticed that unlike other logged postgres errors (e.g. when the password or port is incorrect but the connection string is valid) that would attempt to obfuscate the password field by only logging `Some(_)` for the value - when the connection string was incorrectly formatted the whole text of the connection string was printed to the log - I believe because of this line in the connection string parsing function that passes that conn_str along with the error.

https://github.com/maplibre/martin/blob/8fe9ee45fe51387235f503ec3d8b8726ef411f30/martin-core/src/tiles/postgres/tls.rs#L52

Happy to make a quick PR for this, but wanted to clarify what the ideal behavior would be. My thought would be to replace the `conn_str` with a generic message, but I want to make sure I'm keeping the change in line with project norms. Let me know what makes sense here!

CommanderStorm:
> Happy to make a quick PR for this

Sure, I would love for this.

> My thought would be to replace the conn_str with a generic message

I don't think that would work.
Lets say we change this to a generic `some postgres connection string is invalid`.
[...]

PR #2898, opened 2026-06-16
Obfuscate password in incorrect postgres connection string logs

planted-sam:
- Added function for detecting/obfuscating password portion of connection string (even if incorrectly formatted)
- Use nicer config error so user can see context/location of problem connection string

Fixes issue here: https://github.com/maplibre/martin/issues/2895

### DIFF
--- a/martin-core/src/tiles/postgres/tls.rs
+++ b/martin-core/src/tiles/postgres/tls.rs
@@ -62,6 +62,16 @@ pub fn parse_conn_str(conn_str: &str) -> PostgresResult<(Config, SslModeOverride
     Ok((pg_cfg, mode))
 }
 
+/// Validate that `conn_str` is a syntactically valid Postgres connection string.
+///
+/// Returns the same password-redacted [`BadConnectionString`](crate::tiles::postgres::PostgresError::BadConnectionString)
+/// error as [`parse_conn_str`] on failure. Intended for validating connection strings early
+/// (e.g. while deserializing a config file) so a malformed value is reported against its source
+/// location, rather than surfacing later — without a location — when the pool is created.
+pub fn validate_conn_str(conn_str: &str) -> PostgresResult<()> {
+    parse_conn_str(conn_str).map(|_| ())
+}
+
 #[derive(Debug)]
 struct NoCertificateVerification;
 
--- a/martin-core/src/tiles/postgres/tls.rs
+++ b/martin-core/src/tiles/postgres/tls.rs
@@ -254,17 +254,14 @@ mod tests {
 
     #[test]
     fn bad_conn_str_error_hides_password() {
-        // A malformed connection string must not leak the password into the error message.
+        // A malformed connection string must carry only the password-redacted form, never the
+        // original password. Assert on the stored string directly (not the rendered message) so
+        // the test doesn't depend on the wording of the underlying tokio-postgres error.
         let err = parse_conn_str("postgres://postgres:testpassword@host.docke???WQD?wq/db")
             .expect_err("malformed connection string should fail to parse");
-        let msg = err.to_string();
-        assert!(
-            !msg.contains("testpassword"),
-            "password leaked in error: {msg}"
-        );
-        assert!(
-            msg.contains("****"),
-            "expected redaction marker in error: {msg}"
-        );
+        let BadConnectionString(_, conn_str) = err else {
+            panic!("expected BadConnectionString, got {err:?}");
+        };
+        assert_eq!(conn_str, "postgres://postgres:****@host.docke???WQD?wq/db");
     }
 }
--- a/martin-core/src/tiles/postgres/tls.rs
+++ b/martin-core/src/tiles/postgres/tls.rs
@@ -67,7 +67,7 @@ pub fn parse_conn_str(conn_str: &str) -> PostgresResult<(Config, SslModeOverride
 /// Returns the same password-redacted [`BadConnectionString`](crate::tiles::postgres::PostgresError::BadConnectionString)
 /// error as [`parse_conn_str`] on failure. Intended for validating connection strings early
 /// (e.g. while deserializing a config file) so a malformed value is reported against its source
-/// location, rather than surfacing later — without a location — when the pool is created.
+/// location, rather than surfacing later -- without a location -- when the pool is created.
 pub fn validate_conn_str(conn_str: &str) -> PostgresResult<()> {
     parse_conn_str(conn_str).map(|_| ())
 }
--- a/martin-core/src/tiles/postgres/tls.rs
+++ b/martin-core/src/tiles/postgres/tls.rs
@@ -22,6 +22,7 @@ use crate::tiles::postgres::PostgresError::{
     InvalidPrivateKey, UnknownSslMode,
 };
 use crate::tiles::postgres::PostgresResult;
+use crate::tiles::postgres::utils::redact_conn_str;
 
 /// A temporary workaround for <https://github.com/sfackler/rust-postgres/pull/988>
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
@@ -49,7 +50,7 @@ pub fn parse_conn_str(conn_str: &str) -> PostgresResult<(Config, SslModeOverride
     } else {
         Config::from_str(conn_str)
     };
-    let mut pg_cfg = pg_cfg.map_err(|e| BadConnectionString(e, conn_str.to_string()))?;
+    let mut pg_cfg = pg_cfg.map_err(|e| BadConnectionString(e, redact_conn_str(conn_str)))?;
     if let SslModeOverride::Unmodified(_) = mode {
         mode = SslModeOverride::Unmodified(pg_cfg.get_ssl_mode());
     }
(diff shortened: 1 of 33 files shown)


martin-core/src/tiles/postgres/tls.rs
```diff
@@ -250,4 +251,20 @@ mod tests {
...
+            .expect_err("malformed connection string should fail to parse");
+        let msg = err.to_string();
+        assert!(
+            !msg.contains("testpassword"),
+            "password leaked in error: {msg}"
+        );
+        assert!(
+            msg.contains("****"),
+            "expected redaction marker in error: {msg}"
+        );
```
CommanderStorm: please `assert_eq` on the string. This is much less brittle.

martin-core/src/tiles/postgres/utils.rs
```diff
@@ -19,3 +20,89 @@ pub fn query_to_json(query: Option<&UrlQuery>) -> Json<HashMap<String, serde_jso
...
+        .expect("the regex is valid");
+    let redacted = url_re.replace_all(conn_str, "${pre}****${at}");
+
+    // Keyword form: `password=...`, optionally single- or double-quoted. Require a separator
+    // (start of string, whitespace, `&` or `?`) before the keyword so we don't match a suffix
+    // like `mypassword=`.
+    let kw_re = Regex::new(r#"(?P<pre>(?:^|[\s&?])password=)(?:'[^']*'|"[^"]*"|[^\s&]*)"#)
+        .expect("the regex is valid");
+    kw_re.replace_all(&redacted, "${pre}****").into_owned()
+}
```
CommanderStorm, line 57: wtf is happening here and why does this need a regex?
I would have expected that we can just use the facilities like `url` to not do url parsing ourselves..

martin-core/src/tiles/postgres/utils.rs
```diff
@@ -19,3 +20,89 @@ pub fn query_to_json(query: Option<&UrlQuery>) -> Json<HashMap<String, serde_jso
...
+
+#[cfg(test)]
+mod tests {
+    use super::redact_conn_str;
+
+    #[test]
+    fn redacts_url_password() {
```
CommanderStorm: please use rstest for repeated tests on different input

martin/tests/binary_diagnostics.rs
```diff
@@ -61,3 +61,64 @@ fn graphical_format_emits_human_diagnostic_on_stderr() {
...
+        "password leaked into diagnostic:\n{stderr}"
+    );
+    insta::assert_snapshot!(stderr.trim(), @"
+    martin::config::postgres::pool_creation (https://maplibre.org/martin/config-file/)
+
+      × Failed to create postgres pool: Unable to parse connection string
+      │ postgres://user:****@bad???host/db: invalid connection string
+       ╭─[<config>:2:22]
+     1 │ postgres:
+     2 │   connection_string: postgres://user:****@bad???host/db
```
CommanderStorm: after my config refactoring is done, you will not be able to leak env vars like this.

-> Please remove the test, I don't think it currently adds coverage, or?

martin/src/config/file/error.rs
```diff
@@ -188,6 +258,69 @@ impl Diagnostic for SubstitutionReport {
...
+/// Returns `None` for an empty needle or when the text isn't found (e.g. unusual YAML quoting),
+/// in which case the caller renders the error without a snippet.
+#[cfg(feature = "postgres")]
+fn locate_span(haystack: &str, needle: &str) -> Option<SourceSpan> {
+    if needle.is_empty() {
+        return None;
+    }
+    let offset = haystack.find(needle)?;
+    Some(SourceSpan::new(offset.into(), needle.len()))
+}
```
CommanderStorm: I think this is wrong, this needs to be done properly end to end, not recovered.

The reason is that I can imagine that there might be a way to have two urls, one invalid (shorter) and one valid (longer).
if valid is before invalid, we now point to valid and say it is invalid.

Not great..

CommanderStorm (changes requested):
I don't think the code is ready just yet.
Could you clean it up a round?

planted-sam:
@CommanderStorm Just pushed my changes to address your feedback, let me know what you think! Thank you for that initial round of feedbcak btw!

martin-core/src/tiles/postgres/utils.rs
planted-sam: Replaced with `url` crate usage to be a bit more straightforward.

martin/src/config/file/error.rs
planted-sam: I think I've simplified this more now, let me know if I should make further edits

martin/tests/binary_diagnostics.rs
```diff
@@ -61,3 +61,28 @@ fn graphical_format_emits_human_diagnostic_on_stderr() {
...
+    let yaml = "postgres:\n  connection_string: postgres://user:hunter2@bad???host/db\n";
+    let stderr = run_with_bad_config(yaml, &[]);
+    assert!(
+        !stderr.contains("hunter2"),
+        "password leaked into diagnostic:\n{stderr}"
+    );
+    assert!(
+        stderr.contains("****"),
+        "expected the redaction marker in diagnostic:\n{stderr}"
+    );
```
CommanderStorm, line 79:
```suggestion
```

martin/src/config/file/tiles/postgres/config.rs
```diff
@@ -499,11 +525,38 @@ mod tests {
...
+        );
+        // Points at the invalid second source (line 3), never the valid first one.
+        assert!(
+            rendered.contains("config.yaml:3:"),
+            "expected the diagnostic to point at line 3:\n{rendered}"
+        );
+        assert!(
+            rendered.contains("postgres://user:****@bad???host"),
+            "expected the redacted invalid connection string:\n{rendered}"
+        );
```
CommanderStorm, line 557: `assert_eq` if the result is stable or `insta` if it might result in changes please

martin/src/config/file/error.rs
```diff
@@ -60,6 +60,14 @@ pub enum ConfigFileError {
     #[error("Failed to create postgres pool: {0}")]
     PostgresPoolCreationFailed(#[source] PostgresError),
 
+    /// A `connection_string` in the config file is syntactically invalid.
+    ///
+    /// Detected while deserializing, so the password is redacted and the exact source location
+    /// is known; rendered without a source snippet so the raw line is never echoed back.
+    #[cfg(feature = "postgres")]
+    #[error("{}", .0.message)]
+    InvalidConnectionString(Box<InvalidConnectionStringDetails>),
```
CommanderStorm, line 69: why Box?

martin/src/config/file/tiles/postgres/config.rs
```diff
@@ -499,11 +525,38 @@ mod tests {
...
     use crate::config::primitives::env::FauxEnv;
+    use crate::config::test_helpers::render_failure;
 
     pub fn parse_cfg(yaml: &str) -> Config {
         parse_config(yaml, &FauxEnv::default(), Path::new("<test>")).unwrap()
     }
 
+    /// A malformed `connection_string` is reported against its own source location, with the
+    /// password redacted and no source snippet -- even when a *valid* connection string precedes
+    /// the invalid one (regression test: a text search could mislocate onto the valid line).
```
CommanderStorm, line 536: Please do another cleanup pass.

There are numerous comments that read like they were generated by an LLM. They add noise rather than clarity and make the code harder to review. Please remove them and keep only comments that provide information not already obvious from reading the code.

martin-core/src/tiles/postgres/tls.rs
```diff
@@ -61,6 +62,16 @@ pub fn parse_conn_str(conn_str: &str) -> PostgresResult<(Config, SslModeOverride
...
 
+/// Validate that `conn_str` is a syntactically valid Postgres connection string.
+///
+/// Returns the same password-redacted [`BadConnectionString`](crate::tiles::postgres::PostgresError::BadConnectionString)
+/// error as [`parse_conn_str`] on failure. Intended for validating connection strings early
+/// (e.g. while deserializing a config file) so a malformed value is reported against its source
+/// location, rather than surfacing later -- without a location -- when the pool is created.
+pub fn validate_conn_str(conn_str: &str) -> PostgresResult<()> {
+    parse_conn_str(conn_str).map(|_| ())
+}
```
CommanderStorm, line 73: Why is this a function? This seems very strange

CommanderStorm (changes requested):
Please do another cleanup pass.

There are a number of comments that read like they were generated by an LLM. They add noise rather than clarity and make the code harder to review. Please remove them and keep only comments that provide information not already obvious from reading the code.

martin-core/src/tiles/postgres/utils.rs
CommanderStorm: you are still using regex..

CommanderStorm:
I think the PR is a bit unclean.

Lets do it this way instead:
- #2944 

Sorry for the wasted effort, but I think should be understandable 😉

PR #2899, opened 2026-06-18
feat: update to mlt-core 0.11.1

michaelkirk:
Note: this breaks the martin config because `allow_fpf` was renamed to `allow_fastpfor`

(In pursuit of my "get martin generated MLT working with native and www")

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -168,4 +168,3 @@ similar.opt-level = 3
 # Make release binaries a bit smaller
 lto = true
 codegen-units = 1
-
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -102,7 +102,7 @@ martin-tile-utils = { path = "./martin-tile-utils", version = "0.7.4" }
 mbtiles = { path = "./mbtiles", version = "0.18.0" }
 md5 = "0.8.0"
 miette = { version = "7.6", features = ["fancy"] }
-mlt-core = { version = "0.11.0", default-features = false }
+mlt-core = { version = "0.11.1", default-features = false }
 moka = { version = "0.12", default-features = false }
 notify = "8.2.0"
 num_cpus = "1"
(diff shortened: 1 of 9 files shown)


martin/src/config/file/process.rs
```diff
@@ -69,7 +69,7 @@ pub struct MltEncoderConfig {
     /// Allow FSST string compression.
     pub allow_fsst: Option<bool>,
     /// Allow `FastPFOR` integer compression.
-    pub allow_fpf: Option<bool>,
+    pub allow_fastpfor: Option<bool>,
```
CommanderStorm, line 73: To make sure that this is non-breaking

```suggestion
    #[serde(alias="allow_fpf")]
    pub allow_fastpfor: Option<bool>,
```

CommanderStorm (approved):
Thanks

Cargo.toml
```diff
@@ -102,7 +102,7 @@ martin-tile-utils = { path = "./martin-tile-utils", version = "0.7.4" }
 mbtiles = { path = "./mbtiles", version = "0.18.0" }
 md5 = "0.8.0"
 miette = { version = "7.6", features = ["fancy"] }
-mlt-core = { version = "0.9.1", default-features = false }
+mlt-core = { version = "0.11.0", default-features = false }
```
CommanderStorm: To also contain your geom fix 😉 

```suggestion
mlt-core = { version = "0.11.1", default-features = false }
```

michaelkirk:
https://github.com/maplibre/martin/actions/runs/27770911380/job/82211262898 failed with

>     /usr/bin/ar: unable to copy file 'libmbgl-core.a'; reason: No space left on device

While building maplibre-native. I'm going to kick it to re-run because that's easy, but I don't actually expect it to fix it.

CommanderStorm:
> maplibre-native

Unrelated fail, at least I don't see how it could be related.
Force merging.

Realistically, we want to get back to Precompiled assets for that. For time sakes alone, since CI currently is very very slow
- https://github.com/maplibre/maplibre-native-rs/pull/242

PR #2901, opened 2026-06-19
feat: propagate cancellation to postgres #2646

### LINKED ISSUE #2646: Implement graceful cancellation to terminate PostgreSQL queries
Currently, interrupting `martin` or `martin-cp` (e.g., via `Ctrl+C`) stops the client process, but it does not terminate the corresponding active queries on the database side. Ideally, a client-side cancellation should propagate to PostgreSQL to stop and clean up the backend processes.

Reference:
This is similar to the behavior reported here: https://github.com/maplibre/martin/issues/6

cc: @nyurik

Weixing-Zhang:
Goal:
- martin-cp no longer hangs on Ctrl+C when copying from PostgreSQL.

Changes:
- Add ActiveQueryRegistry to track in-flight Postgres queries and cancel them via CancelToken
- Register/unregister tokens in PostgresSource::get_tile; expose registry through Source::cancel_registery()
- Refactor martin-cp tile copy into producer/consumer tasks; race tile fetching against Ctrl+C with tokio::select!
- On interrupt: cancel active Postgres queries, drain queued tiles (10s timeout), skip metadata updates

Performance Benchmark:
Running `just bench-http`
| # | Experiment | main RPS | branch RPS | main p50 | branch p50 | main p99 | branch p99 |
|---|------------|---------:|-----------:|---------:|---------:|---------:|---------:|
| 1 | PG default (`function_zxy_query`) | 43,490 | 44,182 | 1.10 ms | 1.10 ms | 2.72ms | 1.44 ms |
| 2 | PG MapLibre (`Accept: application/vnd.maplibre-tile`) | 44,827 | 44,791 | 1.08 ms | 1.08ms | 1.52 ms | 1.43 ms |


CommanderStorm:
Can you benchmark (the throughput based benchmark via oha in the justfile) what the effect of this change is on perf?
Adding a multiple central points does not sound like a very performant idea..
Not sure.

Weixing-Zhang:
@CommanderStorm 

I ran `just bench-http` with a running bench server via `just bench-server`.

| # | Experiment | main RPS | branch RPS | main p50 | branch p50 | main p99 | branch p99 |
|---|------------|---------:|-----------:|---------:|---------:|---------:|---------:|
| 1 | PG default (`function_zxy_query`) | 43,490 | 44,182 | 1.10 ms | 1.10 ms | 2.72ms | 1.44 ms |
| 2 | PG MapLibre (`Accept: application/vnd.maplibre-tile`) | 44,827 | 44,791 | 1.08 ms | 1.08ms | 1.52 ms | 1.43 ms |

I do not see any meaningful performance regression from the CancelToken registry. Throughput is within noise (~±1–2%), p50 is unchanged, and p99 is equal or better on the branch for both scenarios.

martin-core/src/tiles/postgres/source.rs
```diff
@@ -140,6 +152,9 @@ impl Source for PostgresSource {
             })?
             .unwrap_or_default();
 
+        // Active query registery cleanup after get_tile() is finished
+        self.pool.active_query_registry().unregister(reg_id).await;
```
CommanderStorm: this `unregister` is not hit in every case because the error cases above may early-abort.

Can you please use the drop-guard pattern to prevent this on a structural level?

martin-core/src/tiles/postgres/pool.rs
```diff
@@ -235,10 +248,61 @@ SELECT (regexp_matches(
...
+        let tokens: Vec<CancelToken> = self.tokens.lock().await.values().cloned().collect();
+        for token in tokens {
+            let _ = token
+                .cancel_query(deadpool_postgres::tokio_postgres::NoTls)
+                .await;
+        }
+    }
+}
+
+#[cfg(all(test, feature = "postgres"))]
```
CommanderStorm: no, the gating to `test-pg` was intentional, otherwise the non-pg tests get the heavy weight pg test, which is not great..

We can move the `test-pg` gate one level down, but we should not remove it

martin-core/src/tiles/postgres/pool.rs
```diff
@@ -235,10 +248,61 @@ SELECT (regexp_matches(
...
+    pub async fn unregister(&self, id: u64) {
+        self.tokens.lock().await.remove(&id);
+    }
+
+    /// Cancels all registered queries.
+    pub async fn cancel_all(&self) {
+        let tokens: Vec<CancelToken> = self.tokens.lock().await.values().cloned().collect();
+        for token in tokens {
+            let _ = token
+                .cancel_query(deadpool_postgres::tokio_postgres::NoTls)
```
CommanderStorm: I think the tls option here will not work on servers with TLS enabled.
We already have config for this, just needs to be plumbed through to here 😉

martin-core/src/tiles/source.rs
```diff
@@ -46,6 +48,14 @@ pub trait Source: Send + Sync + Debug {
         false
     }
 
+    /// Returnes the cancellation registry for queries
+    ///
+    /// Only works for postgresql sources for now
+    #[cfg(feature = "postgres")]
+    fn cancel_registery(&self) -> Option<ActiveQueryRegistry> {
```
CommanderStorm: nit: I think this is a typo

```suggestion
    fn cancel_registry(&self) -> Option<ActiveQueryRegistry> {
```

martin/src/bin/martin-cp.rs
CommanderStorm, line 1: changes here LGTM, but it is hard to keep the refactorings from the actual logic changes.

Splitting the refactoring into another PR would help me review this for the future, but is not nessary if you don't want to.

martin-core/src/tiles/postgres/pool.rs
```diff
@@ -235,10 +248,61 @@ SELECT (regexp_matches(
...
+                .await;
+        }
+    }
+}
```
CommanderStorm: So the counter is only ever accessed directly before aquiring the mutex, so we can move it into the mutex anyhow.

Also, the access to the mutex should be very slow and is not held across await points, so I would prefer the simpler sync mutex since the reasons for tokios mutex are not here..

Also, lets use a `JoinSet` for canceling all and RAII for the registering.

Below is approximately the code that would do this, I have not tested it, it won't compile unless the other aspects are also changed accordingly. ^^

```suggestion
#[derive(Default)]
struct ActiveQueryRegistryInner {
    next_id: u64,
    tokens: HashMap<u64, CancelToken>,
}

#[derive(Clone, Default)]
pub struct ActiveQueryRegistry(Arc<std::sync::Mutex<ActiveQueryRegistryInner>>);

impl ActiveQueryRegistry {
    /// Registers a query's [`CancelToken`] and returns its registry id.
    #[must_use]
    pub fn register(&self, token: CancelToken) -> ActiveQueryGuard {
        let id = {
          let mut g = self.0.lock().expect("poisoned");
          let id = g.next_id;
          g.next_id += 1;
          g.tokens.insert(id, token);
          id
        };
        ActiveQueryGuard { registry: Arc::clone(&self.0), id }
    }

    /// Cancels all registered queries.
    pub async fn cancel_all(&self) {
        let tokens = { 
           let g = self.tokens.lock().expect("poisoned");
           std::mem::take(g.tokens)
        };
        let mut set = tokio::task::JoinSet::new();
        for t in tokens.into_values() {
            set.spawn(async move { 
              let _ = t.cancel_query(deadpool_postgres::tokio_postgres::NoTls).await; 
            });
        }
        set.join_all().await;
    }
}

/// Removes its query from the [`ActiveQueryRegistry`] when dropped.
pub struct ActiveQueryGuard {
    registry: Arc<Mutex<Inner>>,
    id: u64,
}

impl Drop for ActiveQueryGuard {
    fn drop(&mut self) {
        self.registry
            .lock()
            .unwrap_or_else(PoisonError::into_inner)
            .tokens
            .remove(&self.id);
    }
}
```

CommanderStorm (changes requested):
Mostly LGTM.
Perf seems to not be bottlenecked by this, which is good

I would like you to consider a sligtly different design to how the registry works to avoid the correctness bugs here regarding unlocking 😉 

I am also wondering if we need to make this more generic, so not "only" propagate this kind of cancelation to PG, but also the other sources.

martin-core/src/tiles/postgres/pool.rs
Weixing-Zhang: Reverted the change. It makes sense to keep the test within the test-pg gate.

martin-core/src/tiles/postgres/source.rs
Weixing-Zhang: Good point. Added an auto-cleanup for the registry.

martin-core/src/tiles/postgres/pool.rs
Weixing-Zhang: Thank you for the reminder. I was gonna do it later. I adjusted the code accordingly.

martin-core/src/tiles/source.rs
Weixing-Zhang: Fixed this typo in three places.

martin/src/bin/martin-cp.rs
Weixing-Zhang: Thanks — that makes sense. I'll try to separate refactors from logic changes in future PRs. For this one I'll leave it as-is since the restructure was needed for interrupt handling and testability.

CommanderStorm (approved):
LGTM, feel free to merge once you make CI go green

Weixing-Zhang:
> LGTM, feel free to merge once you make CI go green @CommanderStorm 

Thank you for all the great suggestions. I will try to address them and fully test the changes this weekend.

Weixing-Zhang:
@CommanderStorm The unit test failures seems to be coming from the recent mvt integration. https://github.com/maplibre/martin/actions/runs/29269169288/job/86885028375?pr=2901#:~:text=Resulting%20diff%20(max,output%3A%20Broken%20pipe 

Should I update the unit test?

CommanderStorm:
Should be in main already.

PR #2902, opened 2026-06-21
refactor: duckdb bounds refactor and sql utils

manbhav234:
#2264 
This is first in a set of PR's to implement support for duckdb at `martin`. I'll close the previous PR ( #2894 ) as it was bad and convoluted (Sorry for this). 
This PR extracts out the bounds computing logic and a couple of SQL helpers alongside unit tests. All of this is gated behind the `unstable-duckdb` feature.
This PR will be followed by  a top to bottom config parsing for geoparquet files (as it is slightly simpler to implement  this and will also lay down some wiring for other sources we are adding support for). I started work on this and will do the next PR in a day or two.


martin-core/Cargo.toml
```diff
@@ -58,7 +58,7 @@ sprites = [
 styles = ["tokio/fs", "dep:dashmap", "dep:walkdir"]
 mbtiles = ["dep:backon", "dep:mbtiles", "dep:tokio", "_tiles"]
 pmtiles = ["dep:pmtiles", "dep:object_store", "_tiles"]
-duckdb = ["dep:duckdb", "_tiles", "dep:deadpool", "dep:url", "dep:tokio", "tokio/rt"]
+unstable-duckdb = ["dep:duckdb", "_tiles", "dep:deadpool", "dep:url", "dep:tokio", "tokio/rt"]
```
CommanderStorm: no need to rename it here, though I am fine with it too. 😉 

We don't have stability guarantees for martin-core, only martin.

martin/src/config/file/tiles/duckdb/resolver/bounds.rs
```diff
@@ -0,0 +1,134 @@
+use martin_core::tiles::duckdb::DuckDBError::PrepareQueryError;
+use martin_core::tiles::duckdb::{DuckDBPool, DuckDBResult};
+use tilejson::Bounds;
+use tokio::time::timeout;
+use tracing::warn;
+
+use super::super::sql_utils::{epsg_crs, escape_identifier};
```
CommanderStorm: Instead super imports, we generally prefer absolute imports. (exception: `use super::*` in tests)

The reason is that with absolute imports if you move something all occurances are apparent, so it is easier to review if something is not in the right place.

```suggestion
use crate::config::file::tiles::duckdb::sql_utils::{epsg_crs, escape_identifier};
```

martin/src/config/file/tiles/duckdb/resolver/bounds.rs
```diff
@@ -0,0 +1,134 @@
...
+                }
+                _ => None,
+            })
+        })
+        .map_err(|source| PrepareQueryError {
+            source: source.into(),
+            source_id,
+            signature: "bounds".to_string(),
+            query,
+        })
```
CommanderStorm: general: Using martin-core error types for this function is a bit weird.

I think a more ergonomic way would be to use a separate error enum which contains the duckdb one.

This way, the seam between the two parts is clearer

martin/src/config/file/tiles/duckdb/resolver/bounds.rs
```diff
@@ -0,0 +1,134 @@
...
+    srid: i32,
+    auto_bounds: BoundsCalcType,
+) -> DuckDBResult<Option<Bounds>> {
+    match auto_bounds {
+        BoundsCalcType::Skip => Ok(None),
+        BoundsCalcType::Calc => calc_exact_bounds(pool, relation, geom_col, srid).await,
+        BoundsCalcType::Quick => {
+            match timeout(
+                DEFAULT_BOUNDS_TIMEOUT,
+                calc_exact_bounds(pool, relation, geom_col, srid),
```
CommanderStorm: I think the different handling here should just be to use `st_extent_approx` instead of `ST_Extent`, or are there issues with that approach?
Does not sound that complicated, so we might just do the 1-2 line change here imo.

https://duckdb.org/docs/current/core_extensions/spatial/functions#st_extent_approx

martin/src/config/file/tiles/duckdb/resolver/bounds.rs
```diff
@@ -0,0 +1,134 @@
...
+            include_str!("../../../../../../../tests/fixtures/duckdb/bounds_point.sql"),
+        );
+        let pool = db.read_only_pool("bounds-test", 1);
+
+        let calc = calc_relation_bounds(&pool, "test_geom", "geom", 4326, BoundsCalcType::Calc)
+            .await
+            .expect("calculate bounds");
+        assert_eq!(calc, Some(Bounds::new(9.0, 19.0, 11.0, 21.0)));
+
+        let skip = calc_relation_bounds(&pool, "test_geom", "geom", 4326, BoundsCalcType::Skip)
```
CommanderStorm, line 218: lets also test the third case here.

For bonus points, we could change this to an `rstest`, but not nessary.

```suggestion
        let calc = calc_relation_bounds(&pool, "test_geom", "geom", 4326, BoundsCalcType::Calc)
            .await
            .expect("calculate bounds");
        assert_eq!(calc, Some(Bounds::new(9.0, 19.0, 11.0, 21.0)));

        let est = calc_relation_bounds(&pool, "test_geom", "geom", 4326, BoundsCalcType::Quick)
            .await
            .expect("approx bounds");
        assert_eq!(est, Some(Bounds::new(9.0, 19.0, 11.0, 21.0)));

        let skip = calc_relation_bounds(&pool, "test_geom", "geom", 4326, BoundsCalcType::Skip)
```

martin/src/config/file/tiles/duckdb/resolver/bounds.rs
```diff
@@ -0,0 +1,134 @@
...
+
+    use super::calc_relation_bounds;
+    use crate::config::args::BoundsCalcType;
+    use crate::test_support::duckdb::TestDatabase;
+
+    #[tokio::test]
+    async fn calc_and_skip_relation_bounds() {
+        let db = TestDatabase::from_sql(
+            "bounds.duckdb",
+            include_str!("../../../../../../../tests/fixtures/duckdb/bounds_point.sql"),
```
CommanderStorm, line 204: I am fine with this.

It would be better to use a line geometry or something which is "not just one point".

That the extent of a point at `10.0, 20.0` is `9.0, 19.0, 11.0, 21.0` is not immediately clear to me.

martin/src/config/file/tiles/duckdb/sql_utils.rs
```diff
@@ -0,0 +1,41 @@
...
+        assert_eq!(escape_sql_string("simple"), "'simple'");
+        assert_eq!(escape_sql_string("O'Brien"), "'O''Brien'");
+    }
+
+    #[test]
+    fn formats_epsg_crs() {
+        assert_eq!(epsg_crs(4326), "'EPSG:4326'");
+        assert_eq!(epsg_crs(3857), "'EPSG:3857'");
+    }
+}
```
CommanderStorm, line 56: nit: I would like to see this written this way.
So no unnessary cfg and using rstest for a bunch more tests.

Is not actually adding any coverage, but is covering our but on an imporant axis, so future refactorings are saver.

Disclaimer: used an LLM to come up with all of these inputs, might not pass the test

```suggestion
#[cfg(test)]
mod tests {
    use super::*;
    use rstest::rstest;

    #[rstest]
    #[case::simple("roads", "\"roads\"")]
    #[case::embedded_quote("my\"table", "\"my\"\"table\"")]
    #[case::empty("", "\"\"")]
    #[case::quote_only("\"", "\"\"\"\"")]
    #[case::multiple_quotes("a\"\"b", "\"a\"\"\"\"b\"")]
    #[case::dot_in_identifier("schema.table", "\"schema.table\"")]
    #[case::unicode("Straße", "\"Straße\"")]
    fn escape_identifier_cases(
        #[case] input: &str,
        #[case] expected: &str,
    ) {
        assert_eq!(escape_identifier(input), expected);
    }

    #[rstest]
    #[case::simple("simple", "'simple'")]
    #[case::embedded_apostrophe("O'Brien", "'O''Brien'")]
    #[case::empty("", "''")]
    #[case::apostrophe_only("'", "''''")]
    #[case::multiple_apostrophes("a''b", "'a''''b'")]
    #[case::double_quotes_preserved("\"quoted\"", "'\"quoted\"'")]
    #[case::unicode("Straße", "'Straße'")]
    fn escape_sql_string_cases(
        #[case] input: &str,
        #[case] expected: &str,
    ) {
        assert_eq!(escape_sql_string(input), expected);
    }

    #[rstest]
    #[case::wgs84(4326, "'EPSG:4326'")]
    #[case::web_mercator(3857, "'EPSG:3857'")]
    #[case::zero(0, "'EPSG:0'")]
    #[case::negative(-1, "'EPSG:-1'")]
    fn epsg_crs_cases(
        #[case] srid: i32,
        #[case] expected: &str,
    ) {
        assert_eq!(epsg_crs(srid), expected);
    }
}
```

martin/src/test_support.rs
```diff
@@ -91,3 +91,45 @@ pub(crate) mod pg {
...
+        pub(crate) fn from_sql(filename: &str, sql: &str) -> Self {
+            let dir = TempDir::new().expect("temporary DuckDB directory");
+            let path = dir.path().join(filename);
+            let conn = Connection::open(&path).expect("writable DuckDB database");
+            conn.execute_batch("INSTALL spatial;")
+                .expect("spatial extension installed");
+            conn.execute_batch("LOAD spatial;")
+                .expect("spatial extension loaded");
+            conn.execute_batch(sql).expect("fixture SQL executed");
+            drop(conn);
```
CommanderStorm: Don't think this is nessary, or?

```suggestion
```

CommanderStorm (approved):
LGTM.

I suggested a few changes, feel free to accept or reject at your option.
You have merge rights, so merge once ready 😉

manbhav234:
there seems to be some errors in unrelated part of code that are causing `just check` to fail or is it something else ? @CommanderStorm

CommanderStorm:
They are not unrelated to the code. If configured a certain way, the code currently "looks buggy" -> gets a warning.
We treat warnings as errors in ci.

That is fine, there either already is an cfg_attr around there where we can expect these warnings.
Alternatively I can have a look in the evening.

Basically the same thing as on your last PR ^^

PR #2908, opened 2026-06-22
feat(martin-core): passthrough tile source

CommanderStorm:


martin-core/src/tiles/passthrough/source.rs
```diff
@@ -0,0 +1,325 @@
...
+
+/// Resolved configuration for a single passthrough source.
+///
+/// The `martin` config layer is responsible for env-var substitution in `urls`/`headers` before
+/// constructing this; `martin-core` consumes the already-resolved values.
+#[derive(Clone, Debug)]
+pub struct PassthroughConfig {
+    /// Upstream URL(s): a single `TileJSON` document URL, a single `{z}/{x}/{y}` template, or a
+    /// list of templates.
+    pub urls: Vec<String>,
```
nyurik: we need this config in the documentation too. Is it now automated?

nyurik (approved):
looks good with a minor question inline.

I wonder how we can make sources "private" - in some cases like contours,  hillshading, possibly post-processing, we may need to define an arbitrary source (could be from a DB/file or a pass-through), but not expose it publically. Instead, another tile source like the one that computes hillshading could use that private source to get the needed data

martin-core/src/tiles/passthrough/source.rs
CommanderStorm: no, this is only martin-core.

And yes, I could model this better. Should have been an enum.

CommanderStorm:
> I wonder how we can make sources "private" - in some cases like contours, hillshading, possibly post-processing, we may need to define an arbitrary source (could be from a DB/file or a pass-through), but not expose it publically. Instead, another tile source like the one that computes hillshading could use that private source to get the needed data

Not nessary.
Just do it like MLT <-> MVT as a processing plugin.

PR #2912, opened 2026-06-23
fix(martin): guard served ETag against invalid entity-tag characters

CommanderStorm:
Adds `Tile::strong_etag` which reuses a valid stored etag and otherwise falls back to a content hash, so `EntityTag::new_strong` can no longer panic on disallowed characters.


CommanderStorm (commented):
@nyurik proxy reviewed this:
- needs to check if this code is dead by invariant, so if the invalid etag is validated by reqwest/...
- if not dead: OK

PR #2913, opened 2026-06-23
feat(martin-core): add Transport::from_string_headers

CommanderStorm:
Adds a validating `Transport` constructor that builds the header map from string pairs, letting callers supply headers without depending on `reqwest`'s header types.


CommanderStorm:
+1 by @nyurik

PR #2916, opened 2026-06-23
fix: gate idr lint expectations on absence of consuming backends

CommanderStorm:
Fixes the unfulfilled `unused_variables`/`unused_mut` lint expectations that broke CI when `unstable-duckdb` is built alongside another tile backend.

### DIFF
--- a/martin/src/config/file/main/lifecycle.rs
+++ b/martin/src/config/file/main/lifecycle.rs
@@ -353,7 +353,12 @@ impl Config {
     #[cfg(feature = "_tiles")]
     #[instrument(skip_all, err(Debug))]
     #[cfg_attr(
-        feature = "unstable-duckdb",
+        not(any(
+            feature = "postgres",
+            feature = "pmtiles",
+            feature = "mbtiles",
+            feature = "unstable-cog"
+        )),
         expect(
             unused_variables,
             reason = "idr is only consumed by the non-duckdb tile backends"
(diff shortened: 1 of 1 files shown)


martin/src/config/file/main/lifecycle.rs
```diff
@@ -353,7 +353,12 @@ impl Config {
     #[cfg(feature = "_tiles")]
     #[instrument(skip_all, err(Debug))]
     #[cfg_attr(
-        feature = "unstable-duckdb",
+        not(any(
```
nyurik, line 356: i think this overcomplicates the code - just use `allow` with the `reason` explanation

martin/src/config/file/main/lifecycle.rs
```diff
@@ -365,7 +370,12 @@ impl Config {
         #[cfg(feature = "pmtiles")] pmtiles_cache: PmtCache,
     ) -> MartinResult<(Vec<Vec<BoxedSource>>, Vec<TileSourceWarning>)> {
         #[cfg_attr(
-            feature = "unstable-duckdb",
+            not(any(
```
nyurik, line 373: same

nyurik (approved):
i disabled automerge due to feedback, but acceepting

martin/src/config/file/main/lifecycle.rs
CommanderStorm: cannot, is disabled in our clippy lints..

CommanderStorm: actually, we just wanted to enable this, so not currently..

CommanderStorm: but still, this expect is temporary and SHOULD fail once we do duckdb here too, so there being an expect is good.

nyurik: can we re-enable it here?  e.g. `#[allow(clippy::... reason="too complex of a condition"] #[allow(unused_variables, ...)]` ?    But whatever

CommanderStorm: Yes, but enabling this here would mean that the `expect` (which will cause an error -> cause this to be removed once duckdb is in) would not fire.

PR #2921, opened 2026-06-24
refactor(martin-core): derive font extensions from FontFormat via strum

CommanderStorm:
Replace the `FONT_EXTENSIONS` constant with `strum::VariantNames` on `FontFormat` so the recognised font extensions and the enum stay in sync.

Claude picked this out of the larger overlay PR


CommanderStorm:
+1 from @nyurik

PR #2922, opened 2026-06-24
feat(martin-core): add static map overlay rendering engine

CommanderStorm:
martin-core part of #2794


martin-core/src/resources/styles/render_pool.rs
```diff
@@ -337,6 +352,46 @@ fn load_style_cached<S>(
     Ok(())
 }
 
+/// Applies an overlay to the renderer's style and removes it again on drop --
+/// even on an early return or panic -- so the cached style returns to a clean
+/// base for the next request.
+struct RenderOverlay<'r> {
```
nyurik: rename to `RendererWithOverlay`

martin-core/src/resources/styles/render_pool.rs
```diff
@@ -383,9 +438,28 @@ impl StaticRenderer {
             .zoom(params.zoom)
             .bearing(params.bearing)
             .pitch(params.pitch);
-        self.renderer
-            .render_static(&camera)
-            .map_err(StyleError::RenderingError)
+
+        let render_once = |renderer: &mut ImageRenderer<Static>| {
```
nyurik: rename `renderer` to `v` perhaps to shrink the size to a oneliner?  If not, don't woryr

martin-core/src/overlay/model.rs
```diff
@@ -0,0 +1,142 @@
+//! Typed boundary IR for static-render overlays.
```
nyurik: IR?

nyurik (approved):
i'm a bit unhappy with the overall approach - i.e. i would prefer to have a more streamlined:

```rust
// on render request
let r = get_renderer_from_pool();
r.init_base_map(style);
r.add_extras(overlay);
let image = r.renderer(size);
release_renderer(r);
return r;
```

PR #2924, opened 2026-06-24
feat(tile-utils): make tile_bbox public

CommanderStorm:
Expose `tile_bbox` so `martin-core` can compute a tile's Web Mercator bounds

Split out of #2538


CommanderStorm:
+1 from @nyurik

PR #2925, opened 2026-06-24
feat(geojson): add GeoJSON tile source to martin-core

CommanderStorm:
Add the `martin-core` GeoJSON-to-MVT tile source (clip, simplify, validate, encode) behind a `geojson` feature

Split out of #2538 and stacked on #2924


martin-core/README.md
```diff
@@ -14,6 +14,7 @@ The library providing the basic building blocks for [Martin](https://maplibre.or
   - [x] mbtiles
   - [x] pmtiles
   - [x] cog
+  - [x] geojson (requires C++ build tools or `libgeos-dev` being installed)
```
nyurik, line 17: fix

nyurik (approved):
i think we should migrate from i_overlay to geo - this way all underlying types going from geojson through slicing into mvt encoding use the same underlying geotypes.

One important thing that you may still need to fix:  `slice + fix invalid geometries` must happen AFTER the polygon is converted to integers - otherwise the polygons will be incorrect in the end.

PR #2931, opened 2026-06-27
feat: geoparquet config and resolver wiring 

manbhav234:
#2264 
This PR implements config parsing and `resolver` wiring for geoparquet sources added via DuckDB.
This is plumbing with top level wiring into `Config/lifecycle` planned for the next PR as it would be easier to review it that way.


martin/src/config/file/tiles/duckdb/resolver/geoparquet/introspect.rs
```diff
@@ -0,0 +1,223 @@
...
+        .await?
+        .map_err(|source| {
+            GeoparquetError::introspection_query(source, source_label, "srid", query_for_error)
+        })?;
+
+    match crs {
+        None => Err(GeoparquetError::SridUnknown { geometry_column }),
+        Some(None) => Err(GeoparquetError::SridUnknown { geometry_column }),
+        Some(Some(crs)) => parse_crs_to_srid(&crs, &geometry_column),
+    }
```
CommanderStorm: should already raise a clippy lint, but I think a more-rusty way of doing this would be

```suggestion
    match crs.flatten() {
        None => Err(GeoparquetError::SridUnknown { geometry_column }),
        Some(crs) => parse_crs_to_srid(&crs, &geometry_column),
    }
```

martin/src/config/file/tiles/duckdb/resolver/bounds.rs
```diff
@@ -188,6 +203,29 @@ pub async fn calc_relation_bounds(
...
+pub async fn calc_from_expr_bounds(
+    pool: &DuckDBPool,
+    from_expr: &str,
+    label: &str,
+    geom_col: &str,
+    srid: i32,
+    auto_bounds: BoundsCalcType,
+) -> BoundsResult<Option<Bounds>> {
+    bounds_with_auto(pool, from_expr, label, geom_col, srid, auto_bounds).await
+}
```
CommanderStorm: these two seem like the code would be simpler to read if inlined.

martin/src/config/file/tiles/duckdb/resolver/geoparquet/introspect.rs
```diff
@@ -0,0 +1,223 @@
...
+        .and_then(|srid| {
+            if srid > 0 {
+                Ok(srid)
+            } else {
+                Err(GeoparquetError::SridUnknown {
+                    geometry_column: geometry_column.to_string(),
+                })
+            }
+        })
+}
```
CommanderStorm: nit: 
I think it would be benefitial to parse to `NonZeroI32` to encode this invariant and to model the error space a bit more explicitly, at least the errors CAN be better here 😉

martin/src/config/file/tiles/duckdb/resolver/geoparquet/introspect.rs
```diff
@@ -0,0 +1,223 @@
...
+mod tests {
+    use super::*;
+
+    #[test]
+    fn parse_crs_to_srid_accepts_epsg_and_crs84() {
+        for (crs, expected) in [
+            ("EPSG:4326", 4326),
+            ("epsg:3857", 3857),
+            ("OGC:CRS84", 4326),
+        ] {
```
CommanderStorm: please use `rstest` for this

martin/src/config/file/tiles/duckdb/resolver/geoparquet/tests.rs
CommanderStorm, line 1: instead of these being unit tests, can we move them to be e2e tests instead? so in tests.sh?
This way, we test against the actual, public interface instead of this semi-internal one.

This way we also prove this works there.
I am not sure if duckdb support in our e2e tests currently works or if this would require some adjustments to how we feature flag here

martin/src/config/file/tiles/duckdb/config.rs
```diff
@@ -0,0 +1,298 @@
...
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        match self {
+            Self::Database(v) => v.serialize(serializer),
+            Self::GeoParquet(v) => v.serialize(serializer),
+        }
+    }
+}
```
CommanderStorm: I think this is what `derive(Serialize)` would produce.
Please do that instead, it is less code

martin/src/config/file/tiles/duckdb/config.rs
```diff
@@ -0,0 +1,298 @@
...
+                .map_err(|e| de::Error::custom(format!("invalid geoparquet source entry: {e}"))),
+            (false, false) => Err(de::Error::custom(
+                "duckdb source entry must contain exactly one of `database` or `geoparquet`",
+            )),
+            (true, true) => Err(de::Error::custom(
+                "duckdb source entry cannot contain both `database` and `geoparquet`",
+            )),
+        }
+    }
+}
```
CommanderStorm: this is a bit weird how you did this.
Why not derive?

martin/src/config/file/tiles/duckdb/resolver/geoparquet/metadata.rs
```diff
@@ -0,0 +1,38 @@
...
+        other: BTreeMap::default(),
+    };
+
+    tilejson! {
+        tiles: vec![],
+        vector_layers: vec![layer],
+        name: source_id.to_string(),
+        description: format!("GeoParquet ({source_label})"),
+        minzoom: entry.minzoom.unwrap_or(0),
+        maxzoom: entry.maxzoom.unwrap_or(20),
```
CommanderStorm: why this default? above this is not set..

martin/src/config/file/tiles/duckdb/resolver/error.rs
```diff
@@ -22,5 +25,78 @@ pub enum BoundsError {
...
+    #[error("GeoParquet column '{column}' is not a geometry column (type {column_type})")]
+    NotGeometryColumn { column: String, column_type: String },
+
+    /// The configured id column does not exist.
+    #[error("GeoParquet id_column '{column}' was not found")]
+    IdColumnNotFound { column: String },
+
+    /// A local GeoParquet path cannot be represented as UTF-8 for SQL usage.
+    #[error("GeoParquet path is not valid UTF-8: {path:?}")]
+    NonUtf8Path { path: PathBuf },
```
CommanderStorm: nit: If by the type passed into it or via context it cannot be something else, lets not do structs because this way the downstream code is a bit less "dense" (=> less complex to read for the next person) ^^

This feedback also applies to a few of the other errors like `IdColumnNotFound`.



```suggestion
    #[error("GeoParquet path is not valid UTF-8: {0:?}")]
    NonUtf8Path(PathBuf),
```

martin/src/config/file/tiles/duckdb/sql_utils.rs
```diff
@@ -16,6 +16,12 @@ pub fn epsg_crs(srid: i32) -> String {
     escape_sql_string(&format!("EPSG:{srid}"))
 }
 
+/// Build a `read_parquet(...)` table-source expression for inline SQL `FROM` clauses.
+#[must_use]
+pub fn read_parquet_from_expr(path_or_url: &str) -> String {
+    format!("read_parquet({})", escape_sql_string(path_or_url))
+}
+
```
CommanderStorm: Can we inline this? I think it makes the code harder to understand ^^

martin/src/config/file/tiles/duckdb/resolver/geoparquet/sql.rs
```diff
@@ -0,0 +1,162 @@
...
+        {extent}::BIGINT, {buffer}::BIGINT, {clip_geom}
+    ) AS geom
+    {id_field}{properties}
+  FROM {from_expr}, tile, bounds
+  WHERE {tile_filter}
+) AS tile;
+"
+    )
+    .trim()
+    .to_string()
```
CommanderStorm: just trim the string ourselves..
It is not like the start/end are user-controlled..

```suggestion
```

martin/src/config/file/tiles/duckdb/resolver/geoparquet/sql.rs
```diff
@@ -0,0 +1,162 @@
...
+        assert!(sql.contains("ST_AsMVT(tile, 'buildings', 4096, 'geom'"));
+        assert!(sql.contains("4096::BIGINT, 64::BIGINT, true"));
+        assert!(sql.contains(
+            "ST_Transform(ST_SetCRS(\"geom\"::GEOMETRY, 'EPSG:4326'), 'EPSG:4326', 'EPSG:3857', always_xy := true)"
+        ));
+        assert!(sql.contains(", \"name\""));
+        assert!(sql.contains(", \"category\""));
+        assert!(sql.contains(from_expr));
+        assert!(sql.contains("ST_Intersects("));
+        assert!(sql.contains("ST_Expand(bounds.envelope,"));
```
CommanderStorm: very brittle test here and below.

Please just use `insta::assert_snapshot`..

martin/src/config/file/tiles/duckdb/resolver/geoparquet/resolve.rs
```diff
@@ -0,0 +1,55 @@
...
+use crate::config::file::tiles::duckdb::resolver::bounds::calc_from_expr_bounds;
+use crate::config::file::tiles::duckdb::resolver::error::GeoparquetResult;
+use crate::config::file::tiles::duckdb::sources::GeoParquetEntry;
+
+use super::introspect::{geoparquet_from_expr, introspect};
+use super::metadata::build_tilejson;
+use super::sql::build_mvt_sql;
+
+/// Introspects geometry metadata, resolves SRID, and builds a tile-ready `DuckDBSource`.
+pub async fn resolve_geoparquet_source(
```
CommanderStorm, line 17: I don't see any usage for this..
Is this dead code by accident? Am I missing something?

martin/src/config/file/tiles/duckdb/config.rs
```diff
@@ -0,0 +1,298 @@
...
+        };
+
+        assert_eq!(db.database, std::path::PathBuf::from("/data/tiles.duckdb"));
+        assert_eq!(
+            gpq.geoparquet,
+            std::path::PathBuf::from("/data/buildings.parquet")
+        );
+        assert_eq!(gpq.layer_id.as_deref(), Some("buildings"));
+        assert_eq!(gpq.geometry_column.as_deref(), Some("geom"));
+        assert_eq!(gpq.srid, Some(4326));
```
CommanderStorm: Instead of all of this work, just `insta::debug_snapshot` the entire `cfg`.
Has the same effect, if not better. 😉

CommanderStorm (commented):
Mostly LGTM.

There are a few pieces that I am not sure if they are plumbed up fully, but if you implement the e2e tests you will notice this 😉

martin/src/config/file/tiles/duckdb/resolver/geoparquet/tests.rs
manbhav234: I will stack up a PR on this one that completes the implementation in `Config` and `lifecycle` and add e2e test support there to prove this works.

manbhav234:
A side question : when i run just commands locally, it takes up around 56GB of my disk space due to building of duckdb C++ lib. I saw similar issue was handled in CI but any solution for this during local dev ?

CommanderStorm:
Not sure how to solve that..

martin/src/config/file/tiles/duckdb/config.rs
manbhav234: This was for raising errors if both database and geoparquet paths are mentioned under the same source like
```
sources:
- database: path/to/db
  geoparquet: path/to/geoparquet
```
for both serialize and deserialize we can do derive with `#[serde(untagged)]` (to match the first variant recognized) but that would lead to the geoparquet key being an unrecognized key if something like above is received instead of an error. but yeah cleans up a lot of weird code if that's acceptable

martin/src/config/file/tiles/duckdb/resolver/geoparquet/resolve.rs
manbhav234: For now yes, should be implemented in the PR that completes the plumbing.

martin/src/config/file/tiles/duckdb/config.rs
CommanderStorm: Yes, it is acceptable I think.. We even warn, so no biggie 😉

CommanderStorm (approved):
LGTM. CI is a bit unhappy, I have not looked ino it.
Some may be flakes, but likely some true positives too

manbhav234:
> LGTM. CI is a bit unhappy, I have not looked ino it. Some may be flakes, but likely some true positives too

Yeah there were some issues from my side. i pushed some changes. will get this into mergable state

manbhav234:
Is the windows test flaky ?

PR #2944, opened 2026-07-02
fix(postgres): redact password in BadConnectionString error

### LINKED ISSUE #2895: Martin potentially leaks sensitive passowrd information when printing BadConnectionString/PostgresError

CommanderStorm:
Wraps the malformed connection string in a `RedactedConnectionString` that strips the password (URL userinfo, `?password=`, and libpq `password=` keyword forms) before it reaches logs, fixing #2895.


nyurik (approved):
lgtm, but please add a test with the password containing special chars - i.e. use url-encoded `&` as `%..` values

PR #2968, opened 2026-07-06
chore(deps): Update hotpath

pawurb:
Hi, this PR updates hotpath version. The primary difference is new wrap api used by `hotpath::channel!`. Instead of inserting proxy channels to observe messages, macro now returns new instrumented types.  You'll now get precise delay values, perf histogram, and massively smaller profiling overhead:

```
tokio channel: 100000 send/recv cycles per mode
  baseline (raw)    1071.2 ns/op
  wrap (default)    1124.9 ns/op  (+53.7 ns/op vs baseline)
  proxy = true      6987.4 ns/op  (+5916.2 ns/op vs baseline)
```

I did not document it yet but there's also a new `HOTPATH_TIME_SAMPLING_RATE=0.5` config, you can use to remove two `Instant` calls per measurement, and further reduce profiling overhead.

As always feedback welcome!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1374,10 +1374,8 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1aa79e62e7697b8e29b513a68abacf485adcd1fe8284a4316c5ae868e6633327"
 dependencies = [
  "iana-time-zone",
- "js-sys",
  "num-traits",
  "serde",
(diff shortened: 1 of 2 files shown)


CommanderStorm (approved):
thank you!

(the other dependency bump that renovate pushed off will also include this bump, lets see which one CI automerges first)

PR #2979, opened 2026-07-08
fix(postgres): clearer error when a source schema exists but has no tile sources

### LINKED ISSUE #969: More helpful error when function is missing
Thanks for developing martin, it is a pleasure to use.

I recently spun up martin, asking it to serve a function that did not exist. An error reported that the schema `public` was not found, with an empty list of possible values:

```
[2023-10-25T05:27:45Z WARN  martin::pg::utils] Unable to configure source foot because schema 'public' was not found.  Possible values are:
```
I suggest that a more descriptive/helpful error be returned, indicating that the function itself may be missing.

CommanderStorm:
Closes #969


martin/src/config/file/tiles/postgres/resolver/scripts/query_schemas.sql
```diff
@@ -0,0 +1,8 @@
+-- Every user-visible schema in the database, used to tell "schema does not exist"
+-- apart from "schema exists but has no tile-serving functions/tables".
+SELECT nspname AS schema
+FROM pg_namespace
+WHERE
+    nspname NOT LIKE 'pg\_%'
```
nyurik: are you sure you want to block all schemas that begin with `pg_` ?  I would rather block well known ones like `pg_catalog`, and add more as needed.

martin/src/config/file/tiles/postgres/builder.rs
```diff
@@ -383,21 +392,17 @@ impl PostgresAutoDiscoveryBuilder {
...
         id: &str,
         function_info_from_config: &FunctionInfo,
     ) -> Result<(FunctionInfo, PostgresSqlInfo), String> {
-        let function_infos_for_schema = find_info(
+        let function_infos_for_schema = find_schema_info(
             function_infos_from_db,
+            all_schemas,
             &function_info_from_config.schema,
-            "schema",
+            "tile-serving functions (e.g. (z,x,y) -> bytea)",
```
nyurik:
```suggestion
            "tile-serving functions (z,x,y) -> bytea",
```

nyurik (approved):
See a few inline comments.

In the future, we may want to introduce/use case-insensitive strings to represent schemas and possibly other PG IDs.  This way, instead of passing strings/&str, we would pass a magical `PgName` that defines its own `PartialEq, Eq, Hash, PartialOrd, Ord` (or a subset of these) - so that it can be stored in the lookups. Moreover, I think user config should be parsed into these types instead of strings - thus we would be able to validate user input better. I wonder if PG crate already defines this type.

martin/src/config/file/tiles/postgres/resolver/scripts/query_schemas.sql
```diff
@@ -0,0 +1,8 @@
+-- Every user-visible schema in the database, used to tell "schema does not exist"
+-- apart from "schema exists but has no tile-serving functions/tables".
+SELECT nspname AS schema
```
CommanderStorm: otherwise this triggers the lint

```suggestion
SELECT nspname
```

martin/src/config/file/tiles/postgres/resolver/query_schemas.rs
```diff
@@ -0,0 +1,22 @@
...
+/// Used to distinguish a truly missing schema from one that merely has no tile-serving functions or tables.
+pub async fn query_schemas(pool: &PostgresPool) -> PostgresResult<BTreeSet<String>> {
+    let schemas = pool
+        .get()
+        .await?
+        .query(include_str!("scripts/query_schemas.sql"), &[])
+        .await
+        .map_err(|e| PostgresError(e, "querying available schemas"))?
+        .into_iter()
+        .map(|row| row.get::<_, String>("schema"))
```
CommanderStorm:
```suggestion
        .map(|row| row.get::<_, String>("nspname"))
```

martin/src/config/file/tiles/postgres/resolver/scripts/query_schemas.sql
CommanderStorm: Fair.

> 5.10.5. The System Catalog Schema 
In addition to public and user-created schemas, each database contains a pg_catalog schema, which contains the system tables and all the built-in data types, functions, and operators. pg_catalog is always effectively part of the search path. If it is not named explicitly in the path then it is implicitly searched before searching the path's schemas. This ensures that built-in names will always be findable. However, you can explicitly place pg_catalog at the end of your search path if you prefer to have user-defined names override built-in names.
>
> Since system table names begin with pg_, it is best to avoid such names to ensure that you won't suffer a conflict if some future version defines a system table named the same as your table. (With the default search path, an unqualified reference to your table name would then be resolved as the system table instead.) System tables will continue to follow the convention of having names beginning with pg_, so that they will not conflict with unqualified user-table names so long as users avoid the pg_ prefix.



```suggestion
    nspname <> 'pg_catalog'
```

PR #2982, opened 2026-07-08
chore(mbtiles): refactor schemas to individual files

nyurik:


mbtiles/src/schemas/flat.rs
CommanderStorm, line 1: at the point where multiple file names have the same prefix like here, I think we might be better off with a module... what do you think?

nyurik: Yep, good idea, I'll fix it when back,  or feel free

CommanderStorm (approved):

PR #2983, opened 2026-07-08
fix(mbtiles): validate min/max zoom metadata against actual tiles

### LINKED ISSUE #1791: mbtiles does not validate `{min,max}zoom` metadata fileds
In https://github.com/maplibre/martin/pull/1787#discussion_r2043007061 we noticed that zoom fields are simply served from the metadata.

This is a bit too simplistic and we should use:
```python
tilejson.vector_layers[].minzoom = min(min(zoom), tilejson.vector_layers[].maxzoom, tilejson.vector_layers[].minzoom)
tilejson.vector_layers[].maxzoom = min(max(zoom), max(min(zoom), tilejson.vector_layers[].maxzoom, tilejson.vector_layers[].minzoom))
```

Generating mbtiles files might help to test this properly

AnayGarodia:
## What

Fixes #1791. mbtiles served `minzoom`/`maxzoom` (and per-`vector_layer` zoom) straight from the metadata table without checking the tiles it actually holds, so a file could advertise a wider zoom range than it contains.

## How

`get_metadata` now reconciles the served zoom against the real tile extent (via the existing `compute_min_max_zoom`) using the formula proposed in the issue:
- `minzoom` expands *down* to include any lower tiles present,
- `maxzoom` is capped at the highest zoom that actually has tiles,
- absent values are populated from the tile extent,
- reconciliation is skipped for empty tilesets.

Applied to both the top-level metadata and per-`vector_layer` min/maxzoom. I used the issue's reconciliation formula rather than aggressive clamping, which preserves existing valid files (the `zoomed_world_cities` fixture stays correct).

## Testing

5 new zoom tests using in-memory SQLite fixtures (as nyurik suggested), plus the existing suite:
```
cargo test -p mbtiles --lib       # 51 passed
cargo test -p mbtiles --doc       # 21 passed
cargo clippy -p mbtiles --all-targets -- -D warnings   # clean
```
Verified all integration snapshot fixtures already match their real tile extent, so no snapshot churn.

### DIFF
--- a/mbtiles/src/metadata.rs
+++ b/mbtiles/src/metadata.rs
@@ -10,8 +10,8 @@ use tilejson::{Bounds, Center, TileJSON, tilejson};
 use tracing::{info, warn};
 
 use crate::MbtError::InvalidZoomValue;
-use crate::Mbtiles;
 use crate::errors::MbtResult;
+use crate::{Mbtiles, compute_min_max_zoom};
 
 /// Tileset metadata combining [MBTiles](https://github.com/mapbox/mbtiles-spec)
 /// and [TileJSON](https://github.com/mapbox/tilejson-spec) specifications.
@@ -252,6 +252,16 @@ impl Mbtiles {
         // Need to drop rows in order to re-borrow connection reference as mutable
         drop(rows);
 
+        // The metadata `minzoom`/`maxzoom` (and per-`vector_layer` zoom ranges)
+        // are not guaranteed to match the tiles actually stored in the file.
+        // For example, a file may declare `maxzoom=14` in its metadata while only
+        // containing tiles up to zoom 6. Reconcile the served zoom levels against
+        // the real zoom extent of the `tiles` table so we never advertise zoom
+        // levels that have no tiles. See <https://github.com/maplibre/martin/issues/1791>.
+        if let Some((min_zoom, max_zoom)) = compute_min_max_zoom(&mut *conn).await? {
+            reconcile_zoom_levels(&mut tj, min_zoom, max_zoom);
+        }
+
         Ok(Metadata {
             id: self.filename().to_string(),
             tilejson: tj,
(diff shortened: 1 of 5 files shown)


AnayGarodia:
CI caught that the zoom reconciliation also flows through `martin-cp` copy output — I've blessed the four affected `tests/expected/martin-cp/*_metadata.txt` files. The changes are exactly the reconciliation:

- `composite`/`flat`/`no-bbox`: previously-absent per-vector-layer `minzoom: 0` / `maxzoom: 6` are now populated from the tile extent.
- `normalized`: top-level `maxzoom` capped `6 → 1`, since that fixture only holds tiles up to z1.

One thing worth your confirmation: this caps a declared `maxzoom` **down** to the highest zoom that actually has tiles (per the formula in the issue). If a tileset intentionally declares a higher `maxzoom` for client-side overzoom, that would now be narrowed — happy to restrict reconciliation to only *expanding* the range, or to the serving path only, if you'd prefer.

mbtiles/src/metadata.rs
```diff
@@ -252,6 +252,16 @@ impl Mbtiles {
         // Need to drop rows in order to re-borrow connection reference as mutable
         drop(rows);
 
+        // The metadata `minzoom`/`maxzoom` (and per-`vector_layer` zoom ranges)
+        // are not guaranteed to match the tiles actually stored in the file.
+        // For example, a file may declare `maxzoom=14` in its metadata while only
+        // containing tiles up to zoom 6. Reconcile the served zoom levels against
+        // the real zoom extent of the `tiles` table so we never advertise zoom
+        // levels that have no tiles. See <https://github.com/maplibre/martin/issues/1791>.
```
CommanderStorm: can you do a pass to de-vibe your comments one round? aka give it a language pass?
Currently very hard to read due to this across the entire changeset

CommanderStorm (commented):
I am a bit tired, I want to look over if I missed something tomorrow, but otherwise LGTM.

Please clean up the comments though, they are rough given how "vibed" they read..

AnayGarodia:
Done - gave the comments a language pass, cut them down to terse one/two-liners throughout `metadata.rs`.

CommanderStorm (approved):
thanks

PR #2984, opened 2026-07-08
feat(docs): page through the guide with left/right arrow keys

### LINKED ISSUE #2788: make left/right arrow keys page through the guide
* [ ] Can we make left/right arrow keys page through the guide?  Seems like a common pattern for left/right sites.
* [x] The guide's first page should have more of the content from the README to describe what Martin can do

AnayGarodia:
## What

Fixes #2788. Left/Right arrow keys now page through the guide via the footer prev/next links — a common paged-docs convenience.

## How

A single document-level `keydown` listener maps `ArrowLeft`/`ArrowRight` to the existing footer `a.md-footer__link--prev`/`--next` anchors, wired through `extra_javascript`. It survives `navigation.instant`, ignores modifier-key combos, and does nothing when a form field is focused.

## Testing

- `uvx zensical build` → "No issues found"; the script is emitted and referenced in the built HTML.
- Live browser check: from `/installation/`, ArrowRight → `/run/`, ArrowLeft → back to `/installation/`; with an input focused, ArrowRight did not navigate (guard works).

### DIFF
--- a/docs/content/javascripts/keyboard-nav.js
+++ b/docs/content/javascripts/keyboard-nav.js
@@ -5,9 +5,7 @@
 // common "prev/next" behavior of left/right paged sites.
 //
 // See https://github.com/maplibre/martin/issues/2788
-(function () {
-  "use strict";
-
+(() => {
   // Don't hijack the arrow keys while the user is typing (e.g. in the search
   // box) or interacting with a form control.
   function isTypingTarget(el) {
@@ -18,13 +16,13 @@
       return true;
     }
     var tag = el.tagName;
-    return tag === "INPUT" || tag === "TEXTAREA" || tag === "SELECT";
+    return tag === 'INPUT' || tag === 'TEXTAREA' || tag === 'SELECT';
   }
 
   // A single document-level listener keeps working across "instant" navigation,
   // where the page body is swapped without a full reload. The footer links are
   // looked up on each key press so they always point at the current page.
-  document.addEventListener("keydown", function (event) {
+  document.addEventListener('keydown', (event) => {
     // Leave modifier combinations alone (e.g. browser back/forward shortcuts).
     if (event.altKey || event.ctrlKey || event.metaKey || event.shiftKey) {
       return;
@@ -34,16 +32,16 @@
     }
 
     var selector;
-    if (event.key === "ArrowLeft") {
-      selector = "a.md-footer__link--prev[href]";
-    } else if (event.key === "ArrowRight") {
-      selector = "a.md-footer__link--next[href]";
+    if (event.key === 'ArrowLeft') {
+      selector = 'a.md-footer__link--prev[href]';
+    } else if (event.key === 'ArrowRight') {
+      selector = 'a.md-footer__link--next[href]';
     } else {
       return;
     }
 
     var link = document.querySelector(selector);
-    if (link && link.getAttribute("href")) {
+    if (link && link.getAttribute('href')) {
       event.preventDefault();
       // Use click() so "instant" navigation (if enabled) can intercept it.
       link.click();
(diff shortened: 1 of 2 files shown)


zensical.toml
```diff
@@ -127,7 +127,8 @@ nav = [
 # The path provided should be relative to the "docs_dir".
 #
 # Read more: https://zensical.org/docs/customization/#additional-javascript
-#extra_javascript = ["javascripts/extra.js"]
+# Enables Left/Right arrow keys to page through the guide (issue #2788).
```
CommanderStorm:
```suggestion
```

CommanderStorm (changes requested):
I tesed it and it does not page through the docs. Am I using it wrong?

AnayGarodia:
weird, it pages fine on my end. built the docs fresh (`uvx zensical build`, and tried `serve` too) and ←/→ walk through prev/next on the guide pages. poked at the built html to be sure: the `<script src=.../keyboard-nav.js>` tag is on every page and the footer links it hooks into (`md-footer__link--prev`/`--next`) render normally.

my hunch is zensical serve doesn't re-read the new `extra_javascript` line without a full restart, so if you had a serve already running (or a warm `target/book`) when you switched to the branch, the script never got injected and the arrows just do nothing. could you kill the serve, re-run `uvx zensical serve`, and hard-refresh? if it's still dead after that, a quick check: does view-source on a page show the keyboard-nav.js script tag? that'd tell us config-loading vs browser.

can drop a short screen capture of it paging if that's easier.

docs/content/javascripts/keyboard-nav.js
```diff
@@ -0,0 +1,42 @@
...
+  function isTypingTarget(el) {
+    if (!el) {
+      return false;
+    }
+    if (el.isContentEditable) {
+      return true;
+    }
+    var tag = el.tagName;
+    return tag === 'INPUT' || tag === 'TEXTAREA' || tag === 'SELECT';
+  }
```
CommanderStorm, line 13: this does not seem to work seme to quite work

Recording 2026-07-10 at 02 21 01

CommanderStorm (changes requested):
Ok, I think I had something in the cache, can reproduce the buttons changing pages now, so it partially works now.

The search bar does not work anymore though...

AnayGarodia:
good catch, reproduced and fixed.

the bug was in the focus check. i only bailed when you were typing in the search box (activeElement is the input), but when you arrow down into the results the theme moves focus onto the result link - so activeElement is an `<a>`, my guard missed it, and left/right got hijacked and navigated the page out from under the open search.

now it bails whenever the search overlay is open at all, keyed on the theme's own `#__search` toggle checkbox rather than the focused element.

verified against the real search UI in a local build:
- open search via the button or the `/` hotkey -> `#__search` is checked
- with it open, left/right no longer navigate, regardless of where focus landed
- close search -> paging works again
- typing in the box, left/right move the cursor with no nav

on the zensical.toml suggestion: `extra_javascript` is the documented way to register a script (the config comment right above it points there), so i kept it. if you'd rather load it through a template override or hook, happy to switch.

AnayGarodia:
Marked ready. Re-verified the search-vs-paging interaction end to end in a fresh local build (`uvx zensical build` + `serve`), driving a real headless browser rather than eyeballing:

- on `/quick-start/`, → navigates to the next guide page, ← back to the previous one
- open the search overlay (`#__search` checked), then ←/→ no longer navigate the page — regardless of whether focus is on the input or an arrowed-into result link
- close search, ←/→ page through the guide again

That covers the regression you hit (paging worked but search broke): the guard now keys on the theme's `#__search` toggle being open, not on which element has focus, so it can't be defeated by focus moving onto a result link.

CommanderStorm:
No, does not work. Please actually test your work..

Recording 2026-07-12 at 09 58 07

CommanderStorm:
Does not currently work I think, so lets close this PR for now.
I would really like to see this fixed though, so another attempt would be greatly appreciated ^^

Issue #1086, opened 2023-12-19
Store hashing algorithm in the metadata table


nyurik:
`mbtiles` and `martin-cp` [standardize](https://maplibre.org/martin/mbtiles-validation.html#per-tile-validation) how a tile data is hashed in the mbtiles. `tilelive-copy` has been generating [normalized](https://maplibre.org/martin/mbtiles-schema.html#normalized) mbtiles schema with MD5 hashes of the tile data as the table foreign key.  So it was possible to validate the content of the tile by re-computing MD5 hash.

[tippecanoe](https://github.com/felt/tippecanoe/pull/48) (cc @bdon) has used a faster fnv1a checksum algorithm, so clearly there is need for more than one algo.

## Proposal
* Add `hash_algorithm` metadata value with the possible values `md5`, `fnv1a`, ...
* If the value is not there, check if [agg_tiles_hash](https://maplibre.org/martin/mbtiles-validation.html#aggregate-content-validation) metadata value is present, and use its length to decide algorithm
* If not, and mbtiles uses [normalized](https://maplibre.org/martin/mbtiles-schema.html#normalized) or [flat-with-hash](https://maplibre.org/martin/mbtiles-schema.html#flat-with-hash) schema, use the size of the hashing keys to decide the algorithm
* If the hashing algorithm is detected, add it to the metadata, possibly after validating that all hashes match their data

### Multiple Algorithms Consideration
MBTiles use hashes for the `tile_data` column and for the `agg_tiles_hash` metadata field.  In theory these could use different algorithms for performance and security(?) reasons.  If so, we could either have two metadata fields (`hash_algorithm` and `agg_hash_algorithm`), or better yet, rename `agg_tiles_hash` into `agg_tiles_hash_md5` (or other algo).  This keeps the algorithm and its value together, and is fairly easy to detect at runtime.

bdon:
> faster fnv1a checksum algorithm

In tippecanoe we needed a content hash that had the smallest implementation to inline, unsure how the speed compares to md5, but if you need speed you'll probably want something like [xxhash](https://github.com/Cyan4973/xxHash) anyways.

nyurik:
@bdon this proposal is to support and standardize all such algorithms, so that if tippecanoe created mbtiles file, other tools can validate it and possibly add more content, while using the same algo as the original

PR #2985, opened 2026-07-08
feat(mbtiles): read and validate the hash_algorithm metadata key

AnayGarodia:
Refs #1086. This is the **read/validation half** of the request (see scope note below).

## What

mbtiles/martin-cp hash tile data with MD5, but other tools (e.g. tippecanoe) use `fnv1a`. Validating such a file recomputed MD5 and reported a misleading `agg_tiles_hash` **mismatch**, giving no hint that the algorithm simply differs.

## How

- Add the `hash_algorithm` metadata key and a `HashAlgorithm` type (currently `Md5`, the only algorithm this build can compute), with case-insensitive parsing.
- `Mbtiles::get_hash_algorithm`: absent key ⇒ `Md5` (backwards compatible), explicit-but-unrecognized value ⇒ a clear new `MbtError::UnsupportedHashAlgorithm` instead of a confusing hash mismatch.
- Wire it into agg-hash validation so an unsupported-algorithm file fails with an actionable message.

## Scope (intentional)

This PR is **purely additive on the read side** — it does not yet *write* `hash_algorithm` on file creation. Doing so changes generated-file output and would churn ~15 integration metadata snapshots (including Postgres-backed `martin-cp` fixtures that need the full Docker/PG stack to regenerate). I kept this slice self-contained and fully unit-tested so it's easy to review; writing the key on creation and adding more algorithms is a natural follow-up. Hence `Refs` rather than `Fixes` — happy to fold in the write-side (with regenerated snapshots) if you'd prefer a single PR.

## Testing

```
cargo test -p mbtiles --lib     # 50 passed (4 new: default-md5, case-insensitive, reject-unsupported, ...)
cargo test -p mbtiles --doc     # 21 passed
cargo clippy -p mbtiles --all-targets -- -D warnings   # clean
```

### DIFF
--- a/mbtiles/src/errors.rs
+++ b/mbtiles/src/errors.rs
@@ -89,7 +89,10 @@ pub enum MbtError {
     #[error(
         "MBTiles file {filepath} declares an unsupported tile hash algorithm `{algorithm}` in its metadata. This build can only validate `md5` hashes."
     )]
-    UnsupportedHashAlgorithm { algorithm: String, filepath: String },
+    UnsupportedHashAlgorithm {
+        algorithm: String,
+        filepath: PathBuf,
+    },
 
     #[error(r#"Filename "{0}" passed to SQLite must be valid UTF-8"#)]
     InvalidFilenameType(PathBuf),
--- a/mbtiles/src/errors.rs
+++ b/mbtiles/src/errors.rs
@@ -86,6 +86,11 @@ pub enum MbtError {
     )]
     AggHashValueNotFound(String),
 
+    #[error(
+        "MBTiles file {filepath} declares an unsupported tile hash algorithm `{algorithm}` in its metadata. This build can only validate `md5` hashes."
+    )]
+    UnsupportedHashAlgorithm { algorithm: String, filepath: String },
+
     #[error(r#"Filename "{0}" passed to SQLite must be valid UTF-8"#)]
     InvalidFilenameType(PathBuf),
 
--- a/mbtiles/src/validation.rs
+++ b/mbtiles/src/validation.rs
@@ -31,34 +31,24 @@ pub const AGG_TILES_HASH_AFTER_APPLY: &str = "agg_tiles_hash_after_apply";
 /// Metadata key for a diff file, describing the expected [`AGG_TILES_HASH`] value of the tileset to which the diff will be applied.
 pub const AGG_TILES_HASH_BEFORE_APPLY: &str = "agg_tiles_hash_before_apply";
 
-/// Metadata key recording the algorithm used to hash tile data and to compute
-/// the [`AGG_TILES_HASH`] value.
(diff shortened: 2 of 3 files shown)


mbtiles/src/errors.rs
```diff
@@ -86,6 +86,11 @@ pub enum MbtError {
     )]
     AggHashValueNotFound(String),
 
+    #[error(
+        "MBTiles file {filepath} declares an unsupported tile hash algorithm `{algorithm}` in its metadata. This build can only validate `md5` hashes."
+    )]
+    UnsupportedHashAlgorithm { algorithm: String, filepath: String },
```
CommanderStorm:
```suggestion
    UnsupportedHashAlgorithm { algorithm: String, filepath: PathBuf },
```

mbtiles/src/validation.rs
```diff
@@ -223,6 +260,28 @@ impl Mbtiles {
...
+    pub async fn get_hash_algorithm<T>(&self, conn: &mut T) -> MbtResult<HashAlgorithm>
+    where
+        for<'e> &'e mut T: SqliteExecutor<'e>,
+    {
+        match self.get_metadata_value(&mut *conn, HASH_ALGORITHM).await? {
+            None => Ok(HashAlgorithm::Md5),
+            Some(value) => {
+                HashAlgorithm::parse(&value).ok_or_else(|| MbtError::UnsupportedHashAlgorithm {
+                    algorithm: value,
+                    filepath: self.filepath().to_string(),
```
CommanderStorm:
```suggestion
                    filepath: self.filepath(),
```

mbtiles/src/validation.rs
```diff
@@ -618,6 +677,10 @@ LIMIT 1;"
         let Some(stored) = self.get_agg_tiles_hash(&mut *conn).await? else {
             return Err(AggHashValueNotFound(self.filepath().to_string()));
         };
+        // Make sure the stored hash was produced with an algorithm we can
+        // recompute. Otherwise comparing an md5 against, say, an fnv1a hash
+        // would surface as a confusing `AggHashMismatch`. See issue #1086.
+        let HashAlgorithm::Md5 = self.get_hash_algorithm(&mut *conn).await?;
```
CommanderStorm, line 669: did not know this sytax is valid..

CommanderStorm (changes requested):
Mostly LGTM, but lets stick to `PathBuf` in the errrors. I think we have somewhat standardised on this.

AnayGarodia:
Thanks! Switched `filepath` to `PathBuf` in `UnsupportedHashAlgorithm` (and the call site uses `self.filepath().into()`). `cargo fmt --check`, `clippy -D warnings`, and `cargo test -p mbtiles` are green.

CommanderStorm (approved):
CI needs a restart of the flaky job to merge, will do that tomorrow

CommanderStorm:
Thanks for the impl of this 👍🏻

PR #2989, opened 2026-07-09
chore: fix clippy lints for 1.97 release

nyurik:


nyurik:
self-merging as this is trivial

PR #2997, opened 2026-07-13
feat: source-location spans for finalize() validation errors

CommanderStorm:
Makes sure that more of the validation errors that occur *after* YAML parsing (during `Config::finalize()`) include source location info, so miette renders a snippet pointing at the offending value in the config file:

```
martin::config::base_path

  × Base path must begin with '/' and be a valid URL path, but got 'not-a-path'
   ╭─[config.yaml:2:12]
 1 │ pmtiles: /tmp
 2 │ base_path: not-a-path
   ·            ─────┬────
   ·                 ╰── invalid base_path
   ╰────
  help: The path must start with '/' and be a valid URI path, e.g. `/tiles` or `/api/v1`.
```

### DIFF
--- a/martin/src/config/file/cors.rs
+++ b/martin/src/config/file/cors.rs
@@ -101,9 +101,12 @@ impl ConfigurationLivecycleHooks for CorsProperties {
 }
 
 impl CorsProperties {
-    pub fn validate(&self) -> ConfigFileResult<()> {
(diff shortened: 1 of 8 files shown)


CommanderStorm:
this is horrible. Needs Spanned<T> is used inside a struct with #[serde(flatten)]

PR #2999, opened 2026-07-13
feat(config): Spanned validation errors

CommanderStorm:
if parsing failed at runtime, we should include the context of where the validation error is coming from if possible.

Likely needs upstream work in serde-saphyr.


CommanderStorm:
not as simple it seems

PR #3002, opened 2026-07-14
fix(martin-ui): center tile inspect map on tileset bounds, not 0/0/0

### LINKED ISSUE #3001: Inspect dialog of `martin-ui` still opens at `0/0/0` despite #2574
## What happens

Opening the Tile Inspect dialog in the Web UI always shows the map at `#map=0/0/0` (null island, zoom 0), regardless of the tileset. This is the exact symptom that #2325 reported and that #2574 was merged to fix, but the fix does not take effect in practice. The TileJSON already carries correct `bounds` and `center`, yet the inspect map ignores them.

Observed on martin v1.10.1 and on current `main`.

## Expected behavior

As intended by #2325 and #2574, the inspect map should open centered on the tileset, restricted to the dataset extent (`bounds`) and recommended zoom (`center`) from the TileJSON, instead of the global 0/0/0 view.

## How to reproduce

1. Serve any tileset with a local (non world) extent whose `minzoom` is 0, for example an OpenMapTiles set produced by Planetiler: `martin ./some.mbtiles --webui enable-for-all`
2. Open `http://localhost:3000/` and go to the Tiles Catalog.
3. Click `Inspect` on the source.
4. The map opens at `#map=0/0/0` rather than at the tileset center or bounds. You have to pan and zoom manually to find the data.

Confirming the metadata is present (so this is a viewer issue, not a data issue): `curl http://localhost:3000/<source>` returns a TileJSON with valid `bounds` and `center`.

## Cause and proposed fix

`martin/martin-ui/src/components/dialogs/tile-inspect-map.tsx` has three separate problems, all present since #2574:

1. Race between the fetch and the map load. `configureMap()` runs only from the map `onLoad` handler, which usually fires before the TileJSON fetch resolves. It then hits the early `return` (no data yet) and never applies the view. Nothing re-runs it once the fetch resolves, so the map stays at 0/0/0.
2. Falsy zero minzoom. `if (tileJson.minzoom)` skips a valid `minzoom` of 0, which is common for world minimum tilesets such as OpenMapTiles.
3. Dropped recommended zoom. `setCenter([center[0], center[1]])` ignores `center[2]` (the recommended zoom) and no zoom is set otherwise, so the zoom stays at 0.

Proposed fix:

* Re-apply the view once the TileJSON arrives, via a `useEffect` keyed on the fetched data, so it works no matter which of the fetch or the map load wins.
* Use `!== undefined` checks for `minzoom` and `maxzoom`.
* Derive both center and zoom from the data extent with `fitBounds`, falling back to `jumpTo` with `center` including `center[2]` for world extent sets.
* Respect an explicit `#map=z/lat/lng` deep link (do not override a shared link; the default `#map=0/0/0` written by the `hash` prop is treated as unset).

I already have a branch implementing this, with unit tests for the three cases above:
- https://github.com/maplibre/martin/compare/main...yuiseki:martin:fix/inspect-respect-tilejson-bounds

Side note: no existing test caught this because the shared test mocks in `vitest.setup.tsx` stub `Popup` and `MaplibreInspect` as arrow functions, which are not constructable, so no test exercises the `onLoad` path that runs `configureMap()`. The branch works around this locally in the new test file and it may be worth fixing the shared mocks too.

## Related issues and PRs

* #2325 (issue, closed)
* #2574 (PR, merged): Closed #2325 but the fix does not take effect (this issue).
* #2326 (PR, closed): earlier abandoned attempt at #2325.
* #2797 (PR, merged): migrated the UI to openapi-fetch and left `configureMap` wired only to `onLoad`, which is where the race surfaces.

**AI Disclosure**

I used Claude Code Opus 4.8 to investigate, reproduce and try to fix this issue.

yuiseki:
- Fixes #3001.

The tile inspect map ignored the TileJSON `center`/`bounds` and always opened at `#map=0/0/0`. As detailed in the issue, three things combined to cause this: `configureMap()` ran only from the map `onLoad` handler and lost the race against the TileJSON fetch, `if (tileJson.minzoom)` skipped a valid `minzoom` of 0, and `setCenter` dropped `center[2]`.

Changes:

* Re-apply the view once the TileJSON arrives, via a `useEffect` keyed on the fetched data, so it no longer depends on which of the fetch or the map load wins.
* Use `!== undefined` checks for `minzoom`/`maxzoom`.
* Derive center and zoom from the extent with `fitBounds`, falling back to `jumpTo` with `center` (including `center[2]`) for world-extent sets.
* Respect an explicit `#map=z/lat/lng` deep link, treating the default `#map=0/0/0` as unset.

Added unit tests for the three cases (bounds fit with zero minzoom, world-extent jumpTo, deep-link respected).

One note: the shared mocks in `vitest.setup.tsx` stub `Popup` and `MaplibreInspect` as arrow functions, which are not constructable, so no existing test exercised the `onLoad` path; the new test file provides constructable stubs locally, and fixing the shared mocks could be a good follow-up.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -137,7 +137,7 @@ jobs:
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with: { persist-credentials: false }
       - uses: ./.github/actions/setup-rust
-        with: { toolchain: nightly, tools: 'cargo-shear,just', rendering: true }
+        with: { toolchain: nightly-2026-07-12, tools: 'cargo-shear,just', rendering: true }
       - run: just shear
   build-aarch64-unknown-linux-musl:
     uses: ./.github/workflows/build-binary.yml
--- a/martin/martin-ui/__tests__/components/dialogs/tile-inspect-map.test.tsx
+++ b/martin/martin-ui/__tests__/components/dialogs/tile-inspect-map.test.tsx
@@ -1,0 +1,189 @@
+import { cleanup, render, waitFor } from '@testing-library/react';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { TileInspectDialogMap } from '@/components/dialogs/tile-inspect-map';
+import { martinClient } from '@/lib/martin-client';
+import type { Catalog } from '@/lib/types.gen';
+
+// A single mock MapLibre instance whose methods we assert against. Hoisted so
+// the vi.mock factory below (also hoisted) can reference it.
+const { mapSpies } = vi.hoisted(() => ({
+  mapSpies: {
+    addControl: vi.fn(),
+    addLayer: vi.fn(),
+    addSource: vi.fn(),
+    fitBounds: vi.fn(),
+    getLayer: vi.fn(),
+    getSource: vi.fn(),
+    getStyle: vi.fn(),
+    isStyleLoaded: vi.fn(),
+    jumpTo: vi.fn(),
+    on: vi.fn(),
+    removeControl: vi.fn(),
+    removeLayer: vi.fn(),
+    removeSource: vi.fn(),
+    setCenter: vi.fn(),
+    setMaxBounds: vi.fn(),
+    setMaxZoom: vi.fn(),
+    setMinZoom: vi.fn(),
+  },
+}));
+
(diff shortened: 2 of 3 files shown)


CommanderStorm (approved):

CommanderStorm:
Thanks for the improvement 🎉 

(CI will need an re-run for the failing run, will do that tomorrow after a night of sleep. not something that I think you can do under GitHubs perimisison model)

PR #3010, opened 2026-07-14
chore: more sqlx precompiled stmts

nyurik:
keeps editors consistent


CommanderStorm (approved):
but have a look at copilot and fix the failing ci.

PR #3037, opened 2026-07-19
chore: use re-exported mvt from mlt

nyurik:


CommanderStorm (approved):

martin/src/config/file/mod.rs
```diff
@@ -12,7 +12,11 @@ pub use error::{ConfigFileError, ConfigFileResult};
 
 pub mod process;
 pub use process::ProcessConfig;
-#[cfg(all(feature = "mlt", feature = "_tiles"))]
+#[cfg(all(
+    feature = "mlt",
+    feature = "_tiles",
```
CommanderStorm:
```suggestion
```

michaelkirk:
Does this imply that you cannot use MVT in martin without also enabling MLT?

CommanderStorm:
Yes.. maybe was not the best choice to merge this.. fair point..
@nyurik should we revert or is the simpler deps management outweighing this?

nyurik:
tbh, there is relatively small price - MLT is pure rust, and if MLT itself is disabled, we would only waste some compilation time, whereas linking would remove the unused code.  On the other hand, I would much rather use MVT via MLT because whenever we do compile with MLT (which is a much more common case going forward), we will always have consistent versioning between the two

CommanderStorm:
> we would only waste some compilation time

Nope, adding `mlt` as a feature adds the accept header re-encoding support.

CommanderStorm:
can be disabled in the config though, so likely not a big deal.

nyurik:
I meant that if we have (introduce?) `mlt` and `mvt` features, they both can depend on the `mlt-core` crate, but the actual `#[cfg(...)]` would be different - thus we won't link the MLT code if we only want MVT.  That said, I think we are optimizing for a relatively narrow usecase - its not like duckdb or MLN -- which require C++ code binding and tons of dependencies with flaky compilation

Issue #749, opened 2023-07-06
Configure Cache Control Expires headers


nyurik:
Martin should set some sort of the expiry HTTP headers, max-age or similar headers, next to the already existing `E-Tag` headers.

See also [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control)

CommanderStorm:
We likely can also make our caches expirable in a similar way -> https://docs.rs/moka/latest/moka/future/struct.CacheBuilder.html#method.expire_after

nyurik:
i don't think we should mix internal cache and the expiration headers -- while they both speed up access, they function very differently, and have significantly different controls.  I would recommend https://www.varnish-software.com/developers/tutorials/object-lifetime/ as a good intro to how a proxy could behave - note that martin is somewhat of a proxy - esp if it gets content from a http-hosted pmtiles.

CommanderStorm:
I guess you might have usecases where your internal cache is higher than the one you give out.
But still, this concept is very related and doing a full varnish configuration is likely overkill for a larger portion of users.

Not everyone is wikipedia and working with VCL is a resonably high learning curve 😉

nyurik:
that's not what i meant.  Martin has many concepts similar to Varnish - so we may need at least some of the same cache controls to configure it:

* internal object cache
  *  TTL, Grace, Keep controls -- determine how long an object stays in Martin's internal cache, and what happens if the TTL has expired, .... This could be as simple as TTL only, but note that for max performance, we want to allow "respond while refreshing" - i.e. if a user asks for a tile, they may get an expired tile while Martin starts refreshing the tile cache in the background.  This way you don't get big delays in responses every time a popular tile is expired.
  * This should be discussed in a separate ticket
* Header values sent to the client - this is much simpler - we simply tell the client when to refresh.  This is what this ticket is about.

CommanderStorm:
I am not sure if everything that you would in an ideal world want of the cache is currently possible with moka, but yea.. different issue.

nyurik:
@CommanderStorm this ticket is not about internal Martin caching, only about what headers it sends to the clients.  Lets split the internal cache configuration into a separate ticket

CommanderStorm:
I opened
- https://github.com/maplibre/martin/issues/2505 to discuss this, if someone wants to look into this 😉

CommanderStorm:
From a chat on slack:

> We likely should have global header config, and per-source header configs
[...]

PR #3042, opened 2026-07-21
feat: add configurable server-level Cache-Control response header

AnayGarodia:
## Summary

- add a top-level `cache_control` server setting
- apply it as the default response header without overriding endpoint-specific policies
- validate invalid header values during server startup
- document the setting in the generated config and JSON schema

This addresses the global configuration portion of #749. Per-source and per-source-type overrides remain separate follow-up design work; this PR does not change Martin's internal tile-cache expiration.

## Testing

- `cargo test -p martin --lib` (359 passed)
- `just clippy -- -D warnings`
- `just test-schemas` (all config fixtures and schemas passed; its final docs extraction command uses GNU-only `mktemp --suffix` on macOS)
- extracted the generated YAML example with a portable temporary path and validated it with `check-jsonschema`
- `just lint` completed the Rust feature matrix, formatting, UI lint, and UI type check; markdown link checking could not start because Docker credential lookup stalled locally

Implementation and tests were prepared with Codex assistance and reviewed against the full issue history and repository contribution guidance.


martin/src/config/file/srv.rs
```diff
@@ -57,6 +57,16 @@ pub struct SrvConfig {
...
     pub preferred_encoding: Option<PreferredEncoding>,
+    /// Set the default `Cache-Control` response header.
+    ///
+    /// The value is used for responses that do not define a more specific cache policy. For
+    /// example: `public, max-age=3600`. Endpoints with an explicit policy, such as the health
+    /// check, keep their own header.
+    #[cfg_attr(
+        feature = "unstable-schemas",
+        schemars(example = &"public, max-age=3600")
+    )]
```
CommanderStorm: We use the schema.json example to document the default.
Not sure what the default should be. @nyurik no default?

```suggestion
    /// Set the default `Cache-Control` response header.
    ///
    /// The value is used for responses that do not define a more specific cache policy.
    /// For example: `public, max-age=3600`.
    /// Endpoints with an explicit policy, such as the health check, keep their own header.
```

martin/src/config/file/srv.rs
```diff
@@ -212,6 +229,33 @@ mod tests {
...
+
+    #[test]
+    fn reject_invalid_cache_control_header() {
+        let config = SrvConfig {
+            cache_control: Some("public\nmax-age=3600".to_string()),
+            ..Default::default()
+        };
+
+        config.cache_control_header().unwrap_err();
+    }
```
CommanderStorm, line 306: We need to validate this at config parsing/maybe finalisation, not runtime. 
First because that would be a bit wastefully, but also because we can have much better error messages pointing to the actual line where it went wrong, if we do it ahead of time. 

This might mean writing a short unit struct or docs.rs/garde clause.

There is for example
https://docs.rs/actix-web/latest/actix_web/http/header/struct.CacheControl.html
But it does not need to be this exact code, there are a few alternatives.

martin/src/srv/server.rs
```diff
@@ -279,3 +293,56 @@ pub fn new_server(
...
+                .route("/content", web::get().to(content))
+                .wrap(cache_control_middleware(None)),
+        )
+        .await;
+
+        let response =
+            test::call_service(&app, test::TestRequest::get().uri("/content").to_request()).await;
+        assert!(!response.headers().contains_key(CACHE_CONTROL));
+    }
+}
```
CommanderStorm: Please add it to the config read in tests/test.sh and re-bless the output (I can also do that by adding a label to the pr if simpler).

This was we don't add another integration test scheme.

CommanderStorm (changes requested):
Mostly LGTM, but I have a few things that I would like to have cleaned up slightly.

martin/src/config/file/srv.rs
AnayGarodia: Removed the schema example/default in 426c5d61. The generated config now shows `cache_control: null`, while the field docs retain `public, max-age=3600` only as an example value.

martin/src/config/file/srv.rs
AnayGarodia: Moved validation into configuration deserialization in 426c5d61. `CacheControlHeader` now validates both the HTTP header value and parsed Cache-Control directives, so invalid YAML fails with the source line/column before server construction. Focused config and server tests pass, and clippy is clean.

martin/src/srv/server.rs
AnayGarodia: Added `cache_control` to `tests/config.yaml` and re-blessed the configured outputs in 426c5d61: the saved config plus all 41 configured response-header snapshots now cover the default header. Schema fixtures and the generated-doc YAML validate against the config schema.

martin/src/config/file/srv.rs
```diff
@@ -212,6 +272,38 @@ mod tests {
...
+            cache_control: max-age=invalid
+        "})
+        .unwrap_err();
+
+        assert!(
+            error
+                .to_string()
+                .contains("invalid Cache-Control header value"),
+            "unexpected error: {error}"
+        );
```
CommanderStorm: lets assert the full error, is tighter.

```suggestion
        insta::assert_snapshot!(error.to_string(), @"");
```

CommanderStorm (approved):
LGTM from my side. @nyurik is the place in the config ok for you?

martin/src/config/file/srv.rs
```diff
@@ -212,6 +273,32 @@ mod tests {
...
+    }
+
+    #[test]
+    fn reject_invalid_cache_control_header() {
+        insta::assert_snapshot!(
+            render_failure(indoc::indoc! {"
+                cache_control: max-age=invalid
+            "}),
+            @r#"");
```
CommanderStorm:
```suggestion
            @"");
```

martin/src/config/file/srv.rs
```diff
@@ -212,6 +273,32 @@ mod tests {
...
+    #[test]
+    fn reject_invalid_cache_control_header() {
+        insta::assert_snapshot!(
+            render_failure(indoc::indoc! {"
+                cache_control: max-age=invalid
+            "}),
+            @"");
+    }
+
     #[test]
```
CommanderStorm, line 308:
```suggestion
    #[test]
    fn reject_invalid_cache_control_header() {
        insta::assert_snapshot!(
            render_failure(indoc::indoc! {"
                cache_control: max-age=invalid
            "}),
            @"");
    }

    #[test]
    fn reject_empty_cache_control_header() {
        insta::assert_snapshot!(
            render_failure(indoc::indoc! {"
                cache_control: ""
            "}),
            @"");
    }

#[test]
```

martin/src/config/file/srv.rs
```diff
@@ -57,6 +104,13 @@ pub struct SrvConfig {
     /// `gzip` is faster, but `brotli` is smaller, and may be faster with caching.
     /// Default could be different depending on Martin version.
     pub preferred_encoding: Option<PreferredEncoding>,
+    /// Set the default `Cache-Control` response header.
+    ///
+    /// The value is used for responses that do not define a more specific cache policy. For
+    /// example: `public, max-age=3600`. Endpoints with an explicit policy, such as the health
+    /// check, keep their own header.
```
CommanderStorm:
```suggestion
    /// The value is used for responses that do not define a more specific cache policy.
    /// For example: `public, max-age=3600`.
    /// Endpoints with an explicit policy, such as the health check, keep their own header.
```

AnayGarodia:
Fixed the CI break in 6b2e92e4. The empty YAML scalar no longer terminates the `indoc!` literal, and both cache-control validation tests now contain their expected inline snapshots.

Verified locally with `cargo fmt --all -- --check`, both validation tests, and `cargo clippy --workspace --all-targets`; all pass.

martin/src/config/file/srv.rs
```diff
@@ -212,6 +273,53 @@ mod tests {
...
+        insta::assert_snapshot!(
+            render_failure(indoc::indoc! {"
+                cache_control: max-age=invalid
+            "}),
+            @r"
+martin::config::yaml (https://maplibre.org/martin/config-file/)
+
+  × invalid Cache-Control header value 'max-age=invalid': no valid directives
+  help: Check the highlighted token in your YAML. The error usually indicates
+        a mismatched type or an unexpected shape.
```
CommanderStorm, line 304: weird that no token is highlighted.
I need to investigate why this is happening, that should not happen.

martin/src/config/file/srv.rs
```diff
@@ -212,6 +273,53 @@ mod tests {
...
+    #[test]
+    fn parse_cache_control() {
+        let config = serde_saphyr::from_str::<SrvConfig>(indoc! {"
+            cache_control: public, max-age=3600, stale-while-revalidate=60
+        "})
+        .unwrap();
+
+        assert_eq!(
+            config.cache_control.as_ref().map(ToString::to_string),
+            Some("public, max-age=3600, stale-while-revalidate=60".to_string())
```
CommanderStorm:
```suggestion
            Some("public, max-age=3600, stale-while-revalidate=60".to_owned())
```

martin/src/config/file/srv.rs
```diff
@@ -1,20 +1,67 @@
...
+impl CacheControlHeader {
+    #[must_use]
+    pub(crate) fn header_value(&self) -> HeaderValue {
+        self.0.clone()
+    }
+}
+
+impl fmt::Display for CacheControlHeader {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str(self.0.to_str().map_err(|_| fmt::Error)?)
```
CommanderStorm:
```suggestion
        #[expect(clippy::map-err-ignore, reason = "cannot be returned and is not usefull to log")]
        f.write_str(self.0.to_str().map_err(|_| fmt::Error)?)
```

martin/src/config/file/srv.rs
```diff
@@ -1,20 +1,68 @@
...
+impl CacheControlHeader {
+    #[must_use]
+    pub(crate) fn header_value(&self) -> HeaderValue {
+        self.0.clone()
+    }
+}
+
+impl fmt::Display for CacheControlHeader {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        #[expect(clippy::map-err-ignore, reason = "cannot be returned and is not usefull to log")]
```
CommanderStorm:
```suggestion
        #[expect(clippy::map_err_ignore, reason = "cannot be returned and is not usefull to log")]
```

martin/src/config/file/srv.rs
```diff
@@ -1,20 +1,71 @@
...
+    }
+}
+
+impl fmt::Display for CacheControlHeader {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        #[expect(
+            clippy::map_err_ignore,
+            reason = "cannot be returned and is not usefull to log"
+        )]
+        f.write_str(self.0.to_str().map_err(|_| fmt::Error)?)
```
CommanderStorm:
```suggestion
        f.write_str(self.0.to_str().map_err(|_e| fmt::Error)?)
```

Issue #3048, opened 2026-07-25
Replace tests/test.sh with a Rust test harness?


johncarmack1984:
I tried to run the integration suite on my Mac this week and bounced off: GNU sed, the brotli curl check, a few `$OSTYPE == linux*` gates. Which sent me back to the Slack thread from last September -- @CommanderStorm wanted tests that are kinder to new contributors, and @nyurik sketched the fix: spawn the martin binary, hit it with reqwest, snapshot with insta, all in a separate non-published crate with no internal APIs.

The thread ended "not urgent," but I had the itch, so I built it: https://github.com/maplibre/martin/compare/main...johncarmack1984:tests/rust-harness-poc. Two groups ported (`auto_mini` and the `--route-prefix` health check), runs in about 2s on my Mac, which had previously been unable to run test.sh. I think #2785 and #2824 were having the same kind of portability problem.

I tripped over two things while porting: (1) an exported `AWS_PROFILE` fails the suite today; martin's warning text changed, and the `validate_log` allowlist still matches the old wording. And (2) `settings.rendering` only exists in Linux builds currently, so expected output on one platform becomes wrong on another. The harness clears the env and asserts that key per platform.

Before I go further:

- Acceptable? Still the direction you want?
- Sync or async preference? I went sync/blocking for simplicity/readability, but easy to change if async preferred.
- If it looks right, I'll keep porting group by group, small PRs, with the Postgres groups on the existing docker-compose db.

If you'd rather keep the shell suite, no hard feelings! :) It was a fun build either way.

CommanderStorm:
> Acceptable? Still the direction you want?

Yes! and much appreciaed!

> Sync or async preference? I went sync/blocking for simplicity/readability, but easy to change if async preferred.

I think if the tests are `tokio::test` this is a bit more ideomatic and maybe efficent, but not entirely set on this.

> If it looks right, I'll keep porting group by group, small PRs, with the Postgres groups on the existing docker-compose db.

Yes, much appreciaed!

PR #3049, opened 2026-07-25
test: start splitting apart the monolythic e2e tests into cleaner rust

johncarmack1984:
Working PoC for #3048 (built so we're working from code instead of a proposal)

New `integration-tests` crate, `publish = false`, no martin internals modified. It spawns the compiled binary and runs HTTP queries, same as test.sh.

Two groups ported: `auto_mini` and the `--route-prefix` health check (#2723).

Every test gets its own martin on a random port, so they run in parallel. Logs get the `test_log_has_str`/`validate_log` treatment, reqwest handles the brotli, and insta replaces `tests/expected` -- bless with `INSTA_UPDATE=always cargo test -p martin-integration-tests`.

`MARTIN_BIN` still works, docker form included.

```
cargo build -p martin
cargo test -p martin-integration-tests
```

Justfile and CI deliberately untouched. Wanted to start with a small diff.

Thoughts? Objections? Things that could be done better or more idiomatically?


integration-tests/src/lib.rs
```diff
@@ -0,0 +1,511 @@
...
+use std::path::{Path, PathBuf};
+use std::process::{Child, Command, Stdio};
+use std::sync::{Arc, Mutex};
+use std::thread::{self, JoinHandle};
+use std::time::{Duration, Instant};
+
+use brotli::Decompressor;
+use flate2::read::GzDecoder;
+use regex::Regex;
+use reqwest::blocking::Client;
```
CommanderStorm: instead of blocking, can you use `tokio::test` and async? I think it might be a bit more "ideomatic"

integration-tests/src/lib.rs
```diff
@@ -0,0 +1,511 @@
...
+            {
+                return;
+            }
+            let early_exit = self.child.try_wait().expect("failed to poll martin");
+            assert!(
+                early_exit.is_none(),
+                "martin exited during startup with {}; log:\n{}",
+                early_exit.map_or_else(String::new, |status| status.to_string()),
+                self.raw_log()
+            );
```
CommanderStorm: this should be returned as a `Result`, so that tests can actually assert that we are erroring for the right reasons 😉

integration-tests/tests/pmtiles.rs
```diff
@@ -0,0 +1,87 @@
+//! Rust port of the "auto configured Martin" pmtiles slice of `tests/test.sh`
+//! (the `auto_mini` and `route_prefix_health` groups).
+
+use std::fs;
+
+use martin_integration_tests::Martin;
+
+/// `tests/test.sh` "Test minimum auto configured Martin": start martin with
```
CommanderStorm: please de-vibe the comments. I don't think most are nessary.
Good tests should not have comments..

integration-tests/src/lib.rs
```diff
@@ -0,0 +1,511 @@
...
+
+/// Log lines which are known to be acceptable in any test.
+///
+/// Mirrors `validate_log` in `tests/test.sh`.
+const ALLOWED_LOG_LINES: &[&str] = &[
+    "Margin parameter in ST_TileEnvelope is not supported",
+    "PostgreSQL is older than the recommended minimum 12.0.0",
+    "In the used version, some geometry may be hidden on some zoom levels.",
+    "Unable to deserialize SQL comment on public.points2 as tilejson",
+    "Environment variable AWS_PROFILE not supported anymore",
```
CommanderStorm: was removed, ups..

```suggestion
```

integration-tests/src/lib.rs
```diff
@@ -0,0 +1,511 @@
...
+use regex::Regex;
+use reqwest::blocking::Client;
+
+const READY_TIMEOUT: Duration = Duration::from_mins(1);
+const READY_POLL_INTERVAL: Duration = Duration::from_millis(100);
+const STOP_TIMEOUT: Duration = Duration::from_secs(10);
+
+/// Log lines which are known to be acceptable in any test.
+///
+/// Mirrors `validate_log` in `tests/test.sh`.
```
CommanderStorm: please trim the comments a bit to what is actually important 😉

```suggestion
```

integration-tests/src/lib.rs
```diff
@@ -0,0 +1,511 @@
...
+    Command::new(bin)
+}
+
+fn free_port() -> u16 {
+    TcpListener::bind("127.0.0.1:0")
+        .expect("failed to bind an ephemeral port")
+        .local_addr()
+        .expect("failed to read the bound address")
+        .port()
+}
```
CommanderStorm, line 78: I think this will be a bit flaky if we scale up to more tests..
Time of check and time of use is sligtly not the same and is not locking..

integration-tests/src/lib.rs
```diff
@@ -0,0 +1,511 @@
...
+        self
+    }
+
+    /// Path polled until it responds with a success status (default `/health`).
+    /// Needed when `--route-prefix` moves the health endpoint.
+    #[must_use]
+    pub fn readiness_path(mut self, path: &str) -> Self {
+        self.readiness_path = Some(path.to_owned());
+        self
+    }
```
CommanderStorm, line 128: lets have this not be a parameter in the builder, but set this up and poll it internally automatically

integration-tests/tests/pmtiles.rs
```diff
@@ -0,0 +1,87 @@
...
+        .remove("settings")
+        .expect("catalog has a settings key");
+    #[cfg(target_os = "linux")]
+    assert_eq!(settings, serde_json::json!({ "rendering": false }));
+    #[cfg(not(target_os = "linux"))]
+    assert_eq!(settings, serde_json::json!({}));
+
+    let mut insta_settings = insta::Settings::clone_current();
+    insta_settings.set_sort_maps(true);
+    insta_settings.bind(|| insta::assert_json_snapshot!("catalog_auto", catalog_json));
```
CommanderStorm: for short snapshots, lets use inline snapshots for better locality of the test 😉 
Applies in other places too.


```suggestion
    insta_settings.bind(|| insta::assert_json_snapshot!("catalog_auto", catalog_json, @"TODO, bless"));
```

integration-tests/tests/harness.rs
```diff
@@ -0,0 +1,21 @@
+//! Tests for the harness's own failure reporting - the part a contributor
+//! depends on when their martin subprocess misbehaves.
+
+use martin_integration_tests::Martin;
+
+/// A martin that exits during startup (unknown CLI flag) must be reported
+/// with the captured log instead of hanging until the readiness timeout.
+#[test]
+#[should_panic(expected = "martin exited during startup")]
```
CommanderStorm: Lets test this explicity instead. this is a bit fuzzy..

CommanderStorm (changes requested):
Thanks for working on this.
Looks like a nice start!

CommanderStorm:
thinking about it, lets also remove the ported tests from tests.sh, so we can known what already has been ported 😉

integration-tests/src/lib.rs
johncarmack1984: Done, the harness and tests are async now (tokio + async reqwest).

integration-tests/src/lib.rs
johncarmack1984: Done. start() returns Result<Martin, StartError> now, and the harness test asserts the EarlyExit variant and the captured log instead of matching a panic message.

integration-tests/tests/pmtiles.rs
johncarmack1984: Done for tests, all but one comment gone from the test files. The remaining doc comments in the `lib.rs` harness were trimmed down, but can be removed completely if preferred.

integration-tests/src/lib.rs
johncarmack1984: Dropped. Kept the env_remove so a shell exporting AWS_PROFILE doesn't trip the new deprecation warning during assert_log_clean.

integration-tests/src/lib.rs
johncarmack1984: Done.

integration-tests/src/lib.rs
johncarmack1984: Makes sense. Binding :0 would fix it properly, but martin logs the configured address, not the resolved one, so the harness can't learn the port. Went with retry instead: if the child exits with the bind error, start() picks a new port and goes again. If you'd take a small patch to log the resolved address, I'll switch to :0 and delete the retry.

integration-tests/src/lib.rs
johncarmack1984: Removed. Readiness is now "any HTTP response at all": martin doesn't bind the socket until every source is configured, so a response means it's up. That also covers a route prefix set through a config file, which sniffing the args would miss.

integration-tests/tests/pmtiles.rs
johncarmack1984: Done, all four were short enough to inline. The snapshots dir is gone.

integration-tests/tests/harness.rs
johncarmack1984: Done. The test destructures StartError::EarlyExit and asserts the exit status and clap's message in the log.

johncarmack1984:
> thinking about it, lets also remove the ported tests from tests.sh, so we can known what already has been ported 😉

Done, auto_mini and route_prefix_health are out of test.sh. That took tests/expected/auto_mini with it, so its line in the justfile schema-validation fixtures went too (auto and configured still cover the schema).

integration-tests/src/lib.rs
```diff
@@ -0,0 +1,502 @@
+//! Test harness that runs the `martin` binary as a subprocess and lets tests
+//! assert on its HTTP responses and log output. Each [`Martin`] instance runs
+//! on its own port, so tests can run in parallel.
+
+#![allow(
```
CommanderStorm: personal preference (I want to eventually enable that lint ^^)

```suggestion
#![expect(
```

integration-tests/src/lib.rs
CommanderStorm: > If you'd take a small patch to log the resolved address, I'll switch to :0 and delete the retry. 

Sure, that seems cleaner 😉
We can also redact on the log with an regex or whatever.
The exact mechanism is up to you, I just care that this does not end up being as mistireously flaky as port races.
Likely the actual race is not as large so it would not come up, who knows.

integration-tests/tests/harness.rs
```diff
@@ -0,0 +1,34 @@
+//! Tests for the harness's own failure reporting.
+
+#![allow(clippy::panic, reason = "tests fail by panicking")]
```
CommanderStorm: personal preference (I want to eventually enable that lint ^^)

```suggestion
#![expect(clippy::panic, reason = "tests fail by panicking")]
```

tests/test.sh
CommanderStorm: Unrelated fix, but this log is never emitted anymore :)

```suggestion
```

CommanderStorm (approved):
Nice work, thanks.

I am going to enable automerge (since I think this is a very nice improvement!), but the cleanup related to ports you noted would also be much appreciated in a second PR.

integration-tests/src/lib.rs
```diff
@@ -0,0 +1,502 @@
+//! Test harness that runs the `martin` binary as a subprocess and lets tests
+//! assert on its HTTP responses and log output. Each [`Martin`] instance runs
+//! on its own port, so tests can run in parallel.
+
+#![expect(
+    clippy::panic,
+    reason = "panicking with rich context is this test harness's failure-reporting mechanism"
+)]
+
+use std::env;
```
CommanderStorm, line 7:
```suggestion

use std::env;
```

integration-tests/tests/harness.rs
```diff
@@ -0,0 +1,34 @@
+//! Tests for the harness's own failure reporting.
+
+#![expect(clippy::panic, reason = "tests fail by panicking")]
+
+use martin_integration_tests::{Martin, StartError};
```
CommanderStorm:
```suggestion
use martin_integration_tests::{Martin, StartError};
```

integration-tests/src/lib.rs
```diff
@@ -0,0 +1,497 @@
+//! Test harness that runs the `martin` binary as a subprocess and lets tests
+//! assert on its HTTP responses and log output. Each [`Martin`] instance runs
+//! on its own port, so tests can run in parallel.
+
+use std::env;
```
CommanderStorm, line 7:
```suggestion
#![expect(clippy::panic, reason = "tests fail by panicking")]

use std::env;
```

johncarmack1984:
Thanks for the feedback and the merge! Follow-up PR is up now: https://github.com/maplibre/martin/pull/3053

PR #3054, opened 2026-07-25
feat: e2e local geoparquet wiring via duckdb

manbhav234:
continuation of #2931 
part of #2264 

This PR complete the end-to-end support for building and serving vector tiles from local geoparquet files via duckdb and sets up a strong base for wiring e2e support for remote geoparquet, duckdb db files and cli support ( to be wired in subsequent PR's)

I haven't added any docs for this and have not updated the CHANGELOG at the moment. Should likely be done. 

The CI was also wired to run the duckdb cargo tests and the e2e test in `tests/test.sh`. This, however, runs on every PR. Is this something that needs to be improved since it takes a lot of time ?

Also, this is still behind `unstable-duckdb`.


CommanderStorm:
> The CI was also wired to run the duckdb cargo tests and the e2e test in tests/test.sh. This, however, runs on every PR. Is this something that needs to be improved since it takes a lot of time ?

The code there is arguably a bit weird, since it busts a few of our patterns (having a lot of logic in the justfile, ...).
Splitting is fine, we have 30 paralell runners, we might as well use them...

I think the best/simplest way to solve this is to check out this PR-stack and merge it into this PR.
THe refactoring in that PR solves a few of the test management issues we have, so rebasing this work ontop of that would be best.

I think adding a feature flag to the PR and splitting it out similarly to test-duckdb would be wise likely, or?

- https://github.com/maplibre/martin/pull/3051

martin-core/src/tiles/duckdb/pool.rs
```diff
@@ -168,12 +168,19 @@ impl DuckDBPoolManager {
...
     fn open_ready_connection(&self) -> Result<Connection, DuckDBPoolManagerError> {
-        let config = Config::default()
-            .access_mode(AccessMode::ReadOnly)
-            .map_err(|source| Open {
-                source: source.into(),
-                target: self.target.clone().into(),
-            })
+        // File-backed pools are opened read-only. In-memory GeoParquet pools cannot use
+        // AccessMode::ReadOnly - DuckDB rejects "in-memory database in read-only mode".
+        let config = match &self.target {
```
CommanderStorm, line 173: this needs a bit better explanation why duckdb rejects it. it is not entirely clear why they would 🤔

martin-core/src/tiles/duckdb/pool.rs
```diff
@@ -244,14 +251,13 @@ impl Manager for DuckDBPoolManager {
         conn: &mut Self::Type,
         _metrics: &Metrics,
     ) -> RecycleResult<Self::Error> {
-        let target = self.target.clone();
-        tokio::task::block_in_place(|| conn.execute_batch("SELECT 1")).map_err(|source| {
+        // Lightweight sync health check; avoid `block_in_place` so this works on actix's per-worker runtimes.
```
CommanderStorm:
```suggestion
```

martin/src/config/file/main/lifecycle.rs
```diff
@@ -415,6 +418,12 @@ impl Config {
             sources_and_warnings.push(Box::pin(val));
         }
 
+        #[cfg(feature = "unstable-duckdb")]
+        if !self.duckdb.is_empty() {
+            let val = self.duckdb.resolve(idr.clone(), self.cache.policy());
+            sources_and_warnings.push(Box::pin(val));
+        }
```
CommanderStorm, line 425: u sure? no `resolve_files`?

martin/src/config/file/tiles/duckdb/resolver/resolve.rs
```diff
@@ -0,0 +1,261 @@
...
+use crate::config::file::tiles::duckdb::{DuckDbConfig, DuckDbSourceEntry};
+use crate::config::file::{CachePolicy, ResolutionResult, TileSourceWarning};
+use crate::config::primitives::IdResolver;
+
+enum GeoParquetLocation {
+    Local(PathBuf),
+    Remote(Url),
+}
+
+fn classify_geoparquet_path(path: &Path) -> Result<GeoParquetLocation, String> {
```
CommanderStorm: I am pretty sure we have similar-ish logic somewhere. is this a dup or just a very bad upstream/not this PR API?

martin/src/config/file/tiles/duckdb/resolver/resolve.rs
```diff
@@ -0,0 +1,261 @@
...
+                Ok(source) => {
+                    info!(source.id = %source_id, "Configured DuckDB GeoParquet source");
+                    Ok(source)
+                }
+                Err(error) => Err(TileSourceWarning::SourceError {
+                    source_id,
+                    error: error.to_string(),
+                }),
+            });
+        }
```
CommanderStorm: okay.. that is a bit messy..

Can't we emit `Result<BoxedSource, TileSourceWarning>` and avoid all of this indexing, some-none, ...?

The reason why I am asking for this is that it is really confusing for the next reader and very easy to get wrong 😉

martin/src/config/file/tiles/duckdb/resolver/resolve.rs
```diff
@@ -0,0 +1,261 @@
...
+
+        let mut sources = Vec::new();
+        let mut warnings = Vec::new();
+        for result in ordered {
+            match result.expect("every DuckDB source index must be filled") {
+                Ok(source) => sources.push(source),
+                Err(warning) => warnings.push(warning),
+            }
+        }
+        Ok((sources, warnings))
```
CommanderStorm: so essentially, we do all of the above only to deconstruct it again?
Is there not a simpler way like directly clollecting to sources, warnings?

martin/src/config/file/tiles/duckdb/resolver/resolve.rs
```diff
@@ -0,0 +1,261 @@
...
+    }
+}
+
+#[cfg(test)]
```
CommanderStorm, line 119: Can we move these to the `integration_tests` crate insted?
They feel simpler to do e2e..

martin/src/config/file/tiles/duckdb/config.rs
```diff
@@ -351,4 +359,78 @@ sources:
...
+            parse_config(yaml, &HashMap::new(), Path::new("<test>")).expect("parse config");
+        let mut keys: Vec<_> = config.get_unrecognized_keys().into_iter().collect();
+        keys.sort_unstable();
+        insta::assert_debug_snapshot!(keys, @r#"
+        [
+            "duckdb.sources[0].bad_option",
+            "duckdb.typo_key",
+        ]
+        "#);
+    }
```
CommanderStorm: We already test this in our derive macro 😉

```suggestion
```

martin/src/test_support.rs
```diff
@@ -131,56 +131,9 @@ pub(crate) mod duckdb {
...
-            drop(conn);
-
-            DuckDBPool::new_database_file(id.to_string(), db_path, pool_size, None, None)
-                .expect("read-only DuckDB query pool")
-        }
+    /// Path to the committed polygon GeoParquet fixture shared by unit tests and the shell E2E.
+    pub(crate) fn polygons_parquet_path() -> PathBuf {
+        PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+            .join("../tests/fixtures/duckdb/geoparquet_polygons.parquet")
     }
```
CommanderStorm: Is this nessary? can't we just do it like for the others?

CommanderStorm (changes requested):
Mostly LGTM, but I think that moving to the new testing paradyme would de-convolute this code sooo much.

Could you do that?
Otherwise I can also do this after my exams next week

martin/src/config/file/main/lifecycle.rs
manbhav234: Yes, `resolve_files` should be better as it would allow auto discovery and would open path for hot reload later. Can refactor this in a PR stacked on top of this one.

manbhav234:
The top level config `geoparquet` key is a `String` with `location` key being typed. This was done to have better resolution into Local - PathBuf and Remote - Url which was not properly wired before e2e. Other issues are also resolved. 

Also, Thanks for moving the tests and cleaning up the cli for this !

manbhav234:
```
#[cfg(feature = "unstable-duckdb")]
if !self.duckdb.is_empty() {
    let val = self.duckdb.resolve(idr.clone(), self.cache.policy());
    sources_and_warnings.push(Box::pin(val));
}
```
Here, if we go with `resolve_files` to treat this like other file backends, it would require changing the config structure of duckdb to match what is required by `FileConfigEnum` which is likely too much of refactoring. Does it make sense to do this ? Or just implement a `*.parquet` directory scan in `finalize` ? The search directory being mentioned in `geoparquet:` config key.

CommanderStorm:
> ```
> #[cfg(feature = "unstable-duckdb")]
> if !self.duckdb.is_empty() {
>     let val = self.duckdb.resolve(idr.clone(), self.cache.policy());
>     sources_and_warnings.push(Box::pin(val));
> }
> ```
> 
> Here, if we go with `resolve_files` to treat this like other file backends, it would require changing the config structure of duckdb to match what is required by `FileConfigEnum` which is likely too much of refactoring. Does it make sense to do this ? Or just implement a `*.parquet` directory scan in `finalize` ? The search directory being mentioned in `geoparquet:` config key.

if it is so different we may need to revisit this. it is a bit weird if one part of the system is configured very differently..
We can do this in a followup or I can do it

justfile
```diff
@@ -615,6 +617,14 @@ test-pg: fetch start
 test-minio: fetch
     cargo test --features test-minio --no-default-features --test pmt_minio_test
 
+# Run DuckDB/GeoParquet tests only, including the end-to-end ones
+test-duckdb: fetch
+    cargo test -p martin --features test-duckdb --no-default-features --lib
+    cargo test -p martin-core --features unstable-duckdb --no-default-features --lib
+    cargo test -p martin-core --features unstable-duckdb --no-default-features --test duckdb_test
+    cargo build --package martin --no-default-features --features unstable-duckdb
+    cargo test --package martin-integration-tests --features duckdb --test duckdb
```
```suggestion
    cargo test --package martin-integration-tests --features duckdb --test test-duckdb
```

CommanderStorm (approved):
LGTM, thanks

justfile
```diff
@@ -615,6 +617,14 @@ test-pg: fetch start
 test-minio: fetch
     cargo test --features test-minio --no-default-features --test pmt_minio_test
 
+# Run DuckDB/GeoParquet tests only, including the end-to-end ones
+test-duckdb: fetch
+    cargo test -p martin --features test-duckdb --no-default-features --lib
+    cargo test -p martin-core --features unstable-duckdb --no-default-features --lib
+    cargo test -p martin-core --features unstable-duckdb --no-default-features --test duckdb_test
+    cargo build --package martin --no-default-features --features unstable-duckdb
+    cargo test --package martin-integration-tests --features duckdb --test test-duckdb
```
```suggestion
    cargo test --package martin-integration-tests --features test-duckdb --test test-duckdb
```

PR #3056, opened 2026-07-26
test: migrate style e2e tests to rust

CommanderStorm:
Moves the style.json serving and relative-URL expansion cases out of tests/test.sh into integration-tests/tests/styles.rs.


nyurik (approved):
please double check that we actually assert content everywhere, not just the fact that something was returned

PR #3059, opened 2026-07-26
chore(deps): Update hotpath and instrument decoders

pawurb:
Hi, I just published a new https://hotpath.rs/ release with IO instrumentation. 

It allows wrapping any IO stream, to collect metrics like bytes read/write, rate, perf percentiles etc. with minimal overhead.

I tested with:

```
oha -n 100000 -H 'Accept-Encoding: br'       http://localhost:3000/world_cities/0/0/0
```

and noticed that brotli compression is significantly slower.

Before:

before-oha
before-rate

After switching to compression level 4: 

after-oha
after-rate

~30x encoding and response time speedup. Brotli codec rate: `27.9kb/s -> 1.6 MB/s`

Not sure if this change makes sense for your project i.e. response speed vs compression ratio tradeoff. But maybe you'll find the new I/O instrumentation useful in other places. It should work for TCP, TLS, files I/O basically any types using `Read`, `Write`, `AsyncRead`, `AsyncWrite` traits.

I had to reorganize encoders usage, so that all data is captured, but final result should be identical.

As always feedback welcome!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3869,6 +3869,15 @@ version = "0.21.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "902f40dea4993d99db66732fddd9143e92657f1d47a642395f92b339b1375659"
 dependencies = [
+ "hotpath-macros 0.21.5",
+]
+
+[[package]]
+name = "hotpath"
+version = "0.22.0"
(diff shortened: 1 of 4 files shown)


CommanderStorm (approved):
👀 You are the best

PR #3061, opened 2026-07-26
perf: Change dynamic brotli encoding level from 11 to 4

pawurb:
Related to https://github.com/maplibre/martin/pull/3059 

It results in ~30x response time speedup for `content_encoding: br` requests, and preserves compression ratio for static content from `src/bindiff.rs`.

### DIFF
--- a/martin/src/srv/tiles/content.rs
+++ b/martin/src/srv/tiles/content.rs
@@ -651,13 +651,16 @@ impl<'a> DynTileSource<'a> {
     }
 }
 
+/// Brotli quality for on-the-fly response compression; q10+ is orders of magnitude slower.
+const BROTLI_ENCODE_QUALITY: u32 = 4;
+
 #[hotpath::measure]
 fn encode(tile: Tile, enc: ContentEncoding) -> ActixResult<Tile> {
     hotpath::dbg!("encode", enc);
     let etag = tile.etag;
     Ok(match enc {
         ContentEncoding::Brotli => Tile::new_with_etag(
-            encode_brotli(&tile.data)?,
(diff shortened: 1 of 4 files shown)


martin/src/srv/tiles/content.rs
```diff
@@ -651,13 +651,17 @@ impl<'a> DynTileSource<'a> {
     }
 }
 
+/// Brotli quality level for on-the-fly response compression; levels 10 and
+/// above are orders of magnitude slower to encode.
+const BROTLI_ENCODE_QUALITY: u32 = 4;
```
CommanderStorm: 4 is a kind of big dip in size.
We cache the output, so while we defintively should not be using 11, maybe moving up 1-2 levels is possible..

The +30% diff is not ideall.

Lets check what 6 is like

```suggestion
/// Brotli quality level for on-the-fly response compression
const BROTLI_ENCODE_QUALITY: u32 = 6;
```

martin/tests/mb_server_test.rs
```diff
@@ -295,7 +295,7 @@ async fn mbt_get_mvt_brotli() {
     );
     assert_eq!(response.headers().get(CONTENT_ENCODING).unwrap(), "br");
     let body = read_body(response).await;
-    assert_eq!(body.len(), 871); // this number could change if compression gets more optimized
+    assert_eq!(body.len(), 1136); // this number could change if compression settings change
```
CommanderStorm:
```suggestion
    assert_eq!(body.len(), 975); // this number could change if compression settings change
```

CommanderStorm (approved):
seems like 6 is better tradeoff (only +10% size, not +30%)

PR #3066, opened 2026-07-27
test: migrate the mbtiles copy, diff and inspection e2e tests to rust

johncarmack1984:
Ports the rest of the mbtiles utility group, picking up where #3064 stopped.

- `mbtiles_inspect.rs`: summary, meta-all, meta-get, validate (15 tests)
- `mbtiles_diff.rs`: cache schema, cache-purge, diff, bin-diff, apply-patch (12 tests)

Runs in ~0.1s. test.sh drops 894 → 715 lines, and `tests/expected/mbtiles`, `MBTILES_BIN`, `compare_sql_dbs` and the sqldiff tooling check go with it. A fresh bin-diff is now compared to the checked-in fixture directly, which is all sqldiff was for.

Based on #3065 for `MbtilesCli` and `mbtiles_fixture`; happy to rebase.

Three assertions turned out to be testing nothing:

- `summary.json` and `summary.pretty.json` were byte-identical after `jq --sort-keys`; the format difference was never checked
- `copy_diff2.txt` got overwritten by a later `copy --diff-with-file`, so `mbtiles diff` output never reached the comparison
- with sqlite3 missing, the apply test copied expected files into the output dir and passed without running anything. The plain-SQL apply needs no external binary, so it now always runs

That's the whole mbtiles utility group out of test.sh; nothing in the script touches the mbtiles binary anymore.


integration-tests/tests/mbtiles_diff.rs
```diff
@@ -0,0 +1,462 @@
...
+    assert!(output.contains("(cache)"), "{output}");
+
+    let summary = MbtilesCli::new("summary")
+        .arg("--format")
+        .arg("json")
+        .arg(&cache)
+        .run_json()
+        .await;
+    assert_eq!(summary["mbt_type"], "Cache");
+    assert_eq!(summary["tile_count"], 8);
```
CommanderStorm: Let's insts snapshot the summary

There are a few other cases that are similar

integration-tests/tests/mbtiles_diff.rs
```diff
@@ -0,0 +1,462 @@
...
+    assert!(
+        output.contains("Skipping per-tile hash validation because this is a cache MBTiles file"),
+        "{output}"
+    );
+    assert!(
+        output.contains("agg_tiles_hash has been verified"),
+        "{output}"
+    );
+}
+
```
CommanderStorm: .contains is almost always very brittle. 
Let's use insta, maybe with redactions. 

There are a few other cases of this

integration-tests/tests/mbtiles_inspect.rs
```diff
@@ -0,0 +1,348 @@
...
+    let source = mbtiles_fixture(dir.path(), "world_cities").await;
+
+    let summary = summary(&source).run_json().await;
+
+    assert_eq!(summary["mbt_type"], "Flat");
+    assert_eq!(summary["tile_count"], 8);
+    assert_eq!(summary["min_zoom"], 0);
+    assert_eq!(summary["max_zoom"], 6);
+    assert_eq!(summary["min_tile_size"], 20);
+    assert_eq!(summary["max_tile_size"], 1107);
```
CommanderStorm: Insta ;)

CommanderStorm (changes requested):
In a few cases, I would prefer insta, such as instead of .contains an such. 
The reason behind this is that it is both a tighter, better debuggable error and also if something needs updating I can just apply the bless label to the pr and everything will autoapply ;)

CommanderStorm:
Other than that, thank you for working on this, LGTM.

integration-tests/tests/mbtiles_inspect.rs
johncarmack1984: Done in f0cd2bc!

PR #3067, opened 2026-07-27
test: migrate the postgres source discovery e2e tests to rust

johncarmack1984:
Ports the Postgres half of the `Test auto configured Martin` group into `integration-tests/tests/postgres.rs`, 11 tests behind the `test-pg` feature: what auto-discovery publishes, and what those sources serve.

Runs in ~2.7s. test.sh drops 894 → 810 lines, `tests/expected/auto` 149 → 30 files.

Based on #3065, which already has everything this needs except one method: `MartinBuilder::with_postgres()`, the server-side twin of your `MartinCp::with_postgres`. Happy to rebase.

These servers ask for `--pool-size 2` because the default of 20 exhausts `max_connections` once they run in parallel.

The dumps come from `mlt_core`'s reader now, not the `mvt` CLI, so the text isn't byte-identical to the old `.mvt.txt`. Confirmed 40 against the files they replace: same layers, same feature counts, same geometries.

Floats get the same ten-digit treatment test.sh gave them, through insta filters. The tightest reprojected value sits 305 ULPs off a truncation boundary; glibc and Apple's libm disagree by about 1.

Tile geometry isn't stable across PostGIS releases. Against `db-legacy` (13-3.0) seven of eleven fail: layer names and feature counts identical, but 3.0 gives a 39-vertex linestring where 3.6 gives 15, and points come back in a different order.

So nothing calls these yet. `test-e2e` runs the crate without `test-pg`, so they sit out; `test-e2e-pg` would pick them up with one more `--test`, but it runs under `test-pg` against three PostGIS versions and would be red on two. Your call where they belong.

Green on macOS and aarch64 Linux. clippy pedantic, rustfmt and `--unreferenced=reject` clean.

Next PRs would be the configured and postprocessing groups.


integration-tests/tests/postgres.rs
```diff
@@ -0,0 +1,365 @@
...
+    assert_eq!(
+        response.header("content-type"),
+        Some("application/x-protobuf")
+    );
+    response.mvt_dump()
+}
+
+/// Ask `source` for every zoom level in [`ZOOMS`] and snapshot each tile.
+async fn assert_tiles_across_zooms(martin: &Martin, source: &str, snapshot_prefix: &str) {
+    for zxy in ZOOMS {
```
CommanderStorm: I'd just inline this, seems simpler and more readable

```suggestion
    for zxy in [
        "0/0/0",
        "6/57/29",
        "12/3673/1911",
        "13/7346/3822",
        "14/14692/7645",
        "17/117542/61161",
        "18/235085/122323",
    ] {
```

integration-tests/tests/postgres.rs
```diff
@@ -0,0 +1,365 @@
...
+fn json_filters() -> Vec<(&'static str, &'static str)> {
+    vec![
+        (r"(-?\d+\.\d{10})\d+", "$1"),
+        (r"(?m)^etag: .*$", "etag: [ETAG]"),
+    ]
+}
+
+/// The warnings auto-discovery emits for this database however it is started: three tables that
+/// have no spatial index, and the six ids it had to rename to keep them unique and printable.
+fn assert_discovery_warnings(martin: &mut Martin) {
```
CommanderStorm: I think can be trimmed 😉 

```suggestion
//! Sources martin discovers by itself in a `PostgreSQL` database

#![cfg(feature = "test-pg")]

use std::io::Cursor;

use martin_integration_tests::{Martin, TestResponse};
use serde_json::Value;

/// A server that publishes everything it finds in the fixture database.
async fn martin_with_postgres() -> Martin {
    Martin::builder()
        .with_postgres()
        // also adopt tables whose geometry column has SRID 0
        .arg("--default-srid") 
        .arg("900913")
        // no fuzzy estimated bounds
        .arg("--auto-bounds")
        .arg("calc")
        // to not exhaust the global pool
        .arg("--pool-size")
        .arg("1")
        .start()
        .await
        .expect("failed to start martin")
}

fn json_filters() -> Vec<(&'static str, &'static str)> {
    vec![
        (r"(-?\d+\.\d{10})\d+", "$1"),
        (r"(?m)^etag: .*$", "etag: [ETAG]"),
    ]
}

fn assert_discovery_warnings(martin: &mut Martin) {
```

CommanderStorm (approved):
Thanks (I am going to do 2 minor slimmenings on the comments though 😉)

PR #3078, opened 2026-07-28
fix: install configuration in /etc for debian package

### LINKED ISSUE #3076: `debsums` complains about changed configuration file for Debian package based installation
In Debian there is a tool called `debsums` which checks package files for integrity. Configuration files are generally excluded from this check, as they are supposed to be changed. But `debsums` complains when changing the file `/usr/local/etc/martin/config.yaml` as `/usr/local/etc` is not a common configuration directory.

```
debsums: changed file /usr/local/etc/martin/config.yaml (from martin package)
```

To solve this issue I would like to suggest the two following methods:

1. Install the configuration file in `/etc/martin/config.yaml` (recommended)
2. Flag `/usr/local/etc/martin/config.yaml` as a configuration file during the package build. See https://www.debian.org/doc/manuals/maint-guide/dother.en.html#conffiles for details.

Many thanks in advance.

hamarituc:
closes #3076

### DIFF
--- a/debian/martin.service
+++ b/debian/martin.service
@@ -2,7 +2,7 @@
 Description = A blazing fast map tile server which can generate and serve vector tiles on the fly from large PostGIS databases, PMTile, and MBTile files
 
 [Service]
-ExecStart = /usr/bin/martin --config /usr/local/etc/martin/config.yaml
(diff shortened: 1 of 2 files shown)


CommanderStorm (approved):
Sure if you think this is better

PR #3104, opened 2026-08-05
test: move the remaining rendering tests to e2e and drop mitmproxy

CommanderStorm:
The in-process render tests become HTTP-level ones on the integrated cassette, so the mitmproxy replay proxy can go.


integration-tests/src/cassette.rs
```diff
@@ -252,7 +281,7 @@ async fn fetch(client: &Client, url: &str) -> Option<Vec<u8>> {
     Some(decompress(&body, encoding.as_deref()))
 }
 
-/// The recordings under `dir`, keyed by the URL path each answers.
+/// The recordings under `dir`, keyed by the URL path each answers, `` for [`INDEX`] files.
```
CommanderStorm:
```suggestion
/// The recordings under `dir`, keyed by the URL path each answers, or empty for `INDEX`
```

PR #3105, opened 2026-08-10
chore(test): migrate the save-config e2e tests to rust and retire tests/test.sh

### LINKED ISSUE #3048: Replace tests/test.sh with a Rust test harness?

CommanderStorm:
The last two `tests/test.sh` groups become snapshot tests, and the script that ran them is gone: the justfile and CI now drive the rust e2e suite against the docker images, the debian package and the release binaries.


integration-tests/tests/save_config.rs
```diff
@@ -0,0 +1,107 @@
+//! What `--save-config` writes down: the sources and resources martin found by itself, and the
+//! settings it read from a config file.
+
+use std::fs;
+
+use martin_integration_tests::{Martin, MartinBuilder};
+use tempfile::TempDir;
+
+/// A config that sets every documented section outside of `postgres`, which
+/// `postgres.rs` covers with a database at hand.
```
CommanderStorm:
```suggestion
use std::fs;

use martin_integration_tests::{Martin, MartinBuilder};
use tempfile::TempDir;

```

.github/workflows/ci.yml
```diff
@@ -352,39 +349,39 @@ jobs:
           labels: ${{ steps.docker_meta.outputs.labels }}
           platforms: linux/arm64,linux/amd64
 
-      - name: Test linux/arm64 Docker image
+      - name: End-to-end tests against the linux/arm64 Docker image
```
michaelkirk: nit: this appears in the GH sidebar right? IIRC a shorter name is nice, but this is a very minor ignorable nitpick.

e2e-tests/tests/pmtiles.rs
```diff
@@ -52,10 +50,12 @@ async fn auto_configured_minimal() {
...
         .remove("settings")
         .expect("catalog has a settings key");
-    #[cfg(target_os = "linux")]
-    assert_eq!(settings, serde_json::json!({ "rendering": false }));
-    #[cfg(not(target_os = "linux"))]
-    assert_eq!(settings, serde_json::json!({}));
+    // `rendering` is reported by a binary compiled with the feature, which the debian package
+    // and the other packaged builds each decide for themselves.
+    assert!(
+        settings == serde_json::json!({}) || settings == serde_json::json!({"rendering": false}),
```
michaelkirk, line 56: The previous test was more specific - was it intentionally loosened? It seems like not a very interesting test at this point.

CLAUDE.md
```diff
@@ -21,8 +21,8 @@ just test
 cargo test --workspace
 # PG tests (needs `just start`)
 just test-pg
-# integration tests (needs docker)
-just test-int
+# end-to-end tests driving the built binaries (needs docker)
```
michaelkirk: Double checking - do we actually still need docker for these tests?

michaelkirk (approved):
LGTM - some minor things which you can probably ignore. 

So long pile of shell scripts. 🫡

johncarmack1984:
I've been keeping track of all of these e2e PRs, ready to pitch in if it looked like one test or another would be a sticky migration, and honestly I'm floored at how quickly and steadily you got all these moved over to the new harness. Hats off @CommanderStorm.

.github/workflows/ci.yml
CommanderStorm:
```suggestion
      - name: Test linux/arm64 Docker image
```

e2e-tests/tests/geojson.rs
johncarmack1984: Looks like Windows is red on this one. The discovered geojson paths come out as tests/fixtures/geojson\bare_geometry.geojson. Can use the same trick the cassette walker uses:

```suggestion
    let saved = fs::read_to_string(&save_config)
        .expect("martin did not write --save-config")
        .replace(std::path::MAIN_SEPARATOR, "/");
```

e2e-tests/tests/pmtiles.rs
CommanderStorm: some of the tests run against this have rendering not compiled in -> {} and some do.. not sure if this is optimal, but that is the current way...

PR #3106, opened 2026-08-10
test: retire tests/test.sh

CommanderStorm:
Deletes the shell harness and points the justfile and CI at the rust e2e suite instead, including running it against the docker images, the debian package and the release binaries.


CommanderStorm:
Folded into #3105: removing the two groups these tests replace empties tests/test.sh, so the script and its justfile/CI callers have to go in the same change.

PR #3130, opened 2026-08-18
refactor(postgres): make the reloader the single writer of PostgreSQL sources

### LINKED ISSUE #2862: Single-writer tile-source catalog: eliminate the startup-vs-reload double-discovery race
(claude generated, to remind me to do this refactoring)

## Problem

At startup, reloadable sources are written to the catalog **twice from two independent database observations**: synchronously by `config.resolve()`, then again by the reloader's background `seed()`. A `DROP` between the two desyncs them permanently — `resolve()` publishes `roads`, the seed's later discovery omits it, and since every reconcile diffs next-vs-baseline (neither containing `roads`), no removal is ever emitted. The source 500s/serves stale until restart. The add-side of the same race is papered over with a 3 s sleep in tests; the drop-side has no workaround. All four reloaders share the mechanism.

## Solution

Make the reloader the **single writer** of reloadable sources, and the `TileSourceManager` (TSM) the one source of truth for both serving and `--save-config`. The reloader populates the TSM via a synchronous `init()` (reconcile-from-empty, awaited before serving), then maintains it via its existing loop. `config.resolve()` stops resolving tile sources and no longer mutates `Config`. One observation sequence ⇒ the race is impossible by construction and the `Baseline` concept disappears. No new orchestration layer — entry points call `init()` inline before serving.

## User Stories

1. As an operator, I want a table dropped during startup reconciled away on the next poll, so I never serve ghosts.
2. As an operator, I want a table created during startup published on the next poll, so timing never swallows a source.
3. As an operator, I want `/catalog` to reflect one consistent DB view, never an artifact of two discoveries disagreeing.
4. As an operator, I want `on_invalid: abort` to still fail startup on a bad initial source, but only `warn`+skip on a bad *live* reload, so a runtime change can't crash a healthy server.
5. As an operator, I want `--save-config` to serialize the sources actually in the catalog, so the file matches the running server.
6. As an operator, I want sources available the instant serving starts, so there's no post-boot 404 window.
7. As an operator with multiple connections, I want each reloader to own only its sources, so reloads don't cross-interfere.
8. As an operator, I want statically-configured single-file/remote sources unchanged, so the refactor only touches reloadable ones.
9. As a developer, I want the reloader to be the single deep module owning `init()` + reconcile (no `Baseline`, no seed branch, one diff/apply path), so startup and reload can't drift.
10. As a developer, I want the 3 s test sleep deleted and `martin-cp`/other entry points still working.

## Implementation Decisions

- **TSM is the single source of truth**, populated only by reloaders; it retains each source's resolved provenance (spec/config) so config can be reconstructed from it.
- **`config.resolve()` shrinks** to caches/fonts/sprites/styles, yielding an **empty** TSM (carrying only tile cache + `on_invalid`); it no longer discovers, instantiates, or mutates `Config`.
- **`TileSourceReloader::init()`** does the first reconcile from empty (additions only), applies into the shared TSM, and returns discovery warnings. `spawn()` then runs the ongoing loop, diffing against last-applied state (= TSM contents for its sources).
- **`Baseline` and `seed()` deleted** — startup is just reconcile-from-empty.
- **`OnInvalid` seam**: `init()` treats discovery warnings as abortable (today's `handle_tile_warnings`); the live loop stays warn-only. Per-source build failures already handled in the apply path.
- **`--save-config` reads the TSM** (after `init()`), not a `resolve()` mutation.
- **File static/reloadable split**: single-file/remote sources stay resolved-once by the file path; only directory/prefix sources move into `init()`.
- **Startup sequence** (inline in entry points): resolve non-tile + empty TSM → build reloaders → `init()` each → apply `on_invalid` → (`--save-config`) → `spawn()` → serve.

## Testing Decisions

Test external behavior (what the catalog serves, what a saved config contains), not call sequencing. Mandated:

- **PG create/alter/drop e2e (HTTP-level):** extend the existing reload integration test; **delete the 3 s sleep** and assert a source dropped around the startup boundary cannot linger.
- **`--save-config` writeback from TSM:** assert the saved config materializes the same `tables`/`functions` and file entries actually being served. Prior art: existing resolved-config snapshot tests.

`ReloadDriver` unit tests are updated through the `Baseline` removal but aren't the load-bearing proof.

## Out of Scope

Standalone orchestrator abstraction; wiring file kinds through `init()` (likely follow-up — PG is the urgent bug); object-store prefix polling semantics; tile-serving/cache/SQL changes; transactional snapshot discovery; CHANGELOG.

## System Learnings (verified)

- **The race is temporal, not pool-related** — two `discover()` calls at different times see different DB states regardless of pool.
- **`fingerprint()` excludes `bounds`**, so `instantiate()`'s bounds write-back doesn't flip versions; an idle re-discover is stable (an `init()`-from-empty baseline stays stable across polls).
- **`OnInvalid` is already two-layered**: build failures in `apply_changes`; discovery warnings via `handle_tile_warnings` (only `resolve()` calls it today — the new `init()` must). Startup-abort vs live-warn is the one new behavioral seam.
- **`--save-config` can't be reconstructed from the config writeback** — function fingerprints need `PostgresSqlInfo` (`sql_query`/`signature`), which `FuncInfoSources` drops; hence reading from the provenance-retaining TSM.
- **File path is asymmetric**: `FsDiscovery` only re-lists directory sources (remote single-files explicitly skipped), driving the static/reloadable split. PMTiles' remote `Empty` path is already conceptually `init()`-from-empty.
- **`TileSourceManager` = `Arc<DashMap<id, (BoxedSource, ProcessConfig)>>`** today — adding retained provenance is the concrete enabler for `--save-config`-from-TSM.

johncarmack1984:
Fixes #2862 (Postgres half).

@CommanderStorm, your issue already had the design and the test plan, so I mostly just followed it: `PostgresReloader::init()` reconciles from empty before we serve, `start()` polls against that same observation, `PostgresConfig::resolve()` is gone, `--save-config` reads the catalog, and `martin-cp` goes through the same `init()`. The 3s sleep is dead.

Per your request this got split. #3143 (discovery warnings) and #3142 (bounded concurrent advisory builds) are merged; #3144 (the per-source `convert_to_*` fix) is open and rides along here as the first commit until it lands. What's left in this PR is the lifecycle change itself.

Two places it still goes past the spec:

- `BuiltSource` carries provenance and the catalog retains it: that's what lets `--save-config` serialize what's actually served instead of a `resolve()` writeback.
- File kinds stay on `Baseline::StartupResolved`, per your out-of-scope note. `Baseline::Initialized` marks the `init()`-fed driver until they move over and `Baseline`/`seed()` can go. Happy to do that as the follow-up if you want it.

Tests: `pg_reload_test` is now HTTP-level with no fixed sleep, including a table dropped between `init()` and the first poll. The driver e2e covers the same boundary, `reconcile.rs` has `init()` unit tests, and the architecture doc's reload section is updated.

Objections welcome, as always. It was a fun one :)


CommanderStorm:
this PR is pretty massitve given the importance of the code you are working on.

Before I review it, are there any refactorings that you can/want to pull into another PR (so we can merge them)

johncarmack1984:
Yeah, I'll pare it down. Converting this to draft and splitting out two sets of changes that can be merged in parallel as precursors.

johncarmack1984:
Added a third precursor in #3144 that relied on #3142. Once merged and this branch is rebased, this PR is down to just the lifecycle change. Will be about 14 files changed and around 640 lines.

docs/content/architecture.md
```diff
@@ -568,9 +569,9 @@ Read it when you're curious **why** certain choices were made.
 
     They interact in this loop:
 
-    1. **Seed**.
-       Run Discovery once to record a baseline, applying nothing.
-       The catalog was already populated during initialization, so applying would double-add.
+    1. **Init or seed**.
+       PostgreSQL drivers `init()`: run Discovery once, apply everything, record it as the baseline before serving.
+       File-based kinds are loaded during configuration resolution, so their drivers only seed: record a baseline, applying nothing.
```
CommanderStorm:
```suggestion
       `fs` sources loaded during configuration resolution, so their drivers only seed (record a baseline, apply nothing).
```

martin/src/config/file/tiles/reload/postgres.rs
```diff
@@ -1,30 +1,37 @@
-//! [`PostgresReloader`]: polls `PostgreSQL` catalog discovery and applies the diff at runtime.
+//! [`PostgresReloader`]: the single writer of one `PostgreSQL` connection's sources - it loads
+//! them at startup via [`init`](PostgresReloader::init) and keeps them current by polling.
```
CommanderStorm:
```suggestion
//! [`PostgresReloader`] writes the `PostgreSQL` connection's sources by
//! - loading them at startup via [`init`](PostgresReloader::init) and
//! - keeps them current by polling.
```

martin/src/config/file/tiles/reload/postgres.rs
```diff
@@ -1,30 +1,37 @@
...
 /// Reloader for `PostgreSQL` sources.
 ///
-/// `PostgreSQL` has no change-notification channel Martin listens to, so this re-runs catalog
-/// discovery on a fixed [`PollTrigger`] interval and applies the diff (adds, updates, removals)
-/// to the [`TileSourceManager`]. A `reload_interval` of `0s` disables it.
+/// [`init`](Self::init) publishes everything the catalog discovers into the [`TileSourceManager`]
+/// before serving starts. `PostgreSQL` has no change-notification channel Martin listens to, so
+/// [`start`](Self::start) then re-runs discovery on a fixed [`PollTrigger`] interval and applies
+/// the diff (adds, updates, removals). A `reload_interval` of `0s` disables the polling.
 pub struct PostgresReloader {
```
CommanderStorm:
```suggestion
/// [`init`](Self::init) publishes everything the catalog discovers into the [`TileSourceManager`] before serving starts.
/// `PostgreSQL` has no change-notification channel Martin could listens to, so [`start`](Self::start) then re-runs discovery on a fixed [`PollTrigger`] interval.
/// The diff (adds, updates, removals) gets applied.
/// A `reload_interval` of `0s` disables polling.
pub struct PostgresReloader {
```

martin/src/config/file/tiles/reload/postgres.rs
```diff
@@ -49,27 +56,63 @@ impl PostgresReloader {
...
         Self {
-            tile_source_manager: tsm,
-            discovery,
+            driver: ReloadDriver::new(discovery, tsm),
         }
     }
 
+    /// Publishes every discovered source into the catalog and returns the discovery warnings for
+    /// the caller's `on_invalid` policy. Fails if the database is unreachable or, under
+    /// `on_invalid: abort`, if any source fails to build.
```
CommanderStorm:
```suggestion
    /// Publishes every discovered source into the catalog and returns the discovery warnings.
    /// Fails if the database is unreachable or, under `on_invalid: abort`, if any source fails to build.
```

martin/src/reload.rs
```diff
@@ -4,8 +4,21 @@ use futures::stream::{self, StreamExt as _};
...
+/// Where a catalog entry came from, in enough detail to write it back to a config file.
+#[derive(Clone, Debug)]
+pub enum SourceProvenance {
+    #[cfg(feature = "postgres")]
+    Postgres {
+        /// The `connection_string` of the `postgres` config entry that discovered it.
+        connection_string: String,
+        spec: SourceSpec,
+    },
+}
```
CommanderStorm, line 20: yea, what the hell?

I am not sure if this is right, seems wrong.
But it works for now.
Seems like a wierd arch.

CommanderStorm (approved):

martin/src/reload.rs
johncarmack1984: Fair. It's there so `--save-config` can write back what the catalog is actually serving without a second `resolve()` pass, and the enum is single-variant because only PostgreSQL discovers sources at runtime today. If the file kinds move onto `init()` too (the `Baseline` follow-up), provenance becomes just "which config entry built this", and it could collapse to a `String`. Happy to take it that way now if you'd rather.

PR #3131, opened 2026-08-18
perf(pmtiles): open file sources concurrently and share one client per store

### LINKED ISSUE #3122: Reduce startup time with thousands of remote PMTiles sources
### Use case

- I have a configuration with 4,000+ PMTiles sources, where each PMTiles file is hosted remotely in S3.
- I'm currently using the Docker image v1.13.0.
- With this configuration, starting Tile Server takes approximately 20 minutes on my MacBook Air M5.

### Problem
- The large number of PMTiles sources makes the initial startup extremely slow.
- For my use case, I don't necessarily need all 4,000+ sources to be fully initialized or validated before the server starts accepting requests.
- It would be useful if TileServer could start quickly and initialize remote PMTiles sources lazily, as they are first accessed.
- I am running the Tile Server in ECR with load balancer and it is not possible for auto scale up, since startup time is too long.

johncarmack1984:
tl;dr: 1,000 remote PMTiles reduced from 95s to 3.7s.

Refs #2937, Fixes #3122.

@CommanderStorm you said feel free to look into where the time goes, so I did. I put 1,000 PMTiles on a local MinIO behind toxiproxy with 40ms of added latency, which is the only way to see this on a laptop: on plain localhost the same 1,000 sources start in 3s and everything looks fine.

Where it goes: startup opens file sources one at a time, and each remote PMTiles open is a few dependent round trips (header, directory, metadata) on a fresh S3 client. So it's latency x source count, whether the sources are listed under `sources:` or discovered from an `s3://bucket/` prefix (the reloader's first tick builds serially too). @mazedm80's 20 minutes for 4,000 sources at real S3 latency lines up with that.

Two changes, they go together:

- Sources are opened with a bounded fan-out (`MAX_CONCURRENT_SOURCE_INITS = 32`) in `resolve_files` and in `ReloadAdvisory` builds. Ids, duplicate handling, and `--save-config` output resolve in the same order as before; only the open is concurrent. That's the 95s -> 3.7s, both paths.
- `PmtConfig` keeps one `ObjectStore` per store root (bucket or host), shared by every source under it. Concurrency alone was a bad idea on its own: with per-source clients I saw 1,000 idle keep-alive sockets held open (main peaks at 339 as they trickle out; 4,000 sources would sail past a default `ulimit`). Shared clients: 32 sockets, and RSS 36MB instead of 76MB on main.

Why 32: enough to hide latency at any realistic RTT, small enough that a big config doesn't look like a burst to S3. Happy to make it a knob if you'd rather, I just couldn't think of a WHY someone would need to turn it.

Not done here: the lazy-init idea from the issue (the catalog needs each source's TileJSON, so something has to open them; making that fast seemed better than making it later), and recursive local discovery from #2937, which I'd like to do next once we settle the source-ID mapping you mentioned. I'll open that conversation on the discussion.

Tests: a unit test pins the store sharing (same bucket/host -> one client, different -> two); the existing file-source and `--save-config` snapshot suites cover the ordering. Local `test-minio` green.

### DIFF
--- a/martin/src/config/file/file_config.rs
+++ b/martin/src/config/file/file_config.rs
@@ -9,6 +9,8 @@ use std::path::Path;
 use std::path::PathBuf;
 use std::time::Duration;
 
+#[cfg(feature = "_tiles")]
+use futures::stream::{self, StreamExt as _};
 pub use martin_config_macros::ConfigurationLivecycleHooks;
 use martin_core::CacheZoomRange;
 #[cfg(feature = "_tiles")]
@@ -454,9 +456,19 @@ async fn resolve_int<T: TileSourceConfiguration>(
         }
     }
 
+    let custom = &cfg.custom;
+    let opened = stream::iter(planned)
+        .map(|p| async move {
+            let result = p.open(custom).await;
+            (p, result)
+        })
+        .buffered(MAX_CONCURRENT_SOURCE_INITS)
+        .collect::<Vec<_>>()
+        .await;
+
     let mut results = Vec::new();
-    for p in planned {
-        match p.open(&cfg.custom).await {
+    for (p, result) in opened {
+        match result {
             Ok(src) => {
                 p.log_configured();
                 if !p.from_sources
--- a/martin/src/config/file/tiles/pmtiles.rs
+++ b/martin/src/config/file/tiles/pmtiles.rs
@@ -10,9 +10,14 @@ use aws_config::profile::ProfileFileRegionProvider;
 use aws_credential_types::provider::{ProvideCredentials as _, SharedCredentialsProvider};
 #[cfg(test)]
 use aws_runtime::env_config::file::EnvConfigFiles;
+use dashmap::DashMap;
 use martin_core::tiles::BoxedSource;
 use martin_core::tiles::pmtiles::{PmtCache, PmtCacheInstance, PmtilesSource};
 use object_store::aws::{AmazonS3Builder, AwsCredential, AwsCredentialProvider};
+use object_store::azure::MicrosoftAzureBuilder;
+use object_store::client::{ClientOptions, HttpClient, HttpConnector, ReqwestConnector};
+use object_store::gcp::GoogleCloudStorageBuilder;
+use object_store::http::HttpBuilder;
 use object_store::{CredentialProvider, ObjectStore, ObjectStoreScheme};
 use serde::{Deserialize, Serialize};
 use tracing::{trace, warn};
@@ -116,6 +121,11 @@ pub struct PmtConfig {
     #[cfg_attr(feature = "unstable-schemas", schemars(skip))]
     pub aws_credentials: Option<AwsCredentialProvider>,
 
+    /// HTTP clients shared by every remote source of this config (internal state, not serialized)
+    #[serde(skip)]
+    #[cfg_attr(feature = "unstable-schemas", schemars(skip))]
+    pub(crate) http_clients: SharedHttpClients,
+
     #[cfg(test)]
     #[serde(skip)]
     #[cfg_attr(feature = "unstable-schemas", schemars(skip))]
@@ -136,6 +146,7 @@ impl Default for PmtConfig {
             unrecognized: UnrecognizedValues::default(),
             pmtiles_directory_cache: PmtCache::default(),
             aws_credentials: None,
+            http_clients: SharedHttpClients::default(),
             #[cfg(test)]
             aws_profile_files: None,
         }
@@ -153,7 +164,7 @@ impl PartialEq for PmtConfig {
         let base = base
             && self.convert_to_mlt == other.convert_to_mlt
             && self.convert_to_mvt == other.convert_to_mvt;
-        // pmtiles_directory_cache is intentionally excluded from equality check
+        // pmtiles_directory_cache and http_clients are intentionally excluded from equality check
         base
     }
 }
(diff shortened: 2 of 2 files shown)


mazedm80:
Thanks  johncarmack1984, once the PR approved I would like to test it.

martin/src/config/file/tiles/pmtiles.rs
```diff
@@ -117,6 +119,12 @@ pub struct PmtConfig {
     #[cfg_attr(feature = "unstable-schemas", schemars(skip))]
     pub aws_credentials: Option<AwsCredentialProvider>,
 
+    /// One client per store root (bucket / host), shared by every source under it so
+    /// connections are reused instead of each source opening its own.
+    #[serde(skip)]
+    #[cfg_attr(feature = "unstable-schemas", schemars(skip))]
+    pub(crate) stores: Arc<DashMap<String, Arc<dyn ObjectStore>>>,
```
CommanderStorm: Links to all object stores being added to every pmtiles config feels rather hacky. Can you think of a better place to put this?

martin/src/config/file/tiles/pmtiles.rs
```diff
@@ -265,26 +274,38 @@ impl PmtConfig {
...
-        if scheme != ObjectStoreScheme::AmazonS3 {
-            return object_store::parse_url_opts(url, &self.options);
+        let root = url
+            .as_str()
+            .strip_suffix(path.as_ref())
+            .unwrap_or(url.as_str())
+            .to_owned();
+        if let Some(store) = self.stores.get(&root) {
+            return Ok((Box::new(Arc::clone(&*store)), path));
         }
```
CommanderStorm: lets hypothetically say we have two AWS buckets side by side.
They have the same URL, but a different path and different access perrmisisons required.

I think this way of keying on the url and discarding the path is not very robust regarding this..
Are you SURE that this is actually correct and does not break things?

martin/src/config/file/file_config.rs
```diff
@@ -392,6 +394,12 @@ pub async fn resolve_files<T: TileSourceConfiguration>(
     resolve_int(config, idr, extension, default_cache).await
 }
 
+/// How many tile sources are opened at once at startup and on reload. Opening a remote source is
+/// a few dependent round trips, so serial opens cost latency times source count; this bounds the
+/// fan-out so thousands of sources do not mean thousands of simultaneous requests.
+#[cfg(feature = "_tiles")]
+pub const MAX_CONCURRENT_SOURCE_INITS: usize = 32;
```
CommanderStorm: A limit maybe.. but 32 seems rather low..
would this not make the problem worse rather than better?

martin/src/config/file/file_config.rs
```diff
@@ -464,73 +502,140 @@ async fn resolve_int<T: TileSourceConfiguration>(
...
 
-/// Resolves a single tile source configuration and returns a boxed source for further processing.
-///
-/// This function processes a tile source configuration using a given custom implementation of
-/// `TileSourceConfiguration` and resolves its ID using `IdResolver`.
-/// It determines if the source is a URL or a file path, configures the source accordingly.
 #[cfg(feature = "_tiles")]
-async fn resolve_one_source_int<T: TileSourceConfiguration>(
-    custom: &T,
+enum Target {
```
CommanderStorm, line 491: this seems like a good refactoring, can you break this out into another PR (away from the perf-relevant changes).

Not sure about `from_paths`, that seems rather suspicious.

CommanderStorm (commented):
not entirely sure if the three changes here are entirely correct. Would be simpler to review if there was only one, focused change in the PR

martin/src/config/file/tiles/pmtiles.rs
johncarmack1984: Definitely, it's gone. What's there now is one `SharedHttpClients` next to `pmtiles_directory_cache`, same kind of per-config runtime state, handed to every builder through `with_http_connector`. Stores stay per source.

martin/src/config/file/tiles/pmtiles.rs
johncarmack1984: Fair, updated in latest push. Nothing is keyed on the URL now. Each source builds its own store from its own URL and credentials exactly like main; the only thing shared is the HTTP client under it, and `object_store` asks for that per `ClientOptions`, so the S3 metadata client (different timeout, `allow_http`) gets its own too. Test covers s3, https, gs, az, and mixed.

martin/src/config/file/file_config.rs
johncarmack1984: On main it's currently 1, I was worried 32 was high 😅. Happy to go to 64, or make it a config knob if you'd rather; I just couldn't think of a clear use case for the knob. 

Each open is about three dependent round trips, so at 40 ms, 32 in flight puts 1,000 sources at 3.3 s and 4,000 at roughly four times that. Unbounded is faster still, but that's 4,000 sockets from one box at once and S3 throttles per-prefix bursts.

martin/src/config/file/file_config.rs
johncarmack1984: Done: #3133. `from_paths` was the same bit as `!from_sources`, so it's gone there.

johncarmack1984:
Split as asked: the refactor is #3133 on its own, and this branch now sits on top of it as two commits (share the HTTP clients, then the bounded fan-out). Until #3133 merges the diff here shows all three, so the two perf commits are the ones to look at.

The per-root store map is gone. `object_store` has a hook for exactly this (`with_http_connector`), so each source still builds its own store from its own URL and credentials, same as main, and only the reqwest client underneath is shared. Nothing is keyed on the URL anymore. Re-measured: 1,000 sources on the 40 ms rig open 33 sockets instead of 343, RSS 40 MB instead of 77 MB, startup 3.3 s vs 91 s.

CommanderStorm:
will not review today, am a bit tired

johncarmack1984:
No worries, it's pro bono open source anyway, if it doesn't make your life easier it isn't worth it. Thanks for the feedback yesterday and happy to make any other changes when you've got the time to review!

johncarmack1984:
Commit 94190fe fixes a lint that popped up in Rust stable about the 3 hours after this morning's merge, with Rust 1.98's release. Happy to split it into a separate PR, it's just the single commit, in a few minutes we'll know for sure whether it fixed this PR's CI.

CommanderStorm:
@johncarmack1984 you closed the PR. intentional?

johncarmack1984:
Intentional but overthought. Didn't want PRs piling up in your queue and wasn't sure I'd have time to implement changes over the next few days, but found some time this morning and finished the work I was worried I'd have to prioritize next week, so... it's open again. :) along with a PR for the docs build.

CommanderStorm (approved):
LGTM, thanks

PR #3139, opened 2026-08-22
docs(postprocesing): refactor docs to accept the new postprocessing section

CommanderStorm:
moves things around to accept the expanded section

PR #3140, opened 2026-08-22
docs(postprocessing): remove dead link to the not-yet-written hillshade guide

johncarmack1984:
The postprocessing index from #3139 links to `hillshade.md`, which doesn't exist yet, so the markdown link check fails on main and on every branch that contains it. This drops the link and keeps the sentence; when the hillshade guide lands, the link can come back with it.


CommanderStorm:
tnx

PR #3141, opened 2026-08-22
feat(config)!: stop reading configuration from the environment

### LINKED ISSUE #1052: Remove automatic ENV var usage except inside config files
Per discussion in #1050, we should scale down on Martin's implicit usage of the environment variables. This is too unpredictable and could result in all sorts of weird behavior. Instead, Martin should only rely on two things:  command line parameters and config files.  The config files will continue supporting explicit environment variables, e.g. `connection_string: ${MY_DATABASE_URL}` or even with a fallback like `connection_string: ${MY_DATABASE_URL:-postgresql://postgres@localhost/db}`

### Tasks

* [ ] Ensure ALL [env vars](https://maplibre.org/martin/env-vars.html) can be set at least with the config file, or best -- with both config file and CLI parameters. (I think they already are, but double check)
* [ ] Detect if any of [these env vars](https://maplibre.org/martin/env-vars.html) are set, and if they are, show a warning with clear instructions on doing it with either CLI parameters or with config (or both, or a link to docs).
  * For example, if `DATABASE_URL` is set, suggest to use `martin "$DATABASE_URL"` instead, or add `postgres.connection_string: ${DATABASE_URL}` to the config file.
* [ ] Remove any magical handling of the env vars from the code (except for the above detection and warnings)
* [ ] Update documentation

### Alternatives
Instead of removing, we could follow other Rust tools pattern, and rename all environment variables to be `MARTIN_*`, e.g. `MARTIN_DATABASE_URL`. In my opinion, we should still disable the existing ones as they create too many conflicts with other tooling that relies on the same env vars.

reddynitish:
Closes #1052

## Summary

Martin no longer implicitly configures itself from:

- `DATABASE_URL`
- `DEFAULT_SRID`
- `PGSSLCERT`
- `PGSSLKEY`
- `PGSSLROOTCERT`

Configuration now comes from CLI arguments and config files.

Explicit environment substitution inside config files remains supported, including `${VAR}`-style values.

Legacy variables are detected and produce migration guidance without printing their values.

## Migration

| Legacy variable | CLI replacement | Config replacement |
| --- | --- | --- |
| `DATABASE_URL` | positional connection string | `postgres.connection_string` |
| `DEFAULT_SRID` | `--default-srid` | `postgres.default_srid` |
| `PGSSLCERT` | `--ssl-cert` *(new)* | `postgres.ssl_cert` |
| `PGSSLKEY` | `--ssl-key` *(new)* | `postgres.ssl_key` |
| `PGSSLROOTCERT` | `--ca-root-file` | `postgres.ssl_root_cert` |

`--ssl-cert` and `--ssl-key` are new CLI flags, added because those two previously had no CLI equivalent; they mirror `--ca-root-file`.

The e2e test harness was updated to explicitly forward SSL cert paths through CLI flags (instead of relying on the removed implicit env reads), so the existing verify-ca/verify-full CI matrix and `just test-ssl-cert` keep working.

The pmtiles `AWS_*` compatibility shim is intentionally untouched — it already warns and migrates into `pmtiles.*` config keys, and CI depends on it. Happy to remove it in a follow-up if that's in scope for this issue.

## Testing

- `cargo check --workspace`
- `cargo fmt --check`
- `cargo clippy --workspace --all-targets` (warnings denied)
- `cargo build -p martin`
- `cargo test --package martin-e2e-tests --test env_vars` (new, no database needed)
- Relevant `martin` crate unit tests (config/CLI/postgres args)
- DB-backed e2e tests (`config_file`, `martin_cp`, `postgres`, `process`) compile successfully

DB-backed e2e tests were not executed locally because Docker/Linux containers were unavailable on the Windows development machine. Those tests should be validated by CI.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -153,7 +153,7 @@ curl http://localhost:3000/catalog
 just start
 PGHOST=localhost PGPORT=5411 PGUSER=postgres PGPASSWORD=postgres PGDATABASE=db tests/fixtures/initdb.sh
 export DATABASE_URL='postgres://postgres:postgres@localhost:5411/db'
-cargo run --bin martin -- --webui enable-for-all
+cargo run --bin martin -- --webui enable-for-all "$DATABASE_URL"
 ```
 
 ### Scenario 3 - CLI Tools
--- a/.github/files/homebrew.martin.rb.j2
+++ b/.github/files/homebrew.martin.rb.j2
@@ -37,7 +37,7 @@ class Martin < Formula
 
   def caveats; <<~EOS
     Martin requires a database connection string.
-    It can be passed as a command-line argument or as a DATABASE_URL environment variable.
+    Pass it as a command-line argument, or set it in a config file.
       martin postgres://postgres@localhost/db
   EOS
   end
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -358,7 +358,11 @@ jobs:
           # directories they build their own in, so the workspace and the temp directory are
           # mounted where the host has them. The container runs as the runner so that the files
           # it writes into those directories stay ours to read and delete.
-          RUN="--rm --net host --platform $PLATFORM --user $(id -u):$(id -g) -e DATABASE_URL -e AWS_REGION=eu-central-1 -e AWS_SKIP_CREDENTIALS=1 -e RUST_LOG_FORMAT=bare -v $PWD:$PWD -v $TMPDIR:$TMPDIR -w $PWD"
+          # DEFAULT_SRID/PGSSL* are forwarded bare (no `=value`) alongside DATABASE_URL so tests
+          # that check the legacy-env-var migration warnings see them inside the container too --
+          # docker only passes through variables explicitly listed here, never the caller's whole
+          # environment.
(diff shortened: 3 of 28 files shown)


CommanderStorm (changes requested):
This is a 3-in-1 PR, as you likely know from this being 600 lines..
Could you split the individual parts?

The reason is primarily, that two of the parts are very uncontroversial, but the third is a semver major breaking change, so something that we can only do if we bump to martin@v2.

Here are the areas that I see:
- adding the --ssl-cert and --ssl-key CLI flags
- documentation and misc other refactorigns how we SHOULD behave
- warning if one of the Env vars is used, but still using them
- ⚠️ the actuall switchover

Also please do a slop pass, there is some stuff in there which I don't fully think you reviewed, such as `Migrating away from the automatic variables` in the middle of our docs.

reddynitish:
Splitting this up per the review above. Opened two focused replacements:

- #3150 — `--ssl-cert`/`--ssl-key` CLI flags
- #3151 — deprecation warnings for the legacy implicit env vars (`DATABASE_URL`, `DEFAULT_SRID`, `PGSSLCERT`, `PGSSLKEY`, `PGSSLROOTCERT`); current implicit behavior is fully preserved

The semver-major removal of implicit env var reads (the third area from the review) is deferred — that's Martin v2 territory and isn't included in either replacement. Closing this one as superseded.

reddynitish:
Superseded by #3150 and #3151.

PR #3142, opened 2026-08-22
refactor(reload): build advisory additions and updates concurrently

johncarmack1984:
First piece split out of #3130.

`ReloadAdvisory::from_sets`/`from_maps` awaited each source build one at a time, so a reconcile tick that picks up several sources pays their build latencies in sequence. This collects the ids first and builds them through a buffered stream, shared between both constructors as `build_all()`. Input order is preserved, and the fan-out is capped at `MAX_CONCURRENT_SOURCE_INITS` (32): #3122 showed what unbounded opens do once you have a few thousand sources.

Both issue-attached PRs sit on this hunk. #3130 needs it to keep `init()`-at-startup at parity with the old concurrent `resolve()` (0.645s → 0.642s on the 35-source test DB). #3131 carries it today and rebases down to just the pmtiles changes once this lands.

Signatures are unchanged; the existing advisory tests cover ordering and content.

### DIFF
--- a/martin/src/config/file/file_config.rs
+++ b/martin/src/config/file/file_config.rs
@@ -393,7 +393,7 @@ pub async fn resolve_files<T: TileSourceConfiguration>(
 
 /// How many tile sources are opened at once at startup and on reload.
 /// Opening a remote source is a few dependent round trips, so serial opens cost latency times source count.
-/// 
+///
 /// FIXME: make this constant dependent on system size (number of cores) and source types (local/remote)
 #[cfg(feature = "_tiles")]
 pub const MAX_CONCURRENT_SOURCE_INITS: usize = 64;
--- a/martin/src/config/file/file_config.rs
+++ b/martin/src/config/file/file_config.rs
@@ -391,6 +391,12 @@ pub async fn resolve_files<T: TileSourceConfiguration>(
     resolve_int(config, idr, extension, default_cache).await
 }
 
+/// How many tile sources are opened at once at startup and on reload. Opening a remote source is
+/// a few dependent round trips, so serial opens cost latency times source count; this bounds the
+/// fan-out so thousands of sources do not mean thousands of simultaneous requests.
+#[cfg(feature = "_tiles")]
+pub const MAX_CONCURRENT_SOURCE_INITS: usize = 32;
+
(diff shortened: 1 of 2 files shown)


martin/src/reload.rs
```diff
@@ -73,7 +73,7 @@ impl ReloadAdvisory {
     /// Generates an advisory for **versioned** sources (e.g., `MBTiles`, COG).
     ///
     /// Compares keys and version values to distinguish between additions, removals,
-    /// and updates (version changed).
+    /// and updates (version changed). Additions and updates are built concurrently.
```
CommanderStorm:
```suggestion
    /// and updates (version changed).
```

martin/src/reload.rs
```diff
@@ -40,6 +41,7 @@ impl ReloadAdvisory {
     ///
     /// Any source that disappears is a removal; any source that appears is an addition.
     /// Sources that remain are left untouched (no update concept without versions).
+    /// Additions are built concurrently.
```
CommanderStorm:
```suggestion
```

martin/src/config/file/file_config.rs
```diff
@@ -391,6 +391,12 @@ pub async fn resolve_files<T: TileSourceConfiguration>(
     resolve_int(config, idr, extension, default_cache).await
 }
 
+/// How many tile sources are opened at once at startup and on reload. Opening a remote source is
+/// a few dependent round trips, so serial opens cost latency times source count; this bounds the
+/// fan-out so thousands of sources do not mean thousands of simultaneous requests.
+#[cfg(feature = "_tiles")]
+pub const MAX_CONCURRENT_SOURCE_INITS: usize = 32;
```
CommanderStorm:
```suggestion
/// How many tile sources are opened at once at startup and on reload.
/// Opening a remote source is a few dependent round trips, so serial opens cost latency times source count.
/// 
/// FIXME: make this constant dependent on system size (number of cores) and source types (local/remote)
#[cfg(feature = "_tiles")]
pub const MAX_CONCURRENT_SOURCE_INITS: usize = 64;
```

CommanderStorm (approved):
LGTM, thanks

PR #3143, opened 2026-08-22
refactor(discovery): return warnings from Discovery::discover

johncarmack1984:
Second piece split out of #3130.

`PostgresDiscovery` logged its discovery warnings inline, so the driver never saw them, and the other kinds had no channel to report any. `discover()` now returns `Discovered { sources, warnings }` and the `ReloadDriver` owns the logging: one place for every kind, same warn-per-tick behavior as before.

The reason beyond tidiness: #3130's `init()` has to hand these warnings to the caller's `on_invalid` policy (abortable at startup, warn-only live), and that's only possible once they're in the return value instead of a log side effect.

No behavior change on main.


CommanderStorm (approved):
LGTM

PR #3144, opened 2026-08-23
fix(postgres): keep per-source convert_to_* overrides on the reload path

johncarmack1984:
Third piece split out of #3130; a fix rather than a refactor. Builds on #3142 and #3143, both merged.

Startup resolution honors per-table and per-function `convert_to_mlt`/`convert_to_mvt` overrides. The reload path didn't: when the reloader added or rebuilt a postgres source, its process config came from the connection level only, so a reload of the same table silently dropped its overrides.

`Discovery::build` now returns `BuiltSource { source, process }`. The postgres discovery fills `process` from the instantiated spec (per-source > connection); kinds without per-source settings convert with `From<BoxedSource>` and keep the kind-level fallback. A unit test pins the override behavior.

After this, #3130 rebases down to the lifecycle change itself.

### DIFF
--- a/martin/src/config/file/tiles/discovery/postgres.rs
+++ b/martin/src/config/file/tiles/discovery/postgres.rs
@@ -115,38 +115,6 @@ fn per_source_process(connection: &ProcessConfig, _spec: &SourceSpec) -> Process
     connection.clone()
 }
 
-#[cfg(all(test, feature = "mlt"))]
-mod process_tests {
-    use super::*;
-    use crate::config::file::postgres::TableInfo;
-    use crate::config::primitives::AutoOption;
-
-    #[test]
-    fn per_source_convert_overrides_the_connection_level() {
-        let connection = ProcessConfig {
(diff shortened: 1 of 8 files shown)


martin/src/config/file/tiles/discovery/postgres.rs
```diff
@@ -80,16 +79,74 @@ impl Discovery for PostgresDiscovery {
...
+    resolve_process_config(connection, &ProcessConfig::default(), &per_source)
+}
+
+#[cfg(not(feature = "mlt"))]
+fn per_source_process(connection: &ProcessConfig, _spec: &SourceSpec) -> ProcessConfig {
+    connection.clone()
+}
+
+#[cfg(all(test, feature = "mlt"))]
+mod process_tests {
```
CommanderStorm: Please move this into the general mod test and cfg the test instead. 
There was some weirdness with coverage that requires this that I don't fully remember.

CommanderStorm (approved):
Lgtm, assuming you move the test to mod test as noted below

Issue #2004, opened 2025-08-04
Decrease code duplication in CI workflow-call action and yaml anchors


CommanderStorm:
GitHub just shipped the feature to have yaml anchors:
- https://github.com/actions/runner/issues/1182

In the following issue we discussed at least one place where this might be benefitial:
- https://github.com/maplibre/martin/pull/1877

There are likely a lot more of these places and the featre can remove lots of the duplicated code

nyurik:
do we really want to use this?? I am not a big fan of "magic" - i.e. code that gets merged dynamically, e.g. combining multiple yaml lists together using anchors.  Instead, I would much rather have dedicated custom github actions - that are reusable across different actions as needed.  See https://github.com/maplibre/maplibre-rs/tree/main/.github/actions for examples

CommanderStorm:
Both have a place imo

nyurik:
i disagree - i think github actions should be as tiny as possible - because otherwise it becomes really hard to simulate them locally. Instead, I prefer most testing logic to be part of a justfile, including packaging and all other steps. Ideally, you should be able to run exactly the same everything locally, perhaps using docker + other stuff

CommanderStorm:
For the `strategy` field of https://github.com/maplibre/maplibre-native-rs/pull/43/files anchors for example make sense

nyurik:
sure, that one i would agree with, assuming we will always want to have the same list of test targets for regular and msrv. What i wouldn't want is for us to have a bunch of action steps shared between multiple workflows using yaml magic -- that should be done using proper "sub-actions"

johncarmack1984:
I'd like to take this one, if that's alright. I've read through this discussion and have a [branch ready with a first pass](https://github.com/maplibre/martin/compare/main...johncarmack1984:martin:john/ci-2004-dedup), following the direction settled above: sub-actions for shared steps, no yaml step magic.

The first pass reduces to:

- A tiny `install-tools` composite action that pins `taiki-e/install-action` once. DRYs up thirteen call sites (plus two inside `setup-rust`), so future version bumps touch one file instead of seven.
- `hotpath-profile.yml`'s repeated 20-line benchmark script becomes one script, written to /tmp, and called from both steps. A just recipe would be the natural home, but the script can't live in a tracked file: the base run happens after the workspace is checked out at the base sha.

actionlint reports the identical finding set as main, minus two shellcheck notes that die with the duplicated script.

[...]

PR #3145, opened 2026-08-23
ci: dedupe repeated action pins and the hotpath benchmark script

johncarmack1984:
Part of #2004, scoped per [the discussion there](https://github.com/maplibre/martin/issues/2004#issuecomment-5385144989). Two dedups, both behavior-preserving:

- **One pin for `taiki-e/install-action`, reusing `setup-rust`.** Per the suggestion to reuse the existing install-step work: `setup-rust` gains a tool-only mode (`toolchain: none`) that skips the disk cleanup and the toolchain install, and the thirteen jobs that only need prebuilt tools now go through it. The pinned SHA lives in one file instead of fourteen call sites across seven, so a version bump touches one place. Jobs keep exactly the steps they had before — `toolchain: none` runs only the install step, so nothing gets a surprise 25GB disk purge or an unused toolchain.
- **One copy of the hotpath benchmark script.** `hotpath-profile.yml` had the same 20-line start/poll/bench/kill script pasted for the head and base runs. It's now written once to `/tmp` and invoked from both steps. It deliberately lives outside the tree (not the justfile or a local action) because the base run happens after `git checkout <base sha>` switches the whole workspace, so a tracked file would not exist for PRs based on older mains.

Verified with actionlint: the finding set is identical to main apart from line shifts (two shellcheck notes disappear along with the duplicated script).

Left for follow-ups to keep review surface small: the postgis `services:` incantation repeated across `ci.yml`/`hotpath-profile.yml` (needs a reusable workflow since `services` can't live in composite actions), and the `package` job's seven download-artifact steps (a `pattern:` glob changes directory names, so it isn't a pure dedup).

### DIFF
--- a/.github/workflows/build-binary.yml
+++ b/.github/workflows/build-binary.yml
@@ -41,13 +41,11 @@ jobs:
       - name: Checkout sources
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with: { persist-credentials: false }
-      - name: Install just
+      - name: Install just (and cargo-zigbuild for musl)
         uses: ./.github/actions/setup-rust
-        with: { toolchain: none }
-      - name: Install cargo-zigbuild (musl only)
-        if: inputs.musl
-        uses: ./.github/actions/setup-rust
-        with: { toolchain: none, tools: 'cargo-zigbuild' }
+        with:
+          toolchain: none
+          tools: ${{ inputs.musl && 'just,cargo-zigbuild' || 'just' }}
       - name: Install zig (musl only)
         if: inputs.musl
         uses: mlugg/setup-zig@d1434d08867e3ee9daa34448df10607b98908d29 # v2.2.1
--- a/.github/workflows/build-binary.yml
+++ b/.github/workflows/build-binary.yml
@@ -42,12 +42,12 @@ jobs:
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with: { persist-credentials: false }
       - name: Install just
-        uses: taiki-e/install-action@6c6fd71fe4fb72c3697d269963d0e15df8adedad # v2.85.10
-        with: { tool: 'just' }
+        uses: ./.github/actions/setup-rust
+        with: { toolchain: none }
       - name: Install cargo-zigbuild (musl only)
         if: inputs.musl
-        uses: taiki-e/install-action@6c6fd71fe4fb72c3697d269963d0e15df8adedad # v2.85.10
-        with: { tool: 'cargo-zigbuild' }
+        uses: ./.github/actions/setup-rust
+        with: { toolchain: none, tools: 'cargo-zigbuild' }
       - name: Install zig (musl only)
         if: inputs.musl
         uses: mlugg/setup-zig@d1434d08867e3ee9daa34448df10607b98908d29 # v2.2.1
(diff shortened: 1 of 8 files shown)


.github/workflows/build-binary.yml
```diff
@@ -42,12 +42,12 @@ jobs:
...
-        uses: taiki-e/install-action@6c6fd71fe4fb72c3697d269963d0e15df8adedad # v2.85.10
-        with: { tool: 'just' }
+        uses: ./.github/actions/setup-rust
+        with: { toolchain: none }
       - name: Install cargo-zigbuild (musl only)
         if: inputs.musl
-        uses: taiki-e/install-action@6c6fd71fe4fb72c3697d269963d0e15df8adedad # v2.85.10
-        with: { tool: 'cargo-zigbuild' }
+        uses: ./.github/actions/setup-rust
+        with: { toolchain: none, tools: 'cargo-zigbuild' }
```
CommanderStorm: running ./.github/actions/setup-rust twice seems rather uncommon.
Can we just make `tools` an dynamic param and template it using githubs syntax for this?

.github/workflows/hotpath-profile.yml
```diff
@@ -63,16 +63,12 @@ jobs:
...
-      - name: Benchmark head
-        env:
-          HOTPATH_OUTPUT_FORMAT: json
-          HOTPATH_OUTPUT_PATH: /tmp/metrics/head_timing.json
-          HOTPATH_ALLOC_SELF: true
-          DATABASE_URL: postgres://${{ env.PGUSER }}:${{ env.PGPASSWORD }}@${{ env.PGHOST }}:${{ job.services.postgres.ports[5432] }}/${{ env.PGDATABASE }}
+      # Written to /tmp (not the justfile or a local action) because the base
+      # benchmark runs after `git checkout <base sha>` switches the whole tree,
+      # so anything added to tracked files would not exist for the base run.
+      - name: Write the benchmark script both runs share
```
CommanderStorm: please move this to the justfile instead.
This will fail at first, that is fine, I can force-merge

CommanderStorm (changes requested):
mostly LGTM, I have two minor cleanup requests regarding the code

.github/workflows/build-binary.yml
johncarmack1984: Yes, done in c819ec6.

.github/workflows/hotpath-profile.yml
johncarmack1984: Done, c819ec6.

justfile
```diff
@@ -82,6 +82,24 @@ build-hotpath: fetch
...
+    just bench-server-hotpath &
+    MARTIN_PID=$!
+
+    for i in {1..1000}; do
+        if curl -sf http://localhost:3000/health > /dev/null 2>&1; then break; fi
+        sleep 1
+    done
+    curl -sf http://localhost:3000/health > /dev/null 2>&1 || { echo "::error::Martin failed to start"; kill "$MARTIN_PID" 2>/dev/null; exit 1; }
+
+    just bench-http 1m 100k
```
```suggestion
    {{just}} bench-server-hotpath &
    MARTIN_PID=$!

    for i in {1..1000}; do
        if curl -sf http://localhost:3000/health > /dev/null 2>&1; then break; fi
        sleep 1
    done
    curl -sf http://localhost:3000/health > /dev/null 2>&1 || { echo "::error::Martin failed to start"; kill "$MARTIN_PID" 2>/dev/null; exit 1; }

    {{just}} bench-http 1m 100k
```

CommanderStorm (approved):
not enabling automerge since hotpath is not marked as required

PR #3147, opened 2026-08-23
docs: fix typo explicitely -> explicitly (source + generated), recomend -> recommend

vaibhav8a:
Two typo fixes.

**`explicitely` → `explicitly`.** The text lives in Rust doc comments:

- `martin/src/config/file/tiles/postgres/config.rs:137`
- `martin/src/config/file/main/config.rs:175`

`docs/content/files/generated_config.md` carries the header *"This file is auto-generated by `just gen-schemas`. Don't edit it, edit the code's doc comments instead"*, so the source comments are the real fix. I've also applied the same one-word substitution to the two derived artifacts (`generated_config.md` and `schemas/config.json`) so the committed tree matches what the generator would now produce — otherwise a generated-file check in CI would flag them as stale. The change is a pure string replacement, so the result is byte-identical to regenerating.

**`recomend` → `recommend`** in `docs/content/sources-geojson.md:34` (a hand-written page).

Happy to drop the generated-file half of the diff if you'd rather regenerate them yourselves.

### DIFF
--- a/docs/content/files/generated_config.md
+++ b/docs/content/files/generated_config.md
@@ -71,7 +71,7 @@ cache_control: null
 # Can be either:
 # - (default) `auto` - we choose defaults which we think work best for most users
 # - `disabled` - no conversion
-# - explicitely configured
+# - explicitly configured
 convert_to_mlt:
   # Allow `FastPFOR` integer compression.
   allow_fastpfor: null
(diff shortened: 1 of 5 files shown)


CommanderStorm (approved):
lgtm

PR #3148, opened 2026-08-23
test(cache-control): cover the cache_control setting end-to-end and document it

johncarmack1984:
### Overview

Refs #749. #3042 added the server-level `cache_control` setting, but #3105's e2e migration dropped its only test fixtures, so the setting currently ships with zero coverage. This PR restores coverage and documents the setting; no behavior changes.

Per [the plan I posted on #749](https://github.com/maplibre/martin/issues/749#issuecomment-5383685995): this is PR 1 of 2. The per-source override half is built on top of this branch and follows separately.

### Testing

- New `e2e-tests/tests/cache_control.rs`: the configured header is sent on every content endpoint (tiles, tilejson, catalog, sprites, fonts, styles), `/health` keeps its `no-cache` policy, nothing is sent unless configured, and an invalid value fails startup naming the bad value
- Save-config round-trip fixture + snapshot re-cover the setting
- `cargo test --workspace --lib` and the full `cargo test -p martin-e2e-tests` suite pass; deny-warnings clippy clean

### Docs

- reverse-proxy page notes Martin can set the header itself
- The Lambda page's "Teach Martin how to set the Cache-Control and Etag headers" TODO is resolved with a pointer to the setting

### DIFF
--- a/docs/content/run-with-lambda.md
+++ b/docs/content/run-with-lambda.md
@@ -147,4 +147,5 @@ AWS Lambda support is preliminary; there are features to add to Martin, configur
 * Set up a CloudFront CDN, this is a whole thing, but explain the motivation and the basics.
 * Grant the execution role permission to read objects from an S3 bucket, and teach Martin how to make authenticated requests to S3.
 * Teach Martin how to serve all PMTiles files from an S3 bucket rather than having to list them at startup.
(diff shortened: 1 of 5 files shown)


CommanderStorm (approved):
thanks

PR #3149, opened 2026-08-23
test(e2e): fix some CI flakes: give QEMU arm64 docker setup a longer timeout

johncarmack1984:
Last new PR today, I swear 😅. Saw the flake on #3147 and remembered it had happened a [few](https://github.com/maplibre/martin/actions/runs/32611644064/job/97126231221) [times](https://github.com/maplibre/martin/actions/runs/32604448434/job/97108088001) lately, so figured a fix would be welcome. 

Summary: The first parallel wave of e2e tests in the arm64 docker CI step starts several cold QEMU-emulated containers at once; if none reaches ready within the 60s READY_TIMEOUT, the whole wave fails with an empty log. This makes the timeout overridable via MARTIN_E2E_READY_TIMEOUT (seconds) and sets it to 300 for that step.


CommanderStorm (approved):
Fsir

PR #3150, opened 2026-08-24
feat(postgres): add --ssl-cert and --ssl-key CLI flags

reddynitish:
## Summary

Adds `--ssl-cert` and `--ssl-key` CLI flags for the Postgres source, mirroring the existing `--ca-root-file` flag. This split out of #3141 following review from @CommanderStorm, who asked for the SSL flags, the legacy env var warnings, and the breaking removal of implicit env var reads to be separated into independent PRs.

This PR is the narrow, non-breaking part: two new flags only.

- `--ssl-cert <FILE>` / `--ssl-key <FILE>` set the client SSL certificate/key for Postgres connections.
- CLI flags override the config file, the same way `--ca-root-file` already overrides `ssl_root_cert`.
- Existing `PGSSLCERT`/`PGSSLKEY` env var fallback behavior is unchanged.

## Testing

- `cargo fmt --check`
- `cargo check --workspace`
- `cargo test -p martin --lib config::args::postgres` (7 passed)
- `cargo clippy -p martin --all-targets -- -D warnings` — passes for the changed crate; a pre-existing `result_large_err` clippy failure in `martin-core::tiles::postgres::errors` (unrelated file, present on `main`) blocks a full clean workspace clippy run.

### DIFF
--- a/martin/src/config/args/postgres.rs
+++ b/martin/src/config/args/postgres.rs
@@ -347,8 +347,6 @@ mod tests {
         args.check().unwrap();
     }
 
-    /// `--ssl-cert`/`--ssl-key` win over `PGSSLCERT`/`PGSSLKEY`, the same way `--ca-root-file`
-    /// already wins over `PGSSLROOTCERT`.
     #[test]
     fn cli_ssl_cert_and_key_override_env() {
         let mut args = Arguments::new(vec![]);
@@ -382,8 +380,6 @@ mod tests {
         args.check().unwrap();
     }
 
-    /// A config file's SSL values must survive; `--ssl-cert`/`--ssl-key` override them like
-    /// `--ca-root-file` already does for `ssl_root_cert`.
     #[test]
     fn override_config_applies_ssl_cli_flags() {
         let mut config = OptOneMany::One(PostgresConfig {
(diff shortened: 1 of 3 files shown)


martin/src/config/args/postgres.rs
```diff
@@ -314,4 +346,70 @@ mod tests {
         );
         args.check().unwrap();
     }
+
+    /// `--ssl-cert`/`--ssl-key` win over `PGSSLCERT`/`PGSSLKEY`, the same way `--ca-root-file`
+    /// already wins over `PGSSLROOTCERT`.
+    #[test]
```
CommanderStorm:
```suggestion
    #[test]
```

martin/src/config/args/postgres.rs
```diff
@@ -314,4 +346,70 @@ mod tests {
...
+                },
+                ..Default::default()
+            })
+        );
+        args.check().unwrap();
+    }
+
+    /// A config file's SSL values must survive; `--ssl-cert`/`--ssl-key` override them like
+    /// `--ca-root-file` already does for `ssl_root_cert`.
+    #[test]
```
CommanderStorm:
```suggestion
    #[test]
```

CommanderStorm (approved):
thanks

PR #3151, opened 2026-08-24
feat(config): warn about legacy Postgres env var usage

reddynitish:
## Summary

Warns once at startup when Martin falls back to one of the five implicit Postgres env vars (`DATABASE_URL`, `DEFAULT_SRID`, `PGSSLCERT`, `PGSSLKEY`, `PGSSLROOTCERT`), naming the config-file key (and CLI flag, where one exists) that replaces it. Split out of #3141 following review from @CommanderStorm, alongside #3150 (the `--ssl-cert`/`--ssl-key` flags).

This is a deprecation/migration-warning PR only — **all five variables still function exactly as before**. The actual breaking removal is deferred to Martin v2 and is not part of this PR.

- A variable the config file already references explicitly as `${VAR}` is not reported — it's doing exactly what it should.
- Variable values are never logged, only the fact that a var is set.
- Warning text names the replacement, it never claims the variable is already ignored.
- `docs/content/env-vars.md` gets a short deprecation note pointing at #1052.

## Testing

- `cargo fmt --check`
- `cargo check --workspace`
- `cargo test -p martin --lib config::file::main::parse` (63 passed)
- `cargo test -p martin --lib config::args::root` (9 passed, including a regression test that `DATABASE_URL` still configures Postgres while also warning)
- `cargo clippy -p martin --all-targets -- -D warnings` — passes for the changed crate; a pre-existing `result_large_err` clippy failure in `martin-core::tiles::postgres::errors` (unrelated file, present on `main`) blocks a full clean workspace clippy run.


docs/content/env-vars.md
CommanderStorm:
```suggestion

| Environment var <br/> Config File key    | Example                                   | Description                                                                                                                                                                                                |
|------------------------------------------|-------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| `AWS_LAMBDA_RUNTIME_API` <br/> -         |                                           | If defined, connect to AWS Lambda to handle requests. The regular HTTP server is not used. See [Running in AWS Lambda](run-with-lambda.md)                                                                 |

!!! warning "Deprecated environemnt variables"
    Reading below environment variables is deprecated and will be removed in a future, major release.
    See [#1052](https://github.com/maplibre/martin/issues/1052)) for further context.
    Use the appropriate CLI flags or our yaml interpolation (`key: ${ENV_VAR}`) support that replaces them.

    | Environment var <br/> Config File key    | Example                                   | Description                                                                                                                                                                                                |
    |------------------------------------------|-------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
    | `DATABASE_URL` <br/> `connection_string` | `postgres://`<br/>`postgres@localhost/db` | Postgres database connection                                                                                                                                                                               |
    | `DEFAULT_SRID` <br/> `default_srid`      | `4326`                                    | If a PostgreSQL table has a geometry column with SRID=0, use this value instead                                                                                                                            |
    | `PGSSLCERT` <br/> `ssl_cert`             | `./postgresql.crt`                        | A file with a client SSL certificate. [docs](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-SSLCERT)                                                                             |
    | `PGSSLKEY` <br/> `ssl_key`               | `./postgresql.key`                        | A file with the key for the client SSL certificate. [docs](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-SSLKEY)                                                                |
    | `PGSSLROOTCERT` <br/> `ssl_root_cert`    | `./root.crt`                              | A file with trusted root certificate(s). The file should contain a sequence of PEM-formatted CA certificates. [docs](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-SSLROOTCERT) |
```

e2e-tests/src/martin.rs
```diff
@@ -49,6 +49,9 @@ const ALLOWED_LOG_LINES: &[&str] = &[
     "Unable to deserialize SQL comment on public.points2 as tilejson",
     "Discovering tables in PostgreSQL database",
     "ST_EstimatedExtent on",
+    // `MartinBuilder::with_postgres` connects via the `DATABASE_URL` env var,
+    // same as `just start`/`just test-pg` do -- expected on every postgres test.
+    "Environment variable DATABASE_URL is deprecated",
```
CommanderStorm:
```suggestion
    "Environment variable DATABASE_URL is deprecated",
```

reddynitish:
@CommanderStorm Thanks for the earlier review. I addressed the feedback and rechecked the final diff/tests; #3151 is ready for another look when you have time.

CommanderStorm (approved):
I cleaned it up a bit, lgtm now

PR #3152, opened 2026-08-24
ci: dedupe the PostGIS services and the package job's artifact downloads

### LINKED ISSUE #2004: Decrease code duplication in CI workflow-call action and yaml anchors

johncarmack1984:
Closes #2004, second half after #3145. Same rule as the thread settled on: shared steps live in sub-actions, and anchors only carry data. 

Changes are net-negative, with repetitions factored out thusly:

- The PostGIS services share their env and options through anchors. The SSL ones keep `-e` inside `options` (it has to come before the image name). The matrixed one in `test-with-svc` stays as is. I said above this needed a reusable workflow. It didn't: a service block is data.
- The job-level `PG*` env is anchored once.
- The `package` job downloads `build-*` in one step and packages whatever arrived. The target list was written out twice (once in `needs:`, once in the script). Now it's once.

Nothing changes at runtime. Every service gets the same env and docker options, and `package-assets` runs per target in the same layout as before (the loop moves `target/build-<target>` to `target/<target>` first). The `docker run` copy in the Ubuntu test job is shell, not YAML, so it stays.

actionlint reports main's set minus the one `- parallel:` note that dies with the download block. zizmor is clean.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -222,14 +222,15 @@ jobs:
         image: postgis/postgis:18-3.6
         ports:
           - 5432/tcp
-        env: &postgis-env
-          POSTGRES_DB: test
-          POSTGRES_USER: postgres
-          POSTGRES_PASSWORD: postgres
-          PGDATABASE: test
-          PGUSER: postgres
-          PGPASSWORD: postgres
+        # The -e flags stay inside options here: they must precede the image name, and the
+        # runner appends a service's env: map after options, i.e. after the image.
         options: &postgis-ssl-options >-
+          -e POSTGRES_DB=test
+          -e POSTGRES_USER=postgres
+          -e POSTGRES_PASSWORD=postgres
+          -e PGDATABASE=test
+          -e PGUSER=postgres
+          -e PGPASSWORD=postgres
           --health-cmd pg_isready
           --health-interval 10s
           --health-timeout 5s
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -222,7 +222,9 @@ jobs:
         image: postgis/postgis:18-3.6
         ports:
           - 5432/tcp
-        options: >-
+        # The -e flags stay inside options here: they must precede the image name, and the
+        # runner appends a service's env: map after options, i.e. after the image.
+        options: &postgis-ssl-options >-
```
```suggestion
        options: &postgis-ssl-options >-
```

.github/workflows/ci.yml
```diff
@@ -860,35 +837,18 @@ jobs:
...
-          - name: Download build artifact build-x86_64-unknown-linux-musl
-            uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
-            with: { name: 'build-x86_64-unknown-linux-musl', path: 'target/x86_64-unknown-linux-musl' }
-          - name: Download build artifact build-x86_64-pc-windows-msvc
-            uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
-            with: { name: 'build-x86_64-pc-windows-msvc', path: 'target/x86_64-pc-windows-msvc' }
-          - name: Download build artifact build-debian-x86_64
-            uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
-            with: { name: 'build-debian-x86_64', path: 'target/debian-x86_64' }
+      # Every build-* artifact from the jobs in `needs` lands in target/build-<target>/.
```
```suggestion
```

CommanderStorm (approved):
thanks 👍🏻

CommanderStorm (approved):
thanks 👍🏻

PR #3160, opened 2026-08-25
fix(reload): keep per-source convert_to_* overrides for file-backed sources

johncarmack1984:
Companion to #3144, which fixed this for Postgres. The file-backed kinds (mbtiles, pmtiles, geojson, cog) have the same gap: `FsDiscovery` only remembers each configured path's cache policy, so when the reload path rebuilds a source, its per-source `convert_to_*` settings are dropped and it falls back to the kind-level `ProcessConfig`.

The fix mirrors the Postgres one. `FsDiscovery` now keeps the whole configured entry per canonical path (cache policy plus the per-source process, resolved against the kind level at construction), and `build()` returns the override through `BuiltSource`. `NewSource` already prefers that over the kind level, so nothing downstream needed to change. A unit test builds an `FsDiscovery` from a config with one overridden and one plain source on disk and checks what it kept for each.

One small side fix: the `resolve_process_config` re-export was gated on `mlt` plus a hand-listed set of kinds that missed `geojson`, which enables `mlt` without any of them. The gate is now `mlt` + `_tiles`, matching the actual use sites.

Verified: workspace lib tests (720 pass), the full e2e suite (250 pass), and `just check` (cargo hack each-feature, 51 combos) under `CARGO_BUILD_WARNINGS=deny`; fmt + clippy + pre-commit clean.

### DIFF
--- a/martin/src/config/file/mod.rs
+++ b/martin/src/config/file/mod.rs
@@ -15,10 +15,7 @@ pub use error::{ConfigFileError, ConfigFileResult};
 
 pub mod process;
 pub use process::ProcessConfig;
-#[cfg(all(
-    feature = "mlt",
-    any(feature = "mbtiles", feature = "pmtiles", feature = "postgres")
-))]
+#[cfg(all(feature = "mlt", feature = "_tiles"))]
 pub(crate) use process::resolve_process_config;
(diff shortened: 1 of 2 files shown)


CommanderStorm (approved):
lgtm

PR #3029, opened 2026-07-18
fix(pmtiles): restore AWS profile support

floze-the-genius:
## Summary

- make top-level configuration finalization async so PMTiles can load AWS SDK profiles
- support `pmtiles.profile` and the `aws_profile` alias with refreshable credentials and profile region discovery
- preserve explicit PMTiles credentials, region, and unsigned-request settings as higher-priority configuration
- reuse the profile-aware S3 builder for initial sources and remote-prefix reloads
- update the PMTiles guide and generated config schema

## Root cause

Martin copied synchronous `AWS_*` values into `object_store` options during configuration finalization. Named profiles require asynchronous AWS SDK loading, and the previous path explicitly warned that `AWS_PROFILE` was unsupported. It also had no place to retain a refreshable credential provider for long-running servers.

This change adapts the AWS SDK credential provider to `object_store::CredentialProvider`, retaining refresh behavior instead of snapshotting temporary credentials.

## Validation

- `cargo check -p martin --no-default-features --features pmtiles`
- `cargo test -p martin --no-default-features --features pmtiles`
  - 201 library tests
  - 2 binary diagnostic tests
  - 6 PMTiles server tests
  - 1 doctest
- `just clippy`
  - workspace, all targets, no warnings
- config JSON Schema metaschema validation
- OpenAPI 3.1 validation
- all three shipped config fixtures validated against the schema
- generated config example validated against the schema

On macOS, `just test-schemas` reaches and passes the schema, OpenAPI, and fixture checks, then stops because the recipe uses GNU-only `mktemp --suffix`. I reran its final generated-doc extraction and `check-jsonschema` command with a macOS-compatible temporary path; it passed.

## AI disclosure

This contribution was implemented and validated by OpenAI Codex under the contributor's direction. The contributor remains responsible for reviewing and maintaining the submitted change.

Fixes #2286.

PR #3165, opened 2026-08-25
fix(pmtiles): pick up ECS and EKS credential discovery env vars for S3 sources

### LINKED ISSUE #3163: AWS credentials cannot be loaded for PMTiles sources from S3 on ECS Fargate
Recently, I reported an issue related to slow startup when using a huge number of S3 PMTiles files. That issue was resolved in [PR #3131](https://github.com/maplibre/martin/pull/3131?utm_source=chatgpt.com), thanks to @johncarmack1984. 

Yesterday, I wanted to test the changes myself using the nightly version: `ghcr.io/maplibre/martin:nightly`
I am running Martin in AWS ECS Fargate with the following configuration:
```yaml
keep_alive: 75

listen_addresses: 0.0.0.0:3000

worker_processes: 8

pmtiles:
  paths: []

  region: eu-central-1

  sources:
    xx_areas1: s3://my-bucket/data/layers/tile_name1.pmtiles
    xx_areas2: s3://my-bucket/data/layers/tile_name2.pmtiles
    xx_areas3: s3://my-bucket/data/layers/tile_name3.pmtiles
```
## Problem

Martin is unable to load the AWS credentials from the ECS Fargate environment.

I get the following error:

`Source xx_areas1: PMTiles error Generic S3 error: Error performing PUT http://169.254.169.254/latest/api/token in 2.090498973s, after 10 retries, max_retries: 10, retry_timeout: 180s - HTTP error: error sending request processing AmazonS3(my-bucket)`

It appears that Martin is trying to retrieve credentials through the EC2 Instance Metadata Service (169.254.169.254), but this does not seem to work in my ECS Fargate setup.

Could this be related to how AWS credentials are discovered/resolved for S3 access in the new PMTiles implementation?

I would appreciate any help troubleshooting this. If you need any additional information about my ECS/Fargate configuration, IAM role, task definition, environment, or logs, I can provide it.

Thank you!

johncarmack1984:
Fixes #3163.

On ECS/Fargate, EKS IRSA and EKS Pod Identity, the runtime tells you where the task role's credentials live through env vars (`AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` and friends). We build the S3 store with `AmazonS3Builder::new()` and only forward what's in `pmtiles.*`, so none of those reach `object_store`. It falls through to the EC2 instance metadata provider, which doesn't exist there: 10 retries against 169.254.169.254 and every source fails. #2251 kept the static-key env vars working through the deprecation shim; these are the ones it missed. #3029 got profiles back, this gets task roles back.

The fix forwards the seven discovery variables to their `object_store` option at finalize time. Explicit `pmtiles.*` keys win, and `profile` skips the import entirely since the SDK chain it loads already covers ECS and EKS. One info line per forwarded variable, so "is Martin seeing my task role?" is answerable from the log.

I know #1052 is the policy on implicit env vars, so a word on why these are different: they aren't user config. The values are per task (a random relative URI, a rotating token path) and set by the platform. That's also why `container_credentials_relative_uri: ${AWS_CONTAINER_CREDENTIALS_RELATIVE_URI}` isn't a real answer: an unresolved `${VAR}` is a config error, so the same file can't run on a laptop and in the cluster. Same category as `AWS_LAMBDA_RUNTIME_API`, which we already read. If you'd rather keep this docs-only, I can cut it down to the `${VAR}` recipe.

Tests cover the forwarding, explicit-config precedence and the profile skip, through `FauxEnv` so nothing touches the real environment. Docs: an ECS/EKS tip in the S3 section and a row in the env vars table.

Verified: `cargo test -p martin --lib` (378 pass), `cargo clippy -p martin --all-targets` and the `--no-default-features --features pmtiles` build, all under `CARGO_BUILD_WARNINGS=deny`; pre-commit clean.


martin/src/config/file/tiles/pmtiles.rs
```diff
@@ -29,12 +29,30 @@ use crate::config::file::{
...
 pub const DEFAULT_RELOAD_INTERVAL: Duration = Duration::from_mins(10);
 
+/// `object_store` options that AWS runtimes set through the environment to say where credentials
+/// come from: ECS/Fargate task roles, EKS IRSA and EKS Pod Identity.
+///
+/// The variable name is the option name upper-cased, which is also how
+/// [`AmazonS3Builder::from_env`] reads them. The values are per task (a random relative URI, a
+/// rotating token path), so they cannot be written into a config file ahead of time the way keys
+/// or a profile can.
+const AWS_CREDENTIAL_DISCOVERY_KEYS: &[AmazonS3ConfigKey] = &[
```
CommanderStorm:
```suggestion
/// Options that AWS runtimes set through env-vars to say where credentials come from
const AWS_CREDENTIAL_DISCOVERY_KEYS: &[AmazonS3ConfigKey] = &[
```

martin/src/config/file/tiles/pmtiles.rs
```diff
@@ -469,6 +482,31 @@ impl PmtConfig {
...
         }
     }
+
+    /// Forwards the credential-discovery variables for [`AWS_CREDENTIAL_DISCOVERY_KEYS`] to the
+    /// S3 client so task roles work without configuration.
+    ///
+    /// Without these, `object_store` falls back to the EC2 instance metadata service, which does
+    /// not exist on ECS/Fargate or EKS. Explicitly configured keys win over the environment, and a
+    /// `profile` is left alone because the AWS SDK chain it loads already covers these sources.
+    fn import_aws_credential_discovery_env(&mut self, env: &impl Env) {
```
CommanderStorm:
```suggestion
    /// Forwards the credential-discovery variables to the S3 client so task roles work without configuration.
    fn import_aws_credential_discovery_env(&mut self, env: &impl Env) {
```

CommanderStorm (approved):
🫶 Thanks.

:throws up hands: I gues, AWS...

PR #3167, opened 2026-08-25
feat(cache-control): per-source Cache-Control overrides

### LINKED ISSUE #749: Configure Cache Control Expires headers

johncarmack1984:
Closes #749

#3042 added the server-level `cache_control` default and #3148 covered it.
This is the per-source half, in the shape #2673 set for per-source `cache`: a global default plus a `cache_control` key on each source.

### DIFF
--- a/docs/content/run-with-reverse-proxy/index.md
+++ b/docs/content/run-with-reverse-proxy/index.md
@@ -20,7 +20,7 @@ Doing so has a few downsides:
     - [Varnish](https://varnish-cache.org/)
     - [Apache](https://httpd.apache.org/)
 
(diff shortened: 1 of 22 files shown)


CommanderStorm (approved):
the entire process config architecutre is getting out of place a bit. Not sure if this is right 100%.

The code is doing weird things where we don't have a proper seam I think.
Lots of None which is supicious, but we can look at it after we have hillshade in there too

Issue #3156, opened 2026-08-24
CI: re-measured sccache vs rust-cache (#950 revisited), plus a cache-quota audit


johncarmack1984:
I wanted to check whether #950's verdict still holds now that martin compiles two big C/C++ trees (bundled DuckDB and maplibre-native), since rust-cache never caches C/C++ objects and sccache does. So I benchmarked all the configurations on my fork before proposing anything.

Setup: the linux-gnu release build (`just build-release`), ubuntu-latest, two rounds per config: first run populates the cache, second run measures. Branch with the workflow: [`john/ci-sccache-bench`](https://github.com/johncarmack1984/martin/tree/john/ci-sccache-bench).

| config | cold | warm | warm build step | cache footprint |
|---|---|---|---|---|
| no cache | 14m46s | 13m06s | 11m33s | — |
| rust-cache | 13m52s | **8m14s** | **6m46s** (12s restore) | 997MB, one entry |
| sccache (GHA backend, `CMAKE_*_COMPILER_LAUNCHER` wired) | 15m05s | 10m09s | 8m31s | 1.7GB across 1,379 entries |
| both stacked | 13m53s | 7m04s | 5m38s | both footprints |

Yuri's verdict in #950 (2023) remains true in 2026: sccache loses to rust-cache by ~2 minutes warm, even with the C/C++ launchers wired (85.7% hit rate, plus 197 misses on a byte-identical rebuild: the keys aren't deterministic). Stacking both looks fastest on paper, but warm it only saw six compile requests: rust-cache's restored target/ leaves sccache nothing to do, so the delta is noise and the quota cost doubles. Case closed, with numbers this time.

What the benchmark DID surface is that the 10GB cache quota is the real constraint, and right now it's not spent where the minutes are:

- The quota sits at 9.9GB/10GB. Four rust entries account for 9.35GB (DuckDB-coverage alone is 4.7GB).
- `test-cog` spends 1.1GB caching a job that takes 1.6 minutes.
- `unit-test-rust` on macOS sets `cache: true`, but there is no macOS entry in the cache. It's being evicted before it can ever hit. That flag is silently doing nothing.
- Meanwhile the multi-OS e2e job compiles `martin-e2e-tests` from scratch every run, uncached: 13.7 minutes of the Windows leg's 17.7. The benchmark's ~45% warm build-step saving applies to exactly this kind of compile.

[...]

PR #3169, opened 2026-08-25
build: drop debuginfo from the bundled DuckDB build

johncarmack1984:
Ref #3156

The DuckDB job's cache entry is 4.9GB (run [32820471902](https://github.com/maplibre/martin/actions/runs/32820471902) saves `4894610292` bytes), half the repo's 10GB quota, so it is the first entry evicted on any Cargo.lock bump and takes the others with it. `libduckdb-sys` builds DuckDB's C++ through `cc`, which reads the profile's debuginfo setting, so the dev profile compiles all of DuckDB with `-g`. Adding `libduckdb-sys.debug = false` to the existing `[profile.dev.package]` overrides takes `libduckdb.a` from 1.75GB to 268MB per copy and the saved cache entry from 4.9GB to 1.85GB (`4894591590` → `1851514928` bytes, measured on my fork: [before](https://github.com/johncarmack1984/martin/actions/runs/32854032587), [after](https://github.com/johncarmack1984/martin/actions/runs/32854035267)). Martin's own code keeps full debuginfo; DuckDB frames lose line info in backtraces and get `NDEBUG`, which is how DuckDB ships its own builds.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -203,7 +203,7 @@ sqlx-macros.opt-level = 3
 # See https://docs.rs/insta/latest/insta/#optional-faster-runs
 insta.opt-level = 3
 similar.opt-level = 3
-# The bundled DuckDB C++ build takes its debuginfo setting from here; with `-g` libduckdb.a is 1.7GB
+# very large debug bundles
 libduckdb-sys.debug = false
 
 [profile.release]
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -203,6 +203,8 @@ sqlx-macros.opt-level = 3
 # See https://docs.rs/insta/latest/insta/#optional-faster-runs
 insta.opt-level = 3
 similar.opt-level = 3
+# The bundled DuckDB C++ build takes its debuginfo setting from here; with `-g` libduckdb.a is 1.7GB
```
CommanderStorm:
```suggestion
# very large debug bundles
```

CommanderStorm (approved):
thanks for looking into this, this will improve CI a bunch ❤️

PR #3178, opened 2026-08-25
refactor(reload): gather the tile reloaders into TileReloaders

johncarmack1984:
Follow-up to #3130, second piece of the split.

martin.rs and martin-cp each wired the reloader-owned kinds by hand, including a duplicated PostgreSQL init dance. TileReloaders now owns that wiring: init() constructs every kind and initializes the PostgreSQL sources through the catalog's on_invalid policy, start() spawns the reload loops. This is the seam the rest of the #2862 follow-up hangs on: routing the file kinds through init() becomes a change inside TileReloaders instead of both binaries. The global ProcessConfig build moves in as-is; the process-config seam stays the follow-up gated on hillshade (#3167).


martin/src/config/file/tiles/reload/mod.rs
```diff
@@ -8,3 +8,158 @@ pub mod mbtiles;
...
+pub use reloaders::TileReloaders;
+
+#[cfg(any(
+    feature = "mbtiles",
+    feature = "unstable-cog",
+    feature = "geojson",
+    feature = "pmtiles",
+    feature = "postgres"
+))]
+mod reloaders {
```
CommanderStorm: please mvoe this to an proper module

CommanderStorm (changes requested):
I like the change, but I hate where it was added....
move it to an proper module and this has my +1

for bonus points:
good docs is short, one sentence per line and has no weird semicolons, colons, em dashes or other artefacts.

martin/src/config/file/tiles/reload/mod.rs
johncarmack1984: Done and done, a48cd72. moved to module and better comments.

CommanderStorm (approved):
Thanks, much cleaner

PR #3180, opened 2026-08-26
chore: implement hillshade-postprocessing

CommanderStorm:
this PR adds hillshade postprocessing which is a fairly common, but still pretty handy optimisation to save on-device resources, network transit but requries a good bit of CPU time on the server.

To allevieate the obvious performance issue, I use LLVM autovectorisation with the multiversion crate and a few tricks to get this down to 10-15ms per tile, despite procesing 9 hillshade tiles.

Here is one of the results (for further details on what the options are, the docs are something I spend a good chunk of time on, so that is a good start)


docs/content/files/generated_config.md
```diff
@@ -148,6 +148,43 @@ keep_alive: 75
...
+    azimuth: null
+    # Separation between lit and shadowed slopes.
+    # Defaults to `1.9`.
+    contrast: null
+    # How strongly high terrain deepens the contrast.
+    # Defaults to `2.5`.
+    elevation_scale: null
+    # Output image format.
+    # Defaults to `png`.
+    format: png
```
nyurik, line 177: use `null` here?  I'm not sure i like the overall approach , lets comment them all out, but set their values to defaults, e.g. `# format: png` - and we could apply this everywhere in config too to keep it both shorter and in default mode by default

docs/content/postprocessing/hillshade.md
```diff
@@ -0,0 +1,127 @@
...
+tags:
+  - hillshade
+  - raster
+  - tile-sources
+  - configuration
+---
+
+# Hillshade
+
+A hillshade is the grey relief image a map multiplies over its basemap to make terrain look three-dimensional.
```
nyurik: "overlay" is more understanable, but we could specify that it is using multiplication as the process(?) of overlaying

docs/content/postprocessing/hillshade.md
```diff
@@ -0,0 +1,127 @@
...
+Martin can bake one from a source that serves *normal maps* in Mapzen's encoding - tiles whose pixels store which way the ground faces instead of a colour - which **spares the client this work** and lets the result be **compressed further**.
```
nyurik: please add something like "(the left image)"

docs/content/postprocessing/hillshade.md
```diff
@@ -0,0 +1,127 @@
...
+## Configuration
+
+Set `convert_to_hillshade` on the source serving the normal maps:
+
+```yaml
+passthrough:
+  sources:
+    terrain:
+      url: https://elevation-tiles-prod.s3.amazonaws.com/normal/{z}/{x}/{y}.png
+      format: png
```
nyurik: lets remove default values like 'png' here

docs/content/postprocessing/hillshade.md
nyurik: COLOR!

docs/content/postprocessing/hillshade.md
```diff
@@ -0,0 +1,127 @@
...
+      maxzoom: 14
+      convert_to_hillshade: auto
+```
+
+`GET /terrain/12/2100/1300` now returns a 512x512 lossless PNG of shaded relief, replacing the normal map it was baked from.
+
+`auto` bakes with every default, a map of settings (f.ex. `convert_to_hillshade: {altitude: 60}`) overrides said defaults, and `disabled` (the default for `convert_to_hillshade`) is the same as omitting the key.
```
nyurik: lets not invent abbr. esp when we have `e.g.`

docs/content/postprocessing/hillshade.md
```diff
@@ -0,0 +1,127 @@
...
+With `allow_request_overrides: true`, a request may override any of the seven lighting parameters by name:
+
+```text
+GET /terrain/12/2100/1300?azimuth=90&toon_bands=0
+```
+
+`padding`, `format`, and `allow_request_overrides` itself are not overridable.
+Values that are out of range or not a number are rejected with `400`, while query parameters that are not hillshade settings are ignored.
+
+!!! warning "This is an expensive feature"
```
nyurik: move the warning up to the top of this section, focusing on "don't use it unless you are doing an internal development and want to figure out which params look best for your specific map"

docs/content/postprocessing/hillshade.md
```diff
@@ -0,0 +1,127 @@
...
+    Baking costs tens of milliseconds of CPU per tile, which makes hillshading an expensive feature.
```
nyurik: just say its an expensive operation

docs/content/postprocessing/hillshade.md
```diff
@@ -0,0 +1,127 @@
...
+    If you need this, this means you should take care in how you configure the CDN in front.
+    A hillshade for fixed settings never changes (+- data changes), so it is near-perfectly cacheable at the edge.
```
nyurik:
```suggestion
    A hillshade for fixed settings never changes (+/- data changes), so it is near-perfectly cacheable at the edge.
```

nyurik (approved):

PR #190, opened 2021-02-27
Jwt

mdtrooper:
Yes.

The JWT is coded into Martin. And it is runs fine.

Regards :beers:  .

PR #3184, opened 2026-08-26
docs(nginx): document request authentication with auth_request

### LINKED ISSUE #78: Add security to API with JWT
First thing, thanks your project, its awesome.

I am working in a small company (its name is [dotGIS](https://www.dotgiscorp.com/) ) and we are using Martin for some projects.

We need in the api an authentication method such as JWT because it is a common standard.

We can help your project...personally I want to touch martin code...I found some [libraries for JWT in Rust](https://docs.rs/jsonwebtoken/7.1.0/jsonwebtoken/)...but I need to check which one is better.

Well. Could I fork Martin and add (as branch) support for JWT?

Regards.

johncarmack1984:
Closes #78

Martin leaves authentication to the reverse proxy, and the closing note on #190 points at the nginx and apache pages for how. At the moment, neither page says a word about it. This adds an "Authenticating requests" section to the nginx page built on auth_request, the recipe gbip left on #78 in 2021, and one line on the reverse proxy index saying Martin does not authenticate requests.

### DIFF
--- a/docs/content/run-with-nginx.md
+++ b/docs/content/run-with-nginx.md
@@ -32,6 +32,38 @@ location ~ /tiles/(?<fwd_path>.*) {
 }
 ```
 
+### Authenticating requests
+
+Martin does not authenticate requests, so put the check in NGINX.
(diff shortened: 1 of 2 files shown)


CommanderStorm (approved):
Thanks

PR #3185, opened 2026-08-27
fix(mbtiles): avoid full tiles scan when computing min/max zoom

### LINKED ISSUE #3134: Slow startup with large Tippecanoe MBTiles since Martin 1.13
Since Martin 1.13, startup takes several minutes while resolving large MBTiles sources. The same files started quickly with previous versions.

Most affected files were generated with Tippecanoe and use a normalized `map`/`images` schema with an indexed `tiles` view.

This appears related to #2983 which validates metadata against:

```sql
SELECT MIN(zoom_level), MAX(zoom_level)
FROM tiles;
```

https://github.com/maplibre/martin/blob/06f73d769cc7a487714878612b9d77d20f2ab14c/mbtiles/src/queries.rs#L130

SQLite scans the entire index and, for normalized files, may execute the map/images join. Adding map(zoom_level) and converting to flat only partially helps.
This equivalent query completes in milliseconds:

```sql
SELECT
  (SELECT MIN(zoom_level) FROM tiles),
  (SELECT MAX(zoom_level) FROM tiles);
```

Environment:
Martin 1.13 and 1.14
MBTiles 0.19.x
Tippecanoe 2.79.0
Archives up to several GB
Could the startup validation use the scalar-subquery form or provide an option to disable it?

johncarmack1984:
Closes #3134. Change was two lines, the rest is the regression test.

Since 1.13 (#2983) startup checks the metadata zoom range against the tiles with `SELECT min(zoom_level), max(zoom_level) FROM tiles`. SQLite only takes the [min/max index shortcut](https://www.sqlite.org/optoverview.html#the_min_max_optimization) when a query has one aggregate, so that form walks every tile. On a normalized Tippecanoe file it walks the `map` join `images` view, which is where the minutes go. Two scalar subqueries get the shortcut each. A test snapshots the plan of the original query, so it fails the day SQLite stops needing the rewrite and the workaround can go.

`compute_min_max_zoom` on 200k synthetic tiles, criterion, local:

| Layout     | main    | this PR |
|------------|---------|---------|
| flat       | 5.7 ms  | 6.1 µs  |
| normalized | 43.7 ms | 7.6 µs  |

Martin startup on 5M tiles, release build, warm cache. 203 ms is the fixed floor.

| Layout                  | main    | this PR |
|-------------------------|---------|---------|
| flat, 480 MB            | 367 ms  | 203 ms  |
| normalized view, 776 MB | 1650 ms | 203 ms  |


mbtiles/src/queries.rs
```diff
@@ -126,22 +126,19 @@ fn validate_zoom(zoom: Option<i64>, zoom_name: &'static str) -> MbtResult<Option
...
-SELECT min(zoom_level) AS min_zoom,
-       max(zoom_level) AS max_zoom
-FROM tiles;"
-    )
-    .fetch_one(conn)
-    .await?;
-
-    let min_zoom = validate_zoom(info.min_zoom, "zoom_level")?;
-    let max_zoom = validate_zoom(info.max_zoom, "zoom_level")?;
+    let info = query(MIN_MAX_ZOOM_SQL).fetch_one(conn).await?;
```
johncarmack1984: The test needs the exact SQL string, and `query!` only takes a literal. So the query lives in a const and goes through plain `query()` now. Happy to keep the macro and duplicate the string in the test if you'd rather.

CommanderStorm: Please inline and use query!, yes

mbtiles/src/queries.rs
```diff
@@ -126,22 +126,19 @@ fn validate_zoom(zoom: Option<i64>, zoom_name: &'static str) -> MbtResult<Option
     }
 }
 
+const MIN_MAX_ZOOM_SQL: &str = "
+SELECT (SELECT min(zoom_level) FROM tiles) AS min_zoom,
+       (SELECT max(zoom_level) FROM tiles) AS max_zoom;";
```
CommanderStorm: I think this is a bug in salute to have to rewrite it like this. 
Please make sure to add a comment stating as such.

mbtiles/src/queries.rs
johncarmack1984: Done, inlined in 63a3937a

mbtiles/src/queries.rs
```diff
@@ -163,3 +160,39 @@ pub async fn action_with_rusqlite(
...
     action(&rusqlite_conn)
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::metadata::anonymous_mbtiles;
+
+    #[actix_rt::test]
+    async fn min_max_zoom_uses_index_seek() {
```
CommanderStorm, line 196: Please document again that this is a workaround test that we will revert once sqlite gets smart enough to do the rewrite itself.

mbtiles/benches/min_max_zoom.rs
CommanderStorm, line 1: Let's remove the bench. I trust you to have tested and benched it. 
It is unlikely to be useful in the future

mbtiles/src/queries.rs
```diff
@@ -163,3 +160,39 @@ pub async fn action_with_rusqlite(
...
+            assert!(
+                plan.iter().any(|step| step.starts_with("SEARCH ")),
+                "no index seek in plan: {plan:?}"
+            );
+            assert!(
+                !plan
+                    .iter()
+                    .any(|step| step.starts_with("SCAN ") && step != "SCAN CONSTANT ROW"),
+                "full scan in plan: {plan:?}"
+            );
```
CommanderStorm: Seems brittle. 
Please insta assert on the explanation. 
We can always delete the test if it turns out to be labour intensive to maintain or require lots of blessing the results.

CommanderStorm (changes requested):
I'd like a bit of cleanup, but otherwise looks... Plausible

mbtiles/src/queries.rs
johncarmack1984: Added, with the sqlite doc link. f783cf1d

mbtiles/src/queries.rs
johncarmack1984: Done, f783cf1d

mbtiles/benches/min_max_zoom.rs
johncarmack1984: Removed. Numbers stay in the body as a local run.

mbtiles/src/queries.rs
johncarmack1984: Switched to inline insta snapshots, one per layout. f783cf1d

mbtiles/src/queries.rs
```diff
@@ -163,3 +165,65 @@ pub async fn action_with_rusqlite(
...
+    use super::*;
+    use crate::metadata::anonymous_mbtiles;
+
+    async fn min_max_zoom_plan(script: &str) -> String {
+        let (_, mut conn) = anonymous_mbtiles(script).await;
+        query(
+            "
+EXPLAIN QUERY PLAN
+SELECT (SELECT min(zoom_level) FROM tiles) AS min_zoom,
+       (SELECT max(zoom_level) FROM tiles) AS max_zoom;",
```
CommanderStorm: as a test, we want this

```suggestion
SELECT min(zoom_level), max(zoom_level) FROM tiles;",
```

mbtiles/src/queries.rs
```diff
@@ -163,3 +165,64 @@ pub async fn action_with_rusqlite(
...
+        .await
+        .unwrap()
+        .iter()
+        .map(|row| row.get::<String, _>("detail"))
+        .collect::<Vec<_>>()
+        .join("\n")
+    }
+
+    /// Workaround test for the query rewrite in [`compute_min_max_zoom`].
+    /// Delete it once SQLite applies the min/max optimization to both aggregates in one query.
```
CommanderStorm:
```suggestion
    /// Delete this test and inline query above once sqlite does not do an entire query scan here..
```

CommanderStorm (approved):
Thanks

PR #3189, opened 2026-08-30
fix(catalog): serialize the catalog in a stable key order

johncarmack1984:
The four catalog maps were `HashMap`s, so `/catalog` serialized its keys in hash order. That order is randomized per process, meaning two servers over the same sources (or the same server before and after a restart) answered the same catalog with the keys shuffled, and byte-level caching of the response could never hit. `BTreeMap` keys them in sorted order instead.


e2e-tests/tests/catalog.rs
```diff
@@ -0,0 +1,69 @@
...
+    let mut first = martin_with_every_source_kind().await;
+    let mut second = martin_with_every_source_kind().await;
+
+    let first_catalog = first.get("/catalog").await;
+    let second_catalog = second.get("/catalog").await;
+    assert_eq!(first_catalog.status(), 200);
+    assert_eq!(second_catalog.status(), 200);
+    assert_sorted_and_populated(&first_catalog);
+    assert_sorted_and_populated(&second_catalog);
+    assert_eq!(first_catalog.text(), second_catalog.text());
```
CommanderStorm: This is the important variant here..
The rest can/should be dropped since very britte testing.
You can/should also insta::assert the actual snapshot output, so that this is also fixed, not only via the other tests. 

```suggestion
    assert_eq!(first_catalog.text(), second_catalog.text());
```

martin-core/src/resources/fonts/mod.rs
```diff
@@ -115,7 +115,7 @@ fn get_available_codepoints(face: &Face) -> Option<GetGlyphInfo> {
 }
 
 /// Catalog mapping font names to metadata (e.g., "Arial" -> `CatalogFontEntry`).
-pub type FontCatalog = HashMap<String, CatalogFontEntry>;
+pub type FontCatalog = BTreeMap<String, CatalogFontEntry>;
```
CommanderStorm, line 118: iteration order is not actually an important invariant for consumers.
So lets not bother with it, instead please use [FXHashMap](https://docs.rs/fxhash/latest/fxhash/type.FxHashMap.html) or an xxhash based variant, so not an randomised variant.

CommanderStorm (commented):
I'd like two minor changes, otherwise LGTM

e2e-tests/tests/catalog.rs
CommanderStorm: possibly using pretty_assertions since if this would create an diff, this is important to know where it happened xD

martin-core/src/resources/fonts/mod.rs
johncarmack1984: Tried it: FxHashMap fails the two-server test 9/10, BTreeMap 15/15. A fixed hasher pins the ideal bucket, but hashbrown breaks ties by insertion order, and these collect from a DashMap that iterates in random order per process. Reproduced with rustc-hash alone (fxhash is unmaintained, so I used its successor), w/ no martin involved, and found the same 8 keys come out in 4 different orders. The randomness just moves into insertion order. 

Sorting was the cheapest fix I found; happy to try another way if you've got one in mind.

e2e-tests/tests/catalog.rs
johncarmack1984: re: brittle: done in e4012dcd: helper gone, byte-equality kept, plus an insta json snapshot of the whole document (insta sorts keys, so it pins content without claiming order).

re: pretty_assertions: done in same commit. incidentally, how I found where FxHashMap diverged :-D

CommanderStorm (approved):
Thanks, LGTM

PR #3190, opened 2026-08-30
refactor(reload): publish local file kinds through init()

johncarmack1984:
Follow-up to #3178, third piece of the split.

The local file kinds (mbtiles, local pmtiles, cog, geojson) were the last sources still written to the catalog from two scans: resolve() enumerated their directories at startup and each reloader seeded its baseline from a second pass. They now publish through ReloadDriver::init() before serving, the shape #3130 gave PostgreSQL for #2862, and the wiring lands inside TileReloaders as #3178 set up. FsDiscovery carries a FileKind and BuiltSource provenance so --save-config keeps serializing discovered files, with paths spelled the way the config spelled them. An unreadable configured directory now warns through on_invalid and drops out of the watch set instead of breaking the notify trigger for its siblings. Remote pmtiles prefixes still start from an empty baseline; moving them through init() and deleting Baseline is the next piece.


martin/src/config/file/file_config.rs
```diff
@@ -1738,40 +1719,28 @@ mod folder_source_tests {
     #[tokio::test]
     async fn one_good_one_bad() {
         let (sources, warnings) = resolve_mixed_dir(1, 1).await;
-        assert_yaml_snapshot!(sources, @"- good_0");
-        assert_yaml_snapshot!(warnings, @r#"- "Path <DIR>/bad_0.tiles: Source path is not a file: <DIR>/bad_0.tiles""#);
+        assert_yaml_snapshot!(sources, @"[]");
+        assert_yaml_snapshot!(warnings, @"[]");
```
CommanderStorm: The testcases don't seem very healthyor intentional.. what is going on here?

martin/src/config/file/tiles/discovery/fs.rs
```diff
@@ -37,29 +38,40 @@ struct ConfiguredSource {
...
+    src: FileConfigSrc,
 }
 
 /// A [`Discovery`] that enumerates source files under the watched directories.
 pub struct FsDiscovery {
-    directories: Vec<PathBuf>,
+    kind: FileKind,
+    /// Pairs of the configured path and its canonical form.
+    /// Scanning uses the canonical path.
+    /// Config write-back uses the configured spelling so saved paths stay relative when the config was.
```
CommanderStorm: The canonical form can be derived from the relative one. 
Let's just store the relative one here and derive the other.

martin/src/config/file/tiles/discovery/fs.rs
```diff
@@ -293,6 +349,26 @@ mod tests {
...
+        ))]
+        return FileKind::GeoJson;
+        #[cfg(all(
+            not(feature = "mbtiles"),
+            not(feature = "pmtiles"),
+            not(feature = "geojson"),
+            feature = "unstable-cog"
+        ))]
+        return FileKind::Cog;
+    }
```
CommanderStorm: No, needs to be inlined into the tests, this looks very wrong. 
Keep tests DAMP please

martin/src/config/file/tiles/discovery/fs.rs
```diff
@@ -426,4 +505,75 @@ mod tests {
...
+        assert_eq!(
+            discovery.directories().len(),
+            1,
+            "the readable sibling stays watched"
+        );
+        let snapshot = discovery.discover().await.expect("discover").sources;
+        assert!(
+            snapshot.contains_key("alpha"),
+            "the readable sibling still serves"
+        );
```
CommanderStorm: No assertions with contains please. The stronger assertion is to insta assert it.

martin/src/reload.rs
```diff
@@ -3,19 +3,42 @@ use std::collections::{BTreeMap, BTreeSet};
...
+pub enum FileKind {
+    #[cfg(feature = "mbtiles")]
+    Mbtiles,
+    #[cfg(feature = "pmtiles")]
+    Pmtiles,
+    #[cfg(feature = "unstable-cog")]
+    Cog,
+    #[cfg(feature = "geojson")]
+    GeoJson,
+}
```
CommanderStorm: No duckdb?

CommanderStorm (changes requested):
Please review the testcases again. This doesn't seem quite right

martin/src/config/file/tiles/discovery/fs.rs
johncarmack1984: Done, only the configured path is stored now and the canonical form is derived where it's needed. cfb397ef

martin/src/config/file/file_config.rs
johncarmack1984: Yep, they were asserting nothing, my bad. The directory scan moved from resolve() into FsDiscovery and the tests didn't follow. Moved them to fs.rs, where they run through init() against a real catalog, and found along the way that the unreadable-dir warnings never actually reached init(). Both fixed in cfb397ef.

martin/src/config/file/tiles/discovery/fs.rs
johncarmack1984: Inlined as FileKind::Mbtiles and the module is cfg'd on mbtiles instead. cfb397ef

martin/src/config/file/tiles/discovery/fs.rs
johncarmack1984: Insta snapshots now, and the early return for the root case is gone too. cfb397ef

martin/src/reload.rs
johncarmack1984: No reloader and no directory scan for DuckDB (GeoParquetLocation rejects directories), so the catalog never holds a discovered DuckDB source for --save-config to write back. The enum doc now says directory-discovered. cfb397ef

CommanderStorm (approved):

CommanderStorm:
thanks ❤️

PR #3193, opened 2026-08-31
refactor(config): resolve the process config once, into ResolvedProcess

### LINKED ISSUE #3192: refactor: resolve the process config once, into a concrete type
@CommanderStorm, you flagged this on #3167 ("lots of None which is suspicious... the entire process config architecture is getting out of place a bit") and asked to wait for hillshade. Hillshade is 2/3 of the way in as of today (#3180, #3181), so here's what I've been thinking in the meantime.

Today `ProcessConfig` plays two roles: the serde type users write (everything `Option`) and the resolved per-source config we pass around (still everything `Option`). Nine files touch it, and each field resolves with its own precedence rule: mlt/mvt override as a unit, `cache_control` walks per-source > source-type > global, `convert_to_hillshade` is per-source only. Every consumer re-learns which `None`s are meaningful, and the feature gates get re-stated at every call site, which is how the import in #3191 drifted.

Proposal: keep `ProcessConfig` as the serde type, resolve it once during `finalize()` into a concrete resolved type (no `Option`s, defaults applied, precedence baked in), and hand only that to the reloaders and tile pipeline. Feature gates then live on the resolved type's fields instead of being re-stated at every call site.

I'd pick this up after #3183 and #3190 land since it rewires the same call sites. Does that match what you had in mind? Happy to adjust before writing any code.

johncarmack1984:
Closes #3192

`ProcessConfig` was both the settings users write and the value the tile pipeline runs on, all `Option` in both roles. Every consumer re-learned which `None` meant unset and which meant off, and hillshade/contour re-resolved their parameters on every request.

Now `ProcessConfig` is only the per-level layer. `Config::finalize` folds global > source-type > per-source once into `ResolvedProcess`, and that is the only type the catalog, the reloaders and `content.rs` see.

- `mlt` / `mvt` are two-state enums with the encoder already built
- `hillshade` / `contour` carry their `Resolved*` types, so the two `validate_*` passes fold into the same walk
- `cache_control` stays `Option`, no header is a real state
- discovery keeps the kind-level layer to fold per-source overrides over and resolves on the way out

No behaviour change intended.

Side fix: the `mlt`-gated PG discovery test was missing `cache_control` and did not compile under `--features test-pg --all-targets`.

### DIFF
--- a/martin/src/srv/tiles/process/mod.rs
+++ b/martin/src/srv/tiles/process/mod.rs
@@ -150,7 +150,6 @@ pub fn apply_pre_cache_processors(
     let tile = if accepted == Some(Format::Mlt) && tile.info.format == Format::Mvt {
         match config.mlt {
             MltConversion::Encode(cfg) => convert_mvt_to_mlt(tile, cfg)?,
-            // Explicitly opted out - serve the original MVT bytes.
             MltConversion::Disabled => tile,
         }
     } else if accepted == Some(Format::Mvt)
--- a/martin/src/srv/tiles/process/mod.rs
+++ b/martin/src/srv/tiles/process/mod.rs
@@ -20,8 +20,6 @@ use martin_core::tiles::hillshade::HillshadeError;
 #[cfg(all(feature = "mlt", feature = "_tiles"))]
 use martin_tile_utils::Format;
 #[cfg(all(feature = "mlt", feature = "_tiles"))]
-use mlt_core::encoder::EncoderConfig;
-#[cfg(all(feature = "mlt", feature = "_tiles"))]
 use to_mlt::convert_mvt_to_mlt;
 #[cfg(all(feature = "mlt", feature = "_tiles"))]
 use to_mvt::convert_mlt_to_mvt;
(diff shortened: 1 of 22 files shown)


martin/src/srv/tiles/process/mod.rs
```diff
@@ -151,23 +148,14 @@ pub fn apply_pre_cache_processors(
...
-            // No level configured anything -> use defaults.
-            None | Some(MltProcessConfig::Auto) => {
-                convert_mvt_to_mlt(tile, EncoderConfig::default())?
-            }
-            Some(MltProcessConfig::Explicit(cfg)) => {
-                convert_mvt_to_mlt(tile, EncoderConfig::from(cfg.clone()))?
-            }
+        match config.mlt {
+            MltConversion::Encode(cfg) => convert_mvt_to_mlt(tile, cfg)?,
             // Explicitly opted out - serve the original MVT bytes.
```
CommanderStorm:
```suggestion
```

CommanderStorm (approved):
Seems like a much cleaner interface 👍🏻

PR #3194, opened 2026-08-31
feat(cache): add an opt-in DELETE /cache/{source_id} purge route

### LINKED ISSUE #1707: Cache invalidation
I would like to know for possible solutions for invalidation of cached tiles. When I update any feature in my DB, it doesn't reflect on UI, even after refresh on UI.

My setup
UI <-> MARTIN <-> POSTGIS

1 solution is setting cache_size to 0, but that impacts performance badly.

johncarmack1984:
Closes #1707

Martin can drop a source's cached tiles when a file source changes, but nothing outside the process can ask for it. A PostgreSQL user whose rows change has only the TTL knob or a full restart.

This adds `DELETE /cache/{source_id}`, behind `purge_endpoint: true` in the config (off by default). It calls the same `invalidate_source` the reloaders use and answers `200` with a body saying whether tiles were dropped or no tile cache is configured, or `404` for an unknown source. Like the rest of the server it carries no auth of its own, so it goes behind the same proxy rules (`auth_request` on `/cache/` with the nginx setup from #3184).

A Postgres `LISTEN/NOTIFY` hook so a trigger can do this in-band is the natural next step, but this small piece works on its own, so I'll save the trigger for a follow-up to close out the ticket.

### DIFF
--- a/docs/content/files/generated_config.md
+++ b/docs/content/files/generated_config.md
@@ -502,8 +502,7 @@ postgres:
 preferred_encoding: brotli
 # Serve `DELETE /cache/{source_id}`, which drops that source's cached tiles. [default: false]
 #
-# The route has no authentication of its own.
-# Put it behind the same proxy rules as the rest of the server.
+# The route has no authentication of its own, so to prevent DOS please put it behind a reverse proxy.
 purge_endpoint: false
 # Set the URL path prefix for all API routes.
 # When set, Martin will serve all endpoints under this path prefix.
--- a/docs/content/files/generated_config.md
+++ b/docs/content/files/generated_config.md
@@ -113,6 +113,12 @@ cors:
   # '*' will use the requests `ORIGIN` header
   origin:
     - https://example.org
+# Optional endpoints, each off unless enabled here.
+endpoints:
+  # Serve `DELETE /cache/{source_id}`, which drops that source's cached tiles. [default: false]
+  #
+  # The route has no authentication of its own, so to prevent DOS please put it behind a reverse proxy.
+  purge_cache: false
 # Font configuration
 fonts:
   # Cache configuration for fonts.
@@ -500,12 +506,6 @@ postgres:
 # `gzip` is faster, but `brotli` is smaller, and may be faster with caching.
 # Default could be different depending on Martin version.
 preferred_encoding: brotli
-# Optional endpoints, each off unless enabled here.
-endpoints:
-  # Serve `DELETE /cache/{source_id}`, which drops that source's cached tiles. [default: false]
-  #
-  # The route has no authentication of its own, so to prevent DOS please put it behind a reverse proxy.
-  purge_cache: false
 # Set the URL path prefix for all API routes.
 # When set, Martin will serve all endpoints under this path prefix.
 # This allows Martin to be served under a subpath when behind a reverse proxy (e.g., Traefik).
--- a/docs/content/files/generated_config.md
+++ b/docs/content/files/generated_config.md
@@ -500,6 +500,11 @@ postgres:
 # `gzip` is faster, but `brotli` is smaller, and may be faster with caching.
 # Default could be different depending on Martin version.
 preferred_encoding: brotli
+# Serve `DELETE /cache/{source_id}`, which drops that source's cached tiles. [default: false]
+#
+# The route has no authentication of its own.
+# Put it behind the same proxy rules as the rest of the server.
+purge_endpoint: false
 # Set the URL path prefix for all API routes.
 # When set, Martin will serve all endpoints under this path prefix.
 # This allows Martin to be served under a subpath when behind a reverse proxy (e.g., Traefik).
--- a/docs/content/files/generated_config.md
+++ b/docs/content/files/generated_config.md
@@ -500,10 +500,12 @@ postgres:
 # `gzip` is faster, but `brotli` is smaller, and may be faster with caching.
 # Default could be different depending on Martin version.
 preferred_encoding: brotli
-# Serve `DELETE /cache/{source_id}`, which drops that source's cached tiles. [default: false]
-#
-# The route has no authentication of its own, so to prevent DOS please put it behind a reverse proxy.
-purge_endpoint: false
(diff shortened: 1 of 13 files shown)


e2e-tests/tests/cache_purge.rs
```diff
@@ -0,0 +1,52 @@
...
+pmtiles:
+  sources:
+    pmt: tests/fixtures/pmtiles/png.pmtiles
+";
+
+const DISABLED: &str = "
+pmtiles:
+  sources:
+    pmt: tests/fixtures/pmtiles/png.pmtiles
+";
```
CommanderStorm: please inline via indoc. Tests are there to be DAMP, not DRY

e2e-tests/tests/cache_purge.rs
```diff
@@ -0,0 +1,52 @@
...
+        .config(ENABLED)
+        .start()
+        .await
+        .expect("failed to start martin");
+
+    assert_eq!(martin.get("/pmt/0/0/0").await.status(), 200);
+    let purged = martin.delete("/cache/pmt").await;
+    assert_eq!(purged.status(), 204);
+    assert!(purged.body().is_empty());
+    assert_eq!(martin.get("/pmt/0/0/0").await.status(), 200);
```
CommanderStorm: I don't think this is actually testing what the testcase is claiming. How about a passtrhough source with an mock http server and asserting the number of http calls?

martin/src/srv/cache.rs
```diff
@@ -0,0 +1,32 @@
...
+/// Answers `404` for an unknown source and `204` otherwise, also when no tile cache is configured.
+#[cfg_attr(
+    feature = "unstable-schemas",
+    utoipa::path(
+        delete,
+        path = "/cache/{source_id}",
+        params(("source_id" = String, Path, description = "Source ID")),
+        responses(
+            (status = 204, description = "The source's cached tiles are gone"),
+            (status = 404, description = "No such source"),
```
CommanderStorm, line 14: I think we need to have a body for the both to communicate what happened.
The reason is simply because otherwise the not configured and no such source case look identical...

docs/content/files/generated_config.md
```diff
@@ -500,6 +500,11 @@ postgres:
 # `gzip` is faster, but `brotli` is smaller, and may be faster with caching.
 # Default could be different depending on Martin version.
 preferred_encoding: brotli
+# Serve `DELETE /cache/{source_id}`, which drops that source's cached tiles. [default: false]
+#
+# The route has no authentication of its own.
+# Put it behind the same proxy rules as the rest of the server.
```
CommanderStorm:
```suggestion
# The route has no authentication of its own, so to prevent DOS please put it behind an reverse proxy.
```

CommanderStorm (changes requested):
mostly LGTM, there is some imprecice testing that is not actually testing what you claim, but otherwise 👍🏻

e2e-tests/tests/cache_purge.rs
johncarmack1984: Inlined via formatdoc in 628bf963.

e2e-tests/tests/cache_purge.rs
johncarmack1984: You're right, it asserted nothing about the cache. Rewrote it in 628bf963: passthrough over the harness's wiremock server, one upstream call for two GETs, purge, then a second upstream call.

martin/src/srv/cache.rs
johncarmack1984: Both answer 200 with a body now in 628bf963: "Purged the cached tiles of source X" vs "Source X has no cached tiles, tile caching is disabled"; unknown source stays 404.

docs/content/files/generated_config.md
johncarmack1984: Applied in 628bf963

martin/src/srv/cache.rs
```diff
@@ -0,0 +1,35 @@
+use actix_web::web::{Data, Path};
+use actix_web::{HttpResponse, Responder, route};
+
+use crate::tile_source_manager::TileSourceManager;
+
+/// Drops every cached tile of one source.
+///
+/// The body says whether tiles were dropped or no tile cache is configured, and an unknown source answers `404`.
```
CommanderStorm:
```suggestion
```

martin/src/srv/cache.rs
```diff
@@ -0,0 +1,35 @@
...
+#[cfg_attr(
+    feature = "unstable-schemas",
+    utoipa::path(
+        delete,
+        path = "/cache/{source_id}",
+        params(("source_id" = String, Path, description = "Source ID")),
+        responses(
+            (status = 200, description = "The source's cached tiles are gone, or there was no tile cache to drop them from. The body says which.", body = String),
```
CommanderStorm:
```suggestion
            (status = 200, description = "The source's cached tiles are gone, or tile cacheing is disabled.", body = String),
```

docs/content/files/generated_config.md
```diff
@@ -500,6 +500,10 @@ postgres:
 # `gzip` is faster, but `brotli` is smaller, and may be faster with caching.
 # Default could be different depending on Martin version.
 preferred_encoding: brotli
+# Serve `DELETE /cache/{source_id}`, which drops that source's cached tiles. [default: false]
+#
+# The route has no authentication of its own, so to prevent DOS please put it behind a reverse proxy.
+purge_endpoint: false
```
CommanderStorm: I wonder if we should move this to `endpoints.purge_cache` instead. this way, we can add `endpoints.rendering` or `endpoints.managemenet` (for maybe an style editing API)...

What do you think 🤔

CommanderStorm (approved):
LGTM with one caveat

docs/content/files/generated_config.md
johncarmack1984: I think you're right, and cheapest to change now before it's already on main. Moved to endpoints.purge_cache in 43a500e, mirroring the observability.metrics nesting.

PR #3195, opened 2026-08-31
docs(lambda): private S3 through the execution role, one-stack template, measured sizing

### LINKED ISSUE #1102: Document using Martin as a AWS Lambda using PMTiles on S3, or another cloud provider

johncarmack1984:
Closes #1102

The Lambda page still carried the 2024 TODO list, two items of which have since shipped: Martin signs S3 requests with whatever credentials the runtime provides, and a `pmtiles.paths` prefix serves a whole bucket without listing files in the config. Rewrote the page around a run I did against a private bucket with the v1.14.0 arm64 release: the config file is the handler, the execution role gets `s3:GetObject`, nothing else is needed. Added the single SAM template that produced that run (layer, function, role, URL), and replaced the sizing guesswork with the measured numbers at the Lambda defaults (300 ms init, 43 MB, 25 ms per warm tile). RDS and CloudFront stay listed as not covered.

### DIFF
--- a/docs/content/run-with-lambda.md
+++ b/docs/content/run-with-lambda.md
@@ -6,10 +6,12 @@ tags:
   - lambda
 ---
 
-# Using with AWS Lambda - v0.14+
+# Using with AWS Lambda
 
 Martin can run in AWS Lambda.
 This is useful if you want to serve tiles from a serverless environment, while accessing "nearby" data from a PostgreSQL database or PMTiles file in S3, without exposing the raw file to the world to prevent download abuse and improve performance.
+The bucket can stay private.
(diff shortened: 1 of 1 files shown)


CommanderStorm (approved):
Thanks

PR #3183, opened 2026-08-26
feat: add contour processing

CommanderStorm:
This PR adds contour postprocessing.
Compared to the source impl these are the diffences:
- smaleer since MLT support
- smaller since i16 instead of f64 for elevation (plus renamed `ele`)
- smaller since boolean instead of string for the major lines.

contour

PR #3202, opened 2026-09-01
fix(contour): run traced tiles through the MLT transcoder

johncarmack1984:
The contour path in `get_tile_content_from_one_source` returned the traced tile before `apply_pre_cache_processors`, so a contoured source with `convert_to_mlt` set answered an MLT `Accept` header with an MVT body labelled `application/x-protobuf`. This routes traced tiles through the same processor step other vector tiles already take. Follow-up to #3183, docs already recommend pairing contours with MLT.


CommanderStorm (approved):
Good catch!

PR #3208, opened 2026-09-01
feat(martin-cp): skip the tiles below an empty tile

### LINKED ISSUE #2212: martin-cp: better nullness/bbox handling for vastly increased performance

johncarmack1984:
`martin-cp` now copies zoom by zoom and skips fetching the tiles below an empty tile, whenever the source promises that an empty tile only has empty tiles below it. PostgreSQL table sources make that promise, since a tile's envelope contains its children's envelopes. Function sources and tile archives cannot: a function may return different rows per zoom, and an archive can have gaps in its pyramid (our own `world_cities` fixture does, and skipping unconditionally dropped 6 of its 8 tiles in the e2e). So the promise is a `Source` trait method defaulting to false, not a flag. Second half of #2212 after #2385 took care of the bounds; ostensibly, closes #2212. Measured on the test database's `table_source`, z0 to z10, world bbox, concurrency 3:

| | before | after |
|---|---:|---:|
| wall time | 243 s | 11 s |
| tiles fetched | 1,398,101 | 15,557 |
| tiles skipped | 0 | 1,382,544 |
| tiles written | 14,619 | 14,619 |


CommanderStorm (changes requested):
skip_empty_subtrees should be `true` and I don't think this is an option.
Just skip them outright, no config nessary for this.

johncarmack1984:
Flag dropped in 81c554d9. Skipping outright loses data though: the `world_cities` e2e copies 8 tiles and unconditional pruning wrote 2 of them, because an archive can have gaps in its pyramid (a function source has the same problem, different rows per zoom). So the guarantee lives on the source now: `Source::empty_tile_implies_empty_children`, default false, true for pg table sources. martin-cp prunes when every source in the copy makes the promise. Zero config either way.

docs/content/martin-cp.md
```diff
@@ -61,6 +61,10 @@ You should also consider
     Use `identity` to disable compression.
     Ignored for non-encodable tiles like PNG and JPEG.
 
+!!! note
+    When the source guarantees that an empty tile only has empty tiles below it, `martin-cp` copies zoom by zoom and never fetches the tiles below an empty tile.
+    PostgreSQL table sources give that guarantee, so a sparse table over a large bbox no longer pays one query per empty tile at every zoom.
```
CommanderStorm:
```suggestion
    When the source (such as PG tables) guarantees that an empty tile only has empty tiles below it, `martin-cp` copies zoom by zoom and never fetches the tiles below an empty tile.
    This means that even sparse sources can usually be fairly performant.
```

martin/src/bin/martin-cp.rs
```diff
@@ -444,28 +446,75 @@ async fn write_tiles_to_mbtiles(
 }
 
 /// Fetches tiles concurrently and sends them to the consumer via `tx`.
+///
+/// With `prune_empty_subtrees`, zooms are copied in ascending order and a tile whose parent came back empty is reported empty without being fetched.
```
CommanderStorm:
```suggestion
```

CommanderStorm (approved):

martin/src/config/file/tiles/postgres/resolver/query_functions.rs
```diff
@@ -105,9 +105,11 @@ pub async fn query_available_function(pool: &PostgresPool) -> PostgresResult<Sql
                 .insert(
                     function.clone(),
                     (
+                        // a function may return different rows per zoom, so an empty tile says nothing about its children
                         PostgresSqlInfo::new(
                             query,
                             input_types.len() == 4,
+                            false,
```
CommanderStorm, line 112:
```suggestion
PostgresSqlInfo::new(
query,
input_types.len() == 4,
                        // a function may return different rows per zoom, so an empty tile says nothing about its children
                            false,
```

martin/src/config/file/tiles/postgres/resolver/query_tables.rs
```diff
@@ -254,7 +254,8 @@ FROM (
 
     Ok((
         id,
-        PostgresSqlInfo::new(query, false, info.format_id()),
+        // a table tile is empty only when no geometry intersects its envelope, which contains the envelopes of its children
+        PostgresSqlInfo::new(query, false, true, info.format_id()),
```
CommanderStorm:
```suggestion
        PostgresSqlInfo::new(query, false,
        
        // a table tile is empty only when no geometry intersects its envelope, which contains the envelopes of its children
          true, info.format_id()),
```

PR #3209, opened 2026-09-01
docs: add a tutorial for self-hosting a complete basemap

johncarmack1984:
Addresses an ask from a Slack thread in April 2023; no issue attached. 

Adds an end-to-end guide from an OSM extract to a styled map in the browser, with the tiles, glyphs, sprites, and style all served by one Martin instance. Each piece was already documented, but no page connected them into a working basemap. Every command was run as written in a clean directory and the screenshot is the map they produce. Also adds `maplibregl` and `geofabrik` to the spellcheck dictionary so the new page passes cspell.


CommanderStorm (approved):
nice work 👍🏻

Issue #1076, opened 2023-12-16
Add support for tile, sprite, and font source aliasing


nyurik:
Martin allows users to specify multiple comma-separated sources and expect it to work like a single source.

type | URL
---|---
[TileJSON](https://maplibre.org/martin/sources-composite.html) | `/{source1},...,{sourceN}`
[Tiles](https://maplibre.org/martin/sources-composite.html) | `/{source1},...,{sourceN}/{z}/{x}/{y}`
[Sprites](https://maplibre.org/martin/sources-sprites.html) | `/sprite/<sprite_id1>,<sprite_id2>,...,<sprite_idN><suffix>`<br>Where the suffix could be `.png`, `.json`, `@2x.png`, `@2x.json`
[Fonts](https://maplibre.org/martin/sources-fonts.html) | `/font/{name1},…,{nameN}/{start}-{end}`

## Proposal
Server configuration should allow administrator to configure "aliases" - predefined composite sources, where the client would not be aware that the source actually consists of multiple sources combined on the fly.

## Configuration
For tile sources, the configuration would look like this:
```yaml
aliases:
  my_source:  # name of the alias
    sources:
      # List of sources that should be combined, in the order they should be combined.
      # Listing just one source effectively creates a redirect to an existing source.
      # The list must not reference any other aliases.
      - source1
      - source2
      - source3
```

For sprites and fonts, the configuration would look like this (note the `aliases` key is under `sprites`). The `fonts` would be similar.
```yaml
sprites:
  aliases:
    my_sprite:  # name of the sprite alias
      sources:
        # list of sprite sources that should be combined.
        # The list must not reference any other aliases.
        - sprite1
        - sprite2
        - sprite3
  # The rest is the regular sprite configuration as exists now
  paths:
    # all SVG files in this dir will be published as a "my_images" sprite source
    - /path/to/my_images
  sources:
    # SVG images in this directory will be published as a "my_sprites" sprite source
    my_sprites: /path/to/some_dir
```

PR #3210, opened 2026-09-01
feat(fonts): add font aliases (named font stacks)

### LINKED ISSUE #2386: Support named font stacks
Would it be possible to create named font stacks in the `config.yaml`? I am having trouble using glyphs served by Martin as it doesn't do any merging of fonts. Currently I have pre-generated the pbf font ranges thus merging fallbacks for each style. With Martin, I would have to adjust all the layers in my stylesheet to support composite fonts, but this also creates an issue with font styles like "Italic" and "Bold Italic". 

For example, when I add the fallback fonts to my stylesheet:
```js
text-font: ['Noto Sans Bold Italic', 'Noto Sans Arabic Bold']
```
MapLibre requests the non-existant: 
```curl
/fonts/Noto Sans Arabic Bold Italic/{range}
```
Rather than:
```curl
/fonts/Noto Sans Bold Italic,Noto Sans Arabic Bold/{range}
```
 
Another problem with specifying multiple languages in the stylesheet is that it could become very verbose especially if you want to support many writing systems. You could end up writing 10-15 font names for each layer. 

My proposition would be to define named stacks in the `config.yaml`. I imagine it would look something like this:
```yaml
fonts:
  sources:
    # Define named font stacks
    "Noto Sans Regular":
      - /fonts/NotoSans-Regular.ttf        # Latin/Greek/Cyrillic
      - /fonts/NotoSansArabic-Regular.ttf  # Arabic fallback
      - /fonts/NotoSansThai-Regular.ttf    # Thai fallback
    
    "Noto Sans Bold":
      - /fonts/NotoSans-Bold.ttf
      - /fonts/NotoSansArabic-Bold.ttf
      - /fonts/NotoSansThai-Bold.ttf
    
    "Noto Sans Italic":
      - /fonts/NotoSans-Italic.ttf
      - /fonts/NotoSansArabic-Regular.ttf  # Arabic and Thai doesn't have italic, use regular
      - /fonts/NotoSansThai-Regular.ttf
    
    "Noto Sans Bold Italic":
      - /fonts/NotoSans-BoldItalic.ttf
      - /fonts/NotoSansArabic-Bold.ttf     # Use bold for bold italic fallback
      - /fonts/NotoSansThai-Bold.ttf
```

Does this make sense, or is there a cleaner way to go about this that is already available?

johncarmack1984:
Closes #2386

Adds `fonts.aliases` to the config: named font stacks a client requests like a single font, served in the listed fallback order. One server-side name replaces listing every fallback font on every layer, and it sidesteps the font-name mangling MapLibre does to multi-font `text-font` values for styled fonts (see #2386).

This is the fonts part of #1076, with two deviations. The map is flat (`alias: [font, ...]`) since aliases carry no other options yet. An alias may shadow a font it references, so an existing name can gain fallbacks without touching styles. Aliases are listed in the catalog, and cache keys store the expanded fonts, so `DELETE /cache/{font}` also evicts entries reached through an alias. Sprites and tiles can follow the same pattern later if this shape works.

### DIFF
--- a/docs/content/sources-fonts.md
+++ b/docs/content/sources-fonts.md
@@ -88,9 +88,12 @@ fonts:
 
 ### Font Aliases
 
-The config file can also define aliases, which are named font stacks that clients request like a single font.
-Each alias combines the listed fonts in the given fallback order, exactly like a [composite font request](#composite-font-request).
-This keeps stylesheets short when text needs many fallback fonts (e.g. for multiple writing systems).
+!!! tip "Supporting multiple writing systems via `aliases`"
+    Working in an global world means that there are multiple languages that you will need to support in modern cartography.
+    Sadly, not every font that you might want to use supports this or looks good in every language.
+    To support this, we have `aliases`.
+    Each alias combines the listed fonts in the given fallback order, exactly like a [composite font request](#composite-font-request).
+
 
 ```yaml
 fonts:
--- a/docs/content/sources-fonts.md
+++ b/docs/content/sources-fonts.md
@@ -94,7 +94,6 @@ fonts:
     To support this, we have `aliases`.
     Each alias combines the listed fonts in the given fallback order, exactly like a [composite font request](#composite-font-request).
 
-
 ```yaml
 fonts:
   paths:
--- a/docs/content/sources-fonts.md
+++ b/docs/content/sources-fonts.md
@@ -85,3 +85,34 @@ fonts:
   - /path/to/font/file.ttf
   - /path/to/font_dir
 ```
+
+### Font Aliases
+
+The config file can also define aliases, which are named font stacks that clients request like a single font.
+Each alias combines the listed fonts in the given fallback order, exactly like a [composite font request](#composite-font-request).
+This keeps stylesheets short when text needs many fallback fonts (e.g. for multiple writing systems).
+
+```yaml
+fonts:
+  paths:
+    - /path/to/font_dir
+  # Each alias can be requested like a font and serves the listed fonts combined, in fallback order.
+  aliases:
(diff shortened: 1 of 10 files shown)


docs/content/sources-fonts.md
```diff
@@ -85,3 +85,34 @@ fonts:
   - /path/to/font/file.ttf
   - /path/to/font_dir
 ```
+
+### Font Aliases
+
+The config file can also define aliases, which are named font stacks that clients request like a single font.
+Each alias combines the listed fonts in the given fallback order, exactly like a [composite font request](#composite-font-request).
+This keeps stylesheets short when text needs many fallback fonts (e.g. for multiple writing systems).
```
CommanderStorm:
```suggestion
### Font Aliases

!!! tip "Supporting multiple writing systems via `aliases`"
    Working in an global world means that there are multiple languages that you will need to support in modern cartography.
    Sadly, not every font that you might want to use supports this or looks good in every language.
    To support this, we have `aliases`.
    Each alias combines the listed fonts in the given fallback order, exactly like a [composite font request](#composite-font-request).

```

docs/content/sources-fonts.md
```diff
@@ -85,3 +85,37 @@ fonts:
...
+### Font Aliases
+
+!!! tip "Supporting multiple writing systems via `aliases`"
+    Working in an global world means that there are multiple languages that you will need to support in modern cartography.
+    Sadly, not every font that you might want to use supports this or looks good in every language.
+    To support this, we have `aliases`.
+    Each alias combines the listed fonts in the given fallback order, exactly like a [composite font request](#composite-font-request).
+
+
+```yaml
```
CommanderStorm, line 97:
```suggestion
    Each alias combines the listed fonts in the given fallback order, exactly like a [composite font request](#composite-font-request).

```yaml
```

PR #3211, opened 2026-09-01
feat(cache): cache tiles in the encoding the client receives

### LINKED ISSUE #1112: Cache tiles in ready-to-serve encoding
## Current State
Since v0.13, Martin caches tile data as it comes back from the individual sources (PG/mbtiles/pmtiles). This means that if mbtiles contains gziped-compressed MVT tiles, cache will contain that tile, and no re-compression will be needed. If PostgreSQL generates an uncompressed tile, the tile will be cached, but each tile request has to gzip that tile anew. This is the usual case unless user has a custom PG function that uses a gzip extension to compress the tile on the database side.

## Goals
* Martin should cache data **after** compressing it
* Martin **could** re-compress responses in Brotli even if the storage format is gzip.

## Solutions
* generic solution - either as a custom function or a proper Actix "filter", capture the entire request (URL path + query + accept-encoding), and cache/send back the most optimal response
* per end-point (tiles vs tilejson vs sprites ...) a more custom implementation that takes into account the specifics of each endpoint. This might be easier to implement, but also means that each endpoint will have to write a bit of extra code.

### Brotli vs Gzip
Per [Brotli wiki](https://en.wikipedia.org/wiki/Brotli), it offers significantly better rate of compression and decompression performance than gzip. Desktop Chrome current default [Accept-Encoding](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding) is `gzip, deflate, br` (tested on Linux with latest Chrome). Reading [Quality_values](https://developer.mozilla.org/en-US/docs/Glossary/Quality_values) seem to indicate that the order of the values is irrelevant as long as they have the same "q" value (default is 1), which means that the `"gzip, deflate, br"` string indicates it is OK to send back Brotli, even though it is 3rd in the list.  Obviously if `br` is not in the list, it will have to be `gzip` or `identity`.

johncarmack1984:
Closes #1112

A tile that its source hands over uncompressed, PostgreSQL by default, was gzipped or brotlied again on every request, because the cache only held the tile as the pre-cache pipeline produced it. That entry stays, and a second one keyed by the negotiated encoding now holds the bytes the client receives. A brotli client arriving after a gzip client re-encodes from the produced entry instead of going back to the source, the two inserts discussed in the issue. A tile served as produced, a composite, and a hillshaded or contoured tile keep one entry as before, and the cache metric still counts one lookup per request. Re-encoding stored gzip into brotli for clients that prefer it stays a follow-up.

Warm cache, PostgreSQL function and table sources, gzip and br clients, `just build-hotpath` binaries on both sides, 401,202 requests per side:

| | main | this PR |
|---|---:|---:|
| `get_tile_content` average | 107 µs | 29 µs |
| `get_tile_content` p95 | 183 µs | 7 µs |
| `encode` calls | 401,202 | 120 |
| throughput, 256 connections | 103k req/s | 114k req/s |
| p99 latency, 256 connections | 6.0 ms | 3.3 ms |

### DIFF
--- a/martin-core/src/cache.rs
+++ b/martin-core/src/cache.rs
@@ -98,6 +98,17 @@ impl<K: CacheKey, V: Cacheable> ResourceCache<K, V> {
         Ok(entry.into_value())
     }
 
+    /// Reads a cached value without computing one on a miss.
+    /// Records no hit or miss, that is left to the caller.
+    pub async fn get(&self, key: &K) -> Option<V> {
+        self.inner.get(key).await
+    }
+
+    /// Stores a value, replacing any entry under `key`.
+    pub async fn insert(&self, key: K, value: V) {
+        self.inner.insert(key, value).await;
+    }
+
     /// Invalidates entries whose key matches `source_id`.
     /// Eviction is asynchronous.
     /// Flush the cache via [`Self::run_pending_tasks`].
--- a/martin-core/src/tiles/cache.rs
+++ b/martin-core/src/tiles/cache.rs
@@ -1,10 +1,11 @@
-use martin_tile_utils::{Format, TileCoord};
+use martin_tile_utils::{Encoding, Format, TileCoord};
 
 use crate::cache::{CacheKey, Cacheable, ResourceCache};
 use crate::tiles::Tile;
 
 /// Tile cache for storing rendered tile data, keyed by source ID, tile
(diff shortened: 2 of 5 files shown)


CommanderStorm (approved):
Thank you!

PR #3212, opened 2026-09-02
test(martin): make stopping martin an on-drop error if logs still exist

CommanderStorm:
We now enforce that martin is actually stopped in the e2e tests


e2e-tests/src/martin.rs
```diff
@@ -405,99 +484,65 @@ impl Martin {
...
+        self.process.take_log_lines(needle)
     }
 
-    /// Assert that at least one log line contains `needle`, and consume every
-    /// matching line. Must be called after [`Martin::stop`].
+    /// Assert that at least one log line contains `needle`, and consume every matching line.
+    ///
+    /// Only the lines the readers have collected so far are searched, so a line martin writes
+    /// while answering a request may not have arrived yet: assert on those after [`Martin::stop`],
+    /// which reads the log to the end, or wait for them with [`Martin::wait_for_log`].
```
CommanderStorm:
```suggestion
```

PR #3214, opened 2026-09-02
feat(duckdb): read GeoParquet from s3 and the other object stores DuckDB supports

CommanderStorm:
`GeoParquetLocation::from_config` accepted only `http` and `https`, so every other
URL fell through to `PathBuf::canonicalize` and failed with `No such file or
directory`. An `s3://` path - the form the Overture Maps documentation gives -
could not be configured at all.

This matters beyond ergonomics: DuckDB refuses globs over plain HTTP
(`Globs ('*') for generic HTTP file is are not supported`) but expands them over
`s3://`, and object-store datasets are routinely split across part files. Overture
alone ranges from 16 files for `places` to 513 for `buildings`, so without a glob
one Martin source is one part file.

`http`, `https`, `s3`, `gs`, `gcs`, `r2`, `az`, `azure`, `abfss` and `hf` are now
read remotely, `file://` resolves to a local path, and any other scheme still falls
through to the local branch so drive letters and relative paths keep working. The
default source id walks past glob segments to the last fixed one, falling back to
the bucket, so a glob source is not named `*`.

The scheme list is deliberately not `is_remote_url`: that one describes what
`object_store` reaches, so it carries Hadoop-style schemes DuckDB does not know
(`s3a`, `adl`, `abfs`) and omits ones it does (`gcs`, `r2`, `hf`).

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/docs/content/sources-duckdb.md
+++ b/docs/content/sources-duckdb.md
@@ -92,7 +92,7 @@ Each GeoParquet source supports:
 
 - **`geoparquet`**
   - a remote URL of the GeoParquet file (`http`, `https`, `s3`, `gs`, `gcs`, `r2`, `az`, `azure`, `abfss` and `hf` URLs). A remote URL may be a glob such as `s3://bucket/prefix/*.parquet`, which DuckDB expands into every matching part file.
-  - a local path or `file://` URLs. Local paths must name a single file. 
+  - a local path or `file://` URLs. Local paths must name a single file.
 - **`layer_id`** - MVT `source-layer` and the base for the source id (defaults to the file or URL stem).
 - **`geometry_column`** - geometry column name. Auto-detected when the file has exactly one geometry column.
 - **`id_column`** - optional table column to use as the MVT feature id.
--- a/docs/content/sources-duckdb.md
+++ b/docs/content/sources-duckdb.md
@@ -90,7 +90,9 @@ The top-level `pool_size`, `threads`, `memory_limit_mb`, and `auto_bounds` apply
 
 Each GeoParquet source supports:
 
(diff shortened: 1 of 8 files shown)


docs/content/sources-duckdb.md
```diff
@@ -86,7 +90,7 @@ The top-level `pool_size`, `threads`, `memory_limit_mb`, and `auto_bounds` apply
 
 Each GeoParquet source supports:
 
-- **`geoparquet`** - local path or `http://` / `https://` URL of the GeoParquet file.
+- **`geoparquet`** - local path, or a remote URL of the GeoParquet file. `http`, `https`, `s3`, `gs`, `gcs`, `r2`, `az`, `azure`, `abfss` and `hf` URLs are read remotely; `file://` URLs and everything else are read as local paths. A remote URL may be a glob such as `s3://bucket/prefix/*.parquet`, which DuckDB expands into every matching part file. Local paths must name a single file.
```
CommanderStorm:
```suggestion
- **`geoparquet`**
  - a remote URL of the GeoParquet file (`http`, `https`, `s3`, `gs`, `gcs`, `r2`, `az`, `azure`, `abfss` and `hf` URLs). A remote URL may be a glob such as `s3://bucket/prefix/*.parquet`, which DuckDB expands into every matching part file.
  - a local path or `file://` URLs. Local paths must name a single file. 
```

PR #3215, opened 2026-09-02
perf(duckdb): prune GeoParquet row groups with the file's covering bbox

CommanderStorm:
A tile request read the entire geometry column of every file it was pointed at.
On the 16-file, 10.5 GB Overture `places` theme, one z14 tile over London took
5m32s, of which 89% was network wait: `EXPLAIN ANALYZE` showed the scan emitting
all 4,599,286 rows, and HTTP logging showed one range request per row group, so
nothing was being pruned at all.

Two things had to change together.

GeoParquet 1.1 files declare a `covering` bounding box in the Parquet `geo`
metadata - four columns carrying each feature's extent, which Parquet keeps
per-row-group statistics for. Introspection now reads that declaration (paths and
all: the spec lets the covering live in any column at any struct depth) and the
tile query compares it against the tile with a four-way one-sided overlap test.
`ST_Intersects` stays as the exact filter, so the new clause only ever pre-filters.

The tile SQL also had to stop joining its coordinates in from a CTE. With the
coordinates behind a cross join the envelope is not constant at plan time and
DuckDB cannot push anything into the Parquet reader, which is where the pruning
happens. Written inline as the named parameters `$z`/`$x`/`$y`, DuckDB folds the
whole envelope expression and the statement stays static, so `prepare_cached`
keeps working unchanged.

Measured on an 81 MB file served over HTTP, one tile:

| | range requests | bytes read |
|---|---|---|
| before | 62 | 43.6 MB |
| after | 8 | 2.9 MB |

producing a byte-identical tile. End to end against Overture on S3, the z14
London `places` tile now answers in 5.1s.

Two things this deliberately does not do. It does not test whether the covering
box is *contained* in the tile: that prunes harder, but it is `ST_Within`, and it
drops every feature straddling a tile edge - invisible on the point layers people
try first, a hole along every tile seam on polygons. And it only prunes for source
SRIDs 4326 and 3857, whose transform to Web Mercator is monotone per axis; under a
projection with curved edges, transforming a tile's four corners can under-cover
it and clip features silently.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


docs/content/sources-duckdb.md
```diff
@@ -112,6 +112,16 @@ Per-source `pool_size`, `threads`, `memory_limit_mb`, and `auto_bounds` override
...
 
+!!! tip "Row-group pruning"
+    A tile request over a large Parquet file spends nearly all of its time waiting on I/O, so what matters is how much of the file it has to read.
+    Martin reads the [GeoParquet 1.1 `covering`](https://geoparquet.org/releases/v1.1.0/#covering) declaration out of the file's `geo` metadata on startup.
+    When a file has one, every tile query compares that bounding box against the tile before touching a geometry, and DuckDB resolves the comparison against per-row-group statistics, skipping whole row groups instead of decoding them.
+    On an 81 MB remote file one tile went from 62 range requests and 43.6 MB read to 8 requests and 2.9 MB, producing a byte-identical tile.
+
+    Files without a `covering` still work; every tile just reads the whole geometry column.
+    Writers such as [Overture Maps](https://overturemaps.org/), GDAL and GeoPandas emit one.
+    Pruning is also skipped when the source SRID is neither `4326` nor `3857`, because transforming a tile envelope into another projection can under-cover it and silently clip features at tile edges.
```
CommanderStorm:
```suggestion
    To reduce the IO nessary on large duckdb queries, we use can use the [GeoParquet 1.1 `covering`](https://geoparquet.org/releases/v1.1.0/#covering) declaration out of the file's `geo` metadata on startup and honor `bbox.{x,y}{min,max}` structs.
    
    Pruning is also skipped when the source SRID is neither `4326` nor `3857`, because transforming a tile envelope into another projection can under-cover it and silently clip features at tile edges.
```

martin-core/src/tiles/duckdb/source.rs
```diff
@@ -103,7 +103,8 @@ impl Source for DuckDBSource {
 #[derive(Clone, Debug)]
 /// SQL query information for `DuckDB` tile sources.
 pub struct DuckDBSqlInfo {
-    /// SQL query string.
+    /// SQL query string. It takes the tile coordinates as the named parameters `$z`, `$x` and
+    /// `$y`, each of which it must reference at least once.
```
CommanderStorm:
```suggestion
    /// SQL query string with named `x`,`y`,`z` params. 
```

martin/src/config/file/tiles/duckdb/resolver/geoparquet/covering.rs
```diff
@@ -0,0 +1,171 @@
...
+use serde::Deserialize;
+use tracing::debug;
+
+use crate::config::file::tiles::duckdb::sql_utils::escape_identifier;
+
+/// SQL accessors for the four corners of a `GeoParquet` 1.1 `covering` bounding box.
+///
+/// The columns behind them carry per-row-group min/max statistics, so a comparison against
+/// them prunes row groups before any geometry is read. The paths come from the file, not from
+/// convention: the spec lets the covering live in any column, at any struct depth.
```
CommanderStorm:
```suggestion
/// SQL accessors for the four corners of a `GeoParquet` 1.1 `covering` bounding box.
```

martin/src/config/file/tiles/duckdb/resolver/geoparquet/covering.rs
```diff
@@ -0,0 +1,171 @@
...
+        ymin: accessor(&paths.bbox.ymin)?,
+        xmax: accessor(&paths.bbox.xmax)?,
+        ymax: accessor(&paths.bbox.ymax)?,
+    })
+}
+
+/// Reads the `covering` declaration out of the Parquet `geo` key, if the file has one.
+///
+/// Pruning is an optimization, so every failure here is reported at debug level and answered
+/// with [`None`]: the tile SQL simply falls back to filtering on geometry alone.
```
CommanderStorm:
```suggestion
/// Reads the `covering` declaration out of the Parquet `geo` key, if the file has one.
```

martin/src/config/file/tiles/duckdb/resolver/geoparquet/resolve.rs
```diff
@@ -52,3 +52,161 @@ pub async fn resolve_geoparquet_source(
...
+            .await
+            .expect("introspect the covering fixture")
+    }
+
+    /// The tile one request produced, and the physical operators `DuckDB` used to produce it.
+    ///
+    /// The operators are the point: when the covering predicate reaches the Parquet reader,
+    /// `DuckDB` resolves it against the file's own statistics and a tile that matches nothing
+    /// becomes an `EMPTY_RESULT` with no scan at all. Without pushdown the same tile has to
+    /// scan the file and filter row by row, which is the 5-minute-tile behaviour this guards.
```
CommanderStorm:
```suggestion
    /// The tile one request produced, and the physical operators `DuckDB` used to produce it.
```

docs/content/sources-duckdb.md
```diff
@@ -112,6 +112,11 @@ Per-source `pool_size`, `threads`, `memory_limit_mb`, and `auto_bounds` override
     SRID auto-detection supports EPSG codes and `OGC:CRS84` only.
     If the file has more than one geometry column, set `geometry_column` explicitly.
 
+!!! tip "Row-group pruning"
+    To reduce the IO nessary on large duckdb queries, we use can use the [GeoParquet 1.1 `covering`](https://geoparquet.org/releases/v1.1.0/#covering) declaration out of the file's `geo` metadata on startup and honor `bbox.{x,y}{min,max}` structs.
```
CommanderStorm:
```suggestion
    To reduce the IO necessary on large duckdb queries, we use can use the [GeoParquet 1.1 `covering`](https://geoparquet.org/releases/v1.1.0/#covering) declaration out of the file's `geo` metadata on startup and honor `bbox.{x,y}{min,max}` structs.
```

PR #3217, opened 2026-09-02
refactor(config): classify source paths through a `SourceLocation` type

CommanderStorm:
This adds `martin::config::file::SourceLocation`:

```rust
pub enum SourceLocation {
    Local(PathBuf),
    ObjectStore(Url),
    Http(Url),
}
```


CommanderStorm:
Superseded by https://github.com/maplibre/martin/pull/3218 — same commits, but the branch needs to live in `maplibre/martin` like the rest of the `stack-*` branches so the follow-up PRs can base on it.

PR #3219, opened 2026-09-02
feat(postgres): serve a function's key column as the tile ETag

### LINKED ISSUE #1917: Less CPU intensive etag implementation
CDNs rely heavily on etag (usually a hash) to identify if the tile has changed or not. In some cases, the tile backend like .mbtiles is already storing tiles with some hash. For example, .mbtiles frequently uses two tables to remove duplicate tiles - one stores `z, x, y, hash` columns, and another table - `hash, data blob`.

Currently, we implement this via running a non-cryptographical hash function over the output.
We do this even in cases where there is a hash that we could serve causing unnessesary CPU usage (latency is basically instant).

=> Implementing this optimisation would net us about 5% cpu time reduction which we could use to serve a few more tiles at the extreme end (likely you are network bound at the 200k/s mark)

Steps involved:
- [x] make mbtiles backend recognize when the `.mbtiles` file has two frequently used tables as described above, and use the hash value as etag https://github.com/maplibre/martin/pull/1787
- [x] in the individual backends like mbtiles to supply a pre-generated etag
- [ ] individual backends like mbtiles supply a pre-generated etag
- [ ] PostgreSQL functions return rows with two columns, treating the second column as the key. OpenMapTiles already generates these types of functions.
- [ ] support etags for `/catalog`

johncarmack1984:
Closes #1917

Functions returning `(bytea, text)` are discovered with their key column (#380), but the generated query only selected the first column and every tile was hashed again with xxh3, which dericke and tuananh0707 both ran into on #580. The query now selects both and the key is the ETag. The hash stays the fallback for empty tiles, a `NULL` or empty key, and plain `bytea` functions. Table sources don't change. A byte hash isn't a stable identity for a PostGIS tile anyway, since PostgreSQL won't fix feature order without an `ORDER BY` (#1783), and a key column lets the function decide what "changed" means.

| Unchanged tile, 20 identical requests (Martin 1.15.0, table sources) | Distinct ETags |
|---|---:|
| buildings z14 (2,123 polygons) | 5 |
| roads z14 | 2 |
| pois z14 | 2 |
| boundaries z9 | 2 |

### DIFF
--- a/docs/content/sources-pg-functions.md
+++ b/docs/content/sources-pg-functions.md
@@ -12,7 +12,7 @@ A Function Source is a database function which can be used to
 query [vector tiles](https://github.com/mapbox/vector-tile-spec). When started, Martin will look for the functions with
 a suitable signature.
 
-A function can be used as a Function Source if it returns either a `bytea` value, or a record with `bytea` and a `text` values.  The `text` value is expected to be a user-defined hash, e.g. an MD5 value, and it will eventually be used as an [ETag](https://developer.mozilla.org/de/docs/Web/HTTP/Reference/Headers/ETag).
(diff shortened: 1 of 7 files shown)


CommanderStorm (approved):
Thanks a lot

PR #3220, opened 2026-09-02
feat(sprites): add sprite aliases

johncarmack1984:
Sprites slice of #1076, same shape as the fonts one in #3210.

Adds `sprites.aliases` to the config: a named combination of sprite sources that a style requests like a single source, and Martin serves like `/sprite/a,b`. One server-side name replaces listing every source in every style, and a source can gain icons without changing the name styles use, since an alias may shadow a source it references. Aliases are listed in the catalog with the images of every source they combine, and cache keys store the expanded sources, so `DELETE /cache/{sprite}` also evicts entries reached through an alias. Same flat map and the same rules as fonts: only configured sources, no alias inside an alias, and an alias name cannot end in `@2x` since that suffix belongs to the request

### DIFF
--- a/docs/content/files/generated_config.md
+++ b/docs/content/files/generated_config.md
@@ -440,6 +440,12 @@ preferred_encoding: brotli
 route_prefix: null
 # Sprite configuration
 sprites:
+  # Named combinations of sprite sources.
+  #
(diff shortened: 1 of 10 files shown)


CommanderStorm (approved):
Thanks, LGTM

PR #3221, opened 2026-09-02
feat(postgres): add a connection-level cache zoom default

johncarmack1984:
Step 3 of #1137, and the postgres half of step 2. Prior art: #2673 (per-source `cache`), #3193 (per-source > source type > global for the process settings).

Adds `postgres.cache`: the cache zoom bounds every table and function on that connection takes unless it sets its own, layered over the top-level `cache` bounds the way `default_srid` sits on the connection. Until now an auto-discovered table or function could only take the global bounds, so keeping a whole discovered schema out of the cache meant declaring every table. The file kinds get the same per-kind default in a follow-up, which is the rest of step 2 and closes the issue

### DIFF
--- a/docs/content/files/generated_config.md
+++ b/docs/content/files/generated_config.md
@@ -356,6 +356,19 @@ postgres:
       # Optionally set how source ID should be generated based on the table's name,
       # schema, and geometry column
       source_id_format: table.{schema}.{table}.{column}
+  # Zoom-level bounds for caching the tiles of this connection.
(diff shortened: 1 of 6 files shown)


CommanderStorm (approved):
Thanks a ton ❤️

johncarmack1984:
Looks like the failing CI was a GitHub infra flake on the prep step, I think a re-run of failed jobs might turn it into a pass

PR #3222, opened 2026-09-02
fix(reload): apply the global cache zoom bounds to discovered files

johncarmack1984:
Since 1.15.0, the top-level `cache.minzoom` and `cache.maxzoom` do nothing for a file that Martin finds by scanning a `paths:` directory, at startup or when the file shows up later, nor for any PMTiles file under a remote prefix. A file given by its own path, or listed under `sources:`, still gets them. #3190 moved directory scanning into the reload discovery, which only knew a file's own `cache` entry and never saw the top-level bounds. The discovery now starts from the top-level bounds and lets a file's own `cache` override them, which is what the directly named files always did.

Checked on the release binaries with `cache: {minzoom: 1}`, one mbtiles file in a `paths:` directory, and its z0 tile requested three times. A z0 tile should never touch the cache with that bound in place.

| Martin | Tile cache requests at z0 |
|---|---|
| 1.14.0 | none |
| 1.15.0 | 1 miss, 2 hits |
| this branch | none |

The per-kind `cache` default for #1137 builds on the same seam, in a follow-up

### DIFF
--- a/e2e-tests/tests/mbtiles.rs
+++ b/e2e-tests/tests/mbtiles.rs
@@ -429,18 +429,15 @@ async fn reload_removes_a_source_when_its_file_is_deleted() {
     martin.assert_startup_warnings();
 }
 
-/// The `martin_tile_cache_requests_total` lines of a metrics scrape, so a test can say what the
-/// tile cache saw without pinning the rest of the scrape.
-fn tile_cache_lines(scrape: &str) -> String {
-    scrape
-        .lines()
-        .filter(|line| line.starts_with("martin_tile_cache_requests_total"))
-        .collect::<Vec<_>>()
-        .join("\n")
-}
-
 #[tokio::test]
 async fn a_file_discovered_in_a_directory_takes_the_global_cache_bounds() {
+    fn tile_cache_lines(scrape: &str) -> String {
+        scrape
+            .lines()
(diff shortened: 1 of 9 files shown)


e2e-tests/tests/mbtiles.rs
```diff
@@ -428,3 +428,51 @@ async fn reload_removes_a_source_when_its_file_is_deleted() {
...
+fn tile_cache_lines(scrape: &str) -> String {
+    scrape
+        .lines()
+        .filter(|line| line.starts_with("martin_tile_cache_requests_total"))
+        .collect::<Vec<_>>()
+        .join("\n")
+}
+
+#[tokio::test]
+async fn a_file_discovered_in_a_directory_takes_the_global_cache_bounds() {
```
CommanderStorm: I don't expect this will be usefull globally

```suggestion
#[tokio::test]
async fn a_file_discovered_in_a_directory_takes_the_global_cache_bounds() {
    fn tile_cache_lines(scrape: &str) -> String {
        scrape
            .lines()
            .filter(|line| line.starts_with("martin_tile_cache_requests_total"))
            .collect::<Vec<_>>()
            .join("\n")
    }
```

CommanderStorm (approved):
Thanks, Ups...

PR #3223, opened 2026-09-02
feat(config): add a per-kind cache zoom default for file and passthrough sources

### LINKED ISSUE #1137: Make the tile source cache configuration per source configurable

johncarmack1984:
Closes #1137

Adds `cache` to the `mbtiles`, `pmtiles`, `cog`, `geojson`, `duckdb` and `passthrough` blocks: the cache zoom bounds every source of that kind takes unless it sets its own, layered over the top-level `cache` bounds the same way `convert_to_mlt` layers since #3193. A source's own `cache` still wins, and a kind without the key behaves as before. With `postgres.cache` (#3221) every source type now has the per-type default Yuri's checklist asked for, and the `enable`/`auto` literals from step 1 have no behaviour left to carry, since omitting the key already means enabled.

| Precedence | Example |
|---|---|
| source | `mbtiles.sources.<id>.cache` |
| kind | `mbtiles.cache` |
| global | `cache.minzoom` / `cache.maxzoom` |

### DIFF
--- a/docs/content/files/generated_config.md
+++ b/docs/content/files/generated_config.md
@@ -148,6 +148,19 @@ geojson:
   # Clip margin kept around each tile edge, in tile units, defaulting to 64.
   # Increase it if you see seam artifacts on line caps/joins or polygon outlines near tile edges.
   buffer: 64
+  # Zoom-level bounds for caching the tiles of every `GeoJSON` source without its own `cache`.
+  # Overrides the top-level `cache` bounds.
+  cache:
+    # Default maximum zoom level (inclusive) for tile caching.
+    # Tiles further zoomed in than this will bypass the cache entirely.
+    # Can be overridden per-source.
+    # default: null (no upper bound, all zoom levels cached)
+    maxzoom: 14
+    # Default minimum zoom level (inclusive) for tile caching.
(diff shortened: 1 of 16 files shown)


CommanderStorm (approved):
Thanks ❤️

CommanderStorm:
(CI has some problems with the PR, lets see if the `bless` label can also fix them)

johncarmack1984:
TIL the bless job skips the DuckDB feature 😄 

Blessed the two DuckDbConfig snapshots manually in 06e257cc, but the push killed auto-merge. Sorry about that!

CommanderStorm:
bless likely should bless all. I want to be able to use the labeling to do this..

Issue #1901, opened 2025-07-03
ERROR:  lwcollection_construct: mixed dimension geometries: 2/0


ognjenantonijevic:
Hi,

This is not a Martin issue per se, but rather a suggestion to add a setting to include a parameter which will be added to each ST_AsMVTGeom function call.

I've been experiencing issues when generating tiles for several layers. After googling a bit I found out this is a known issue with ST_AsMVTGeom, which is (reportedly) solved in some other tile servers by adding a ST_Force2D function inside the function call.

Error message:
ERROR:  lwcollection_construct: mixed dimension geometries: 2/0

Relevant issues:
https://trac.osgeo.org/postgis/ticket/4690
https://github.com/CrunchyData/pg_tileserv/issues/39
https://github.com/submarcos/django-vectortiles/pull/57

The setting could be something like force_2d=true and it would influence the format of the ST_AsMVTGeom function call in file https://github.com/maplibre/martin/blob/cce9d08f14fe5dfff01804000f341917bb6adf09/martin/src/pg/query_tables.rs#L198

ognjenantonijevic:
Actually, it seems I've managed to go around this problem by turning off geometry clipping in settings.yaml. Leaving this here in case someone else has the same problem.

CommanderStorm:
From the issue you mentioned:

> ST_Force2D seems to fix this problem, but at a steep cost in time to generate MVTs

This seems like not such a nice tradeoff.. 🤔
If we support this this must be only on columns which actually need it and with a warning that this degrades performance.

I would like to push this down to the data generation level.
=> How do you generate the data? Having mixed dimension seems kind of unnessesary

PR #218, opened 2021-06-07
Possible fix for complex geoms

christianversloot:
As far as I can see this fixes #215 out-of-the-box for complex geometries (`CURVEPOLYGON`, `MULTISURFACE`, ...) without spatial inequality for simple types (`POINT`, `LINESTRING`, ...) occurring because of it (see #215).

Issue #215, opened 2021-06-04
Martin throwing HTTP 500 on complex geometry types


christianversloot:
I am currently trying to implement a PoC using Martin with Leaflet VectorGrid for visualizing the Basisregistratie Grootschalige Topografie (BGT), which is a standardized dataset representing a variety of objects in the Dutch public area.

I am using the following POSTGIS dump for the PoC which represents a small slice from the entire BGT (all object types in one area): https://geocatalogus.nl/dataset/bgt-imgeo-database-dumps/resource/1ae231d7-5753-4e3b-afe8-e1b2636c474d

Martin finds all table sources nicely:

```
chris@DESKTOP-PJKJ0UE:/mnt/c/Users/chris/Downloads/martin-Windows-x86_64$ ./martin.exe postgres://postgres:pgpg@localhost:5433/bgt
[2021-06-04T08:10:13Z INFO  martin] Starting martin v0.5.0
[2021-06-04T08:10:13Z INFO  martin] Config is not set, scanning database
[2021-06-04T08:10:14Z INFO  martin] Connected to postgres://postgres:pgpg@localhost:5433/bgt
[2021-06-04T08:10:14Z INFO  martin::table_source] Found latest.bak table source
[2021-06-04T08:10:14Z INFO  martin::table_source] Found latest.bakactueel table source
[2021-06-04T08:10:14Z INFO  martin::table_source] Found latest.bakactueelbestaand table source
[2021-06-04T08:10:14Z INFO  martin::table_source] Found latest.begroeidterreindeel table source
[2021-06-04T08:10:14Z INFO  martin::table_source] Found latest.begroeidterreindeel table source
[2021-06-04T08:10:14Z INFO  martin::table_source] Found latest.begroeidterreindeelactueel table source
[2021-06-04T08:10:14Z INFO  martin::table_source] Found latest.begroeidterreindeelactueel table source
[2021-06-04T08:10:14Z INFO  martin::table_source] Found latest.begroeidterreindeelactueelbestaand table source
[2021-06-04T08:10:14Z INFO  martin::table_source] Found latest.begroeidterreindeelactueelbestaand table source
[2021-06-04T08:10:14Z INFO  martin::table_source] Found latest.bord table source
[2021-06-04T08:10:14Z INFO  martin::table_source] Found latest.bordactueel table source
[2021-06-04T08:10:14Z INFO  martin::table_source] Found latest.bordactueelbestaand table source
[...]

PR #3224, opened 2026-09-02
perf(postgres): linearize only geometry columns that can hold arcs

johncarmack1984:
Every table query wraps the geometry in `ST_CurveToLine` since #218, which fixed the HTTP 500 that curve columns raised in #215. PostGIS does not skip that call for arc-free input, it deserializes, linearizes and reserializes every polygon, so a dense tile pays for a full copy of its geometry for nothing. Columns whose `geometry_columns.type` is one of the six linear types now skip the wrapper, the same column-type gating you asked for on `ST_Force2D` in #1901. Generic `GEOMETRY`, collections, curve types and unknown types keep it, and the tiles are byte-identical either way.

Measured on a Berlin OSM buildings table (2,123 polygons and 28,000 vertices in the z14 tile), first in pgbench with the query alone, then through Martin with oha.

| | With `ST_CurveToLine` | Without |
|---|---:|---:|
| pgbench, 1 client, per tile | 10.86 ms | 10.69 ms |
| pgbench, 8 clients, saturated | 458.6 tiles/s | 474.5 tiles/s |
| Martin, 1 client, per tile | 12.32 ms | 12.25 ms |
| Martin, 10 clients | 484 to 488 tiles/s | 494 tiles/s |


martin/src/config/file/tiles/postgres/resolver/query_tables.rs
```diff
@@ -363,3 +387,44 @@ pub fn polygon_to_bbox(polygon: &ewkb::Polygon) -> Option<Bounds> {
...
+            "CURVEPOLYGON",
+            "MULTICURVE",
+            "MULTISURFACE",
+            "",
+        ] {
+            assert!(may_contain_arcs(Some(t)), "{t}");
+        }
+        assert!(may_contain_arcs(None));
+    }
+}
```
CommanderStorm: instead of unit tests against internal functions, can you please test this via the public api, so via integration or e2e tests?

It is hard to know if we are handling things correctly otherwise just because we might be wrongly handling one of them and just have a testing gap.

johncarmack1984: Makes sense to me! 

Done in 5b0b374. The helper's unit tests are gone, replaced by two fixture tables rendered end to end. `curves` is a typed CURVEPOLYGON column, `curves_untyped` is a plain GEOMETRY column holding a CIRCULARSTRING and a COMPOUNDCURVE. The existing tables cover the linear side; no snapshots moved.

PR #3226, opened 2026-09-02
perf(reload): start the PostgreSQL poll one interval after init

johncarmack1984:
Follows #3130. `PostgresReloader::init` publishes everything the catalog discovers before the server starts, and `start` then spawns a `PollTrigger` whose first tick is immediate, a leftover from when that first pass was the seed. The same discovery therefore ran a second time a few milliseconds later, while the first tile requests were arriving and sharing the pool with it. The remote PMTiles path still needs the immediate tick, since nothing else loads remote prefixes, so `PollTrigger::new` keeps it and the PostgreSQL poll uses `PollTrigger::after_interval`, which waits one interval first.

Measured server-side with `log_min_duration_statement`, four tables, `auto_publish.tables` only, default 10 minute interval.

| Per start | main | this branch |
|---|---:|---:|
| Discovery passes | 2 | 1 |
| Catalog time, all discovery queries | 31 to 42 ms | 18 to 26 ms |

With the sibling catalog-skip PR on top, the one remaining pass is 6 to 8 ms for this config

### DIFF
--- a/martin/src/config/file/tiles/driver/trigger.rs
+++ b/martin/src/config/file/tiles/driver/trigger.rs
@@ -65,16 +65,27 @@ impl Trigger for NotifyTrigger {
     }
 }
 
-/// Fires on a fixed interval, first tick immediate. Never ends the loop.
+/// Fires on a fixed interval. Never ends the loop.
 pub struct PollTrigger {
     ticker: tokio::time::Interval,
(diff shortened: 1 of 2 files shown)


CommanderStorm (approved):
Thanks 👍🏻

PR #3229, opened 2026-09-03
feat: add tile source aliases

### LINKED ISSUE #1076: Add support for tile, sprite, and font source aliasing

johncarmack1984:
Closes #1076

Last slice of #1076 after fonts (#3210) and sprites (#3220). A top-level `aliases` block names a combination of tile sources that clients request like a single source: `/{alias}` and `/{alias}/{z}/{x}/{y}` serve exactly what `/{a},{b}` serves, so a style no longer carries the composite list. Aliases may only reference tile sources, an alias may shadow the source it extends, and every alias is checked at startup. The catalog lists an alias under its own name with its members' format. Tile cache keys stay per member source, so purging a source also drops what its aliases served.

The shape is Yuri's proposal from the issue body, flattened to `name: [members]` like the two merged slices.

### DIFF
--- a/docs/content/files/generated_config.md
+++ b/docs/content/files/generated_config.md
@@ -11,6 +11,13 @@
 ```yaml title="config.yaml"
 # yaml-language-server: $schema=https://raw.githubusercontent.com/maplibre/martin/main/schemas/config.json
 
+# Named combinations of tile sources.
+#
+# Each alias can be requested like a tile source and serves the listed sources combined,
+# exactly like the composite request `/{source1},{source2}`.
+# Aliases may only reference tile sources, not other aliases.
+# An alias sharing the name of a tile source takes precedence over it.
+aliases: {}
 # Set `TileJSON` URL path prefix.
 # This overrides the default path prefix for URLs in `TileJSON` responses.
 # If both `route_prefix` and `base_path` are set, `base_path` takes priority for `TileJSON` URLs.
--- a/docs/content/sources-composite.md
+++ b/docs/content/sources-composite.md
@@ -25,3 +25,30 @@ curl localhost:3000/points,lines
 # Whole world as a single tile
 curl localhost:3000/points,lines/0/0/0
 ```
+
+## Source Aliases
+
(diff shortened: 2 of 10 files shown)


CommanderStorm (approved):
Nice work!

PR #3231, opened 2026-09-03
feat(postgres): retry the first database connection at startup

### LINKED ISSUE #1539: Add an ability to wait for the PostgreSQL to become available before starting

johncarmack1984:
Closes #1539

Martin exits at once when PostgreSQL is not accepting connections yet, which is the normal state for a few seconds after a container stack or a host boots. The first connection is now retried every half second for up to 30 seconds by default before startup fails. A refused, reset or not-yet-ready connection is retried. A wrong password, an unknown database or a failed TLS handshake still fails at once. After two seconds a warning names the way out.

`postgres.retry_timeout` (or `--pg-retry-timeout`) takes a duration, `0s` to fail at once, or `infinite` to wait until the database answers.

| retry_timeout | first refused connection | log |
|---|---|---|
| `0s` | fails at once | error only |
| `3s` (default `30s`) | retried every half second until it passes | warning after two seconds, then the error |
| `infinite` | waits | warning every ten seconds |


martin/src/config/file/tiles/postgres/config.rs
```diff
@@ -102,6 +103,12 @@ pub struct PostgresConfig {
         schemars(with = "crate::config::file::CachePolicyShape")
     )]
     pub cache: CachePolicy,
+    /// How many times the first connection is retried, one second apart, before startup fails \[default: 30\]
+    ///
+    /// A number, or `infinite` to wait until the database answers.
+    /// `0` fails on the first refused connection.
+    #[cfg_attr(feature = "unstable-schemas", schemars(with = "Option<String>", example = &"infinite"))]
```
CommanderStorm:
```suggestion
    #[cfg_attr(feature = "unstable-schemas", schemars(with = "Option<String>", example = &"30"))]
```

martin/src/config/file/tiles/postgres/builder.rs
```diff
@@ -108,6 +108,7 @@ impl PostgresAutoDiscoveryBuilder {
             config.ssl_certificates.ssl_key.as_ref(),
             config.ssl_certificates.ssl_root_cert.as_ref(),
             config.pool_size.unwrap_or(DEFAULT_POOL_SIZE).get(),
+            config.connection_retries.unwrap_or_default(),
```
CommanderStorm: does this not mean default=0? I think 30 is more sensible..

martin-core/src/tiles/postgres/connection_retries.rs
```diff
@@ -0,0 +1,126 @@
...
+        #[derive(Deserialize)]
+        #[serde(untagged)]
+        enum Raw {
+            Count(u32),
+            Text(String),
+        }
+        match Raw::deserialize(deserializer)? {
+            Raw::Count(n) => Ok(Self::Count(n)),
+            Raw::Text(s) => s.parse().map_err(serde::de::Error::custom),
+        }
```
CommanderStorm: otherwise the error message of the ? will be pretty terrible.

```suggestion
        match Raw::deserialize(deserializer) {
            Ok(Raw::Count(n)) => Ok(Self::Count(n)),
            Ok(Raw::Text(s)) if s.as_str().eq_ignore_ascii_case("infinite") => Ok(Self::Infinite),
            Ok(Raw::Text(_)) | Err(_) => Err(serde::de::Error::custom(format!("expected either `infinite` or a number")),
        }
```

martin-core/src/tiles/postgres/connection_retries.rs
```diff
@@ -0,0 +1,126 @@
...
+        let s = s.trim();
+        if s.eq_ignore_ascii_case("infinite") {
+            return Ok(Self::Infinite);
+        }
+        s.parse::<u32>().map(Self::Count).map_err(|_not_a_number| {
+            format!("expected a number of retries or `infinite`, got {s:?}")
+        })
+    }
+}
+
```
CommanderStorm: Seems like the second impl of this

```suggestion
```

martin-core/src/tiles/postgres/pool.rs
```diff
@@ -54,12 +62,37 @@ impl PostgresPool {
...
         ssl_root_cert: Option<&PathBuf>,
         pool_size: usize,
+    ) -> PostgresResult<Self> {
+        Self::new_with_retries(
+            connection_string,
+            ssl_cert,
+            ssl_key,
+            ssl_root_cert,
+            pool_size,
+            ConnectionRetries::DEFAULT,
```
CommanderStorm:
```suggestion
            ConnectionRetries::default(),
```

martin-core/src/tiles/postgres/pool.rs
```diff
@@ -54,12 +62,37 @@ impl PostgresPool {
     /// - `ssl_key`: Same as PGSSLKEY ([docs](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-SSLKEY))
     /// - `ssl_root_cert`: Same as PGSSLROOTCERT ([docs](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-SSLROOTCERT))
     /// - `pool_size`: Maximum number of connections in the pool
+    ///
+    /// The first connection is retried with [`ConnectionRetries::DEFAULT`].
```
CommanderStorm:
```suggestion
```

martin-core/src/tiles/postgres/pool.rs
```diff
@@ -54,12 +62,37 @@ impl PostgresPool {
...
+            ConnectionRetries::DEFAULT,
+        )
+        .await
+    }
+
+    /// Same as [`new`](Self::new), retrying the first connection `retries` times, one second apart.
+    ///
+    /// Only a refused, dropped or not yet accepting connection is retried.
+    /// A wrong password or database name fails at once.
+    pub async fn new_with_retries(
```
CommanderStorm: why do we need to have new vs new_with_retries? lets just add an param to new.

martin-core/src/tiles/postgres/connection_retries.rs
```diff
@@ -0,0 +1,126 @@
...
+    /// Retry this many times, so `0` fails on the first refused connection.
+    Count(u32),
+    /// Retry until the database answers.
+    Infinite,
+}
+
+impl ConnectionRetries {
+    /// Retries before startup fails when nothing is configured.
+    pub const DEFAULT: Self = Self::Count(30);
+
```
CommanderStorm: pls inline

martin-core/src/tiles/postgres/pool.rs
```diff
@@ -31,6 +33,12 @@ const ST_TILE_ENVELOPE_POSTGIS_VERSION: Version = Version::new(3, 1, 0);
 const MISSING_GEOM_FIXED_POSTGIS_VERSION: Version = Version::new(3, 5, 0);
 /// Minimum version of postgres required for [`RECOMMENDED_POSTGIS_VERSION`] according to the [Support Matrix](https://trac.osgeo.org/postgis/wiki/UsersWikiPostgreSQLPostGIS)
 const RECOMMENDED_POSTGRES_VERSION: Version = Version::new(12, 0, 0);
+/// Pause between two attempts at the first connection.
+const RETRY_INTERVAL: Duration = Duration::from_secs(1);
```
CommanderStorm: lets do this every 500ms or so, this means retries are a bit faster, but not undue so.

martin-core/src/tiles/postgres/pool.rs
```diff
@@ -172,6 +205,32 @@ impl PostgresPool {
...
+                Err(error) => return Err(error),
+            };
+            failed += 1;
+            let warn_due = last_warning.is_none_or(|at| at.elapsed() >= RETRY_WARN_EVERY);
+            if started.elapsed() >= RETRY_WARN_AFTER && warn_due {
+                warn!(
+                    source.id = %self.id,
+                    postgres.attempts = failed,
+                    postgres.retries = %retries,
+                    "PostgreSQL is not accepting connections yet, retrying every second. Set `postgres.connection_retries: 0` or `--pg-connection-retries 0` to fail at once: {error}"
```
CommanderStorm:
```suggestion
                    postgres.retries = %retries,
                    error = %error,
                    "PostgreSQL is not accepting connections yet, retrying every second. Set `postgres.connection_retries: 0` or `--pg-connection-retries 0` to fail at once"
```

e2e-tests/tests/postgres_unreachable.rs
```diff
@@ -0,0 +1,56 @@
+//! Startup against a `PostgreSQL` that is not there: the first connection is retried, or not.
+
+use martin_e2e_tests::{Martin, StartError};
+
+/// Nothing listens on port 1, so every attempt is refused at once.
+const CONFIG: &str = "
```
CommanderStorm: pls inline, and please with the appropriate indoc macro

martin/src/config/args/postgres.rs
```diff
@@ -27,6 +28,9 @@ pub struct PostgresArgs {
     pub default_srid: Option<i32>,
     #[arg(help = format!("Maximum Postgres connections pool size [DEFAULT: {DEFAULT_POOL_SIZE}]"), short, long)]
     pub pool_size: Option<NonZeroUsize>,
+    /// How many times the first PostgreSQL connection is retried, one second apart, before startup fails: a number or `infinite`. [DEFAULT: 30]
+    #[arg(long)]
+    pub pg_connection_retries: Option<ConnectionRetries>,
```
CommanderStorm: I am not sure if the amount of retries is the quanity that you as an operator want to control.
I think you waould rather like to bound the time that we may retry given that we have no clue on what the time to connect is going to be and how this plays in here... 🤔

CommanderStorm (changes requested):
Overall mostly LGTM, there are a few artefacts to clean up though.

CommanderStorm (approved):
thanks

Issue #1749, opened 2025-03-13
Rework/simplify Martin demo site


nyurik:
The site at https://martin.maplibre.org/ is really long and verbose. We should simplify it, possibly removing a lot of things, and focus on a short summary of what's available plus the demo.  Brevity – the sister of talent...

CommanderStorm:
It also heavily focusses on postgis support, while not even mentioning that we added support mbtiles/pmtiles/cog 😅

Emanuelmsjv:
Hi i would like to work on this issue.

CommanderStorm:
Awsome

Emanuelmsjv:
I've reviewed the site. Could you clarify which sections you'd like removed or shortened? I want to make sure my changes align with your vision before submitting a PR. @CommanderStorm

CommanderStorm:
In general, there are a few sections that are really not usefull.

Here is an example how the competition is presenting themselves -> https://tileserver.app/
So more direct and less hand-wavy marketing.

Emanuelmsjv:
Hi @CommanderStorm! I've reviewed both the current Martin demo site and tileserver.app as a reference.

Here's what I'm thinking:

***Remove:***
- The 3 feature headline sections ("Turning Data into Vector Tiles", "Generating Tiles with Functions", "Filtering and Aggregating Data on the Fly") — these repeat what the docs already cover

***Shorten***
- NYC taxi dataset explanation → trim to one line

***Keep:***
- Hero title + one-liner description
- The interactive demo itself
- GitHub + Docs CTA links

This would make the page demo-first and much more direct. Does this align with your vision before I open a PR?

CommanderStorm:
Yea, that sounds about what we want.

maybe a small installer ("install via -> tabs (docker, homebrew, cargo-install, cargo-binstall,. ...)") box would be nice too next to the docs CTA

PR #3232, opened 2026-09-03
docs(demo): trim the demo site to the demo and add an install box

johncarmack1984:
works towards #1749

As discussed in the issue:

- The three feature sections that repeat the docs before a visitor reaches the map have been removed
- The taxi text has been cut to one line, naming PostGIS, PMTiles, MBTiles and GeoJSON
- An install box with tabs for Docker, Homebrew, cargo binstall, cargo install and the Debian package has been added. 

A creative liberty I took:
- Each command block has a copy button, using the same Lucide icons as martin-ui.

### L: Before. R: After

demo-1749-before-after


CommanderStorm (approved):
LGTM, but I think we can rebuild this even stronger so I am not going to close the issue ✔️ 

- https://github.com/maplibre/martin/pull/2605

Especially showing the globe, MLT, hillshading, ... there is so much functionality that we have that the demo does not cover at all..

PR #3234, opened 2026-09-03
feat(mbtiles): warn when a served file has no tile index

### LINKED ISSUE #741: Add more validations to .mbtiles detection

johncarmack1984:
Closes #741

Martin serves an MBTiles file whose tile table has no index on `(zoom_level, tile_column, tile_row)` without a word, and every tile request then scans the table. The report on the issue saw request times past a second on such a file. `mbtiles validate` already refuses these files, but the server never runs that check. On startup Martin now looks for any index covering the three coordinate columns on the table it reads from and logs a warning naming the table and the `CREATE UNIQUE INDEX` that fixes it. The file is still served, `validate` is unchanged, and the validation levels sketched on the issue are not part of this.

### DIFF
--- a/e2e-tests/tests/mbtiles.rs
+++ b/e2e-tests/tests/mbtiles.rs
@@ -577,3 +577,31 @@ async fn a_kind_level_cache_bound_covers_a_directory_and_a_source_can_override_i
     "#);
     overridden.stop().await;
 }
+
+#[tokio::test]
+async fn a_source_without_a_tile_index_is_served_with_a_warning() {
+    use sqlx::Connection as _;
+
+    let dir = tempfile::tempdir().expect("failed to create a temp dir");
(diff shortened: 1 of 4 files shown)


CommanderStorm (approved):
Thanks

PR #3235, opened 2026-09-03
feat(mbtiles): show progress for the long commands

### LINKED ISSUE #1024: Use indicatif or prodash crates for CLI status reporting
I think we should switch our output formatting to either [indicatif crate](https://docs.rs/indicatif/latest/indicatif/) or [prodash](https://github.com/Byron/prodash#readme).   `indicatif` supports progress bars and number formatting, so we could probably remove the `size_format` dependency. `prodash` seem to support logging better?

We will probably need a single status bar, as we don't have well defined parallelization.

  - [x] `martin-cp` while generating tiles - this is the most important target
  - [ ] `mbtiles copy` - not certain if this is possible to track the progress of a single SQLite `INSERT INTO ...` sql statement.

### Indicatif Demo
demo

### Prodash Demo
[asciicast](https://asciinema.org/a/315956)

johncarmack1984:
Closes #1024

The `mbtiles` binary gives no sign of life during `copy`, `validate`, `pack` and the other commands that take a while on a large file. The martin-cp half of the issue landed earlier, and this is the mbtiles half, drawn through the same tracing-indicatif layer martin-cp uses, so no new crate enters the lock file. `unpack` shows a progress bar and `pack` a running tile count, both with a rate, and `copy`, `diff`, `apply-patch` and `validate` show a spinner with the elapsed time, since SQLite exposes no progress for the statements they run. Log lines and bars share one writer so they do not interleave, and nothing is drawn when stderr is not a terminal, which is also why the e2e suite has no snapshot of a bar. Log lines emitted inside a command now carry its name, `INFO validate: ...`, and that prefix is the only change in the moved CLI snapshots.

demo-1024-terminal


CommanderStorm (approved):
thanks

Issue #565, opened 2023-02-01
Function Source: getmvt (openmaptiles) - Unimplemented type: 3


dlimkin:
Hello I want use `martin` to display basemap with postgis.
I'm trying to use the `getmvt` function generated by [openmaptiles](https://github.com/openmaptiles/openmaptiles), but I'm getting an error in js libraries MapLibre, MapBox, Deckgl, Leaflet

```
index.js:191 Uncaught Error: Unimplemented type: 3
    at Pbf.skip (index.js:191:20)
    at Pbf.readFields (index.js:41:45)
    at new VectorTileLayer$2 (vectortilelayer.js:20:9)
    at readTile (vectortile.js:13:21)
    at Pbf.readFields (index.js:39:13)
    at new VectorTile$1 (vectortile.js:8:23)
    at FileReader.<anonymous> (Leaflet.VectorGrid.Protobuf.js:124:22)
```
`getmvt` return code `200` and `content-encoding: gzip`

```
$$ 
SELECT mvt, md5(mvt) AS key FROM (SELECT GZIP(STRING_AGG(mvtl, '')) AS mvt FROM (
  SELECT COALESCE(ST_AsMVT(t, 'water', 4096, 'mvtgeometry'), '') as mvtl FROM (SELECT id, ST_AsMVTGeom(geometry, ST_TileEnvelope(zoom, x, y), 4096, 64, true) AS mvtgeometry, class, intermittent, brunnel FROM layer_water(ST_Expand(ST_TileEnvelope(zoom, x, y), 626172.1357121641/2^zoom), zoom)) AS t
    UNION ALL
  SELECT COALESCE(ST_AsMVT(t, 'waterway', 4096, 'mvtgeometry'), '') as mvtl FROM (SELECT ST_AsMVTGeom(geometry, ST_TileEnvelope(zoom, x, y), 4096, 64, true) AS mvtgeometry, name, name_en, name_de, NULLIF(tags->'name:en', '') AS "name:en", NULLIF(tags->'name:ru', '') AS "name:ru", NULLIF(tags->'name:uk', '') AS "name:uk", NULLIF(tags->'name_int', '') AS "name_int", NULLIF(tags->'name:latin', '') AS "name:latin", NULLIF(tags->'name:nonlatin', '') AS "name:nonlatin", class, brunnel, intermittent FROM layer_waterway(ST_Expand(ST_TileEnvelope(zoom, x, y), 626172.1357121641/2^zoom), zoom)) AS t
[...]

PR #3236, opened 2026-09-03
feat(postgres): serve gzip-compressed tiles returned by functions

### LINKED ISSUE #566: Support PG functions that return gzipped values
If a postgres function return gzipped values, we should be able to detect that and treat it accordingly.  The current content detection system is not too expensive to do it on every reply.

See also
- #565

johncarmack1984:
Closes #566

A PostgreSQL function that returns gzip-compressed tiles is served as if the bytes were plain. Martin then compresses them a second time for clients that accept gzip, and hands the compressed bytes untouched to clients that do not, which is the "Unimplemented type: 3" failure in #565. The source now checks the magic bytes of every tile the function returns and reports gzip or zlib on that tile, as the passthrough source already does, so the server passes the bytes through to clients that accept the encoding and decompresses them for the rest. Tables are unaffected since ST_AsMVT never compresses.


martin-core/src/tiles/postgres/source.rs
```diff
@@ -104,24 +104,38 @@ impl Source for PostgresSource {
...
+    fn tile_info_for(&self, data: &[u8]) -> TileInfo {
+        let encoding = if data.starts_with(b"\x1f\x8b") {
+            Encoding::Gzip
+        } else if data.starts_with(b"\x78\x9c") {
+            Encoding::Zlib
+        } else {
+            return self.tile_info;
+        };
+        TileInfo::new(self.tile_info.format, encoding)
+    }
```
CommanderStorm, line 133: Can you move this to a variant of 
https://github.com/maplibre/martin/blob/aa882ab264f2f080887e0e282fadd88a1ed9c87c/martin-tile-utils/src/lib.rs#L260

Likely we can do a bit better on that code though since decompressing fully to know if it is a cartain format seems.. wastefull..

johncarmack1984: Moved: `Encoding::detect` in martin-tile-utils reads the magic bytes only, `TileInfo::detect` now uses it for its own gzip/zlib check, and the source calls it. d9975370

The full decompression in `TileInfo::detect` is untouched here. Happy to take that as a follow-up.

martin-tile-utils/src/lib.rs
```diff
@@ -241,6 +241,18 @@ impl Encoding {
             Self::Gzip | Self::Zlib | Self::Brotli | Self::Zstd => true,
         }
     }
+
+    /// The compression the magic bytes of `value` announce, without decompressing anything
+    #[must_use]
+    pub fn detect(value: &[u8]) -> Option<Self> {
+        if value.starts_with(b"\x1f\x8b") {
+            Some(Self::Gzip)
+        } else if value.starts_with(b"\x78\x9c") {
```
deniial00: This currently recognizes only the default-compression zlib header `78 9c`. Per [RFC 1950 §2.2](https://www.rfc-editor.org/rfc/rfc1950.html#section-2.2), the second byte contains `FLEVEL` plus `FCHECK`, so with the normal `CMF = 0x78`, valid headers include `78 01`, `78 5e`, `78 9c`, and `78 da`. Streams produced at other compression levels would therefore remain marked uncompressed and hit the original double-compression/invalid-tile behavior. Please validate the complete CMF/FLG header and add coverage for at least `78 01` and `78 da`, or narrow the documented support to gzip (docs).

johncarmack1984: Nice catch!

johncarmack1984: Good catch, thanks. `Encoding::detect` now checks the CMF/FLG pair per RFC 1950 §2.2 (deflate, window ≤ 32K, FCHECK) instead of the one byte pair, with the four levels, a smaller window, and three invalid headers covered. 48373dd5

johncarmack1984:
@CommanderStorm let me know what else you'd like done before pulling this one from draft. I ran some benchmarks to make sure this was worth it. The reason to return gzip from a function is that pre-rendered, pre-compressed tiles in a table are the cheapest thing PostgreSQL can hand back, and the numbers bear that out. Same 200 tiles, three function sources, gzip-accepting client, 32 connections, tile cache off, release builds of main and this branch. "Valid" means the client gunzipped what it was sent and got an MVT.

| Function source | valid, main | valid, #3236 | req/s, main | req/s, #3236 | p50 ms, main | p50 ms, #3236 |
|---|---|---|---|---|---|---|
| live `ST_AsMVT` | 200/200 | 200/200 | 4728 | 4885 | 6.00 | 5.79 |
| stored MVT, martin gzips it | 200/200 | 200/200 | 12135 | 13341 | 2.43 | 2.19 |
| stored gzip, passed through | 0/200 | 200/200 | 18122 | 19252 | 1.71 | 1.62 |

The fastest row is the one that does not work on main, due to the bytes getting compressed a second time on the way out. 

The throughput with the check is 4x the live path and 1.4x faster than storing raw MVT, since martin never touches the bytes. 

The two rows that already worked are unchanged within noise. 

Clients that refuse gzip now get the tile decompressed, which costs a gunzip per request where main was returning the compressed bytes as if they were plain.

CommanderStorm (approved):
Thanks

PR #3238, opened 2026-09-03
feat(postgres): add a CQL2 filter to table sources

### LINKED ISSUE #1605: Table sources, support WHERE conditions
That would be very useful to have "where" in table sources:

```yaml
  tables:
    tab_filtered:
      layer_id: table_source

      # Table name (required)
      table: table_source
      
      # SQL where conditions
      where: (TYPE = 'A') and (Pop > 100)
```

johncarmack1984:
Closes #1605

A table source serves every row of its table, and narrowing it to a subset means writing a function source that re-declares by hand what table discovery already knows about columns, ids and bounds. A configured table now takes a `filter` written in CQL2 text, which cql2-rs validates at startup and translates to SQL for the tile query, so a typo fails at load rather than on the first tile, nothing but a CQL2 expression can reach the query, and the same key can later mean the same thing on DuckDB sources. The filter also narrows the bounds calculation, and two entries on the same table with different filters are distinct sources rather than the duplicate they used to warn about.

Builds on #3241, which turned on the serde_json `preserve_order` that cql2 forces. One small thing rides along: license exceptions for two permissive crates cql2 pulls in.


CommanderStorm (changes requested):
I am not entirely sure if adding a where clause is:
- entirely correct
- what we want to do

I would find CQL2 (cql2-rs) a lot cleaner, since we can also use this for duckdb/...

johncarmack1984:
I measured both, and for good measure, current main using existing methods. 

Performance is not affected either way. Pure design discussion, I've got no dog in this fight. 

Here's outcomes:

| Filtered by | Serves the same tiles as this PR | Throughput relative to this PR |
|---|---|---|
| `where` (this PR) | yes, 200/200 | 1.00 |
| CQL2 `filter` through cql2-rs (prototype) | yes, 200/200 | 1.00 |
| A PostgreSQL view, no Martin change | yes, 200/200 | 0.92 |
| A function source, no Martin change | no, 196/200 | 1.04 |
| Nothing (main, all rows) | no, 0/200 | 0.27, tiles 10× larger |

The four filtered variants are the same query to PostgreSQL and land within run-to-run noise of each other. What separates them is not speed:

| | `where` | CQL2 `filter` | View | Function source |
|---|---|---|---|---|
| Needs only a config change | yes | yes | no | no |
| A typo fails at startup | no, on the first tile | yes, at parse | n/a | n/a |
| Same key usable for DuckDB later | no | yes, cql2-rs renders DuckDB SQL | no | no |
| Arbitrary PostgreSQL expressions | yes | no, CQL2 grammar only | yes | yes |
| New dependencies | none | sqlparser, jsonschema, jiff, wkt and a few small crates | none | none |
| Matches Martin's own tile envelope | yes | yes | yes | no, 4 tiles drifted |

So pushed in c1a20e9: `filter` in CQL2 text, parsed and translated by cql2-rs at startup, everything else as it was. 

As for all the changes in 7395f08, c53f06d, 8da6cff, and 60738b6: cql2 enables serde_json's `preserve_order`, so every JSON map martin writes keeps insertion order now, meaning the generated config doc and three snapshots. Only way to undo it would be to PR a feature flag upstream.

Either way, again, I've got no dog in this fight; happy to revert to `where`, leave `cql2`, try the upstream PR, or even close the PR if this functionality isn't wanted anymore.

CommanderStorm (approved):
Very nice.

Instead of linking to ogc, could you steal the docs from for example here? https://docs.3dcitydb.org/1.3/citydb-tool/cql2/

We can then add this as tabs below this section so that what cql2 actually is is more clear.

docs/content/files/generated_config.md
CommanderStorm, line 1: the order here seems to be a bit weird now. intentional?

In any case, I think we should pull the `preserve_order` to another PR since it seems to make reviewiing the PR pretty hard. Good stuff otherwise!

e2e-tests/tests/postgres.rs
```diff
@@ -847,3 +861,89 @@ async fn a_function_returning_gzip_compressed_tiles_is_served_in_the_encoding_th
...
+        .await
+        .expect_err("martin must not start with a filter it cannot parse");
+    let StartError::EarlyExit { status, log } = error else {
+        panic!("expected an early exit, got: {error}");
+    };
+    assert!(!status.success(), "exit status must be a failure: {status}");
+    assert!(
+        log.contains("Filter 'gid <=' is not valid CQL2"),
+        "log must name the filter; log:\n{log}"
+    );
```
CommanderStorm: I think at this point you should know the drill, or? 😉 
insta::assert the log pls


```suggestion
    insta::assert_snapshot!(log, @"");
```

martin/src/config/file/tiles/postgres/config_table.rs
```diff
@@ -93,6 +93,10 @@ pub struct TableInfo {
     #[cfg_attr(feature = "unstable-schemas", schemars(example = &true))]
     pub clip_geom: Option<bool>,
 
+    /// CQL2 filter that limits which rows are served, translated to SQL and added to the tile query with `AND`
+    #[cfg_attr(feature = "unstable-schemas", schemars(example = &"population > 100000"))]
```
CommanderStorm: examples are always the default.. a bit sneaky.. 

```suggestion
```

docs/content/sources-pg-tables.md
```diff
@@ -58,6 +58,101 @@ The TileJSON:
...
+
+    ```yaml
+    # Rows within a bounding box, for a column in SRID 4326
+    filter: S_INTERSECTS(geom, st_setsrid(BBOX(5.9, 45.8, 10.5, 47.8), 4326))
+    # Rows inside a polygon
+    filter: S_WITHIN(geom, st_setsrid(POLYGON((5.9 45.8, 10.5 45.8, 10.5 47.8, 5.9 47.8, 5.9 45.8)), 4326))
+    ```
+
+A function CQL2 does not define is passed on to PostgreSQL by name, which is how `st_setsrid` works above.
+Write such functions in lower case, since `st_area(geom) > 10` calls PostGIS while `ST_Area(geom) > 10` looks for a function named `"ST_Area"` and fails.
```
CommanderStorm:
```suggestion
=== "Spatial and other functions"

    `S_INTERSECTS`, `S_WITHIN`, `S_CONTAINS`, `S_DISJOINT` and the other spatial operators compare the geometry column with a `BBOX(xmin, ymin, xmax, ymax)` or a WKT literal such as `POLYGON((...))`.
    The literal reaches PostGIS without an SRID, and PostGIS refuses to compare geometries whose SRIDs differ, so give the literal the column's SRID with `st_setsrid`.

    ```yaml
    # Rows within a bounding box, for a column in SRID 4326
    filter: S_INTERSECTS(geom, st_setsrid(BBOX(5.9, 45.8, 10.5, 47.8), 4326))
    # Rows inside a polygon
    filter: S_WITHIN(geom, st_setsrid(POLYGON((5.9 45.8, 10.5 45.8, 10.5 47.8, 5.9 47.8, 5.9 45.8)), 4326))
    ```

    A function CQL2 does not define is passed on to PostgreSQL by name, which is how `st_setsrid` works above.
    Write such functions in lower case, since `st_area(geom) > 10` calls PostGIS while `ST_Area(geom) > 10` looks for a function named `"ST_Area"` and fails.
```

CommanderStorm (approved):
minor nits

johncarmack1984:
Failure on this and #3244 are both a CI flake; holding a PR to fix until the renovate PRs finish their CI so we're not racing for the same runners

PR #3239, opened 2026-09-03
feat(mbtiles): copy several source files into one destination

### LINKED ISSUE #965: Allow multiple files to be copied with mbtiles (merge)
When running `mbtiles copy <src> <trg>` we should also support multiple source files, e.g.`mbtiles copy <src1> <src2> <src3> <trg>`

johncarmack1984:
Closes #965

`mbtiles copy` takes one source, so merging tiles from several files means a chain of copies into the same destination by hand. It now takes any number of source files before the destination and copies them into it in order, using the existing `--on-duplicate` mode to decide which file wins where two hold the same tile, which is why that flag is required for more than one source. `--diff-with-file` and `--apply-patch` keep exactly one source, as the issue settled. The `mbtiles` library API is unchanged.

### DIFF
--- a/e2e-tests/tests/mbtiles_cli.rs
+++ b/e2e-tests/tests/mbtiles_cli.rs
@@ -382,10 +382,7 @@ async fn copying_several_sources_needs_on_duplicate() {
         .run_failing()
         .await;
 
-    assert!(
-        output.contains("needs --on-duplicate"),
-        "unexpected output:\n{output}"
-    );
+    insta::assert_snapshot!(output, @"ERROR copying 2 sources into one file needs --on-duplicate to say what happens when two of them hold the same tile");
     assert!(!merged.exists(), "nothing is copied before the check");
 }
 
@@ -406,8 +403,5 @@ async fn copy_with_a_diff_file_takes_one_source() {
         .run_failing()
         .await;
 
-    assert!(
-        output.contains("take exactly one source file"),
-        "unexpected output:\n{output}"
-    );
+    insta::assert_snapshot!(output, @"ERROR --diff-with-file and --apply-patch take exactly one source file, got 2");
 }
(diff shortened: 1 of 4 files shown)


e2e-tests/tests/mbtiles_cli.rs
```diff
@@ -320,3 +320,94 @@ async fn pack_rejects_inconsistent_tile_formats() {
...
+        .arg(&second)
+        .arg("--on-duplicate")
+        .arg("override")
+        .run_failing()
+        .await;
+
+    assert!(
+        output.contains("take exactly one source file"),
+        "unexpected output:\n{output}"
+    );
```
CommanderStorm: instead of .contains(...), please use insta

mbtiles/src/bin/mbtiles.rs
```diff
@@ -176,8 +176,9 @@ enum Commands {
 pub struct CopyArgs {
     /// `MBTiles` file to read from
     src_file: PathBuf,
-    /// `MBTiles` file to write to
-    dst_file: PathBuf,
+    /// Further `MBTiles` files to read from, followed by the file to write to
+    #[arg(required = true, num_args = 1.., value_name = "FILE")]
+    files: Vec<PathBuf>,
```
CommanderStorm: this is super akward for docs and code...
Please flip it so that src_files is the Vec and dst_file is the one.

CommanderStorm (changes requested):
mostly lgtm.
I have two minor cleanup requests

e2e-tests/tests/mbtiles_cli.rs
johncarmack1984: Done, both refusals are inline snapshots now. fc398ac8

mbtiles/src/bin/mbtiles.rs
johncarmack1984: Flipped: `<SRC_FILES>... <DST_FILE>`. fc398ac8

One consequence worth your call: clap cannot resume the list once an option interrupts it, so `copy src --diff-with-file x dst` no longer parses and options have to follow the files. Ten test invocations in mbtiles_diff.rs had a flag between the two files, so I moved them. If that ordering matters to existing users, the fixed-first-source shape from before is the alternative.

CommanderStorm (approved):
tnx

PR #3230, opened 2026-09-03
feat(config): publish each subdirectory of a collection as its own source

johncarmack1984:
Closes #1829

Adds `collections` to every file-backed section, next to `paths` and `sources`. A collection is a directory of project directories, and what a project holds is published under the project's name. Sprites publish each project directory as a source, styles and the tile kinds publish `<project>.<file>`, and fonts load the fonts inside. The tile reloaders watch a collection, so a project that appears later is published without a restart. `paths` and `sources` keep their meaning.

```yaml
sprites:
  collections: /projects/sprites   # /projects/sprites/project1/*.svg -> /sprite/project1.png
styles:
  collections: /projects/styles    # /projects/styles/project1/basic.json -> /style/project1.basic
mbtiles:
  collections: /projects/tiles     # /projects/tiles/project1/roads.mbtiles -> /project1.roads
```

PR #3240, opened 2026-09-04
fix(reload): fix recursive-paths CI flake, recheck when a new dir appears 

johncarmack1984:
The recursive-paths e2e test from #3230 times out on CI now and then, most recently on #3200 and #3236: a file moved into a directory created a moment earlier is never published. The watcher only starts covering a new directory once it has processed the event that created it, so a file that lands inside during that gap raises no event and the reconciler never runs for it. The trigger now schedules one more pass a second after any directory creation, which finds whatever arrived in the gap. Steady state is unchanged, one extra reconcile per new directory, and a unit test pins the second pass


CommanderStorm (approved):
tnx

Issue #875, opened 2023-09-11
Cloud Optimized GeoTIFF (COG) support


sharkAndshark:
Add support for [Cloud Optimized GeoTIFF (COG)](https://www.cogeo.org/).

## CLI
```shell
martin  /path/to/cogs/file.tif  /path/to/directory
```

## Conf file
```yaml
...
cogs:
  paths:
    # scan this whole dir, matching all *.tif files
    - /dir-path
    # specific cog file will be published as cog source
    - /path/to/file1.tif
  sources:
   # named source matching source name to a single file
    cog-src1:  /path/to/file.tif
```

It seems that the conf sections about PMTiles and MBTiles are very similar.  I don't know maybe we could merge them into a single section? 
```yaml
files:
  paths:
    - /dir1
    - /path/to/file1.ext
  sources:
    source1: /path/to/file.ext
```

Things to track:

- [x] Rgb/Rgba with 8 bits per sample
      done by #1590
- [x] Auto-discovery for both `*.tif` and `*.tiff` 
      done by #1635 
- [ ] Gray and 8 bits per sample
- [ ] Other combination of color-type and bits per sample
- [ ] Common compression method which [tiff crate](https://crates.io/crates/tiff) not supported. More research needs to do here.
  - [x] NONE
  - [x] PACKBITS
  - [x] Deflate
  - [X] LZW
  - [X] ModernJPEG
  - [x] ZSTD (since https://github.com/image-rs/image-tiff/pull/256)
  - [ ] LZMA
- [ ] Do not ignore sub-file which is mask, more research needs to do here
- [ ] Remote COG file like S3

nyurik:
I like the idea, thanks!  One aspect that I am not too certain about is combining things into `files` -- one of highly requested features for files it to serve them from `http` -- in order words, pmtiles and i suspect COG files will reside in some cloud bucket, and Martin will access them with the HTTP range requests.

This means that we would need to tell the type of that file without relying on file extension. We could try to auto-detect it - e.g. get first few bytes and detect the type, but it won't be relevant to mbtiles, so in short - much easier to keep things clean and separate.

sharkAndshark:
Some refs:
* [TIFF spec](https://download.osgeo.org/libtiff/doc/TIFF6.pdf)
* [The BigTIFF File Format](https://www.awaresystems.be/imaging/tiff/bigtiff.html)
* [GeoTIFF spec](https://docs.ogc.org/is/19-008r4/19-008r4.html)
[...]

PR #3242, opened 2026-09-04
refactor(pmtiles): extract shared object-store configuration

deniial00:
## Summary

This is the first PR in the stacked implementation for #875.

It extracts PMTiles object-store configuration and remote discovery into reusable components. This prepares the shared infrastructure needed for remote HTTP/S3 COG support without enabling that behavior yet.

## Changes

- Extract PMTiles object-store configuration, credential resolution, client sharing, and URL parsing into `ObjectStoreConfig`.
- Make remote object discovery storage-neutral by injecting:
  - supported extensions
  - object-store parser
  - source builder
  - polling interval and log label
- Preserve explicitly configured, non-secret object-store options in `--save-config`.
- Preserve boolean and numeric option types when saving.
- Redact credentials, SSE-C customer encryption keys, and proxy user information.
- Add an E2E test that restarts Martin from a saved S3-compatible configuration.
- Keep the abstraction gated behind `pmtiles`; COG dependency wiring comes in the next PR.
- Leave generated schemas unchanged for this refactoring PR.

## Scope

This PR does not add remote COG support. It is intended to keep existing PMTiles behavior intact while establishing the shared object-store layer used by the follow-up PRs.

## Verification

- `cargo +1.98 fmt --all -- --check`
- `cargo +1.98 build -p martin --no-default-features --features pmtiles`
- `cargo +1.98 check -p martin --no-default-features --features unstable-cog`
- `cargo +1.98 check -p martin --no-default-features`
- `cargo +1.98 check -p martin --no-default-features --features mbtiles`
- Object-store serialization and redaction unit tests
- Clippy with PMTiles and all targets
- Generated PMTiles schema compared with the committed schema
- Testing with fixtures as well as PMTiles at hand

The mock-S3 E2E test compiles but could not be executed successfully on my local host: both the new test and the existing mock-S3 control test time out during Martin startup. CI should provide the authoritative E2E result.

## Stack

1. **This PR:** shared object-store extraction for PMTiles
2. Async COG reader and explicitly configured HTTP/S3 objects
3. Conditional replacement detection and reload


e2e-tests/tests/pmtiles.rs
```diff
@@ -386,6 +386,40 @@ async fn a_configured_source_is_read_from_an_s3_bucket() {
...
+
+    assert_eq!(martin.get("/s3/1/0/0").await.status(), 200);
+    martin.stop().await;
+    assert_the_aws_environment_was_overridden(&mut martin);
+    martin.assert_startup_warnings();
+
+    let saved = fs::read_to_string(&save_config).expect("martin did not write --save-config");
+    assert!(saved.contains("allow_http: true"));
+    assert!(saved.contains("skip_signature: true"));
+    assert!(saved.contains("virtual_hosted_style_request: false"));
```
CommanderStorm: instead of `.contains`, please use insta::assert.. since less brittle.
Also, the latter auto-applies with me adding the bless label, the first needs actual debugging if it fails.

martin/src/config/file/tiles/discovery/object_store.rs
```diff
@@ -180,9 +190,82 @@ async fn list_remote_prefix(
...
+    result.push_str(url.path());
+    result
+}
+
+#[cfg(test)]
+mod tests {
+    use object_store::memory::InMemory;
+    use object_store::{ObjectStoreExt as _, PutPayload};
+
+    use super::list_remote_prefix;
```
CommanderStorm:
```suggestion
    use super::*;
```

martin/src/config/file/tiles/discovery/object_store.rs
```diff
@@ -1,55 +1,74 @@
...
+    Pin<Box<dyn Future<Output = SourceBuildResult<BuiltSource>> + Send>>;
+pub type ObjectStoreSourceBuilder =
+    Box<dyn Fn(String, Url, CachePolicy) -> ObjectStoreSourceFuture + Send + Sync>;
+pub type ObjectStoreParser = Box<
+    dyn Fn(
+            &Url,
+        )
+            -> object_store::Result<(Box<dyn object_store::ObjectStore>, object_store::path::Path)>
+        + Send
+        + Sync,
```
CommanderStorm: uuuh... that is a bit nasty...

Could you have a look if we move this to an enum (to model the state space..) if we can clean up this with less dyn-pin-box-ing around the place?
If not, fine too..

CommanderStorm (changes requested):
Could you change the `.contains` to an `insta::assert..`?
I know claude likes to do that, but this kind of test is not super maintainable longer term and insta tests are much nicer.

Otherwise LGTM

CommanderStorm:
Plus CI is red...

e2e-tests/tests/pmtiles.rs
deniial00: The test now parses the saved YAML and uses an inline Insta snapshot, while separately checking the dynamic endpoint value. It also restarts Martin from the saved configuration.

martin/src/config/file/tiles/discovery/object_store.rs
deniial00: the test module now uses use super::*;

martin/src/config/file/tiles/discovery/object_store.rs
deniial00: Done in e944c6a. I replaced the boxed async source-builder closure with an explicit ObjectStoreSourceBuilder enum and a Pmtiles variant.

martin/src/config/file/tiles/discovery/object_store.rs
```diff
@@ -146,30 +155,31 @@ fn version_from_meta(meta: &object_store::ObjectMeta) -> Version {
...
-        .map_err(|e| ConfigFileError::ObjectStoreList(e, prefix.to_string()))?
+        .map_err(|error| ConfigFileError::ObjectStoreList(error, sanitized_url(prefix)))?
     {
-        if !meta.location.as_ref().ends_with(PMTILES_EXT_DOT) {
+        let Some(filename) = meta.location.filename() else {
+            continue;
+        };
+        let Some((stem, extension)) = filename.rsplit_once('.') else {
+            continue;
+        };
```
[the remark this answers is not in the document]
deniial00: Fixed in e944c6a. Entries with an empty stem are skipped, and the discovery test now includes .tif coverage.

martin/src/config/file/tiles/pmtiles.rs
```diff
@@ -393,157 +210,48 @@ impl PmtConfig {
...
+                );
+            } else {
+                warn!(
+                    "Environment variable AWS_PROFILE is deprecated. Please use pmtiles.profile in the configuration file instead."
+                );
+                self.profile = Some(profile);
             }
-            info!("Using {env_key} from the environment as pmtiles.{bare} for S3 credentials.");
-            self.options.insert(bare.to_owned(), value);
         }
```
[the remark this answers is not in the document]
deniial00: Fixed in e944c6a. The legacy environment migration now explicitly reads and writes self.object_store.profile.

deniial00:
@CommanderStorm  Thanks for the review. I addressed all comments in e944c6a: switched the saved-config assertions to parsed Insta snapshots, replaced the boxed async builder with an enum, fixed the profile field access, adopted use super::*, added empty-stem handling, and corrected the CI warning mismatch. Fresh checks are running now.

e2e-tests/tests/pmtiles.rs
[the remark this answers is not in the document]
deniial00: Fixed in e944c6a. The test now parses the saved YAML with serde_saphyr into a semantic value and uses a sorted Insta JSON snapshot, so YAML formatting and ordering changes do not affect the assertion.

CommanderStorm (approved):
Thanks

PR #3243, opened 2026-09-04
feat(postgres): publish every overload of a function

### LINKED ISSUE #803: Allow overloaded functions (name+param combinations)
PR https://github.com/maplibre/martin/pull/543 adds a few more unit tests:

```
"MixedCase".function_dup(int, int, int)
"MixedCase"."function_Dup"(int, int, int)
"MixedCase".function_dup(int, int, int, json)
"MixedCase".function_dup(int, int, int, jsonb)
```

Martin should be able to create these with de-duplicated names, e.g. `function_dup.1`, `function_dup.2`, etc.

### Implementation notes
* The complex part is that configuration must be able to support multiple argument types. [Config spec](https://maplibre.org/martin/config-file.html) may need to be expanded, and we would need to support function argument parsing. Parsing should also do some basic PostgeSQL type normalization like `int -> int4` and `bigint -> int8`

```yaml
postgres:
  functions:
    function_source_id:
      schema: "public"
      # Idea 1: more ergonomic, but harder to implement:
      function: "function_zxy_query(int4, int4, int4)"
      # Idea 2: less ergonomic but simpler
      function: function_zxy_query
      args: ["int4", "int4", "int4"]
```

* `fn query_available_function(...) -> InfoMap<InfoMap<(PgSqlInfo, FunctionInfo)>>`, which is a `HashMap<String, HashMap<String, (PgSqlInfo, FunctionInfo)>>` - in other words a map of schema name to a map of function name (unique) to all the data about the given function.  In order to support multiple names, it would have to be changed to something like this: `InfoMap<InfoMap<HashMap<Vec<String>, (PgSqlInfo, FunctionInfo)>>>` (list of arguments is used as a key to arguments).

johncarmack1984:
Closes #803

PostgreSQL lets one function name carry several signatures, and discovery already returns each of them, but the catalog kept the last row per name, so an overloaded function was served by whichever variant the query listed last, behind a warning nobody could act on. A name with a queryless and a query-taking variant is now one source that routes on the request: a bare URL runs the queryless function and a URL with a query string runs the other, with the comments of both merged into its TileJSON. A further variant of the name, such as a `jsonb` twin of the `json` one, becomes its own source with the `.1` suffix the id resolver already hands to tables, and a configuration picks it by writing its signature in `function`, which is what `--save-config` writes back. The discovery query joined comments by name, which would have cross-joined the variants, so it joins by the specific name now.

Measured locally: a million random points, 200 tiles at z8 to z11, a release build with the cache off, oha for 20 s per row. The routed pair answers a bare URL byte-identical to a lone queryless function on all 200 tiles, and a URL with a query string identical to a lone json function, apart from feature order on the tiles a parallel scan returns in a different order between two calls of the same URL. Every row sits within run-to-run noise of its neighbour, so the routing costs nothing measurable.

| Request (c=1) | req/s | p50 ms | p99 ms |
|---|---|---|---|
| routed pair, bare URL (runs the queryless variant) | 491 | 1.4 | 6.2 |
| lone queryless function, bare URL | 543 | 1.2 | 5.8 |
| lone json function, bare URL (what the pair ran before) | 521 | 1.3 | 6.1 |
| routed pair, ?min_pop=500000 (runs the json variant) | 736 | 0.9 | 4.1 |
| lone json function, ?min_pop=500000 | 735 | 0.9 | 4.1 |

| Request (c=64) | req/s | p50 ms | p99 ms |
|---|---|---|---|
| routed pair, bare URL (runs the queryless variant) | 4775 | 12.4 | 28.3 |
| lone queryless function, bare URL | 4683 | 12.7 | 29.7 |
| lone json function, bare URL (what the pair ran before) | 4460 | 13.3 | 30.3 |
| routed pair, ?min_pop=500000 (runs the json variant) | 6183 | 9.5 | 22.6 |
| lone json function, ?min_pop=500000 | 5673 | 10.3 | 25.7 |


docs/content/sources-pg-functions.md
```diff
@@ -174,6 +174,23 @@ END
...
+
+```yaml
+postgres:
+  functions:
+    with_query:
+      schema: public
+      function: function_zxy(integer, integer, integer, json)
+```
+
+A configuration that names only `function_zxy` is refused while the name is overloaded, and the error lists the signatures to choose from.
```
CommanderStorm: > PostgreSQL lets several functions share a name as long as their argument types differ, such as a `function_zxy(z, x, y)` next to a `function_zxy(z, x, y, query json)`.

Would it not be better to let query containing vs queryless functions route to the appropriate function instead?
Are there tradeoffs here?

johncarmack1984: Done in 4e61184c. A name with a queryless and a query-taking variant is one source now. A bare URL runs the queryless function, a URL with a query string runs the other, and the cache key already carried the query string only when the source takes one, so the two do not collide. 

The comments of both are merged into the TileJSON, the query variant's winning. 

A third variant, `jsonb` next to `json`, keeps the numbered suffix and the signature in config, since "has a query string" cannot tell those two apart. 

Fixture has all three now, and the tests ask the same tile with and without `?answer=42`.

Yes, small flexibility tradeoffs in this case: a config can no longer expose only one half of a pair, and a json variant that used to answer a bare URL with `{}` now hands that URL to its queryless twin when one exists.

tests/fixtures/functions/function_dup.sql
CommanderStorm, line 1: okay, I think there are multiple interesting variants to test:

- no json + json => can serve query + ...
- no json + jsonb => can serve query + ...
- json + jsonb => two sources
- no json + json + jsonb + different comments on all => how comments are handled

Could you split the test apart? currently it is not really testing all the cases and it is important that those cases not change because I think that is something people would depend on implicitly

martin/src/config/file/tiles/postgres/resolver/query_functions.rs
```diff
@@ -2,135 +2,226 @@
...
+        }
+    } else {
+        debug!(
+            "Unable to find a SQL comment on {schema}.{function}, a default function description will be used"
+        );
+        None
+    };
 
-    Ok(res)
+    assert!(input_types.len() >= 3 && input_types.len() <= 4);
```
CommanderStorm: What an confusing way to write this..

```suggestion
    assert!(input_types.len() == 3 || input_types.len() == 4);
```

martin/src/config/file/tiles/postgres/resolver/query_functions.rs
```diff
@@ -2,135 +2,226 @@
...
+            );
+        }
+    }
+    assert!(output_type == "bytea" || output_type == "record");
+
+    let mut query = function_call(&schema, &function, &input_types);
+
+    // TODO: Rewrite as a if-let chain:  if Some(names) = output_record_names && output_type == "record" { ... }
+    let mut has_etag_column = false;
+    let ret_inf = if let (Some(names), "record") = (output_record_names, output_type.as_str()) {
```
CommanderStorm: lets fix the todo


```suggestion
    let mut has_etag_column = false;
    let ret_inf = if Some(names) = output_record_names && output_type == "record" {
```

martin/src/config/file/tiles/postgres/resolver/query_functions.rs
```diff
@@ -2,135 +2,226 @@
...
+/// Reads one row of the discovery query into the variant it describes.
+fn parse_variant(row: &Row) -> Variant {
+    let schema: String = row.get("schema");
+    let function: String = row.get("name");
+    let output_type: String = row.get("output_type");
+    let output_record_types = jsonb_to_vec(row.get("output_record_types"));
+    let output_record_names = jsonb_to_vec(row.get("output_record_names"));
+    let input_types = jsonb_to_vec(row.get("input_types")).expect("Can't get input types");
+    let input_names = jsonb_to_vec(row.get("input_names")).expect("Can't get input names");
+    let tilejson = if let Some(text) = row.get("description") {
```
CommanderStorm: lets move all the .get(..) into the fn signature, so that `parse_variant` gets to operate on `schema: String, function: String, ...` which will be a bit more readable and remove the need to expose the `Row` type

martin/src/config/file/tiles/postgres/resolver/scripts/query_available_function.sql
CommanderStorm, line 1: the change here seems suspicious.

Please verify that we still support multiple namespaces

CommanderStorm (commented):
Much better, but I would like this code tested a bit more thoroughly, since I suspect that at least the namespace change looks hella suspicious..

CommanderStorm (approved):
Thanks ❤️

PR #3257, opened 2026-09-04
ci(docker): flake fix: forward RUST_LOG into the image the e2e tests drive

johncarmack1984:
The docker e2e steps pass a fixed list of variables into the container. A test that lowers `RUST_LOG` for its log snapshot, the CQL2 startup-failure test on #3238, gets the whole startup log there and only there, so that leg is red while every other leg is green. `RUST_LOG` is now forwarded when a test sets it and stays unset otherwise, the same way `DEFAULT_SRID` already is.


CommanderStorm (approved):
tnx

PR #3262, opened 2026-09-05
feat(mbtiles): record and honour the tile hash algorithm

### LINKED ISSUE #1086: Store hashing algorithm in the metadata table

johncarmack1984:
Closes #1086

Every hash in an MBTiles file, the per-tile `tile_hash` or `tile_id` and the `agg_tiles_hash`, is now computed with the algorithm the file records in `hash_algorithm`, and every file the tool creates records one. A file without the key is `md5`, so nothing existing changes meaning. The read half landed in #2985; this is the write half plus the algorithms tippecanoe and friends use: `fnv1a` and `xxh64`/`xxh3` as hex through the `sqlite-hashes` features the crate already ships, and `fnv1a-decimal`, which is what tippecanoe actually writes into `tile_id` (64-bit FNV-1a rendered as decimal digits, so its length says nothing about the algorithm and the thread's detect-by-length idea does not work). `mbtiles copy` gives the destination the source's algorithm and `--hash-algorithm` picks another; `bindiff` hashes in Rust with the same function the SQL uses, pinned by a test that runs both on the same bytes against each algorithm's published test vector. A tippecanoe-shaped fixture fails per-tile validation as `md5` and passes once its key says `fnv1a-decimal`. The 45 snapshot changes and the checked-in bindiff fixture are one added `hash_algorithm: md5` metadata row each.


CommanderStorm (approved):
Thanks 🙏

PR #3264, opened 2026-09-05
chore: apply a less strict fmt width

CommanderStorm:
I feel that my monitor is not 60char wide, so I can use up to 100


CommanderStorm:
too much churn

PR #3266, opened 2026-09-05
feat(cog): stream configured remote objects

deniial00:
## Summary

Second PR in the stacked implementation for #875, building on the merged object-store configuration (#3242).

It adds an async, range-based COG reader and lets COG sources be served from explicitly configured remote HTTP and S3-compatible locations, reusing the shared `ObjectStoreConfig`.

## Changes

- Add a storage-neutral `CogReader` trait with two implementations:
  - `LocalFileCogReader`: reads local COGs via blocking I/O off the async runtime, retaining the original OS path representation (non-UTF-8 filenames keep working).
  - `ObjectStoreCogReader`: reads COGs from any configured `object_store` using ranged `GET`s.
- Wire `ObjectStoreConfig` into `CogConfig`, so COG URLs support the same authentication, endpoint, proxy, and client options as PMTiles, including AWS environment/profile discovery.
- Enable `parse_urls` for COG CLI arguments and config parsing; a URL whose path ends in `.tif`/`.tiff` becomes a remote COG source.
- Clamp async-tiff metadata readahead to the known object size, while tile reads keep strict range validation.
- Contain `async-tiff` parser panics on malformed metadata and return them as errors.
- Validate TIFF tile tables (overflow, missing offsets/counts) instead of silently serving empty tiles; sparse tiles (zero byte counts) still return empty responses.
- Redact credentials, query strings, and fragments from COG locations in errors, logs, and `--save-config` output; local `FileConfigSrc` serialization is sanitized the same way.
- Default `cog.allow_http` to `true` so plain HTTP COG URLs work without extra options.
- Teach the E2E mock server `HEAD` and `ETag`s, add large-metadata and sparse COG fixtures, and add direct HTTP and S3-compatible E2E coverage, including a restart from the saved configuration and redaction assertions.

## Scope

This PR does not add remote prefix discovery or reload-driven replacement; conditional object replacement detection and reload come next.

## Verification

- `cargo +1.98 test -p martin --features unstable-cog --no-default-features --lib` (245 passed)
- `cargo +1.98 test -p martin-core --features unstable-cog --no-default-features --lib` (48 passed)
- `cargo +1.98 test --package martin-e2e-tests --features test-cog --test cog` (21 passed)
- `cargo +1.98 test --package martin-e2e-tests --test pmtiles` (16 passed) and `--test config_file` (5 passed)
- `cargo +1.98 fmt --all -- --check`
- `cargo +1.98 clippy -p martin --all-targets -- -D warnings` and `-p martin-core` with `unstable-cog`
- `cargo +1.98 check -p martin --no-default-features` / `--features mbtiles` / `--features pmtiles` / `--features unstable-cog`

## Stack

1. ~~Shared object-store extraction for PMTiles (#3242)~~
2. **This PR:** async COG reader and explicitly configured HTTP/S3 objects
3. Conditional replacement detection and reload


martin-core/src/tiles/cog/image.rs
```diff
@@ -1,179 +1,137 @@
...
+use crate::tiles::cog::{CogError, CogReader};
 
-/// WEBP compression code (not in standard TIFF, registered by GDAL)
-pub const COMPRESSION_WEBP: u16 = 50001;
-
-/// Image represents a single image in a COG file. A tiff file may contain many images.
-/// This struct contains information and methods for taking tiles from the image.
+/// Image represents a single image in a COG file. A TIFF may contain many overview images.
 #[derive(Clone, Debug)]
 pub struct Image {
```
CommanderStorm: please revert the docs changes you did here

CommanderStorm, line 17: please revert the changes to this struct. they seem unrelated mostly

martin-core/src/tiles/cog/image.rs
```diff
@@ -1,179 +1,137 @@
...
-        if let Some(tables) = jpeg_tables {
-            return Ok(merge_jpeg_tables_with_tile(&tables, &tile_data));
+        if self.is_passthrough_compression() {
+            let CompressedBytes::Chunky(bytes) = tile.compressed_bytes() else {
+                return Err(CogError::AsyncTiff(
+                    async_tiff::error::AsyncTiffError::General(
+                        "planar TIFFs are not supported".to_owned(),
+                    ),
+                    location.to_owned(),
+                ));
```
CommanderStorm: lets model the state space properly, ok?
The string returns are harder for consumers to handle than nessary..

```suggestion
                return Err(CogError::UnsipportedPlanarLayout(location.to_owned()));
```

martin-core/src/tiles/cog/image.rs
```diff
@@ -185,178 +143,146 @@ impl Image {
...
-///
-/// In TIFF JPEG compression, the quantization and Huffman tables are often stored
-/// separately in the `JPEGTables` tag and shared across all tiles. Each tile then only
-/// contains the frame data without these tables.
-///
-/// `JPEGTables` format: SOI (FFD8) + tables (DQT, DHT, etc.) + EOI (FFD9)
-/// Tile data format: SOI (FFD8) + frame header + scan data + EOI (FFD9)
-///
-/// To merge: Take tables (without SOI/EOI) and insert after tile's SOI, before frame data.
 fn merge_jpeg_tables_with_tile(jpeg_tables: &[u8], tile_data: &[u8]) -> Vec<u8> {
```
CommanderStorm, line 204: please revert the changes to thus fns docs

deniial00:
Follow-up pushed as d12c966b on top of the branch (no history rewrite):

- clippy is clean for `-p martin-core -p martin-e2e-tests --all-targets`
- the e2e GeoJSON conversion now emits `_layer` as the first property and catalog JSON normalization sorts object keys, making test output deterministic
- the replacement-detection reader test now clones the store as an explicit `Arc<InMemory>`

Scoped validation on the pushed head:
- `just test-cog`: 245 `martin` lib + 49 `martin-core` lib + 21 COG e2e tests pass
- `cargo test -p martin-e2e-tests --test catalog`: both tests pass against the default-features `martin` binary

CI was re-triggered by the push and is running.

martin/src/config/file/object_store.rs
```diff
@@ -612,3 +612,87 @@ impl CredentialProvider for AwsSdkCredentialProvider {
...
+            "http://localhost:9000"
+        );
+
+        let state = config
+            .resolve(&IdResolver::new(RESERVED_KEYWORDS))
+            .await
+            .unwrap();
+        let saved = serde_saphyr::to_string(&config.with_catalog(&state.tile_manager)).unwrap();
+        let saved: serde_json::Value = serde_saphyr::from_str(&saved).unwrap();
+        assert_eq!(saved["cog"]["aws_endpoint"], json!("http://localhost:9000"));
```
CommanderStorm:
```suggestion
        insta::asssert_json_snapshot!(saved["cog"], @"");
```

martin/src/config/file/object_store.rs
```diff
@@ -612,3 +612,87 @@ impl CredentialProvider for AwsSdkCredentialProvider {
...
+        assert_eq!(value["allow_http"], true);
+        assert_eq!(value["skip_signature"], false);
+        for key in [
+            "aws_access_key_id",
+            "secret_access_key",
+            "google_bearer_token",
+            "azure_storage_sas_key",
+        ] {
+            assert!(value.get(key).is_none(), "{key} must be redacted");
+        }
```
CommanderStorm, line 642:
```suggestion
        let value = serde_json::to_value(config).unwrap();
        insta::asssert_json_snapshot!(value, @"");
```

e2e-tests/tests/cog.rs
```diff
@@ -129,6 +129,141 @@ cog:
...
+
+    let requests = statics.request_log().await;
+    assert!(requests.lines().any(|line| line.starts_with("HEAD ")));
+    assert!(
+        requests
+            .lines()
+            .filter(|line| line.starts_with("GET "))
+            .all(|line| line.contains("bytes=")),
+        "every COG GET must be a byte-range request:\n{requests}"
+    );
```
CommanderStorm:
```suggestion
    insta::assert_snapshot!(requests, @"");
```

e2e-tests/tests/cog.rs
```diff
@@ -129,6 +129,141 @@ cog:
...
```
CommanderStorm:
```suggestion
    insta::assert_snapshot!(requests, @"");
```

e2e-tests/tests/cog.rs
```diff
@@ -244,9 +379,9 @@ async fn the_shape_of_a_tile_response() {
 
     let tile = martin.get("/usda_naip_128_none_z2/18/42712/97343").await;
     insta::assert_snapshot!(tile.headers_snapshot(), @r#"
```
CommanderStorm, line 378: why did the size change here? THis seems suspicious..

CommanderStorm (changes requested):
I have a few requests for how to clean this up, but it boils down to
- revert the two cases where the previously helpfull docs were purged
- investgage and tell my why the cog size changed
- use insta instead of brittle string contains in tests

e2e-tests/tests/cog.rs
CommanderStorm: Investigated: not a regression. The old `tiff`-crate reader corrupted the padding region of partial/overview edge tiles (this tile comes from an overview page where the real content is 64x128, padded into a 128x128 tile). The new async-tiff reader decodes it correctly — verified pixel-for-pixel against `tifffile`'s reference decode — and the genuinely-black padding compresses smaller (5286 -> 5132 bytes). Blessed the snapshot.

CommanderStorm (approved):

PR #3268, opened 2026-09-05
feat(cog): detect and reload replaced remote objects

deniial00:
## Summary

Third PR in the stacked implementation for #875. Contains a single commit on top of #3266 (stacked locally on its `feat/cog-remote-reader` branch; GitHub does not yet support fork-internal base branches, so the PR lists both layers — the second commit is the increment).

Configured remote COG objects were loaded once at startup and never replaced. This PR wires them into the existing reload machinery with **conditional replacement detection**: each pass sends one `HEAD` per object, derives a version from its ETag or last-modified timestamp, and the generic `ReloadDriver` diff rebuilds only the objects that actually changed.

## Changes

- Add a storage-neutral `ConfiguredObjectDiscovery` alongside the PMTiles prefix discovery. It covers the `sources` map entries and any `paths` entry that names one remote object; local files stay with the notify-driven `FsDiscovery`, and prefix URLs remain out of scope for the next PR.
- Drive the remote half with a `PollTrigger` and a new `cog.reload_interval` setting — default `10m`, `0s` disables — mirroring `pmtiles.reload_interval` semantics.
- Seed the remote baseline at startup without applying, since `config.resolve()` already loaded these sources; only later version changes trigger rebuilds.
- On a failed check, retain the object's last-known version instead of dropping it, so a transient store outage cannot surface as a source removal. (Prefix listing in the PMTiles reloader currently warns-and-skips, which the version diff reads as removals; this PR does not change that, but the same pattern can be adopted there.)
- Keep per-source cache bounds, `convert_to_*`/`cache_control` overrides, and `--save-config` provenance for reloaded remote objects.
- E2E: the mock object store now serves mutable content with content-hash `ETag`s, and a new test covers both directions — an unchanged object costs only `HEAD` requests and never rebuilds, while a replaced object is reloaded within the poll window.

## Scope

No prefix listing (remote `paths` directories) and no generated-schema changes; remote prefix discovery and polling, then documentation and schema updates, follow in the stack.

## Verification

- `cargo +1.98 test --package martin-e2e-tests --features test-cog --test cog` (22 passed, including the new replacement test)
- `cargo +1.98 test -p martin --features unstable-cog --no-default-features --lib` (249 passed, 3 new discovery unit tests)
- `cargo +1.98 test -p martin-core --features unstable-cog --no-default-features --lib` (48 passed)
- `cargo +1.98 test --package martin-e2e-tests --test pmtiles` (16 passed) and `--test config_file` (5 passed)
- `cargo +1.98 fmt --all -- --check`; clippy `-D warnings` on `martin`, `martin-core`, `martin-e2e-tests` (default features); `cargo check` on `martin --no-default-features --features unstable-cog` and `--features pmtiles`

## Stack

1. ~~Shared object-store extraction for PMTiles (#3242)~~
2. ~~Async COG reader and explicitly configured HTTP/S3 objects (#3266)~~
3. **This PR:** conditional replacement detection and reload


martin/src/config/file/tiles/discovery/object_store.rs
```diff
@@ -43,11 +52,239 @@ impl ObjectStoreSourceBuilder {
...
+            && let Some(sources) = &cfg.sources
+        {
+            for (id, src) in sources {
+                let Ok(SourceLocation::ObjectStore(url) | SourceLocation::Http(url)) =
+                    SourceLocation::classify_path(src.get_path())
+                else {
+                    // Local sources belong to the file-based discovery.
+                    continue;
+                };
+                let resolved = id_resolver.resolve(id, url.to_string());
```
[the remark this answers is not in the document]
deniial00: Fixed in 046432f: both discovery resolve sites now use the sanitized URL (scheme, host, port, path only). I also switched the two startup resolution sites in `file_config.rs` to `sanitize_url` so the ids stay identical between startup and reload, and the collision warnings there cannot leak credentials either.

martin/src/config/file/tiles/reload/cog.rs
```diff
@@ -24,44 +33,81 @@ impl CogReloader {
...
-        let trigger = NotifyTrigger::new(&directories, recursive)?;
-        self.driver.spawn(trigger, Baseline::Initialized);
+
+        if has_remote {
+            if interval.is_zero() {
+                tracing::info!(
+                    "CogReloader: remote object polling disabled (reload_interval = 0s)"
+                );
+            } else {
+                let trigger = PollTrigger::new(interval);
```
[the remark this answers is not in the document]
deniial00: Fixed in 046432f: the remote half now uses `PollTrigger::after_interval`, so the first post-seed poll happens after the configured interval.

martin/src/config/file/tiles/cog.rs
```diff
@@ -36,14 +31,97 @@ pub struct CogConfig {
...
+        }
+        if !self.cache.is_empty() {
+            map.serialize_entry("cache", &self.cache)?;
+        }
+        if self.reload_interval != DEFAULT_RELOAD_INTERVAL {
+            map.serialize_entry(
+                "reload_interval",
+                &humantime_serde::re::humantime::format_duration(self.reload_interval).to_string(),
+            )?;
+        }
```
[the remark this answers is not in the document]
deniial00: Fixed in 046432f: `reload_interval` is serialized with `humantime_serde::Serde::from`, matching the PMTiles and Postgres configs.

deniial00:
Synced with the reader branch and pushed without rewriting history:

- 2e37dea4 merges `feat/cog-remote-reader` in, carrying its lint and snapshot-stabilization fixes (deterministic e2e GeoJSON property order, sorted catalog JSON keys)
- dc1a5f23 replaces the fully qualified `crate::tiles::cog::CogObjectMeta` return type with a regular import

Scoped validation on the pushed head:
- `cargo clippy -p martin-core -p martin-e2e-tests --all-targets` is clean
- `just test-cog`: 247 `martin` lib + 49 `martin-core` lib + 22 COG e2e tests pass, including the replacement-detection scenario

CI was re-triggered by the push and is running.

deniial00:
## Pre-existing flaws surfaced by the review — should they be fixed here?

The review flagged three P2 findings that this stack does not address. I checked attribution: **none of the three behaviors is introduced by this PR** — all exist unchanged on `main` (they come from the PMTiles reload machinery, #3242). This PR only makes them reachable for remote COGs.

1. **A failed live update is treated as successfully reconciled**
   `tile_source_manager.rs` + `config/file/tiles/driver/reconcile.rs`: under `on_invalid: warn`, `apply_changes` skips the failed source and returns `Ok(())`, so the driver advances the baseline and the broken version is never retried. The commit-on-`Ok(())` logic is untouched by this PR — the diff only rewires how the *baseline* is seeded.

2. **HTTPS query authentication is silently discarded**
   `config/file/object_store.rs`: the HTTP builder receives only the URL origin, so `?token=...` on presigned URLs never reaches `HEAD`/range `GET` requests. The production logic predates this PR; the stack's E2E appends `?token=secret-query` but only asserts the token is *redacted* from saved config, not that it is *sent*.

3. **Weak version fingerprints can make an object permanently unreloadable**
   `config/file/tiles/discovery/object_store.rs`: versions derive from `ETag`, else `last_modified` — `ObjectMeta::version` and `size` are ignored; HTTP stores without `Last-Modified` report epoch, producing a constant tracked version. The fingerprint logic is inherited from the pre-existing PMTiles discovery (renamed/refactored here, not changed).

Two smaller related items in the same review, also pre-existing: the case-sensitive/inconsistent URL extension classifier (`file_config.rs` vs. the discovery's `eq_ignore_ascii_case`), and the metadata-readahead clamp edge case that can produce an inverted range when the requested start lies past EOF (`cog/reader.rs`).

**Question for the maintainers:** given these predate this stack and the PR only exposes them to the new remote-COG path — should I fix them as part of this work, and if so, as a separate follow-up PR rather than growing this one? Happy to split either way.

CommanderStorm (changes requested):
Could you rebase/merge with main? It is not very reviewable currently. Will do initial vreview after

martin/src/config/file/tiles/driver/reconcile.rs
CommanderStorm, line 1: the changes here do seem like an merge conflict gone arie.
Why are we removing the `Baseline::StartupResolved`

martin/src/config/file/tiles/discovery/object_store.rs
```diff
@@ -162,15 +417,28 @@ impl Discovery for ObjectStoreDiscovery {
...
+        meta.e_tag.as_deref(),
+        Some(meta.last_modified.timestamp_millis()),
+    )
+}
+
+#[cfg(feature = "unstable-cog")]
+pub(crate) fn version_from_cog_meta(meta: &CogObjectMeta) -> Version {
+    version_from_parts(meta.e_tag.as_deref(), meta.last_modified_millis)
+}
+
```
CommanderStorm: please inline these two, and also why is this pub(crate)?

martin/src/config/file/tiles/discovery/object_store.rs
```diff
@@ -43,11 +54,255 @@ impl ObjectStoreSourceBuilder {
...
+    objects: Vec<ConfiguredObject>,
+    reload_interval: Duration,
+    /// Last-known version per object id, for the retention-on-failure behavior.
+    last_versions: Mutex<BTreeMap<String, Version>>,
+    parser: ObjectStoreParser,
+    build: ObjectStoreSourceBuilder,
+    process: ResolvedProcess,
+}
+
+impl ConfiguredObjectDiscovery {
```
CommanderStorm, line 97: why is there so much code added here?
It feels like I have seen below code already.
Can you run an duplicate search and try to find an better abstraction than this?

If there is no better abstraction, that would be fine, but this seems kind of strange.

CommanderStorm (changes requested):
Seems fairly straitforward.

I have two architectureal questons, but otherwise LGTM

martin/src/config/file/tiles/discovery/object_store.rs
deniial00: Agreed. I will remove the crate-visible helper and keep the conversion local instead.

martin/src/config/file/tiles/driver/reconcile.rs
deniial00: You are right—the rebase replayed an older driver refactor over the newer main API. I will restore `Baseline::StartupResolved` and adapt the COG reloader to it.

martin/src/config/file/tiles/discovery/object_store.rs
deniial00: Agreed. I will compare the existing discovery paths with a duplicate scan and consolidate the shared configured-source planning rather than keep parallel parsing.

CommanderStorm (approved):
LGTM

PR #3269, opened 2026-09-05
test: make the Windows-only unit test job pass again

johncarmack1984:
The `Unit test the server (windows-latest)` job runs on pushes to main only, and it has been red on every push since aa20fd32: #3260 got it past the CRLF failures, and the next problems were waiting behind them. Five snapshot tests build a path with `Path::join` or take one from a temp file and pin it as written with forward slashes, so on Windows the value comes back with backslashes: `style_external` in martin-core, the three style catalog snapshots in `styles.rs`, the discovery error in `fs.rs`, and the JSON diagnostic in `binary_diagnostics.rs`, where the temp path arrives with its backslashes escaped and the mask never matches. Behind those, the e2e crate imports two helpers only its `cfg(not(windows))` tests use, which the job's deny-warnings mode turns into an error. The tests now replace the platform separator before snapshotting, the way the e2e catalog and geojson tests already do, the diagnostic test masks the escaped form as well, and the two imports carry the same cfg as their users. Verified by running the job's recipe on a Windows machine with warnings denied: every package, the e2e crate and the doc tests pass.

### DIFF
--- a/e2e-tests/tests/mbtiles.rs
+++ b/e2e-tests/tests/mbtiles.rs
@@ -1,6 +1,8 @@
 //! `MBTiles` sources, both configured up front and served from a watched directory.
 
-use martin_e2e_tests::{Martin, WatchedDir, fixture, mbtiles_fixture, mbtiles_from_sql};
+use martin_e2e_tests::{Martin, WatchedDir, mbtiles_fixture};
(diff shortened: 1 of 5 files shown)


CommanderStorm (approved):
thanks 👀

PR #3271, opened 2026-09-05
test: experiment if using nextest can speed up CI

CommanderStorm:


CommanderStorm:
not actually faster, but I still like it more

PR #3255, opened 2026-09-04
ci: split CI into rust, ui and docs workflows gated on what changed

CommanderStorm:
Splits `ci.yml` into an orchestrator plus `rust.yml`/`ui.yml`/`docs.yml`/`schemas.yml`/`release.yml`, running each suite only when a change can affect it and deferring the Windows/macOS builds to the top of a stack.

PR #3282, opened 2026-09-09
ci: attach the binaries before the release is published

johncarmack1984:
`martin-v1.16.0` went out with no binaries, and main's docs job has been red on every push since: the draft was published by hand before its upload ran, and that run's docs job 404s on every `releases/latest/download/*` link while an empty release is latest, which since #3255 skips the release job (https://github.com/maplibre/martin/actions/runs/34157558587).

release-plz.toml describes draft, attach, attest, then publish, but release.yml only attached on `release: published`, so a hand publish was always the first step. Now the push run that drafted the release attaches, attests and publishes it, with the release-plz PAT so the event still fires for the docker tags and the Homebrew PR. The release-event run uploads only for a draft someone published by hand, and takes the Homebrew checksums from the published assets, so attested files are never clobbered.

The publish loop and the asset lookup were run locally against 1.15.0 and 1.16.0 with `gh release view`; the workflow itself only runs on a real release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -64,6 +64,36 @@ jobs:
             mv "$dir" "target/$target"
             just package-assets "$target"
           done
+      - name: Save the packaged binaries for release-plz
+        if: github.event_name == 'push'
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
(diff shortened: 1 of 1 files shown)


johncarmack1984:
(note: CI won't pass until the binaries are attached to the release, this PR just automates that process in future releases)

CommanderStorm (approved):
Lets try it

PR #3291, opened 2026-09-10
chore: add a benchmark to make sure we don't regress on cache usage ops

CommanderStorm:


martin/benches/sources.rs
```diff
@@ -1,10 +1,13 @@
+use std::hash::{BuildHasher, RandomState};
```
CommanderStorm:
```suggestion
use std::hash::{BuildHasher as _, RandomState};
```

PR #3293, opened 2026-09-10
perf(pmtiles): read local files in place instead of through object_store

johncarmack1984:
A local PMTiles source under concurrent load holds one blocking thread per in-flight request, and in a CPU-limited container the scheduler throttles the whole process for most of every second while the same archive as MBTiles runs unthrottled.

Local files went through object_store's `LocalFileSystem`, which takes two `spawn_blocking` hops per read, one to open and stat the file and one to read the range. Local paths and `file://` URLs now go through the pmtiles crate's `MmapBackend`, as they did before #2251, and remote stores are unchanged. Each read stamps the file's inode, mtime and size as the data version, so a replaced file still surfaces as `SourceModified` and reloads as in #2858. Each read stats the file before touching the mapping and fails with `SourceModified` when it changed, so a file rewritten in place is detected before any page of it is read; only a rewrite racing the read itself can still fault, the same exposure as before #2251. Windows refuses to write a mapped file in place, so replacement there is by rename, which works.

Measured on the 40 MB Berlin archive from the warefeats tile server benchmark, one z14 tile, `cache: disable`, 4 workers, Docker with `cpus: 4`, oha inside the container's network namespace, 12 s passes, threads sampled mid-pass. The MBTiles rows serve the same tiles from the MBTiles copy of the archive and are not changed by this PR.

| Source | Build | Connections | Threads | req/s | p50 ms | p99 ms | Throttled periods | Memory after load |
|---------|---------|------------:|--------:|-------:|-------:|-------:|------------------:|------------------:|
| MBTiles | main | 10 | 20 | 17,340 | 0.54 | 1.09 | 78 of 174 | 21 MB |
| PMTiles | main | 10 | 43 | 19,764 | 0.46 | 0.88 | 149 of 174 | 36 MB |
| PMTiles | this PR | 10 | 10 | 39,746 | 0.25 | 0.39 | 11 of 175 | 14 MB |
| MBTiles | main | 100 | 20 | 12,995 | 7.61 | 9.58 | 0 of 173 | 95 MB |
| PMTiles | main | 100 | 165 | 17,718 | 4.39 | 35.4 | 166 of 174 | 130 MB |
| PMTiles | this PR | 100 | 10 | 33,646 | 2.69 | 5.05 | 32 of 175 | 26 MB |

Throttled periods are `nr_throttled` of `nr_periods` from the container's `cpu.stat`, memory is the `docker stats` usage.

### DIFF
--- a/martin-core/src/tiles/pmtiles/backend.rs
+++ b/martin-core/src/tiles/pmtiles/backend.rs
@@ -1,80 +1,41 @@
 //! Byte-range backends behind a [`PmtilesSource`](super::PmtilesSource).
 
-use std::fs::{File, Metadata};
-use std::future::ready;
-use std::io;
+use std::fs::Metadata;
 #[cfg(unix)]
-use std::os::unix::fs::{FileExt as _, MetadataExt as _};
-#[cfg(windows)]
-use std::os::windows::fs::FileExt as _;
+use std::os::unix::fs::MetadataExt as _;
 use std::path::PathBuf;
 use std::time::SystemTime;
 
-use pmtiles::{AsyncBackend, BackendResponse, ObjectStoreBackend, PmtResult};
+use derive_debug::Dbg;
+use pmtiles::{AsyncBackend, BackendResponse, MmapBackend, ObjectStoreBackend, PmtResult};
 
-/// Reads a local `PMTiles` file in place with positional reads on one open handle.
+/// Reads a local `PMTiles` file through a memory map.
 ///
 /// Every read reports the file's inode, size and modification time as its data version.
 /// A replaced file therefore surfaces as [`pmtiles::PmtError::SourceModified`].
-#[derive(Debug)]
+#[derive(Dbg)]
 pub(crate) struct PmtFileBackend {
-    file: File,
+    #[dbg(skip)]
+    mmap: MmapBackend,
     path: PathBuf,
 }
 
 impl PmtFileBackend {
-    /// Opens `path` for reading.
-    pub(crate) fn open(path: impl Into<PathBuf>) -> io::Result<Self> {
+    /// Maps `path` for reading.
+    pub(crate) async fn open(path: impl Into<PathBuf>) -> PmtResult<Self> {
         let path = path.into();
-        let file = File::open(&path)?;
-        Ok(Self { file, path })
-    }
-
-    /// Reads `length` bytes at `offset`, clamped to the end of the file.
-    fn read_blocking(&self, offset: usize, length: usize) -> PmtResult<BackendResponse> {
-        let meta = std::fs::metadata(&self.path)?;
-        let available = meta.len().saturating_sub(offset as u64);
-        let length = usize::try_from(available).map_or(length, |a| length.min(a));
-        let mut buf = vec![0_u8; length];
-        read_exact_at(&self.file, &mut buf, offset as u64)?;
-        Ok(BackendResponse::new_with_version(
-            buf.into(),
-            data_version(&meta),
-        ))
+        let mmap = MmapBackend::try_from(&path).await?;
+        Ok(Self { mmap, path })
     }
 }
 
 impl AsyncBackend for PmtFileBackend {
-    fn read(
-        &self,
-        offset: usize,
-        length: usize,
-    ) -> impl Future<Output = PmtResult<BackendResponse>> + Send {
-        // A page cache read is cheaper than a hop to the blocking pool, so it runs inline.
-        ready(self.read_blocking(offset, length))
+    async fn read(&self, offset: usize, length: usize) -> PmtResult<BackendResponse> {
+        let version = data_version(&std::fs::metadata(&self.path)?);
+        let mut response = self.mmap.read(offset, length).await?;
+        response.data_version_string = Some(version);
+        Ok(response)
     }
-}
-
-#[cfg(unix)]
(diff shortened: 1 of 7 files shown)


martin-core/src/tiles/pmtiles/backend.rs
```diff
@@ -0,0 +1,110 @@
...
+    /// Opens `path` for reading.
+    pub(crate) fn open(path: impl Into<PathBuf>) -> io::Result<Self> {
+        let path = path.into();
+        let file = File::open(&path)?;
+        Ok(Self { file, path })
+    }
+
+    /// Reads up to `length` bytes at `offset`.
+    /// Returns fewer bytes at the end of the file.
+    fn read_blocking(&self, offset: usize, length: usize) -> PmtResult<BackendResponse> {
```
CommanderStorm: why `read_blocking` instead of the respective `tokio` facilities?

CommanderStorm (changes requested):
Instead of the manually async `read_blocking`, could you please use the tokio facilities to do IO? that is much more efficient and straightforward than the blocking manual buffering approach you used.

johncarmack1984:
Yep, refactoring now.

johncarmack1984:
So, not to be difficult, but rather to provide full context for decision-making: I did the refactor both ways and ran benchmarks on each so we could see what it costs.

`tokio::fs::File` has no positional read, so the literal version is `try_clone`, `seek` and `read_exact` per tile; the other keeps one handle and does one `spawn_blocking` per read with `read_exact_at`, which is what tokio's own file IO does underneath (its docs: "Tokio's file uses spawn_blocking behind the scenes, and this has serious performance consequences", https://docs.rs/tokio/latest/tokio/fs/index.html). Same container and rig as the PR table, oha inside the namespace, one 12 s pass each:

| Build | Connections | Threads | req/s | p50 ms | p99 ms | Throttled periods | Memory after load |
|-------|------------:|--------:|------:|-------:|-------:|------------------:|------------------:|
| main | 100 | 160 | 13,745 | 5.97 | 30.0 | 163 of 174 | 132 MB |
| `tokio::fs::File` | 100 | 121 | 20,617 | 4.25 | 24.8 | 131 of 174 | 140 MB |
| one `spawn_blocking` per read | 100 | 110 | 20,913 | 4.41 | 23.9 | 111 of 174 | 209 MB |
| this PR, inline | 100 | 10 | 25,486 | 3.66 | 9.42 | 17 of 174 | 18 MB |
| main | 10 | 42 | 14,834 | 0.61 | 1.61 | 54 of 174 | 36 MB |
| `tokio::fs::File` | 10 | 30 | 20,141 | 0.44 | 1.29 | 3 of 175 | 30 MB |
| one `spawn_blocking` per read | 10 | 35 | 25,273 | 0.37 | 0.78 | 79 of 174 | 86 MB |
| this PR, inline | 10 | 10 | 30,963 | 0.31 | 0.69 | 4 of 174 | 12 MB |

Both tokio versions beat main, and both keep the bottleneck this PR found for local files under load: a thread per in-flight request, so 100 connections still put over 100 threads on 4 CPUs and the cgroup throttles them through most of the run. The inline read is the only row that stays at 10 threads. A pread of this tile out of the page cache measures 8 us on this rig, and the rest of the gap between those rows is the cost of handing that read to another thread and back, which is why the mmap backend before #2251 read inline too.

The loop you flagged is gone in both variants, `read_exact_at` does it (Windows keeps a small shim since it has no `read_exact_at`), and that change will get pushed either way (edit: done in a1ce1798).

My preference would be the inline read on this data. If you want file IO kept off the workers as a policy, I'll gladly defer and push the `spawn_blocking` version.

CommanderStorm:
🤔 How about we go back to `MmapBackend` in that case for local files?
I don't like the blocking since that is not very efficient..

I only migrated because I thought that simplifiying the architecture would be more important I think and did not reaaaly consider the perf impacts of that switch in particular.

johncarmack1984:
Done in eec11364: local files go through `MmapBackend` again, with the data version stamped from a `stat` per read so #2858's reload still fires (edit: d6c5d570 moves that stat ahead of the mapping read and swaps the same-fixture test for two that replace the archive with different contents; a same-size copy is invisible on Windows, where the version has no inode and `CopyFile` keeps the mtime). Same rig and run as before, 100 connections:

| Build | Threads | req/s | p50 ms | p99 ms | Throttled periods | Memory after load |
|-------|--------:|------:|-------:|-------:|------------------:|------------------:|
| main | 165 | 17,718 | 4.39 | 35.4 | 166 of 174 | 130 MB |
| inline `read_exact_at` | 10 | 28,180 | 3.02 | 6.50 | 41 of 175 | 64 MB |
| `MmapBackend` | 10 | 33,646 | 2.69 | 5.05 | 32 of 175 | 26 MB |

mmap beats the inline read on throughput, both latencies and memory, so that's the PR now and the body says so. One thing on record there: a file rewritten in place is detected by the stat before any page of the mapping is read, and only a rewrite racing the read itself can still fault, the same exposure as before #2251 (edit: before d6c5d570 this said the process aborts). Windows refuses to write a mapped file at all and rename-over works there, so replacement by rename is the path everywhere.

And honestly, #2251 makes perfect sense to me. I always prefer the simpler mental model and might have done the same. I only caught this because of some benchmarks I was running in a side project. Meanwhile I'd been perfectly content with the object_store version ^_^

CommanderStorm (approved):
LGTM

PR #3287, opened 2026-09-09
feat(cog): discover and poll remote prefixes

deniial00:
## Summary

Fourth PR in the stacked implementation for #875. Contains two commits on top of #3268 (stacked locally on its `feat/cog-configured-reload` branch; GitHub does not yet support fork-internal base branches, so this PR lists both layers — the final two commits are the increment).

Remote COG prefix URLs in `cog.paths` are now discovered and kept current through the existing object-store reload machinery. Each poll lists the prefix, derives versions from ETags or last-modified timestamps, and applies additions, replacements, and removals without downloading complete objects.

## Changes

- Extend the storage-neutral `ObjectStoreDiscovery` to build COG sources under remote prefixes while preserving the existing PMTiles integration.
- Recognize `.tif` and `.tiff` objects case-insensitively, derive stable source IDs from object names, and preserve URL authority, port, path, query, and fragment internally while redacting credentials from diagnostics and saved configuration.
- Poll configured COG prefixes through a dedicated `ReloadDriver`; `cog.reload_interval` controls both configured-object checks and prefix listing, with `0s` disabling remote polling.
- Retain the last successful prefix listing when a store temporarily fails, preventing an outage from appearing as mass source removal.
- Preserve query parameters on HTTP object requests so query-token and presigned URLs remain usable at runtime; fragments and credentials remain excluded from diagnostics and persisted configuration.
- Make reload application report the IDs actually applied. Under `on_invalid: warn`, failed startup sources, additions, and replacements retain retryable baseline state instead of becoming permanently missing or stale.
- Expand the HTTP/S3-compatible E2E server with mutable listings, strict query authentication, and controllable read failures.
- Add E2E coverage for prefix discovery, replacement, addition, removal, query-authenticated HTTP reads, startup recovery, and repeated replacement retries.

## Scope

No documentation or generated-schema updates; those remain the final PR in the stack. Local COG directories remain notify-driven and unchanged.

## Verification

- `just test-cog` (254 Martin tests, 50 martin-core tests, 25 COG E2E tests)
- COG E2E suite passed twice consecutively (25/25 each run)
- `cargo test -p martin --features unstable-cog config::file::tiles::driver::reconcile::tests` (14 passed)
- PMTiles E2E suite (16 passed)
- MBTiles E2E suite (62 passed)
- `cargo clippy -p martin --features unstable-cog --all-targets -- -D warnings -A unfulfilled-lint-expectations`
[...]

PR #3340, opened 2026-09-17
fix(reload): retry sources skipped by warn policy

deniial00:
## Summary

- report source builds skipped under `on_invalid: warn`
- advance reload baselines only for successfully applied additions and updates
- seed startup baselines from sources actually present in the live manager

Split from #3287 to keep the reload retry fix focused.

## Tests

- `cargo build --workspace`
- `cargo test -p martin --features unstable-cog config::file::tiles::driver::reconcile::tests`
- `cargo test -p martin --features unstable-cog tile_source_manager::tests::watcher_loop_around_persistent_bad_file -- --exact`


CommanderStorm (approved):
LGTM

PR #3341, opened 2026-09-17
fix(object-store): retain listings after transient failures

deniial00:
## Summary

- cache the last successful listing for each remote object-store prefix
- reuse retained entries when a later listing fails transiently
- keep first-listing failures non-fatal and covered by the existing skip behavior

## Tests

- `cargo clippy --workspace --all-targets`
- `cargo test -p martin discovery::object_store::tests`


CommanderStorm (approved):
makes senese to me

PR #3342, opened 2026-09-17
fix(object-store): preserve remote URL components

deniial00:
## Summary

- preserve HTTP query parameters when constructing the remote object store
- retain scheme, user info, port, query, and fragment while deriving listed object URLs
- require the configured query in the HTTP COG end-to-end fixture

Split from #3287 to keep the URL handling fix focused.

## Tests

- `cargo build --workspace`
- `cargo test -p martin --features unstable-cog config::file::tiles::discovery::object_store::tests::prefix_discovery_filters_extensions_and_preserves_object_urls -- --exact`
- `just test-cog`


martin/src/config/file/tiles/discovery/object_store.rs
```diff
@@ -449,7 +441,8 @@ mod tests {
             ))
         });
         let entries = list_remote_prefix(
-            &Url::parse("s3://bucket/imagery/").unwrap(),
+            &Url::parse("https://user:secret@example.com:8443/imagery/?token=secret#fragment")
```
CommanderStorm, line 444: just fyi, this is an correct change, but we may happily log this, so maybe not great for your infra...

CommanderStorm (approved):
LGTM

martin/src/config/file/tiles/discovery/object_store.rs
deniial00: Hmm, good catch - still only happens when a file needs to be renamed. So probably won't happen often. I don't need support for query params but I can submit a follow-up PR.

PR #3344, opened 2026-09-17
feat(cog): discover and poll remote COG prefixes

deniial00:
## Summary

Remote COG prefixes in `cog.paths` are now discovered and kept current through the object-store reload machinery, completing the COG layer of #875 alongside the existing configured-object and directory handling.

- `CogReloader` gains a third `ReloadDriver` that lists each remote prefix (`s3://`, `https://`, …) on a `PollTrigger` at `cog.reload_interval` and applies additions, replacements, and removals; `0s` disables remote polling.
- The e2e object-store stub becomes mutable (insert/remove/replace, transient GET failures) and answers S3 `ListObjectsV2`, so prefix re-listing is observable.
- New e2e coverage: prefix discovery + replacement + removal, startup reads that fail and heal, and failed updates retried until the replacement reads.

Supersedes #3287; the independent fixes split out of it were merged as #3340, #3341, #3342.

## Tests

- `cargo build --workspace`
- `cargo clippy --workspace --all-targets`
- `just test-cog`: 316 martin unit tests, martin-core unit tests, 44 COG e2e tests (incl. the three new ones)


e2e-tests/tests/cog.rs
```diff
@@ -37,11 +37,9 @@ async fn tilejson(martin: &Martin, id: &str) -> Value {
 fn assert_remote_reads_use_ranges(
     requests: &str,
     path: &str,
-    query: Option<&str>,
     expected_reads: usize,
     ranges: &[&str],
 ) {
-    let path = query.map_or_else(|| path.to_owned(), |query| format!("{path}?{query}"));
```
CommanderStorm, line 44: did we not add this last PR? is this a merge issue?

e2e-tests/tests/cog.rs
```diff
@@ -253,12 +264,143 @@ cog:
...
+    .expect("the source must load once reads heal");
+
+    // The failure was expected; consume the warnings the harness would otherwise flag. Every
+    // failure line carries `error=`, regardless of the shape the log takes.
+    martin.assert_log_contains("Tile source resolution warning");
+    let drained = martin.take_log_lines("error=");
+    assert!(
+        !drained.is_empty(),
+        "the failing reads must have been logged"
+    );
```
CommanderStorm: please actually assert thwat the log is, not just that it is non empty. same as the other PRs. use insta

e2e-tests/tests/cog.rs
```diff
@@ -253,12 +264,143 @@ cog:
...
+        .start()
+        .await
+        .expect("failed to start martin with a read-failing remote COG");
+
+    // Reads fail, so the object must not be served; the source is absent, not stale.
+    let catalog = async || {
+        martin.get("/catalog").await.json()["tiles"]
+            .get("remote")
+            .is_some()
+    };
```
CommanderStorm: requests like this please always inline. Tests are supposed to be DAMP, not DRY because otherwise they get hard to debug

e2e-tests/tests/cog.rs
```diff
@@ -253,12 +264,143 @@ cog:
...
         "/cogtest/usda_naip_128_none_z2.tif",
-        None,
         2,
         &["bytes=0-32767", "bytes=1284-66819"],
     );
 }
 
+/// A configured COG whose object is served but whose range reads fail at startup must load the
+/// moment reads heal, even though the object's version never changes. This is the P1 regression
+/// for "warn-policy sinks advance the reload baseline past a source they skipped".
```
CommanderStorm: please remove all comments. tests need to be named and wrtitten in a way that it is obvious what we are doing without comments

e2e-tests/tests/cog.rs
```diff
@@ -333,6 +475,73 @@ cog:
...
+    assert_eq!(martin.get("/second/19/85424/194685").await.status(), 200);
+
+    statics.remove(first_key);
+    martin.wait_for_source_removed("first").await;
+    martin.stop().await;
+
+    let requests = statics.request_log().await;
+    let list_count = requests
+        .lines()
+        .filter(|request| {
```
CommanderStorm: please insta snapshot the lines instead

CommanderStorm (changes requested):
Mostly LGTM, but the tests need a bit of cleanup to make them stronger because currently they have a bit of the usual LLMy style that is not quite testing what we want.

e2e-tests/tests/cog.rs
deniial00: Done. The exact ListObjectsV2 request lines are now covered by an inline insta snapshot.

e2e-tests/tests/cog.rs
deniial00: Yes, it was stale after the split. Rebased onto main and restored the query-aware helper from #3342.

e2e-tests/tests/cog.rs
deniial00: Done. The catalog requests are now inlined at each assertion.

e2e-tests/tests/cog.rs
deniial00: Done. The expected error lines now use inline insta snapshots with only endpoint and timing values filtered.

e2e-tests/tests/cog.rs
deniial00: Done. Removed the explanatory comments and kept the behavior explicit in test names and assertions.

PR #3345, opened 2026-09-17
feat(duckdb): accept .parquet and .duckdb paths as CLI positional args

CommanderStorm:
`martin data.parquet` / `martin tiles.duckdb` now work like the mbtiles/pmtiles shorthands


martin/src/config/args/root.rs
```diff
@@ -192,6 +195,27 @@ impl Args {
             config.cog = parse_file_args(&mut cli_strings, &["tif", "tiff"], true);
         }
 
+        #[cfg(feature = "unstable-duckdb")]
+        if !cli_strings.is_empty() {
+            let geoparquet = parse_file_paths(&mut cli_strings, &["parquet"], false, false)
```
CommanderStorm:
```suggestion
            let geoparquet = parse_file_paths(&mut cli_strings, &["parquet", "geoparquet"], false, false)
```

PR #3349, opened 2026-09-18
test(e2e): sort JSON keys so the snapshots pass in a workspace-wide build

johncarmack1984:
`just test` and `cargo nextest run --workspace` fail 26 e2e snapshots on a clean main. CI stays green because every workflow picks its packages, so this only bites locally. Since #3241 the e2e crate gets serde_json's `preserve_order` only when cargo builds it alongside martin and mbtiles, so its JSON keys are sorted in one build and in wire order in the other, and no snapshot can pass both. The harness now sorts keys where it parses JSON and where it writes the PMTiles fixture's metadata, the way #3241 did for `generated_config.md`. No snapshot changes.

As a side fix this drops a pending `.snap.new` that was committed in wire order and ignores `*.snap.new` next to `*.pending-snap`. insta deletes that file on every passing run, which leaves the tree dirty.


e2e-tests/src/pmtiles.rs
```diff
@@ -56,6 +56,7 @@ fn coord(z: i64, x: i64, row: i64) -> TileCoord {
 ///
 /// The `json` row is itself an object holding `vector_layers`, which readers expect alongside the
 /// other rows rather than nested, so it is merged in.
+/// The keys are sorted so the archive has the same bytes in every build.
```
CommanderStorm:
```suggestion
```

CommanderStorm (approved):
thanks!

PR #3351, opened 2026-09-18
feat!: drop the single-colon `${VAR:default}` substitution syntax

johncarmack1984:

`${VAR:default}` has warned since #2943 that it was going away, and now it's gone. Use `${VAR:-default}`. A config still on the old form fails at startup with the substitution error, and its help line already names the `:-` form.


docs/content/config-file/index.md
```diff
@@ -18,9 +18,9 @@ default `connection_string: ${MY_DATABASE_URL:-postgres://postgres@localhost/db}
...
 ```
 
-!!! warning "Deprecation of single-colon interpolation"
-    The legacy single-colon default `${MY_DATABASE_URL:postgres://postgres@localhost/db}` is still
-    accepted for backward compatibility, but is deprecated in favor of the `:-` form shown above.
+!!! warning "Single-colon defaults were removed in v2.0"
+    The single-colon default `${MY_DATABASE_URL:postgres://postgres@localhost/db}` is no longer accepted.
+    Use the `:-` form shown above.
 
 !!! tip "auto-generate a config file with `--save-config`"
```
CommanderStorm:
```suggestion
!!! tip "auto-generate a config file with `--save-config`"
```

CommanderStorm (approved):
Thanks!

PR #3352, opened 2026-09-18
feat!: default `allow_http` to `false`

johncarmack1984:

`allow_http` now defaults to `false` for PMTiles and COG, as the 0.20.0 changelog promised. An `http://` source without it is refused at startup with an error that names the option. Left alone, that failure reads `HTTP error: builder error`.

`--config` and positional sources are exclusive, so `martin http://host/tiles.pmtiles` can't opt in anymore and needs a config file. Happy to add an `--allow-http` flag if you'd rather keep that working.

COG carried its own silent `true` to match PMTiles, so it flips with it.


CommanderStorm (approved):
❤️

PR #3353, opened 2026-09-18
feat!: drop the deprecated cache keys

johncarmack1984:

`cache_size_mb`, `tile_cache_size_mb`, `directory_cache_size_mb` and `pmtiles.dir_cache_size_mb` are no longer migrated. They now get the same unrecognized-key warning as any typo, and the cache falls back to its default size. Use `cache.size_mb`, `cache.tile_size_mb` and `pmtiles.directory_cache.size_mb`.

One thing to know: the migration only warned when both spellings were set. A config with just the old key moved silently, so v2 is the first those users hear of it. Probably worth a line in the migration guide.


CommanderStorm (approved):
🎉

PR #3354, opened 2026-09-18
feat(rendering): mark server-side style rendering as stable

CommanderStorm:
Server-side style rendering is no longer experimental: it runs on a configurable worker pool now, so this drops the `EXPERIMENTAL` labels, the "Expect breaking changes in upcoming releases" startup warning, and the docs notes saying its behavior may change in patch releases.


docs/content/sources-styles/rendering.md
```diff
@@ -39,6 +35,17 @@ styles:
     rendering: true
 ```
 
+Renders run on a dedicated thread pool. `rendering: true` sizes it from the
+logical CPU count, clamped to 2..=8; the long form sets the number of render
+threads explicitly:
```
CommanderStorm:
```suggestion
Renders run on a dedicated thread pool.
`rendering: true` sizes it from the logical CPU count, clamped to `2..=8`.
The long form sets the number of render threads explicitly:
```

PR #3362, opened 2026-09-19
refactor: dispatch tile sources through a closed AnySource enum

CommanderStorm:
Replaces the `Arc<dyn Source>` trait object with a closed `AnySource` enum

Stacked on #3359


martin-core/src/tiles/any.rs
```diff
@@ -0,0 +1,205 @@
+//! [`AnySource`], the closed set of tile sources the server dispatches on.
+//!
+//! Built-in sources are reached by a `match` rather than a vtable, so their
+//! `get_tile_with_etag` is a direct call instead of an `#[async_trait]` boxed
+//! future.
```
CommanderStorm:
```suggestion
//! [`AnySource`], the closed set of tile sources the server dispatches on.
```

PR #3363, opened 2026-09-19
feat!: drop the legacy AWS env vars and config keys

johncarmack1984:

`AWS_S3_FORCE_PATH_STYLE`, `AWS_SKIP_CREDENTIALS`, `AWS_NO_CREDENTIALS` and `AWS_PROFILE` are no longer read, as the 0.20.0 changelog promised. `aws_s3_force_path_style`, `force_path_style`, `aws_skip_credentials` and `aws_no_credentials` are no longer migrated and get the same unrecognized-key warning as any typo. Use `virtual_hosted_style_request` (inverted), `skip_signature` and `profile`.

Three things to know:

- A leftover env var is ignored silently. Off AWS, a public bucket that leaned on `AWS_SKIP_CREDENTIALS=1` then fails at startup on the instance metadata lookup, and the error doesn't name the variable. Probably worth a line in the migration guide.
- COG read `AWS_PROFILE` too, silently, through the shared object-store config (#3242). That read goes with the rest.
- `martin s3://public-bucket/tiles.pmtiles` can't ask for unsigned requests anymore, so it needs a config file with `skip_signature: true`. Same shape as `allow_http` in #3352.


e2e-tests/src/martin.rs
```diff
@@ -126,9 +126,7 @@ impl MartinBuilder {
         cmd.current_dir(workspace_root())
             // The environment the justfile runs the tests under.
             .env_remove("DATABASE_URL")
-            .env_remove("AWS_PROFILE")
             .env("RUST_LOG_FORMAT", "bare")
-            .env("AWS_SKIP_CREDENTIALS", "1")
```
johncarmack1984, line 131: The harness set `AWS_SKIP_CREDENTIALS=1` on every start and 67 tests asserted its warning. Most of the diff is those assertions going away.

CommanderStorm (approved):
Thanks!

PR #3364, opened 2026-09-19
fix(mbtiles): compute tile hashes when copying from a dedup-id file

johncarmack1984:
`mbtiles copy` from a planetiler file (`tiles_shallow` + `tiles_data`) into `normalized` or `flat-with-hash` wrote the integer `tile_data_id` where the tile hash goes. The copy then fails `mbtiles validate`:

```
ERROR At least one tile has mismatching hash: stored value is `1` != computed value `E65F5FDBC963CBE2E6112C787C9B7BBD` in MBTile file out.mbtiles
```

`mbtiles diff` and `mbtiles copy --apply-patch` did the same whenever one of their inputs was such a file. A dedup-id file now gets its hashes computed from the tile data, the way a flat file does.

With no `--dst-type`, `mbtiles diff` and `mbtiles copy --apply-patch` from such a file died with `no such table: images`. They now default to `map` + `images`, as `mbtiles copy` already did.

Found while benchmarking the dedup-id default for #3333.


mbtiles/src/copier.rs
```diff
@@ -836,7 +836,7 @@ fn get_select_from_apply_patch(
     } else {
         fn get_tile_hash_expr(tbl: &str, typ: MbtType, algorithm: HashAlgorithm) -> String {
             match typ {
-                Flat | Cache => {
+                Flat | Cache | MbtType::DEDUP_ID => {
                     let hash = algorithm.sql_hash(&format!("{tbl}.tile_data"));
                     format!("IIF({tbl}.tile_data ISNULL, NULL, {hash})")
```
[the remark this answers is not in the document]
johncarmack1984: Fair. Both now take the same default as `copy`, and the test runs without a `--dst-type`. 9d1e810e

mbtiles/src/validation.rs
```diff
@@ -244,6 +244,12 @@ pub enum MbtType {
 }
 
 impl MbtType {
+    /// What [`Mbtiles::detect_type`] reports for a `tiles_shallow` + `tiles_data` file
+    pub(crate) const DEDUP_ID: Self = Self::Normalized {
```
CommanderStorm: please inline these, I think it makes the code simpler to read 😉 
We can add method to construct it though to reduce boilerplate by maybe a line or so.

CommanderStorm (approved):
LGTM assuming you inline 😆

mbtiles/src/validation.rs
johncarmack1984: Done in 7520109!

PR #3371, opened 2026-09-22
feat!: stop reading the Postgres env vars

### LINKED ISSUE #1052: Remove automatic ENV var usage except inside config files

johncarmack1984:
Closes #1052

Last of the v2.0 musts from #3333. Martin no longer reads `DATABASE_URL`, `DEFAULT_SRID`, `PGSSLCERT`, `PGSSLKEY` or `PGSSLROOTCERT` from its environment. #3150 added the missing flags, #3151 added the warning, this is the removal both promised. Sources come from the command line or the config file, and a config file can still pull any variable in with `${DATABASE_URL}`.

Three things to know:

- `martin` with only `DATABASE_URL` exported now stops with "No tile sources found". The other four go quiet instead: a table with SRID 0 gets skipped with its usual warning, and a client certificate that used to ride in on `PGSSLCERT` is not sent. Worth a line in the migration guide.
- Docker users put the connection string after the image name, or keep `-e DATABASE_URL` and write `connection_string: ${DATABASE_URL}` into a mounted config. The docs examples moved with it, the demo compose and the install snippet on the demo site too.
- `--ssl-cert`, `--ssl-key` and `--ca-root-file` replace the three `PGSSL*` files, on the command line and on top of a config.


e2e-tests/src/lib.rs
```diff
@@ -61,6 +61,19 @@ pub fn workspace_root() -> PathBuf {
         .to_path_buf()
 }
 
+/// The client certificate flags for the test database, taken from the `PGSSL*` variables in the harness's environment.
+fn pg_ssl_args() -> Vec<OsString> {
```
johncarmack1984, line 65: `just test-ssl-cert` handed the client certificate to the binaries through the removed variables. The harness now turns `PGSSL*` in its own environment into the flags, on the config path too, which the old read never reached.

e2e-tests/src/martin.rs
```diff
@@ -140,6 +139,10 @@ impl MartinBuilder {
         }
         if let Some(url) = &self.database_url {
             cmd.env("DATABASE_URL", url);
+            if !self.args.iter().any(|arg| arg == "--config") {
```
johncarmack1984, line 142: Without a config the URL goes on the command line, since martin refuses positional connections next to `--config`. `DATABASE_URL` stays in the subprocess environment for the configs that expand it. Same in `martin_cp.rs`.

martin/src/config/args/postgres.rs
```diff
@@ -48,21 +47,21 @@ pub struct PostgresArgs {
...
-        self,
-        cli_strings: &mut Arguments,
-        env: &impl Env,
-    ) -> OptOneMany<PostgresConfig> {
-        let connections = Self::extract_conn_strings(cli_strings, env);
-        let default_srid = self.get_default_srid(env);
-        let certs = self.get_certs(env);
+    pub fn into_config(self, cli_strings: &mut Arguments) -> OptOneMany<PostgresConfig> {
+        let connections = Self::extract_conn_strings(cli_strings);
+        let certs = PostgresSslCerts {
```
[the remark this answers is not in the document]
johncarmack1984: Added `an_exported_database_url_alone_is_not_a_source`: `DATABASE_URL` set, nothing else, martin has to stop with the no-sources line. 5f42c473. The other four have nothing to show without a source, and the `env` parameter is gone from the whole args path, so a read cannot come back by accident.

CommanderStorm (approved):
tnx

